PR #232, opened 2018-09-17
Python 3.5 compatibility

adocherty:
As `random.choices` is unavailable for python 3.5 I've replaced it with `random.choice` and also introduced random states to allow the walkers to operate isolated from the default random state.

Also fixed an issue which prevented random walks not being repeatable with different python versions when they relied on GraphWalk.adj -- this was being created directly from the graph adjacency which seemed to change in different python/networkx versions.


adocherty:
I also get the same warnings. The tensorflow warnings must be from Keras, as we don't call tensorflow directly. I also get the same warnings with Python 3.6 on the develop branch when I update to the latest tensorflow - so they are not to do with this pull request I think.

Issue #125, opened 2018-07-19
Determine performance difference between stellar-ml and biased random walk reference implememtation


PantelisElinas:
### Description

The Node2Vec reference implementation (RI) employs two tricks to speed up random walks: (a) it uses alias sampling, and (b) pre-computes transition probabilities for all nodes and edges. The current implementation of BiasedRandomWalk does not use the above. The RI uses more memory (to store the transition probabilities and alias tables) and also requires more memory to store the latter, in order to speed up sampling. The stellar implementation uses less memory and does no pre-processing but might pay a penalty for computing the discrete distributions and sampling from them using standard method at every step of a walk.

This issue is about having a closer look at the two implementations, evaluating the differences in performance and memory requirements as a function of the graph size and node degree distributions so that we can choose one of the two methods to go into the library or provide both implementations and advise the user when to use each of them.

### User Story
As a: Research Engineer
I want: to know the differences in computational performance between the two different approaches
so that: I can implement the one most suitable for the stellar-ml library.

### Done Checklist (Development)
- [ ] Implementation of RI method for random walks
- [ ] Performance evaluation of two algorithms
- [ ] Pull request

PR #236, opened 2018-09-18
Fix sampling bug in BiasedRandomWalk and make it 6-7x faster (& add benchmarks)

### LINKED ISSUE #234: BiasedRandomWalk handles the d_tx = 1 "stay in the neighbourhood" case using the first node of a walk, not the previous node
**Describe the bug**

The BiasedRandomWalks use the neighbours of the first node in the walk, for the d_tx = 1 case from the node2vec paper. They should be using the neighbours of the second-to-last ("previous") node in the walk. 

The `previous_node_neighbours` variable never gets updated:
https://github.com/stellargraph/stellargraph/blob/537bd252e59fe72a44fb5aa8c6056fff4f05ed54/stellargraph/data/explorer.py#L275

**To Reproduce**

Run a random walk on a graph where there are squares, like

```
0--1
|  |
2--3
```

If `p = q = 1e10` as above, and a walk goes 0-2-3 or 0-1-3, it should have equal chance of continuing to either 1 or 3 (the d_tx = 0 and d_tx = 2 case have unnormalised transition probabilities 1/p = 1/q), but the walks will always continue either 0-2-3-1 or 0-1-3-2.

```python
import stellargraph as sg
import stellargraph.data.explorer as sge
import networkx as nx

graph = nx.Graph()

# 0-2
# | |
# 1-3

graph.add_edges_from([
    (0, 1),
    (0, 2),
    (2, 3),
    (1, 3),
])


graph = sg.StellarGraph(graph)

walker = sge.BiasedRandomWalk(graph)

[walk for walk in walker.run([0], 100, 1e10, 1e10, 4) if walk[:3] == [0, 1, 3]]
```
```
[[0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2],
 [0, 1, 3, 2]]
```

See https://colab.research.google.com/drive/1m_ctgC2_m8o0n5ZhXl1Gxnv5AdBuAJTR

**Expected behavior**

Half the walks should be 0-1-3-2 and half should be 0-1-3-1 in the example above.

**Screenshots**

NA.

**Desktop (please complete the following information):**

Stellargraph 0.4.0.

huonw:
This does a lot of tweaks to `BiasedRandomWalk`.

It first fixes #234 by updating `previous_node_neighbours`.

It then adds benchmarks for all the explorers (using http://pytest-benchmark.readthedocs.io/) and uses these to prove some optimisations to BiasedRandomWalk are actually optimisations (which is related to #125). The sequence of performance is:

```
Name (time in us)                                       Min                Mean              Median                   Max             StdDev                IQR            Outliers
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_biasedrandomwalk (0010_3088fcb)      62.4390 (1.0)       73.5903 (1.0)       70.9465 (1.0)        251.1770 (1.0)      12.2770 (1.0)      11.3350 (1.0)       541;310
test_benchmark_biasedrandomwalk (0011_16f289f)     416.4540 (6.67)     459.5375 (6.24)     434.2510 (6.12)       996.6660 (3.97)     59.4566 (4.84)     38.4620 (3.39)      160;145
test_benchmark_biasedrandomwalk (0009_dc88536)     422.3280 (6.76)     475.0922 (6.46)     446.2395 (6.29)     1,285.6580 (5.12)     64.1259 (5.22)     51.8620 (4.58)      198;142
```

The first (fastest) one corresponds to the last commit, the second to the second last and the third to the third last.

If you'd prefer to not have the (dev) dependency on `pytest-benchmark`, I can remove the benchmarks. Please also see the individual commit messages for more details.


adocherty:
I like the benchmarking! We can use that elsewhere to check for unexpected (or expected!) performance changes 🕙

huonw:
Hm, do I also need to submit this `develop`, or are we going to automatically merge hotfix-0.4.1 into `develop` (which I don't think is in the gitflow model)? Also, by switching this to the other graph, it included https://github.com/stellargraph/stellargraph/commit/99444ba0246dfeea1da20767d70bfd4ff5b9b7c1, which probably shouldn't be in a 0.4 branch?

PR #273, opened 2018-10-11
Fixed the bug in LinkSequence that threw an error when no link targets were given

youph:
- Fixed the bug in `LinkSequence` class that occurred when no link targets were given to the link iterator.
- Added a test, `test_GraphSAGELinkGenerator_no_targets`, covering the case of link generator with no targets (e.g., for prediction)

### DIFF
--- a/tests/mapper/test_link_mappers.py
+++ b/tests/mapper/test_link_mappers.py
@@ -374,3 +374,17 @@ class Test_HinSAGELinkGenerator(object):
 
         with pytest.raises(IndexError):
             nf, nl = mapper[2]
+
+    def test_HinSAGELinkGenerator_no_targets(self):
+        """
+        This tests link generator's iterator for prediction, i.e., without targets provided
+        """
+        G = example_HIN_1(self.n_feat)
+        links = [(1, 4), (1, 5), (0, 4), (0, 5)]  # selected ('movie', 'user') links
+        data_size = len(links)
+
+        gen = HinSAGELinkGenerator(
+            G, batch_size=self.batch_size, num_samples=self.num_samples
+        ).flow(links)
+        for i in range(len(gen)):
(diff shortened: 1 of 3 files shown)


tests/mapper/test_link_mappers.py
```diff
@@ -241,6 +241,17 @@ def test_GraphSAGELinkGenerator_no_samples(self):
         with pytest.raises(ValueError):
             nf, nl = mapper[0]
 
+    def test_GraphSAGELinkGenerator_no_targets(self):
+        """
+        This tests link generator's iterator for prediction, i.e., without targets provided
+        """
+        G = example_Graph_2(self.n_feat)
+        iter = GraphSAGELinkGenerator(
```
adocherty: Shadowing built-in names is not a good idea, let's rename this variable.

tests/mapper/test_link_mappers.py
```diff
@@ -241,6 +241,17 @@ def test_GraphSAGELinkGenerator_no_samples(self):
...
+            G, batch_size=self.batch_size, num_samples=self.num_samples
+        ).flow(G.edges())
+        for i in range(len(iter)):
+            assert iter.__getitem__(i)[1] is None
```
adocherty: Woudn't `iter[i][1]` be more readable?

adocherty:
Hi Yuri, I think we should also check the HinSAGE link generator in the same way and I've added such a test.  I also looked at the node_generators and I was not testing them with targets well. I've added some tests for those as well to this PR.

youph:
@adocherty thanks for adding the tests!

youph:
@adocherty strange: I've marked your comments as resolved, but github still says I need to address the change request. Does it wait for you to mark the comments as resolved as well?

PR #278, opened 2018-10-19
Added GraphSAGE MaxPoolAggregator and MeanPoolAggregator and tests

adocherty:
Two aggregators have been implemented:

MaxPoolAggregator (from the GraphSAGE paper)

MeanPoolAggregator (This is not quite the GCN aggregator, but similar)

### DIFF
--- a/stellargraph/layer/graphsage.py
+++ b/stellargraph/layer/graphsage.py
@@ -19,7 +19,7 @@
 GraphSAGE and compatible aggregator layers
 
 """
-__all__ = ["GraphSAGE", "MeanAggregator"]
+__all__ = ["GraphSAGE", "MeanAggregator", "MaxPoolAggregator", "MeanPoolAggregator"]
 
 import numpy as np
 from keras.engine.topology import Layer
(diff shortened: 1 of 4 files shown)


stellargraph/layer/graphsage.py
```diff
@@ -119,9 +119,300 @@ def call(self, x, **kwargs):
...
+        assert output_dim % 2 == 0
+        self.half_output_dim = int(output_dim / 2)
+
+        # TODO: These should be user parameters
+        self.hidden_dim = output_dim
+        self.hidden_act = activations.get("relu")
+
+        self.has_bias = bias
+        self.act = activations.get(act)
+        self.w_neigh = None
```
youph: self.w_neigh doesn't seem to be used

stellargraph/layer/graphsage.py
```diff
@@ -119,9 +119,300 @@ def call(self, x, **kwargs):
...
+          x: Keras Tensor
+
+        Returns:
+            Keras Tensor representing the aggregated embeddings in the input.
+
+        """
+        # x[0]: self vector (batch_size, head size, feature_size)
+        # x[1]: neighbour vector (batch_size, head size, neighbours, feature_size)
+        xw_neigh = K.dot(x[1], self.w_pool) + self.b_pool
+        xw_self = K.dot(x[0], self.w_self) + self.b_self
```
youph: I'd expect `MaxPoolAggregator` and `MeanPoolAggregator` to have similar structure, differing only in the `K.max` or `K.mean` operation applied to the neighbours' features. Here, the two classes are also different in parameters: for example, the `MeanPoolAggregator` class has `self.b_self bias`, which has no counterpart in the `MaxPoolAggregator` class.

stellargraph/layer/graphsage.py
```diff
@@ -119,9 +119,300 @@ def call(self, x, **kwargs):
 
         from_self = K.dot(x[0], self.w_self)
         from_neigh = K.dot(neigh_means, self.w_neigh)
-        total = K.concatenate([from_self, from_neigh], axis=2)
+        h_out = K.concatenate([from_self, from_neigh], axis=2)
 
-        return self.act(total + self.bias if self.has_bias else total)
+        if self.has_bias:
```
youph: I see you've replaced the return statement with a more readable one here and in the other two GraphSAGE aggregators, but left the original return statement in HinSAGE's mean aggregator (line 138).

adocherty: I was a bit confused by the return statement and wasn't sure it was doing what I expected. I think the more verbose statement is clearer.

I think graphsage and hinsage aggregators should be synced – I will do this when I implement the pooling aggregators for hinsage :-)

stellargraph/layer/graphsage.py
adocherty: I've modified the meanpool class to act the same as the maxpool class.

stellargraph/layer/graphsage.py
adocherty: True! I renamed what I used in `build(...)` and forgot to update the code in `__init__`.

adocherty:
Addressed review comments and checked the original GraphSAGE code. The implementations of MaxPoolAggregator and MeanPoolAggregator should now match the reference code, although the reference code allows multiple Dense layers to be stacked when calculating the neighbour activations, as they mention in the paper.

I don't think we should replicate this behaviour here.  I have an idea to make our GraphSAGE more flexible and allow all aggregators to be implemented with less work.  I'll make a new issue for it.

stellargraph/layer/graphsage.py
youph: So now that MeanPoolAggregator is changed to match the structure of MaxPoolAggregator, this no longer applies, right?

stellargraph/layer/graphsage.py
youph: I've also renamed these aggregators to MeanPoolingAggregator and MaxPoolingAggregator

stellargraph/layer/graphsage.py
youph: Ok, sounds good.

youph:
FYI: When experimenting with new aggregators, you've introduced a mismatch in the aggregators used in train() and test() in `demos/node-classification-graphsage/graphsage-cora-example.py`, which would throw an error upon loading a saved model. I've fixed that.

stellargraph/layer/graphsage.py
adocherty: Yes – this no longer applies.

adocherty:
Thanks for catching the aggregator mismatch!

youph:
@adocherty 
- please delete the merged branch
- move the issue to Done (if it's really done)

PR #276, opened 2018-10-16
add github merged pr plugin

TheDen:
PR to use the plugin https://github.com/stellargraph/github-merged-pr-buildkite-plugin (forked from seek, and adapted for our needs)

What it does/features

* Runs build on the merge state pull requests, rather than pre-merge
* If a merge isn't doable, it'll revert to running the build of the `HEAD` of the PR
* Merge happens outside the scope of the steps

PR #283, opened 2018-11-06
Revert "add github merged pr plugin"

youph:
Reverts stellargraph/stellargraph#276

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ scikit_learn==0.19.2
 matplotlib==2.2.3
 pandas==0.23.4
 gensim==3.4.0
-pytest>=3.6.0
+pytest==3.9.3
 pytest-benchmark>=3.1
 pytest-cov>=2.6.0
 coveralls>=1.5.1


TheDen:
Given `pytest` in `requirements.txt` is set to `pytest>=3.6.0`, it's now pulling version `3.10.0` (released a few days ago)—[changelog](https://docs.pytest.org/en/latest/changelog.html)

There seems to be a regression, since the `-p no:cacheprovider` now fails, with the error 
```AttributeError: 'Config' object has no attribute 'cache'```

Removing the flag, the test run again (done locally), but we also get new `DeprecationWarnings`, e.g.,

```
/usr/local/lib/python3.6/site-packages/tensorflow/python/util/tf_inspect.py:75: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
```
and
```/usr/local/lib/python3.6/site-packages/tensorflow/python/framework/ops.py:923
  /usr/local/lib/python3.6/site-packages/tensorflow/python/framework/ops.py:923: DeprecationWarning: builtin type EagerTensor has no __module__ attribute
```

We can either pin `pytest` to the previous version, or move forward with the latest release. What are your thoughts @youph @adocherty?

adocherty:
I'm happy with either way – we don't need the latest features of `pytest` currently, so pinning may be the easiest solution right now.

youph:
I guess pinning `pytest` to the previous version is better, since if we don't, the `-p no:cacheprovider` flag will fail depending on which version of `pytest` is installed, right?

TheDen:
Yep, and you'll get all the extra warnings

youph:
@TheDen I was just doing that, you beat me by 30 sec :)

PR #286, opened 2018-11-12
Added weighted biased random walker.

habiba-h:
Added weighted random walk.
Added unit tests for the weighted walks.

Input: weighted edge list <u,v,w>
It can be a simple or multi weighted graph.
If weights are negative, the method terminates with a message.
If different multi-edge weights between a pair of nodes terminates with a message.

### DIFF
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -274,7 +274,15 @@ class BiasedRandomWalk(GraphWalk):
     #           print("its a weighted graph!")
 
     def run(
-        self, nodes=None, n=None, p=1.0, q=1.0, length=None, seed=None, weighted=False
+        self,
+        nodes=None,
+        n=None,
+        p=1.0,
+        q=1.0,
+        length=None,
+        seed=None,
+        weighted=False,
+        weight="weight",
     ):
 
         """
@@ -288,14 +296,21 @@ class BiasedRandomWalk(GraphWalk):
             length: <int> Maximum length of each random walk
             seed: <int> Random number generator seed; default is None
             weighted: <False or True> Indicates whether the walk is unweighted or weighted
-
+            weight: <string> Label of the edge.
 
         Returns:
             <list> List of lists of nodes ids for each of the random walks
 
         """
         self._check_parameter_values(
-            nodes=nodes, n=n, p=p, q=q, length=length, seed=seed, weighted=weighted
+            nodes=nodes,
+            n=n,
+            p=p,
+            q=q,
+            length=length,
+            seed=seed,
+            weighted=weighted,
+            weight=weight,
         )
 
         if seed:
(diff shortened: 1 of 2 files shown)


PantelisElinas:
Hi @habiba-h 

I want to add one more item to my review.

When you check if some of the edge weights are negative or for multiple edges with different weights, I noticed that after you print a message you immediately 'return'.

I want to suggest 2 changes.

1) The message should include the name of the object that produced the same way we include the name in the WARNING messages when we check the method parameters. 

2) I think it is not a good idea to just return on error. It would be best to throw a ValueError exception with a meaningful message. The problem with just returning is that people do tend to ignore printed messages or might just not even be looking at the terminal when they are running the code and so they may never see the error and continue thinking that they have calculated a proper set of random walk data. Negative weights and multiple edges with weights are significant problems that halting execution is more appropriate than just returning with a warning message.

Regards,

P.

habiba-h:
Hi @PantelisElinas, 
1. The formatting issue has been fixed.
2. Incorporated the suggestions about checking weights. The run method returns with appropriate message if edge weights are missing, NaN, inf, negative, or multiedges with different weights.
3. Removed the default weight assignment of 1. If the weight is missing it gets caught earlier on.
4. The 'weight' label is left as it is. Its a tricky one. Networkx by default name the 'value' of an edge as its 'weight'. I could not find a way to change the name to a custom one. However, Stellargraph object as an edge label attribute which is set to label = 'default'. In my opinion this is what should assigned the proper label according to whatever the edge is about.  For example, if the edges 'represent' probabilities, then the label of the edge should be 'probability' and its value/weight is what is called the weight.

PantelisElinas:
Hi @habiba-h 

here is an example on how to create a graph with edge property that is not 'weight'. 

>> g2 = nx.Graph()
>> g2.add_edges_from([(1,2), (1,3)])
>> g2.edges(data=True)
EdgeDataView([(1, 2, {}), (1, 3, {})])
>> g2[1][2]['w'] =1
>> g2.edges(data=True)
EdgeDataView([(1, 2, {'w': 1}), (1, 3, {})])
>>from stellargraph import StellarGraph
Backend TkAgg is interactive backend. Turning interactive mode on.
>>sg2 = StellarGraph(g2)
>>sg2.edges(data=True)
MultiEdgeDataView([(1, 2, {'w': 1, 'label': 'default'}), (1, 3, {'label': 'default'})])
>> sg2[1][2][0].get('w')
1

Notice that in the above example, I created a graph with 2 edges and 3 nodes and I assigned a 'w' attribute to one of the edges. There is no edge attribute called 'weight' associated with any of the edges. If I give the random walker an instance of a graph where I stored edge weights under the label 'w', then the current implementation won't work even though the data is there. There is nothing in StellarGraph that creates a 'weight' attribute for edges by default. 

We leave it up to the user to create the networkx object and then ask her to cast to StellarGraph type but the latter operation does not relabel any of the edge attributes or even create one called 'weight' by default. The user might as well call it 'w' as I did in the example above.

Please make it a parameter for run() so that callers can specify the edge attribute name to be used; the default can be 'weight'.

@youph what do you think? 

Regards,

P.

youph:
@PantelisElinas @habiba-h indeed, one can create any numerical edge property representing edge weight, and call it whatever, so it seems like a good idea to pass the name of the edge attribute, that is to be regarded as weight in the weighted RW, to the `.run()` method of the walker. Plus, it's easy to implement.

Another question is: what if there are no edge attributes at all? No "weight", no "w", etc. Should we then assume equal weights of 1? I'd say yes to this. What do you think?

habiba-h:
@PantelisElinas and @youph  updated the run method. It can take any user specified string as the weight label. Default is set to weight.  Although I would argue that the weight(value) should be called weight or value, as the word weight is generally associate with the value of the edge. In my opinion what you want is the 'label' to reflect what kind of edge it is. For example, the kind of edge is 'rating', 'citation', 'probability' etc. The 'weight' of rating, probability etc, is the numeric value over which the weighted walk happens.    The Stellargraph graph already has a label associated with each edge which is set to 'default'. I think this is the attribute that should be given the right label and weight should be left as weight. For example, for the Cora dataset, the label of the edge should be  cite/citation and its weight is a numeric value (although in the case of cora dataset the weights of citation don't make sense but hope it is clear what I am saying). 
Another aesthetic reason is that this a 'weighted random walk' so weight is a natural argument for it. 
Anyways, the run method can be passed another string to give whatever name to the weight. 
@PantelisElinas  Thanks for pointing out how the weight label gets updated. 
@PantelisElinas Also, updated the Valueerror strings.

PantelisElinas:
@youph 

Your question relates to the last paragraph in my original review. 

I think that in the current implementation, we should not make any assumptions if weighted=True and no 'weight' attribute is given we should raise an error.  If weighted=False, then the code never tries to access any edge attributes and all is well even if the edges have weights. 

As I said above, it would simplify the code if we assumed that if the default label 'weight' is missing then we can just treat the graph as unweighted or equivalently one that has equal weights for each edge; any non-zero number will do but obviously 1.0 makes perfect sense. If we go this direction, then we don't really need the weighted flag. If a user specifies a label other than the default then if it is missing, we should raise an error because we can take that to mean that the user is trying to operate on a weighted graph but she has provided a graph with missing data.

My suggestion is to have a careful think about this and implement it later if we think it is more beneficial than the current approach.

Regards,

P.

youph:
@habiba-h I see what you mean. We can enforce usage of 'weight' edge attribute (with edges differing by type) in all cases, or we can keep the current flexible implementation where it is suggested rather than enforced - so the user can use some other name for edge weights.

habiba-h:
@PantelisElinas 
1. Changed the edge weight parameter from 'weight' to 'edge_weight_label'. Default value of it is 'weight'
2. Fixed the check for correct type of value (string) being passed through the 'edge_weight_label'.
3. Removed the tests which checked for the correct graph being given to the run method for the biased weighted random walk as they do not test the run method directly but just the correctness of the graph.
4. Incorporated the assert statement in the test for correct execution of the run method.

youph:
@habiba-h this issue has been closed, but the PR is not merged. Is this how you intended it?

habiba-h:
@youph Sorry rookie mistake. Thought I had merged the changes before closing. @PantelisElinas showed me what I was supposed to do. Its fixed and closed now.

youph:
I've replaced multiple calls to `v.get(edge_weight_label)` with just one call, to optimise the checks a little.

PR #290, opened 2018-11-14
Added GraphSAGE node embedding extraction and visualisation

youph:
Added node embedding extraction and visualisation to the node classification demo with graphsage:
demos/node-classification-graphsage/graphsage-cora-node-classification-example.ipynb


adocherty:
I've made a change to the notebook to fix saving the *graphml* file. This causes an error when using the default options due to NetworkX not expanding user paths (unlike Pandas which does):

The `data_dir` is now specified as:
```
data_dir = os.path.expanduser("~/Data/Cora/cora")
```

The embeddings are extracted by searching for the required output layer using it's name `reshape_4`. This only works the first time the model is run, when the model is recreated the names will be different. We already have the correct embedding outputs in the `x_out` variable, why not use these directly?

e.g.
```
embedding_model = Model(inputs=x_inp, outputs=x_out)
```

I made these changes and committed them – hope that's OK!

youph:
@annitrolla I've implemented your suggestions, please have a look

Issue #285, opened 2018-11-09
EdgeSplitter: introduce a switch for keeping the reduced graph connected


youph:
### Description
Currently `EdgeSplitter` ensures that sampled positive edges are not from the graph's spanning tree, which is done to ensure that removing these edges does not increase the number of connected components in the reduced graph. 

We should allow users to switch this behaviour on or off with a switch (e.g., `keep_connected`) passed to the `train_test_split` method.

### Done Checklist (Development)
- [x] Produced code for required functionality
- [x] Branch and Pull Request build on CI
- [x] Branch and Pull Request pass unit tests on CI
- [x] Peer Code Review Performed
- [x] Code well commented
- [x] Documentation in repo

youph:
This is also needed to simplify @lichaoir 's comparison with platform.

PR #292, opened 2018-11-20
Added keep_connected flag for EdgeSplitter train_test_split method

PantelisElinas:
Hi @youph 

This relates to issue #285.

Regards,

P.


PantelisElinas:
Hi @youph 

I have updated the EdgeSplitter so that the number of edges to sample is a function of the total number of edges for both homogeneous and heterogeneous graphs.

Also, we now raise ValueError if it is not possible to sample the required number of positive or negative edges.

Regards,

P.

PantelisElinas:
Hi @youph 

with regards to the docstring for __init__ you might notice that we don't provide docstring for the __init__ functions anywhere in the code; we provide the necessary information in the corresponding class' docstring. I have update the description of EdgeSplitter class to include descriptions of the parameters for its __init__. We should discuss with the team if we want to change how we do the documentation and update across the library.

Although I have updated the expression on lines 516 and 571, the same result is achieved whether we check for equality or less than. I've updated to your suggestion but makes little difference in terms of functionality.

Regards,

P.

Issue #62, opened 2018-06-04
PoC for Unsupervised GraphSAGE


arc0:
### Description
Currently the GraphSage unsupervised method is not in the library. This task is to add a simple unsupervised GraphSage module in the stellar-ml library.

### User Story
As a: Data Scientist
I want: everything that GraphSAGE offers
so that: I have freedom for my unsupervised method experiments

### Done Checklist (Development)
- [ ] Branch and Pull Request build on CI
- [ ] Code well commented

youph:
Please update the Done Checklist

PR #295, opened 2018-11-28
Feature/weightedwalks

habiba-h:
Hi Yuriy,

will you be able to review #286 which is connected to this issue. This is only a simple demo for the weighted random walks.  I replicated the simple random walks for link prediction in node2vec.  Thanks!
Habiba


habiba-h:
@adocherty its the issue #62 that is meant to be looked at by you. Thanks!

adocherty:
I've had a look at the notebook `unsupervisedGraphSAGE`.

As we discussed, we need to shuffle the positive and negative edges together.  Perhaps we should have an option for the `GraphSAGELinkGenerator`  and `GraphSAGENodeGenerator` to shuffle the data provided to it, perhaps at each epoch? Keras has this option but I don't think it can do global data shuffling when using a generator.

Also, I think we should remove the `G_train` and `G_test` if they are the same.

Everything else looks good.

habiba-h:
Thanks @youph for the helpful comments.  I have fixed the notebook. It should run now. Tried to give a better explanation of each step. Compared the weighted and unweighted random walks on the Cora dataset in the form of visualising the embeddings and comparing them in a downstream classification task.  The weighted embeddings don't look as compact as the unweighted ones. Moreover, the accuracy on the classification task is very poor.  One possible explanation is that the weights based on Jaccard similarity are not necessarily the right way to weigh the edges between nodes.

youph:
New comments are in commit f1e371b

youph:
@habiba-h I've fixed the notebook to give identical results (including embeddings, plots, and accuracies) for the last 2 cases. The conclusion needs re-writing (see my comments in the notebook, marked with YT).

I have a question that's beyond the scope of this demo: when I run the same notebook twice, I get different results, despite the fact that we set all seeds and random states everywhere, and the graph is the same. Is there another source of randomness that causes results to change between runs?

habiba-h:
@youph Updated the notebook. Modified the start and conclusion of the notebook to (hopefully) better reflect the purpose of the notebook in terms of demonstrating how Node2Vec can be applied to weighted biased random walks. As well as,  comparing the weighted walks to unweighted walks.

habiba-h:
@youph Fixed the black formatting issue.  All checks pass. Can you please approve the review? Thanks.

PR #299, opened 2018-12-19
Issue/isolated nodes

adocherty:
Fixed the handling of isolated nodes in GraphSAGE node and link generators.

Isolated nodes were already handled by the HinSAGE generators, so this PR gives GraphSAGE generators the same behavior. Now, when an isolated node is a head node for a generator the random samples of neighbors will be filled with the null-node ID. This is currently the Python `None`. Next the `StellarGraph.get_features_for_nodes` method will get the null-node features (currently all zeros) and put these in the batch features.  In future, the null-node features could be configurable, and even trainable.

Also introduced are unit tests to test for isolated head nodes in node and link generators for both GraphSAGE and HinSAGE generators.

Now, running the graphsage link prediction demo using the `EdgeSplitter` with `keep_connected=False` works and gives similar error to the `keep_connected=True` setting.

### DIFF
--- a/demos/node-classification-graphsage/graphsage-cora-example.py
+++ b/demos/node-classification-graphsage/graphsage-cora-example.py
@@ -101,7 +101,7 @@ def train(
 
     # Create graph schema
     sample_edges = [e for ii,e in enumerate(G.edges(keys=True)) if ii < 5]
-    schema = G.create_graph_schema(nodes=node_ids[:1], edges=sample_edges)
+    schema = G.create_graph_schema(create_type_maps=False, nodes=node_ids[:1], edges=sample_edges)
 
     # Create mappers for GraphSAGE that input data from the graph to the model
     generator = GraphSAGENodeGenerator(G, batch_size, num_samples, schema=schema, seed=42)
--- a/demos/node-classification-graphsage/graphsage-cora-example.py
+++ b/demos/node-classification-graphsage/graphsage-cora-example.py
@@ -99,8 +99,12 @@ def train(
         test_nodes, test_targets, train_size=500, test_size=None, random_state=523214
     )
 
+    # Create graph schema
+    sample_edges = [e for ii,e in enumerate(G.edges(keys=True)) if ii < 5]
+    schema = G.create_graph_schema(nodes=node_ids[:1], edges=sample_edges)
+
     # Create mappers for GraphSAGE that input data from the graph to the model
-    generator = GraphSAGENodeGenerator(G, batch_size, num_samples, seed=42)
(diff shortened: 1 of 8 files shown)


demos/node-classification-graphsage/graphsage-cora-example.py
```diff
@@ -99,8 +99,12 @@ def train(
         test_nodes, test_targets, train_size=500, test_size=None, random_state=523214
     )
 
+    # Create graph schema
```
youph: why do we need to create the graph schema and pass it to the node generator? This extra step was not there before, what is the justification for putting it in the demo, making the workflow more complicated?

stellargraph/mapper/link_mappers.py
```diff
@@ -215,21 +215,31 @@ def sample_features(self, head_links, sampling_schema):
...
-                walks_at_level = list(it.chain(*[w[loc:end_loc] for w in walks]))
-                if len(samples_per_hop) < 1:
-                    return [walks_at_level]
-                return [walks_at_level] + get_levels(
-                    end_loc, lsize * samples_per_hop[0], samples_per_hop[1:], walks
-                )
+            # Isolated nodes will return only themselves in the sample list
+            # let's correct for this by padding with None (the dummy node ID)
+            node_samples = [
+                ns + [None] * num_full_samples if len(ns) == 1 else ns
```
youph: nice!

demos/node-classification-graphsage/graphsage-cora-example.py
```diff
@@ -99,8 +99,12 @@ def train(
...
+    sample_edges = [e for ii,e in enumerate(G.edges(keys=True)) if ii < 5]
+    schema = G.create_graph_schema(create_type_maps=False, nodes=node_ids[:1], edges=sample_edges)
```
youph: why do we need to create the graph schema and pass it to the node generator? This extra step was not there before, what is the justification for putting it in the demo, making the workflow more complicated?

adocherty: You're right .. this shouldn't be there. It was test code that somehow got checked in. I'll remove it.

stellargraph/mapper/link_mappers.py
youph: I wonder whether this can be modified to something like
`node_samples = [[None] if len(ns) == 0 else ns + [None] * num_full_samples if len(ns)==1 else ns]`
to allow for the case of num_samples = [0, 0, ..., 0] where no neighbours should be sampled?

youph:
I've added one more comment

PR #308, opened 2019-01-09
Build MLP when zero neighbours are provided

adocherty:
The GraphSAGE and HinSAGE classes when given 0 for the number of samples in all layers build a MLP on the self features of the nodes.  This currently means that the neighbour weights are not included in the model and the neighbour features are replaced with a vector of all zeros at each layer.

This doesn't currently build the same MLP as would be done manually as the neighbour features (all zeros) are still concatenated with the self features. However, the accuracy on CORA is around 55% the same as a manual MLP, and we can build an identical MLP if we introduce a summation option in the aggregator layers that adds the self and neighbour features.

Unit tests have been added for the aggregators and the models for both GraphSAGE and HinSAGE.


stellargraph/layer/graphsage.py
```diff
@@ -541,54 +482,52 @@ def __call__(self, x: List):
             Output tensor
         """
 
-        def compose_layers(_x: List, layer: int):
+        def apply_layer(x: List, layer: int):
             """
-            Function to recursively compose aggregation layers. When current layer is at final layer, then length of _x
-            should be 1, and compose_layers(_x, layer) returns _x[0].
+            Compute the list of output tensors for a single HinSAGE layer
```
PantelisElinas: Shouldn't this say 'GraphSAGE layer'?

stellargraph/layer/graphsage.py
```diff
@@ -541,54 +482,52 @@ def __call__(self, x: List):
...
 
-        assert isinstance(x, list), "Input features must be a list"
-        assert (
-            len(x) == self.n_layers + 1 > 1
-        ), "Length of input features should match the number of GraphSAGE layers"
+        if not isinstance(xin, list):
+            raise TypeError("Input features to GraphSAGE must be a list")
 
-        return self._normalization(compose_layers(x, 0))
+        if len(xin) != self.n_layers + 1:
```
PantelisElinas: Should the message for the ValueError exception say that '...the number of GraphSAGE layers + 1'?

tests/mapper/test_node_mappers.py
```diff
@@ -529,6 +545,29 @@ def test_hinnodemapper_manual_schema():
...
+    schema = G.create_graph_schema()
+    sampling_adj = schema.type_adjacency_list(["t2"], 2)
+
+    assert len(mapper) == 1
+
+    # Get a batch!
+    batch_feats, batch_targets = mapper[0]
+    assert len(batch_feats) == len(sampling_adj)
+
+    print(sampling_adj)
```
PantelisElinas: Do we need this print statement here?

tests/mapper/test_node_mappers.py
```diff
@@ -529,6 +545,29 @@ def test_hinnodemapper_manual_schema():
...
+
+    print(batch_feats)
```
PantelisElinas: Do we need this print statement here?

stellargraph/layer/graphsage.py
adocherty: Yes, I've fixed this.

stellargraph/layer/graphsage.py
adocherty: True ... I've changed the error message to reflect this.

tests/mapper/test_node_mappers.py
adocherty: Not really, although there are quite a few print statements we have left in the tests. Pytest suppresses these by default.

tests/mapper/test_node_mappers.py
adocherty: I have commented them out.

adocherty:
Also, I just caught & fixed a bug that made the GraphSAGE & HinSAGE models not able to be saved. I've added a couple of unit tests to check that Keras serialization works with our GraphSAGE & HinSAGE models.

adocherty:
The ordering of the Dropout and Reshape layers have been changed to match the previous implementation. As we suspected, there is no significant change to the performance of the model when run with fixed train/test split and fixed generator seeds.

Issue #216, opened 2018-09-10
Add GAT to the stellargraph library


youph:
### Description
We need to populate the library with algorithms. Graph Attention networks (GAT) is a good candidate. 
Reference Keras implementation of GAT: https://github.com/danielegrattarola/keras-gat 

### Done Checklist (Development)
- [x] GAT "layer" added to the library
- [x] Demo of using GAT added in `/demos`
- [x] Branch and Pull Request build on CI
- [x] Branch and Pull Request pass unit tests on CI
- [ ] Branch and Pull Request pass integration tests on CI
- [ ] Version number reflects new status
- [ ] Peer Code Review Performed
- [x] Code well commented
- [x] Documentation/reference in repo

PR #315, opened 2019-01-22
Feature/gat

youph:
This PR addresses #216 and includes the following:
- [x] Graph Attention layer implementation
- [x] GAT class implementation (a stack of Graph Attention layers)
- [x] FullBatchNodeGenerator class
- [x] Demos:
  - [x] python script
  - [x] notebook
- [x] Unit tests
- [x] README updated
  - [x] Reference to GAT paper added
  - [x] GAT algorithm added
- [x] AUTHORS updated with GAT authors


stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+        Args:
+            x_inp (Tensor): input of the 1st GAT layer in the stack
+
+        Returns: Output tensor of the GAT layers stack
+
+        """
+
+        assert isinstance(x_inp, list), "input must be a list, got {} instead".format(
+            type(x_inp)
+        )
```
PantelisElinas: See my comment above about assert vs raising ValueError.

stellargraph/mapper/node_mappers.py
```diff
@@ -448,3 +454,132 @@ def flow_from_dataframe(self, node_targets):
...
+        self.name = name
+
+        # Check if the graph has features
+        G.check_graph_for_ml()
+
+        # Create sparse adjacency matrix
+        self.node_list = list(G.nodes())
+        self.Aadj = nx.adjacency_matrix(G, nodelist=self.node_list)
+
+        # We need a schema to check compatibility with GraphSAGE, GAT, GCN
```
PantelisElinas: Maybe remove reference to GraphSAGE? Can this generator be used with GraphSAGE?

stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+        in_dropout=0.0,
+        attn_dropout=0.0,
+        normalize="l2",
+        generator=None,
+    ):
+        """
+
+        Args:
+            layer_sizes: list of output sizes of GAT layers in the stack
+            attn_heads: number of attention heads
```
PantelisElinas: Maybe explain that this specifies the number of attention heads for all layers expect the last one that always has 1 head.

stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+            attn_heads_reduction:
```
PantelisElinas: Missing description.

stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+        bias=True,
+        in_dropout=0.0,
+        attn_dropout=0.0,
+        normalize="l2",
+        generator=None,
+    ):
+        """
+
+        Args:
+            layer_sizes: list of output sizes of GAT layers in the stack
```
adocherty: It's a good idea to specify the accepted types of the arguments. For example (list of int) or (List[int]).

stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+            ), "{}: generator must be of type FullBatchNodeGenerator or None; received object of type {} instead".format(
+                type(self).__name__, type(generator).__name__
+            )
+
+        if attn_heads_reduction is None:
+            # default head reductions, see eqs 5-6 of the GAT paper
+            self.attn_heads_reduction = ["concat"] * (len(layer_sizes) - 1) + [
+                "average"
+            ]
+        else:
```
adocherty: Perhaps a check that the length of `attn_heads_reduction` is correct would be a good idea.

stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+        self.attn_dropout = attn_dropout
+        self.generator = generator
+
+        if generator is not None:
+            assert isinstance(
+                generator, FullBatchNodeGenerator
+            ), "{}: generator must be of type FullBatchNodeGenerator or None; received object of type {} instead".format(
+                type(self).__name__, type(generator).__name__
+            )
+
```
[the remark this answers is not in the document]
adocherty: I agree!

tests/layer/test_graph_attention.py
```diff
@@ -0,0 +1,375 @@
...
+            kernel_initializer="ones",
+        )
+        x_inp = [Input(shape=(self.F_in,)), Input(shape=(self.N,))]
+        x_out = gat(x_inp)
+
+        model = keras.Model(inputs=x_inp, outputs=x_out)
+        assert model.output_shape[-1] == self.F_out
+
+        X = np.ones((self.N, self.F_in))  # features
+        A = np.eye(self.N)  # adjacency matrix with self-loops only
```
adocherty: Perhaps a test with a non-identity adjacency matrix would be useful?

tests/mapper/test_node_mappers.py
```diff
@@ -559,3 +559,88 @@ def test_hinnodemapper_no_neighbors():
...
+
+        assert len(gen) == 1
+        assert gen.A.shape == (self.N, self.N)
+        assert gen.features.shape == (self.N, self.n_feat)
+        assert sum(gen.sample_weight) == len(node_ids)
+        assert gen.targets is None
+
+        [X, A], y, sample_weights = gen.__getitem__(0)
+        assert np.array_equal(X, gen.features)  # X should be equal to gen.features
+        assert (A != gen.A).nnz == 0  # A should be equal to gen.A
```
adocherty: `len(np.nonzero(A != gen.A)[0])` would work for both dense and sparse matrices :-)

demos/node-classification-gat/gat-cora-example.py
```diff
@@ -0,0 +1,360 @@
...
+
+    print("\nBest model's Test Set Metrics:")
+    for name, val in zip(model.metrics_names, test_metrics):
+        print("\t{}: {:0.4f}".format(name, val))
+
+    # Get predictions for all nodes
+    if args.interface == "fit":
+        all_predictions = model.predict(x=[X, A], batch_size=N)
+    else:
+        all_predictions = model.predict_generator(generator.flow(node_ids))
```
adocherty: The `predict` or `predict_generator` function now operates differently to the `GraphSAGE` model in that if you give it less than the complete set of nodes, it will still return all predictions and in a fixed order.

This is confusing for users – whatever is given inthe `flow` argument will be ignored by the `predict_generator` as it's just used for masking.

stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Definition of Graph Attention Network (GAT) layer and GAT class that is a stack of GAT layers
+"""
```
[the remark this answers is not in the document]
youph: Reference added

stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+        bias_initializer="zeros",
+        attn_kernel_initializer="glorot_uniform",
+        kernel_regularizer=None,
+        bias_regularizer=None,
+        attn_kernel_regularizer=None,
+        activity_regularizer=None,
+        kernel_constraint=None,
+        bias_constraint=None,
+        attn_kernel_constraint=None,
+        **kwargs
```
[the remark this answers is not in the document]
youph: Added "Must be a Keras constraint https://keras.io/constraints/" to the `Args` docstring

stellargraph/layer/graph_attention.py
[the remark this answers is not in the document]
youph: Good point! Thanks for the link to the discussion.

stellargraph/mapper/node_mappers.py
youph: Fixed.

stellargraph/layer/graph_attention.py
youph: Fixed

stellargraph/mapper/node_mappers.py
youph: fixed.

stellargraph/layer/graph_attention.py
youph: description added

stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+        # Convert A to dense tensor - needed for the mask to work
+        if K.is_sparse(A):
+            A = tf.sparse_tensor_to_dense(A, validate_indices=False)
+
+        # For the GAT model to match that in the paper, we need to ensure that the graph has self-loops,
+        # since the neighbourhood of node i in eq. (4) includes node i itself.
+        # Adding self-loops to A via setting the diagonal elements of A to 1.0:
+        # N = kwargs.get("num_nodes")
+        # get the number of nodes from inputs[1] directly, rather than passing it via kwargs
+        N = inputs[1]._keras_shape[-1]
```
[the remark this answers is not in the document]
youph: No, this doesn't work, as this gives a tensor `Tensor("graph_attention_1/strided_slice:0", shape=(), dtype=int32)` instead of an integer.

stellargraph/layer/graph_attention.py
youph: fixed

stellargraph/layer/graph_attention.py
youph: done, done.

stellargraph/layer/graph_attention.py
youph: check is added

tests/mapper/test_node_mappers.py
youph: fixed

demos/node-classification-gat/gat-cora-example.py
youph: I've added a comment about this in both the demos

tests/layer/test_graph_attention.py
youph: what would that test check?

stellargraph/layer/graph_attention.py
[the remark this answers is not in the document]
adocherty: Ah, if you need the int shape you can use `K.int_shape(...)`

tests/layer/test_graph_attention.py
adocherty: Check the application of the attention layer to a graph with some neighbours.

stellargraph/layer/graph_attention.py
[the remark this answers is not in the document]
youph: thanks!

stellargraph/layer/graph_attention.py
```diff
@@ -0,0 +1,481 @@
...
+        if N is not None:
+            # create self-loops
+            A = tf.linalg.set_diag(A, K.cast(np.ones((N,)), dtype="float"))
```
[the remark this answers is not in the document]
youph: I've added an argument "add_self_loops" to `GAT.node_model` method, which toggles adding self-loops in the `GraphAttention` layers of the GAT model. The default value is set to `True` so that the user does not have to worry about adding self-loops to make sure the resulting GAT layers match those in the paper, yet still allowing advanced users to switch self-loops off if needed.

tests/layer/test_graph_attention.py
youph: added one such (simple) test with nodes 0-1 connected

youph:
> * The predictions obtained using `predict_generator` ignore the arguments to the `.flow` method of the generator. This could be confusing for users, and is different from the behaviour of the GraphSAGE model.
I've added a comment to both demos, explaining this difference.

> * The number of attention heads at the final layer is set to 1 always.
Fixed: now it can be either a single int (in which case it's that number for all but the last layer, and one for the last layer), or a list of ints.

> * I like the model checkpointing and early stopping in the python script, but I think it's overkill for the notebook – it slows down the execution of the notebook and also prints out a lot of epochs when running it.
True, although I wanted to have few more new things in the notebook demo, compared to those using GraphSAGE for the same task. Otherwise the difference boils down to 2 lines of code: different generator and different model. Kind of boring. This way people who only look at notebooks (and not the scripts) would see how to organise early stopping and checkpointing :)

> * Also note that the model is different from the published GAT model: they have the final layer output the number of classes and apply a softmax over that. We add a dense layer. The accuracy isn't much different between either option, but maybe we can point out that we are doing it differently.
I've changed the model architecture in the notebook, to match that of the GAT paper.

youph:
> Hi @youph
> 
> it looks good. Have a look at my comments.
> 
> I noticed that running the notebook does not produce the same result as the reference paper. I get about 80% accuracy and the paper states 83%. It looks like the parameter setting are the same. Maybe later we should have a look at why there is a difference.
> 
> Regards,
> 
> P.

@PantelisElinas that's true; one possible reason is a somewhat different model architecture. I've updated the model architecture in the notebook demo, to match that of the paper more closely.

adocherty:
Hi @youph .. The changes look good to me. 

In terms of documentation, we'll need to add the GAT class to the `docs/api.txt` file in a similar way to GraphSAGE or GCN.  To make the automatic API documentation work, all documentation for a class needs to be under the main class definition, not in the `__init__` method.

You can check the docs locally by doing the following:
```
pip install sphinx_rtd_theme sphinx
cd docs
make html
open _build/html/index.html
```

youph:
@adocherty thanks for the instructions about the docs; I've updated the docs api, and brushed the doc strings.

youph:
@wangzhen263 Please modify the doc strings in GCN layer and class as well, to make sure they are displayed nicely in the docs, as per @adocherty's instructions above. In particular, to make the automatic API documentation work, all documentation for a class needs to be under the main class definition, not in the `__init__` method.

PR #316, opened 2019-01-23
Feature/graphsage attention agg

adocherty:
Added an GAT-style attention aggregator for GraphSAGE.  This performs the same attention computation as GAT on a batch of input nodes. Dropout is handled externally in the same way as the default GraphSAGE code (different to GAT which would require dropout within the aggregator).

Currently the aggregator is single-headed only. We can extend this to multi-headed in a more general way by adding the multi-head support in the GraphSAGE model class, enabling multiple-head models with different aggregators than just the attention aggregator. 

Additional changes are to build a more flexible MLP model when there are zero samples in all aggregators (required as the attention aggregator wasn't compatible with how I did this previously).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,9 +1,6 @@
 steps:
   - label: ":python:"
-    command:
-      - pip install -q --no-cache-dir -r requirements.txt -e .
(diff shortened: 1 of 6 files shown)


adocherty:
Don't use this pull request – I've rebased the branch on develop and created a new pull request #317

PR #331, opened 2019-02-18
Unsupervised GraphSAGE 

habiba-h:


stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+import numpy as np
+
+from stellargraph.core.utils import is_real_iterable
+from stellargraph.core.graph import StellarGraphBase
+from stellargraph.data.explorer import UniformRandomWalk
+
+
+class UnsupervisedSampler:
+    """
+        This class does unsupervised samlping in a graph.
```
adocherty: Keep comments to around 100 characters max, so they are more easily visible in editors.

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+        The positive samples are all the (target, context) pairs from the walks generated by walker. Negative samples are contexts generated for each target based on a sampling strategy.
+    """
+
+    def __init__(
+        self, G, walker=None, nodes=None, length=2, number_of_walks=1, seed=None
+    ):
+
+        """
+           Initializing all the paramters for the class instance. A Stellargraph is a must. 
+           The root nodes and walker are required. However, if not provided they default to all nodes in the graph and uniform random walker respectively.
```
adocherty: The documentation for a class should be under the class definition, not in `__init__` (this is so the API docs are generated properly)

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+               number_of_walks <int>: Number of walks from each root node. Note, at the moment this parameter is only used to calculate the number of batches in each epoch,
+                                      as we do not have a pregenerated sample list, we need to estimate given a batch_size, how many batches to generate in each epoch.  
+               
+        
+        """
+        # Initialize the random state
+        self._random_state = random.Random(seed)
+
+        # Initialize a numpy random state (for numpy random methods)
+        self._np_random_state = np.random.RandomState(seed=seed)
```
adocherty: You never use these random seeds. Remove them if unneeded.

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+                )
+            else:
+                self.walker = UniformRandomWalk(G)
+        else:
+            self.walker = UniformRandomWalk(G)
+
+        if nodes is None:
+            self.nodes = (
+                G.nodes()
+            )  # if no root nodes are provided for sampling defaulting to using all nodes as root nodes.
```
adocherty: It might be good to make `self.nodes` a list

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+
+        self.seed = seed
+
+    def generator(self, batch_size):
+
+        """
+        This method yields a batch_size of +/- samples when it is called. The batch_size must be provided. It should be a positive integer. Moreover, for now we make the batch_size even by
+        unit incrementing it, incase it is odd. This is because we generate one negative sample for each positive sample.
+        This method:
+            1. Iterates over the root nodes.
```
adocherty: Don't use numbered lists in API docstrings as Sphinx doesn't work well with them.

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+
+        Returns:
+            A tuple of lists: (target, context) pairs and their labels. 
+        """
+
+        self._check_parameter_values(batch_size=batch_size)
+
+        # making the batch_size even since we generate 1 negative sample for each positive one.
+        if batch_size % 2 != 0:
+            batch_size += 1
```
adocherty: I think I'd prefer raising an error here, so that people aren't confused by their batch sizes being different than expected.

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+                walk = self.walker.run(
+                    nodes=[node],  # root nodes
+                    length=self.length,  # maximum length of a random walk
+                    n=1,  # number of random walks per root node
+                    seed=None,
+                )
+                # (target,contect) pair sampling - GraphSAGE way
+                target = walk[0][0]
+                context_window = walk[0][1:]
+                if len(context_window) >= 1:
```
adocherty: We don’t need this check, the for loop will not run if `len(context_window)<1` so I'd remove this if statement.

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+                    for context in context_window:
+                        # Don't add self pairs
+                        if context != target:
+                            positive_pairs.append((target, context))
```
adocherty: To avoid deep indentation, you can also use
```
if context == target:
    continue

postive_pairs.append(...)
...
```

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+
+        # making the batch_size even since we generate 1 negative sample for each positive one.
+        if batch_size % 2 != 0:
+            batch_size += 1
+
+        positive_pairs = list()
+        negative_pairs = list()
+
+        positive_samples_counter = 0
+        negative_samples_counter = 0
```
adocherty: Do we need both counters?

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+                            ) >= batch_size:
+                                all_pairs = positive_pairs + negative_pairs
+                                all_targets = [1] * len(positive_pairs) + [0] * len(
+                                    negative_pairs
+                                )
+                                edge_ids_labels = list(zip(all_pairs, all_targets))
+                                random.shuffle(edge_ids_labels)
+                                edge_ids, edge_labels = [
+                                    [z[i] for z in edge_ids_labels] for i in (0, 1)
+                                ]
```
adocherty: This can be done with another `zip`:
```
edge_ids, edge_labels = zip(*edge_ids_labels)
```

stellargraph/mapper/link_mappers.py
```diff
@@ -19,46 +19,48 @@
 for link prediction/link attribute inference problems using GraphSAGE and HinSAGE.
 
 """
-__all__ = ["LinkSequence", "GraphSAGELinkGenerator", "HinSAGELinkGenerator"]
+__all__ = ["OnDemandLinkSequence", "GraphSAGELinkGenerator", "HinSAGELinkGenerator"]
```
adocherty: Add "LinkSequence" here as well.

stellargraph/mapper/link_mappers.py
```diff
@@ -163,6 +162,120 @@ def on_epoch_end(self):
...
+
+        if batch_num >= self.__len__():
+            raise IndexError(
+                "Mapper: batch_num larger than number of esstaimted  batches for this epoch."
+            )
+        # print("Fetching {} batch {} [{}]".format(self.name, batch_num, start_idx))
+
+        # Get head nodes and labels
+        head_ids, batch_targets = next(self._gen)
+        self.ids = list(head_ids)
```
adocherty: Do we need to set `self.ids` for each batch?

stellargraph/mapper/link_mappers.py
```diff
@@ -163,6 +162,120 @@ def on_epoch_end(self):
...
+
+        # Get head node types from all src, dst nodes extracted from all links,
+        # and make sure there's only one pair of node types:
+        self.head_node_types = self._infer_head_node_types(self.generator.schema)
```
adocherty: Inferring the head node types & sampling schema every time a batch is generated is not the best. Is there a way we can do this just once?

adocherty:
1. Use lower lower caps with underscores for all file names, e.g. `unsupervised_sampler.py`. More information here: https://www.python.org/dev/peps/pep-0008/#naming-conventions

2. Keep comments to around 100 characters max, so they are more easily visible in editors. All documentation for a class should be under the class definition, not in `__init__` (this is so the API docs are generated properly, see below)

3. Add API documentation by adding the classes you want exposed to the `docs/api.txt` file. For example,
```
Unsupervised Sampling
------------------------

.. automodule:: stellargraph.data.unsupervisedSampler
  :members: UnsupervisedSampler
```

Install `sphinx` and check if the documentation renders correctly, for example:
```
pip install sphinx_rtd_theme sphinx
cd docs
make html
open _build/html/index.html
```

4. Why does the `UnsupervisedGenerator` only work with a `UniformRandomWalk` class

5. 
I think we should check if the `UnsupervisedSampler` class is passed to `HinSAGE` and raise a warning if it is.

stellargraph/data/unsupervisedSampler.py
youph: actually a NodeView might be better in terms of memory footprint. Turning it into a list "unwraps" the NodeView into the actual list, while NodeView is just an iterator, right?

stellargraph/mapper/link_mappers.py
youph: agreed

stellargraph/mapper/link_mappers.py
youph: I agree, this should be done at construction time (in __init__)

stellargraph/data/unsupervisedSampler.py
```diff
@@ -0,0 +1,229 @@
...
+
+        done = False
+        while not done:
+            for node in self.nodes:  # iterate over root nodes
+                # Get 1 walk at a time. For now its assumed that its a uniform random walker
+                walk = self.walker.run(
+                    nodes=[node],  # root nodes
+                    length=self.length,  # maximum length of a random walk
+                    n=1,  # number of random walks per root node
+                    seed=None,
```
youph: why don't you pass the seed from the constructor to the `walker.run` method?

stellargraph/data/unsupervisedSampler.py
habiba-h: Removed them.

stellargraph/data/unsupervisedSampler.py
habiba-h: Fixed the comments.

stellargraph/data/unsupervisedSampler.py
habiba-h: Fixed this. Now, the generator works with only even sized batches. Added unit test as well.

stellargraph/data/unsupervisedSampler.py
habiba-h: Removed this check.

stellargraph/data/unsupervisedSampler.py
habiba-h: Now using only one sample counter.

stellargraph/data/unsupervisedSampler.py
habiba-h: Fixed.

stellargraph/mapper/link_mappers.py
habiba-h: Added LinkSequence to the string.

stellargraph/data/unsupervisedSampler.py
habiba-h: Now the self.seed value is being passed to the run method.

stellargraph/data/unsupervisedSampler.py
habiba-h: I have changed this to a list so that we can shuffle the nodes before every set of walks.

stellargraph/mapper/link_mappers.py
habiba-h: Now, the head_node_types and sampling schema are inferred only once at the time when the first batch is sampled.

stellargraph/mapper/link_mappers.py
habiba-h: Removed the self.ids variable. 
This was the class variable that was being passed to the method that infer the head node types. Now, the head node types are inferred only when the first batch that is generated. I pass the head_ids of the first batch for head node type inference.

habiba-h:
@youph and @adocherty, just wanted to let you know that I have made the changes suggested and committed them. Thanks for the constructive feedback and endless discussions :-)!

adocherty:
I've updated a few things in the `UnsupervisedSampler` class. In particular I use a random state object for the randomness which allows it to be reproducible but not use the same seed every time.

Also, I've updated the docstrings for the class and methods.

PR #335, opened 2019-02-22
Unsupervised graph sage demo

habiba-h:
This is a demo of learning unsupervised GraphSAGE based embeddings of the CORA network. This notebook is a simple step by step demonstration of how Unsupervised GraphSAGE is implemented.


adocherty:
A few comments:

1. As I mentioned, I think the demo notebook might be better placed under the embeddings directory – but this can be done later when you reorganize the `demos` directory.
2. Related to @youph 's comment about link prediction you use the terms 'true links' and 'false links' as well as positive and negative links. I don't think we should use the terms true and false.

For example, instead of saying:
> by feeding it a training sample of true and false links from the underlying graph that are extracted without any true class label information. 

I think we should say: "by feeding it a training sample of node pairs from random walks on the underlying graph and negative node pairs from a global node distribution"

3. I think we should note that this method is a reimplementation of the algorithm in the GraphSAGE paper adding a citation (in the top section).

4. Under "Unsupervised GraphSAGE with on demand sampling" I don't think we need to specify the parameters of the class – this is given in the API documentation.

5. Currently we don't specify a `walker` object so references to this in the text should be removed.

6 . I don't think we need to get to this level of detail – most users won't care:
> regardless of whether the negative sample is the target itself or one of the contexts.

7. This is no longer needed:
> **2. The random walker used in sampling.**

8. Use standard Python code style (https://www.python.org/dev/peps/pep-0008/). For example, this:
```
unsupervisedSamples = UnsupervisedSampler(G, nodes = G.nodes(),length = length, number_of_walks = number_of_walks )
```
is better written:
```
unsupervised_samples = UnsupervisedSampler(G, nodes=G.nodes(), length=length, number_of_walks=number_of_walks )
```
Note the spaces, the use of underscores not camelCase.

9. Under "Create a link mapper" can we simplify this explanation and use the term "node pairs" instead of "links".

10. Under "Data Splitting", this looks like debugging code:
```
print("Array shapes:\n X_train = {}\n y_train = {}\n X_test = {}\n y_test = {}".format(X_train.shape, y_train.shape, X_test.shape, y_test.shape))
```

11. "### Thoughts about why we should care?" might be better phrased as "Uses for unsupervised embeddings"?

habiba-h:
@adocherty and @youph, 
i have updated the unsupervised graphsage notebook. 
Following are the list of changes. I tried to keep the numbering consistent with @adocherty recommendations for changes.
1.  The notebook is now under the embedding folder. The name of the notebook has been updated to better reflect its purpose. The readme in the embedding folder has been updated with the relevant details. 
2.  True and false in connection to positive and negative links has been changed. Now, it should be consistently true and false. 
About the text in the introduction, I gave it another shot to get rid of the ambiguities regarding links and pair terminology. I will have another go on it with a fresher mind. @youph I will check the language in the blog you sent me the link of and see if I can refine the text. 
3.  Added reference to the original paper.
4.  Removed the list of parameters under Unsupervised GraphSAGE with on demand sampling. Described what goes into the class as free text.
5.  Removed reference to walker object.
6. Removed "regardless of whether the negative sample is the target itself or one of the contexts."
7. Removed the walker object creation part.
8. Fixed the python coding style and naming of the variable.
9. Under "Create a link mapper"  changed links to pairs.
10. Removed the train test data sizes.  
11. Changed the heading of the conclusion.

Also, updated the readme in the demo folder. Added unsupervised graphsage reference under embeddings.

Also, added a couple of words in the readme in the main stellargraph directory, mentioning unsupervised graphsage in the graphsage reference.

youph:
@habiba-h I made further clarifications of the workflow, please have a look.

habiba-h:
@youph I looked over the notebook and all looks good. So will merge it now.

PR #337, opened 2019-02-22
fixed a bug where seed was not passed to sampler in GraphSAGELinkGenerator.__init__()

youph:
This PR should fix the bug

### DIFF
--- a/stellargraph/mapper/link_mappers.py
+++ b/stellargraph/mapper/link_mappers.py
@@ -201,7 +201,7 @@ class GraphSAGELinkGenerator:
         self.name = name
 
         # The sampler used to generate random samples of neighbours
-        self.sampler = SampledBreadthFirstWalk(G)
+        self.sampler = SampledBreadthFirstWalk(G, seed=seed)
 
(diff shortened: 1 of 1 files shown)


youph:
@adocherty while fixing this, I noticed that we call the sampler before we get the schema, and we don't pass the schema to the sampler. Should it be the other way around, i.e., instead of 
```
# The sampler used to generate random samples of neighbours
self.sampler = SampledBreadthFirstWalk(G, seed=seed)

# We need a schema for compatibility with HinSAGE
self.schema = G.create_graph_schema(create_type_maps=True)
```
do 
```
# We need a schema for compatibility with HinSAGE
self.schema = G.create_graph_schema(create_type_maps=True)

# The sampler used to generate random samples of neighbours
self.sampler = SampledBreadthFirstWalk(G, graph_schema=self.schema, seed=seed)
```
?

adocherty:
Good catch @youph. I think we should be doing this as otherwise the sampler will create the schema from the graph again, thus slowing things down. We can also make `HinSAGEGenerator` the default for both GraphSAGE and HinSAGE as it can handle both.

PR #338, opened 2019-02-24
Updated generators to pass graph schema to walkers

adocherty:
Fixed a bug noted in PR  #337:

> I noticed that we call the sampler before we get the schema, and we don't pass the schema to the sampler. Should it be the other way around, i.e., instead of

```
# The sampler used to generate random samples of neighbours
self.sampler = SampledBreadthFirstWalk(G, seed=seed)

# We need a schema for compatibility with HinSAGE
self.schema = G.create_graph_schema(create_type_maps=True)
```

> do

```
# We need a schema for compatibility with HinSAGE
self.schema = G.create_graph_schema(create_type_maps=True)

# The sampler used to generate random samples of neighbours
self.sampler = SampledBreadthFirstWalk(G, graph_schema=self.schema, seed=seed)
```


adocherty:
Yes, sorry I don't know how I missed that. The errors were because we currently require a "type map" in the schema to map each node to its type. This is hard to do manually and I think we should avoid this in the future and push it to the graph class. I created an issue for doing this: https://github.com/stellargraph/stellargraph/issues/347

youph:
@adocherty should this be merged?

adocherty:
I forgot about this one.  I've updated the node mappers to raise an error if the type maps are not supplied.  I also merged develop into this branch ready to merge the changes into develop.

PR #341, opened 2019-03-04
Hotfix/gcn-doc

wangzhen263:
1. Fixed the doc style of GCN model
2. Updated CSIRO DATA61 copyright year "2018" -> "2018-2019"


wangzhen263:
> * Add expected argument types for all arguments (currently there are expected types listed for less then half of the args)

Added

stellargraph/layer/gcn.py
```diff
@@ -149,8 +188,20 @@ def get_config(self):
...
+    activation functions for each hidden layers, and a generator object.
+
+    Args:
+        layer_sizes (list of int): list of output sizes of GCN layers in the stack
+        activations (list of str): list of activations applied to each layer's output
+        generator (FullBatchNodeGenerator): an instance of FullBatchNodeGenerator class constructed on the graph of interest
+        bias (bool): toggles an optional bias in GCN layers
+        dropout (float): dropout rate applied to input features of each GCN layer
+        kernel_regularizer (str): normalization applied to the kernels of GCN layers
+        kwargs: additional parameters for chebyshev or localpool filters
```
youph: is there a description of what these additional parameters are, and how to use them? If not, it should be provided.

stellargraph/layer/gcn.py
```diff
@@ -27,8 +27,28 @@
...
+        use_bias (bool): toggles an optional bias
+        kernel_initializer (str): name of layer bias f the initializer for kernel parameters (weights)
+        bias_initializer (str): name of the initializer for bias
+        attn_kernel_initializer (str): name of the initializer for attention kernel
+        kernel_regularizer (str): name of regularizer to be applied to layer kernel. Must be a Keras regularizer.
+        bias_regularizer (str): name of regularizer to be applied to layer bias. Must be a Keras regularizer.
+        activity_regularizer (str): not used in the current implementation
+        kernel_constraint (str): constraint applied to layer's kernel
+        bias_constraint (str): constraint applied to layer's bias
+        **kwargs:
```
youph: description of **kwargs?

stellargraph/layer/gcn.py
wangzhen263: Thanks, added.

stellargraph/layer/gcn.py
wangzhen263: Based on the practices of keras layer code https://github.com/keras-team/keras/blob/master/keras/layers/core.py, we should remove it from comments.

stellargraph/layer/gcn.py
youph: You've given an example of `filter="localpool", max_degree=2` as additional arguments to the `GCN` class constructor, but in `gcn-cora-example.py` the argument `filter="localpool"` is passed to the `FullBatchNodeGenerator`, not to `GCN`! This is quite confusing and in fact misleading. Please fix properly.

wangzhen263: GCN class constructor also has `filter="localpool"` in the `gcn-cora-example.py` in the VERY right in the L61 because L61 is quite long and is hidded by the browser.  please see https://github.com/stellargraph/stellargraph/blob/hotfix/gcn-doc/demos/node-classification-gcn/gcn-cora-example.py#L61.

youph: ok, thanks, didn't notice. This however raises other questions, e.g., are the `filter` arguments passed to `FullBatchNodeGenerator` and to `GCN` required to be the same? What if they are not? Can the `filter` type be inferred from the `generator` passed to `GCN`, rather than be specified as an optional argument? Think of a "fool proof" way of making this work.

youph: Also, if `filter` argument can be passed to `FullBatchNodeGenerator` as a kwarg, a description should also be added in file:///Users/tys017/Projects/Graph_Analytics/code/stellar-ml/docs/_build/html/api.html#stellargraph.mapper.FullBatchNodeGenerator

wangzhen263: Updated

PR #368, opened 2019-04-12
dockerise stellargraph and publish images via CI

TheDen:
PR:

* Dockerises `stellargraph` (image name `stellargraph/stellargraph`
* Dockerises the `stellargraph` demos in a docker image with jupyter notebook (image name `stellargraph/stellargraph-demos`), with the included datasets—built from the base image `stellargraph/stellargraph`
* Both images are public
* Updates the buildkite pipeline so that the docker images are built and pushed via the CI
   * PRs build and push docker images with the tag `testing`
   * `develop` branch builds and pushes the docker images with the tag `latest`
   * As discussed with Andrew, for now we don't build and push on `master` since it gets merged back into `develop`
* Updates the README.md with information on running the docker image + adds the "docker pulls" badge for `stellargraph`

The demos can be tested by running
```docker run -it -p 8888:8888 stellargraph/stellargraph-demos```
and opening http://localhost:8888

### CODE REVIEW

File: docker/stellargraph-demos/Dockerfile [hunk 1]
```diff
@@ -0,0 +1,32 @@
+FROM python:3.6-slim as base
```
huonw: Why this as a base?

README.md
```diff
@@ -152,10 +157,29 @@ pip install -r requirements.txt
 pip install .
 ```
 
+## Docker Images
+
+There are two docker images
+
+* [stellargraph/stellaraph](https://hub.docker.com/r/stellargraph/stellargraph): Docker image with `stellargraph` installed.
```
huonw:
```suggestion
* [stellargraph/stellargraph](https://hub.docker.com/r/stellargraph/stellargraph): Docker image with `stellargraph` installed.
```

- [ ] And the same on line 167

docker/stellargraph/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.6-slim
+
+RUN adduser --disabled-password --gecos '' stellar
+
+USER stellar
+WORKDIR /home/stellar
+
+ENV PATH=${PATH}:/home/stellar/.local/bin
+
+COPY --chown=stellar . /home/stellar/stellargraph/
```
huonw: According to dive, this has copied 114MB of data into the image, e.g.:

- 40MB in `.git`
- 14MB in `.tox` folder (including a 3.6MB python3.6 binary!)
- 44MB in `tests/resources`

I believe these are all not truly needed (e.g. `.tox` is temporary), and all this gets deleted in the next `RUN` line.

Then again... the `RUN` line installs 900MB of packages (of which stellargraph itself is <1MB) into `~/.local/lib/python3.6/site-packages`, so maybe this doesn't matter.

.buildkite/pipeline.yml
```diff
@@ -18,7 +18,7 @@ steps:
     agents:
       queue: 't2medium'
 
-  - label: "style"
+  - label: ":python-black: style"
```
huonw: (This and the 3.6 above probably didn't need to happen in this PR 😄 )

docker/stellargraph/Dockerfile
TheDen: 🤔 nice catch! I can do a `rm -rf` on `.tox`+`.git` in `RUN` or add them in a [.dockerignore](https://docs.docker.com/engine/reference/builder/#dockerignore-file) file

.buildkite/pipeline.yml
TheDen: I know, I couldn't resist!

docker/stellargraph/Dockerfile
TheDen: Actually may not be needed since like you mentioned it's removed so it won't really add to the final image size

docker/stellargraph-demos/Dockerfile
TheDen: I'm using a multi-stage build—so the first intermediate image install the packages needed to download the files and the 2nd actual image (which the the `stellargraph/stellargraph` image just built) just copies the files from the first

huonw: Right, but why python:3.6-slim? This stage doesn't seem to use any python?

docker/stellargraph/Dockerfile
huonw: It does add to the size of the layers, meaning the size of all the data that gets downloaded. Is it possible to somehow reuse the gitignore file as the dockerignore? The former already filters out uninteresting/temporary files, and will be kept up to date better.

docker/stellargraph-demos/Dockerfile
TheDen: Ah I see—I guess it's to keep it consistent. The image would already live in the agent from the `stellargraph/stellargraph` build—and it's going to be used for the actual build, so it makes sense to use it. What are your thoughts? We can update it to use `alpine`?

I also want to update the images used in the BK pipeline from `python3.6` to the slim version in a different PR

docker/stellargraph/Dockerfile
TheDen: We could potentially create a symbolic link?

docker/stellargraph-demos/Dockerfile
huonw: Ah makes a lot of sense. Maybe a `# image needed anyway, so also use for downloading` comment or something?

TheDen: Yeah sure, I'll add a comment 👍

docker/stellargraph/Dockerfile
huonw: It looks like the syntax is not quite a perfect match https://github.com/LinusU/gitignore-to-dockerignore

TheDen: 🤔 I guess we can use that tool to autogenerate it for now

TheDen:
I've added a `.dockerignore` but I'm taking a whitelist approach via negation (the `README.md` seems to be required by `seup.py` for building stellargraph).

I also thought about the base image and decided to switch to `alpine`—reason being is that if someone builds locally and they won't have `python:3.6-slim` (having `stellargraph/stellargraph` locally won't count AFAIK). It's also easier to pull a 3MB image + alpine it has more of the packages required built in.

What do you think @huonw?

huonw:
> I've added a .dockerignore but I'm taking a whitelist approach via negation (the README.md file seems to be required by seup.py for building stellargraph).

Seems fragile, but I guess the installed may start failing if some required files are missed (e.g. if someone has a change that adds new required ones, they'll be forced to update them here). The most concerning thing would be potential files that are required for running but aren't required for installation, meaning the docker image is built successfully, but can't actually run anything.

> I also thought about the base image and decided to switch to alpine—reason being is that if someone builds locally and they won't have python:3.6-slim (having stellargraph/stellargraph locally won't count AFAIK). It's also easier to pull a 3MB image + alpine it has more of the packages required built in.

Docker's layer model means `stellargraph/stellargraph` should count? It would have to check docker hub to find the tag, but it should find out that it's already downloaded all the layers.

TheDen:
Yep it would use the layer hash if it exists—will need to test it out locally to be certain it works. Agreed on the potential issues with the whitelist approach (I already ran into it with the latest build). I'll roll it back to using blacklist ignores, so at the very least we can speed up the build context generation and reduce the size. Using `.gitignore` also isn't enough since it doesn't capture everything that one would want to ignore for the image. 

I'm thinking doing blanket `COPY . /dest` is considered harmful or at least problematic, since you end up managing a `.dockerignore` file for everything under the root directory. We could alternatively just have `COPY` of the folders + files that are actually needed and not worry about  a monolithic `.dockerignore` (keep it only for `.gitignore` type files)?

huonw:
> Using .gitignore also isn't enough since it doesn't capture everything that one would want to ignore for the image

It isn't enough to be perfect, but it covers a lot, e.g. the `.tox` folder.

> We could alternatively just have COPY of the folders + files that are actually needed and not worry about a monolithic .dockerignore (keep it only for .gitignore type files)?

I think that's quite similar to the whitelist approach (it's a whitelist of files to explicitly copy), and so has the same downsides?

TheDen:
I added a `.dockerignore` (generated from `.gitignore`). To avoid the whitelist/blacklist approach, the latest commit does a `COPY .` as before—however using multi-stage builds, the final image copies the installed packages from `/home/stellar/.local` so it's clean

docker/stellargraph/Dockerfile
```diff
@@ -1,14 +1,18 @@
...
+
+COPY . /build/
+
+RUN pip install --no-cache-dir -r /build/requirements.txt /build/ --user
+
 FROM python:3.6-slim
 
 RUN adduser --disabled-password --gecos '' stellar
 
+COPY --chown=stellar --from=builder /root/.local /home/stellar/.local
```
huonw: Could you add a comment or two about how this all works? E.g. referencing that `--user` will install to `.local`?

.buildkite/pipeline.yml
```diff
@@ -26,14 +26,29 @@ steps:
...
           - PYTHONDONTWRITEBYTECODE=1
 
+  - label: ":docker: :hadolint:"
+    command: "find -name 'Dockerfile*' -print0 | xargs -0 hadolint"
+    plugins:
+       docker#v3.0.1:
+          image: "hadolint/hadolint:latest-debian"
+
+  - label: ":docker: publish numbered images"
+    command: ".buildkite/steps/manage-images.sh build-publish-numbered"
```
huonw: What's the motivation for doing this separation? For the other places, it's been motivated by wanting to run tests on the temporary images but that's not currently the case here (it would probably good here as a future enhancement... it'd be nice to check that `stellargraph`/`tensorflow`/etc. all get installed right).

docker/stellargraph-demos/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+      && unzip BlogCatalog-dataset.zip
+
+# hadolint ignore=DL3006
+FROM stellargraph/stellargraph
+
+ENV JUPYTER_VERSION="1.0.0" \
+  PATH=${PATH}:/home/stellar/.local/bin
+
+# hadolint ignore=DL3013
+RUN pip install --upgrade pip --user \
```
huonw: What's the reason for updating pip here?

docker/stellargraph-demos/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+      && pip install jupyter=="${JUPYTER_VERSION}" --user
```
huonw: Should this use `--no-cache-dir` and/or multistage builds?

.buildkite/pipeline.yml
TheDen: Currently it's so the image can be built in parallel with the other steps, so the build is faster—but also in case other tests may be used

docker/stellargraph-demos/Dockerfile
TheDen: To keep it up to date with the deployed jupyter notebook. Running `!pip install` inside a notebook will ensure there aren't any warnings. We can remove this if you want?

docker/stellargraph/Dockerfile
```diff
@@ -5,12 +5,14 @@ WORKDIR /build/
...
 COPY . /build/
 
+# installs the packages to /root/.local
 RUN pip install --no-cache-dir -r /build/requirements.txt /build/ --user
 
 FROM python:3.6-slim
 
 RUN adduser --disabled-password --gecos '' stellar
 
+# copy the installed packagres (/root/.local) from the builder
```
huonw:
```suggestion
# copy the installed packages (/root/.local) from the builder
```

docker/stellargraph-demos/Dockerfile
huonw: Seems like a good reason (maybe a comment?). 👍

docker/stellargraph-demos/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+      && curl -sLO "http://socialcomputing.asu.edu/uploads/1283153973/BlogCatalog-dataset.zip" \
+      && unzip BlogCatalog-dataset.zip
+
+# hadolint ignore=DL3006
+FROM stellargraph/stellargraph
+
+ENV JUPYTER_VERSION="1.0.0" \
+  PATH=${PATH}:/home/stellar/.local/bin
+
+# keep pip up to date for the end to user
```
huonw:
```suggestion
# keep pip up to date for the end user
```

huonw:
It seems like this was superseded by #386? Can it be closed?

TheDen:
Yep, closing

PR #377, opened 2019-04-29
Missing paramiko ps

jessclmoore:
Added missing paramiko package to a requirements.txt for link-prediction/random-walks and node-classification/node2vec, and updated README.md in these directories.
Please review


adocherty:
A first comment is that we should also add this package to the `setup.py` requirements list. The default `pip` tool installation does not use the `requirements.txt` .

This should be an optional requirement under "demos" (as it is not required by the base library). For example:
```
EXTRAS_REQURES = {"demos": ["numba",  "paramiko"], "test": ["pytest", "pandas", "pytest-benchmark"]}
```

youph:
@jessclmoore none of the examples in `demos/link-prediction/random-walks` require `paramiko` package. Moreover, the description of the package http://www.paramiko.org suggests it has nothing to do with our demos. Are you sure it's needed?

youph:
@PantelisElinas when you check this PR, could you please also take a look at whether we still need a separate `demos/link-prediction/random-walks/requirements.txt` file, or should we instead include those requirements in `setup.py` under "demos"? Also, the scipy version in `demos/link-prediction/random-walks/requirements.txt` mismatches that in the main requirements.

youph:
> A first comment is that we should also add this package to the `setup.py` requirements list. The default `pip` tool installation does not use the `requirements.txt` .
> 
> This should be an optional requirement under "demos" (as it is not required by the base library). For example:
> 
> ```
> EXTRAS_REQURES = {"demos": ["numba",  "paramiko"], "test": ["pytest", "pandas", "pytest-benchmark"]}
> ```

@adocherty this should ONLY be added as an optional requirement under "demos" in `setup.py`, not "ALSO", right?

demos/node-classification/node2vec/requirements.txt
```diff
@@ -0,0 +1,2 @@
+paramiko==2.4.2
```
youph: The demo runs fine in an environment without `paramiko` installed. I don't understand why `paramiko` is needed, especially since it's a python package for ssh2.

jessclmoore:
@youph I received the following warning when running demos/embeddings/stellargraph-metapath2vec.ipynb

UserWarning: paramiko missing, opening SSH/SCP/SFTP paths will be disabled.  `pip install paramiko` to suppress

jessclmoore:
@youph The same paramiko missing user warning also occurred when running these demos:
node-classification/node2vec/*
link-prediction/random-walks/*

adocherty:
> > A first comment is that we should also add this package to the `setup.py` requirements list. The default `pip` tool installation does not use the `requirements.txt` .
> > This should be an optional requirement under "demos" (as it is not required by the base library). For example:
> > ```
> > EXTRAS_REQURES = {"demos": ["numba",  "paramiko"], "test": ["pytest", "pandas", "pytest-benchmark"]}
> > ```
> 
> @adocherty this should ONLY be added as an optional requirement under "demos" in `setup.py`, not "ALSO", right?

It depends what we want `requirements.txt` to be.  It is not used in our setup instructions, but is kept around for two purposes:
* Backward compatibility for systems/people who are used to using the requirements file. 
* Specifying the environment for buildkite tests.

So, there is a question of whether we want only what's needed for the library and unit tests or everything required for the demos as well.

PantelisElinas:
> @PantelisElinas when you check this PR, could you please also take a look at whether we still need a separate `demos/link-prediction/random-walks/requirements.txt` file, or should we instead include those requirements in `setup.py` under "demos"? Also, the scipy version in `demos/link-prediction/random-walks/requirements.txt` mismatches that in the main requirements.

@youph we can remove the requirements.txt from this directory. The only package we need is chardet so we can move it to setup.py and delete the file here. The newer version of scipy works just as well so we can go with the version in setup.py.

PantelisElinas:
Hi @jessclmoore 

I can't replicate your error with the missing paramiko package. I've tried it on several clean installations. I don't understand why you get this error.

Regards,

P.

PantelisElinas:
> @youph I received the following warning when running demos/embeddings/stellargraph-metapath2vec.ipynb
> 
> UserWarning: paramiko missing, opening SSH/SCP/SFTP paths will be disabled. `pip install paramiko` to suppress

Hi @jessclmoore this being a warning, do the demos still run correctly? If it is just a warning generated because of some strange dependency among other packages you have installed, then we probably should not worry about it, especially since none of us can replicate this error.

Regards,

P.

PR #378, opened 2019-04-30
Updated README.md to include additional setup instructions needed pri…

jessclmoore:
…or to stellargraph install on mac osx

Suggest useful to add section on Mac OSX Setup prior to installation instructions, as users need xcode command line tools and a ~/.matplotlib/matplotlibrc. The command line tools are required by pip to install packages from wheels, e.g. for igraph. The backend setting in matplotlibrc was needed to use matplotlib in python interpreter installed with Anaconda.
Please review changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@
    * [Guiding Principles](#guiding-principles)
    * [Getting Started](#getting-started)
    * [Installation](#installation)
+       * [Mac OSX System Setup](#mac-osx-system-setup)
        * [Install StellarGraph using pip](#install-stellargraph-using-pip)
        * [Install StellarGraph from Github source](#install-stellargraph-from-github-source)
    * [Running the examples](#running-the-examples)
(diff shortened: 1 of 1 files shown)


youph:
Thanks for this PR @jessclmoore! Although the instructions for Mac OS installation might be useful for some users, I feel they have little to do with the library, rather with making python work on Mac OS, and hence I tend to think these details are excessive. What do you think @adocherty @PantelisElinas?

adocherty:
> Thanks for this PR @jessclmoore! Although the instructions for Mac OS installation might be useful for some users, I feel they have little to do with our library, but rather with making python work on Mac OS, and hence I tend to think these details are excessive. What do you think @adocherty @PantelisElinas?

I think that these instructions could be very useful for our partners in installing the library, but I tend to agree with @youph that it is not really about our library. 

Perhaps, we can put this detail in the main docs (say `docs/quickstart.txt`) so they are available but we can keep the front-page readme cleaner?

adocherty:
@youph I suggest we close this?

PR #380, opened 2019-05-01
Refer to the GAT paper, not the GaAN one

huonw:
The GAT paper is https://arxiv.org/abs/1710.10903, while
https://arxiv.org/abs/1803.07294 is Zhang, Jiani, et al. "GaAN: Gated
attention networks for learning on large and spatiotemporal graphs."


huonw:
I don't know if this is the right fix (as in, maybe the title/author should be changed?), but either way, the links and the text are currently inconsistent.

PR #383, opened 2019-05-03
Demo/saliency maps

sktzwhj:
Hi @youph ,

I have made the demo of interpretability for GCN. Please have a look at it. Any comments and suggestions are appreciated.

-Huijun


sktzwhj:
> The demo notebook didn't work as is, had to fix it.
> The initial comments below refer to cell numbers in the newly pushed demo notebook `node-link-importance-demo-gcn.ipynb`
> 
> * cell 24: why do we see "Converting sparse IndexedSlices to a dense Tensor of unknown shape"?
This is the problem caused by using K.gather (see https://stackoverflow.com/questions/35892412/tensorflow-dense-gradient-explanation). One possible fix is to use tf.dynamic_partition (see https://www.tensorflow.org/api_docs/python/tf/dynamic_partition). However, the latter is tf-specific. I am not sure we want to keep the code Keras-compatible for different backends. If not, we may fix by using the referenced solution.
 
> * cell 30: what do these sanity checks verify, and what is the expected behaviour? Also, we should probably compare "Number of non-zero elements in integrate_link_mask" with number of edges in G_ego; the former should be less or equal the latter, right?

You are correct. I have added more descriptions here. 

> * cell 31: do the node ids shown in the top-10 most important links correspond to real node ids in the graph, or to node indices in the graph? I.e., what does node 0 correspond to? Also, wouldn't we expect at least some of the top-10 links to include the target node 127?

Well, you are correct. Fixed. Also, remember that there is a hyper-parameter to tune here - the number of steps for integration. A too large/small step may lead to inaccurate importance.

> * cell 32: needs a comment of what's happening there

Comments added.

> * cell 33: the plot looks good, but I don't understand why the link from target to the large hub node has small importance? I'd expect it to have a high positive importance. Moreover, in another run I saw this link to have a small _negative_ importance, which is even more strange.

I would suggest starting from a small step when you choose a different node given the fact that the step number for integration is a hyper-parameter to tune. i.e., starting from a small step and increase it gradually until we get a stable result.

>   This plot made for a different target node, `target_idx = 1` (node 1033), aligns more with my intuition:
>   image
> * cell 35:
>   
>   * only the target node is displayed, at least in this run. Not sure why...
>   * `prediction_change = clean_prediction[1] - predict_after_perturb[1]` has hard-wired class of interest. I think you want `prediction_change = clean_prediction[class_of_interest] - predict_after_perturb[class_of_interest]`. Please fix.

Fixed. The above issue for only displaying the target is likely to be caused by this bug as well.

PR #386, opened 2019-05-06
add stellargraph docker image

TheDen:
To help us move forward, this PR is split from #368 so that it's only building the `stellargraph/stellargraph` docker image, until we can confirm hosting datasets (since [stellargraph-evaluation](https://github.com/stellargraph/stellargraph-evaluation/) is now using the `stellargraph/stellargraph` image)


.buildkite/steps/manage-images.sh
```diff
@@ -0,0 +1,88 @@
...
+        echo "successfully deleted 'stellargraph/${image_name}:${version}'"
+      else
+        echo "failed to delete 'stellargraph/${image_name}:${version}' (probably never uploaded)"
+      fi
+      fi
+      #done
+    done
+}
+
+images=stellargraph
```
huonw: Should this still be a list for ease of extension?

```suggestion
images=(
  stellargraph
)
```

.buildkite/steps/script.sh
```diff
@@ -11,5 +11,5 @@ py.test -ra --cov=stellargraph tests/ --doctest-modules --doctest-modules --cov-
 coveralls
 
 if [ "${BUILDKITE_BRANCH}" = "develop"  ] || [ "${BUILDKITE_BRANCH}" = "master" ] ; then
-	upload_tests
+  upload_tests
```
huonw: (Why did this need to change?)

.buildkite/pipeline.yml
```diff
@@ -26,8 +26,32 @@ steps:
         environment:
           - PYTHONDONTWRITEBYTECODE=1
 
+  - label: ":docker: :hadolint:"
```
huonw: Maybe this could be a bit more descriptive, since hadolint isn't entirely obvious (I know platform currently isn't, but maybe it should be)?

```suggestion
  - label: ":docker: :hadolint: linting Dockerfiles"
```

.buildkite/steps/script.sh
TheDen: It was a `\t` character I noticed that was left over in a script that uses spaces

.buildkite/steps/manage-images.sh
TheDen: Good idea

.buildkite/pipeline.yml
TheDen: Okay I've pushed the suggested changes 👍

PR #391, opened 2019-05-08
Added seaborn

jessclmoore:
The seaborn package was required ensembles/ensemble-node-classification-example.ipynb and similar demos. I've added it to the overall requirements.txt as required by more than one demo.


youph:
@jessclmoore instead of adding `seaborn` to `requirements.txt`, I've added it to the `demos` requirements specified in `setup.py`, as this package is only required by the demos. Could you please check whether this solves the issue?

requirements.txt
```diff
@@ -12,3 +12,5 @@ pytest==3.9.3
 pytest-benchmark>=3.1
 pytest-cov>=2.6.0
 coveralls>=1.5.1
+seaborn>=0.9.0
```
youph: this should not be in base library requirements

jessclmoore: Yep worked. I uninstalled seaborn, removed seaborn from requirements.txt and added seaborn to setup.py demos, and installing with
pip install stellargraph[demos]
then tested:
-> ensemble-node-classification-example.ipynb worked with seaborn 

I think this change just needs the README changed to tell people who install with git that they still have to run:
pip install stellargraph[demos]

jessclmoore: @youph I've made the changes. and pushed it up.

PR #392, opened 2019-05-08
Python igraph

jessclmoore:
The python-igraph package required for the community detection demo is no longer available in v0.7.1  on PyPl. As only required in this demonstration, I've removed the version number from this package in the requirements.txt of the community-detection folder.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,5 +12,4 @@ pytest==3.9.3
 pytest-benchmark>=3.1
 pytest-cov>=2.6.0
 coveralls>=1.5.1
-seaborn>=0.9.0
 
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -12,3 +12,5 @@ pytest==3.9.3
 pytest-benchmark>=3.1
 pytest-cov>=2.6.0
 coveralls>=1.5.1
+seaborn>=0.9.0
```
youph: this shouldn't be here, as it's not required by the library itself. This should go into "demos" requirements in `setup.py`

youph: @jessclmoore Are you sure `seaborn` is required? For which demo?

jessclmoore: On this demo: ensemble-node-classification-example.ipynb with model_type=’graphsage’ on the Cora dataset, I got the error: ModuleNotFoundError: No module named 'seaborn'

jessclmoore: Thanks @youph. I've fixed this by removing seaborn from the requirements.txt and pushed back  up.

PR #394, opened 2019-05-09
use python:3.6-slim docker image

TheDen:
PR:

* Updates the docker image that's used for the tests to `python:3.6-slim` from `python:3.6`
* `python:3.6` is 354MBs compressed
* `python:3.6-slim` is 50MBs compressed
* This means that the buildkite agents will pull down the image faster since it's significantly smaller
* Since the `stellargraph/stellargraph` docker image uses `python:3.6-slim`, the agent only needs to pull down one image

The CI passes with the slim image, however I'm now hesitant to use it given the recommendation from https://hub.docker.com/_/python?tab=description

>This image does not contain the common packages contained in the default tag and only contains the minimal packages needed to run python. Unless you are working in an environment where only the python image will be deployed and you have space constraints, we highly recommend using the default image of this repository.

What are your thoughts?

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -3,7 +3,7 @@ steps:
     command: ".buildkite/steps/script.sh"
     plugins:
       docker#v3.2.0:
-        image: "python:3.6"
+        image: "python:3.6-slim"
         environment:
           - PYTHONDONTWRITEBYTECODE=1
           - GIT_BRANCH=$BUILDKITE_BRANCH


huonw:
Seems fine to me; we're not interacting with this image directly as it's only on CI, and, if there is something that needs more than the minimal image, then it can be changed back to the non-`slim` one then.

However... it might not be worth optimising for "only" 300MB of downloads on these build machines. 

To put some numbers (but only one sample for each...), the test step takes:

- ~5:40 when it has to pull the `python:3.6` images https://buildkite.com/stellar/stellargraph/builds/1209#752797a8-5cd9-4341-9caf-0613b0281651
- ~5:10 when it has to pull the `python:3.6-slim` images https://buildkite.com/stellar/stellargraph/builds/1218#d4f7ed30-b5d4-4c89-8f36-5f5041cfc2c5
- ~5:05 when it doesn't have to pull either https://buildkite.com/stellar/stellargraph/builds/1212#d125715c-40d9-4047-8af5-c8671f53a671

TheDen:
Thanks for the timings breakdown @huonw! I'll open the PR to be ready for review, and you guys can approve if you're okay with this change. Agreed that we can always roll back to not using slim. Another aspect I forgot is that the disk space usage on the agents will be less (but that's also negligible).

PR #405, opened 2019-05-16
Fixing GCN_Aadj_feats_op function passing in FullBatchNodeGenerator

habiba-h:


stellargraph/mapper/node_mappers.py
```diff
@@ -575,7 +576,10 @@ class FullBatchNodeGenerator:
             `gcn-cora-example.py <https://github.com/stellargraph/stellargraph/blob/master/demos/node-classification-gcn/gcn-cora-example.py>`_
     """
 
-    def __init__(self, G, name=None, func_opt=None, k=None, **kwargs):
+    # def __init__(self, G, name=None, func_opt=None, k=None, **kwargs):
+
+    def __init__(self, G, name=None, k=None, **kwargs):
```
adocherty: Here I was thinking that we would keep a function parameter. I'd prefer something like 'A_transform' or just 'transform'. Then, if this option is not specified, you can apply the standard function.

I'd also prefer there to be a 'method' parameter which then supplies the options similar to the 'filter'  parameter, for example:
*  "gcn" (now it's localpool but I think gcn would be more understandable)
* "chebyshev" 
* "sgcn" (now smoothed)
*  None, "none" or "gat" where the adjacency matrix will be 

The user-supplied function would only take A and k as parameters (currently it is passed the `**kwargs`, but I think this is unnecessarily complicated as the user can always create a closure).

Therefore, I'm thinking of the method to be defined as:
```
 def __init__(self, G, name=None, method="gcn", k=1, transform=None):
```
so that when you don't specify anything you get the standard GCN preprocessing.

Alternatively, we can set the default option as `method=None` which will be the preprocessing needed for GAT.

habiba-h: Hi @adocherty & @youph , 
i have made the following changes based on @adocherty comments above and our supplementary discussion on Wednesday regarding this. 
1. The `FullBatchNodeGenerator` definition has been updated. 
However, instead of 
`def __init__(self, G, name=None, method="gcn", k=1, transform=None):`
it's
`def __init__(self, G, name=None, method="gcn", k=1, transform=None, sparse = True):` 
(whether to convert to a dense graph (specially when method = None meaning GAT). I think Andrew's working on an update about whether to create a dense matrix or not so this part may need to be revisited. 
2. The `GCN_Aadj_feats_op` method definition is modified to take in a method as a string  (valid options: 'gcn', 'chebyshev', 'sgcn') and `k` to be used as the power of adjacency matrix incase of sgcn or maximum order of Chebyshev polynomials.   
3. In class `GCN` `__init__` and `node_model`  are updated based on FullBatchNodeGenerator parameters.
4. The `node_mappers` and `utils` tests are updated for `FullBatchNodeGenerator` and `GCN_Aadj_feats_op` changes respectively.
5. Demos:  The `FullBatchNodeGenerator` call is `gcn-cora-example.py` is updated according to the new definition of of the method.

adocherty:
This looks good now.  I've changed the docstring of `FullBatchNodeGenerator` and added a test for `k<2` for the Chebyshev method. If you're happy with the changes, go ahead and merge this.

You'll need to merge the develop branch into this branch and resolve the merge conflicts first.

adocherty:
Actually, one more thing: After you merge develop, please update the CHANGELOG.md file with a quick one-line description of the changes in this PR (under enhancements at the top of the file).

PR #416, opened 2019-06-05
use python 3.7-slim docker image

TheDen:
Similarly to #394 this PR updates the Buildkite docker image to use `3.7-slim`. It also uses the minor version `3.7` which means the latest patch versions are always used (rather than pinning to `3.7.2`)


adocherty:
This looks good to me!  Also, @youph is away so don't wait for a review.  I think you are good to merge this.

PR #417, opened 2019-06-05
Added indexed targets for GAT & GCN

adocherty:
The GCN & GAT layers and classes have been updated to return only outputs for the nodes provided to the .flow method of the `FullBatchNodeGenerator` object, and in the same order.

Furthermore, there is no need to use the weighted metrics or losses for the method, the same accuracy and other metrics can now be used as for the GraphSAGE/HinSAGE models.
If we would like to pass weights in the training functions to correctly deal with weighted problems this can be done in the same way as other models by using the Keras `.fit` or `.fit_generator` methods with one of the `class_weight=` or `sample_weight=` argument supplied.

### Review Checklist:
- [x] Required functionality implemented
- [x] Unit tests run
- [x] Documentation matches implementation (doc strings, README, Sphinx)
- [x] Python notebooks in demos directory run (GCN/SGC & Ensemble demos are affected, new notebook for GCN is implemented.)


adocherty:
Thanks for the review @habiba-h see below for responses to your comments:

> Something very simple: The GCN demo runs like a top however in the accompanying text for explanations and references, it is referred to as GAT. Just some updates in the notebook text.

Good catch. I've edited the GCN demo and replaced all instances of GAT with GCN.

> And this other issue is not related to the indexed targets functionality but just something that would need to checked for. So in the FullBatchNodeGenerator for the gat method if the sparse = True (or left out completely, since default is true) the gat model breaks down. I understand for `gat` there has to be a dense matrix but the parameter doesn't get checked in the code so a dense matrix isn't created and the `fit_generator` with `gat` breaks down.

Good point, the GAT model does not work with sparse matrices in this PR. However, the following PR implements this feature:
https://github.com/stellargraph/stellargraph/pull/420

Also, as we create the `FullBatchNodeGenerator` before we create the model, we can't check that the model can accept sparse matrices at instantiation.  It could be a problem going forward as it means we require all full-batch methods to support sparse matrices.  We may need to revisit this later.

adocherty:
An issue raised by @PantelisElinas:

When using a binary classification the output of the GCN model stack has an extra dimension, namely the final output will be of shape 1 x N x 1.  To feed a binary vector of labels into the `.fit` method we need to reshape it to N x 1.

To fix this we can reshape the output predictions of the layer:
```
x_out = Reshape((-1,))(x_out)
```

This is done internally in GraphSAGE (for a different reason).  I have added a new issue to investigate this: https://github.com/stellargraph/stellargraph/issues/425

PR #420, opened 2019-06-12
Implemented GAT using a sparse adjacency matrix

adocherty:
GAT has been implemented using sparse matrix operations. This only works with a tensorflow backend.

Now, GCN and GAT can both use either dense or sparse adjacency matrices.  The dense adjacency matrices are currently required for the saliency map calculations.

### Review Checklist:
- [x]  Required functionality implemented
- [x]  Unit tests run
- [x]  Documentation matches implementation (doc strings, README, Sphinx)
- [x]  Python notebooks in demos directory run (GAT & Ensemble demos are affected)


demos/node-classification/gat/gat-cora-example.py
```diff
@@ -61,7 +61,7 @@ def train(
     target_name="subject",
 ):
     """
-    Train a GAT model on the specified graph G with given parameters, evaluate it, and save the model.
+    Train a GAT model on the specified graph G with given parameters, zevaluate it, and save the model.
```
PantelisElinas: Spelling error, "zevaluate" --> "evaluate"

stellargraph/mapper/node_mappers.py
```diff
@@ -607,24 +714,27 @@ def __init__(self, G, name=None, method="gcn", k=1, sparse=True, transform=None)
...
+        self.use_sparse = sparse
 
-        if not self.sparse:
-            self.Aadj = self.Aadj.todense()
+        # If sparse is specified, check that the backend is tensorflow
+        if self.use_sparse and K.backend() != "tensorflow":
+            warnings.warn(
+                "Sparse adjacency matrices are only supported in tensorflow."
+                " Falling back to using a dense adjacency matrix."
+            )
```
PantelisElinas: Should self.use_sparse be set to False after the warning?

stellargraph/layer/graph_attention.py
```diff
@@ -208,38 +232,37 @@ def build(self, input_shape):
...
+        Notes:
+            This does not add self loops to the adjacency matrix.
+            The output indices are only used when `final_layer=True`
 
+        Args:
+            inputs (list): list of inputs with 3 items:
+            node features (size b x N x F),
+            output indices (size b x M),
+            graph adjacency matrix (size b x N x N),
+            where N is the number of nodes in the graph,
```
sktzwhj: It might be clearer to say what b represents here.

stellargraph/layer/graph_attention.py
```diff
@@ -297,36 +320,217 @@ def call(self, inputs, **kwargs):
...
+            kernel = self.kernels[head]  # W in the paper (F x F')
+            attention_kernel = self.attn_kernels[
+                head
+            ]  # Attention kernel a in the paper (2F' x 1)
+
+            # Compute inputs to attention network
+            features = K.dot(X, kernel)  # (N x F')
+
+            # Compute feature combinations
+            # Note: [[a_1], [a_2]]^T [[Wh_i], [Wh_2]] = [a_1]^T [Wh_i] + [a_2]^T [Wh_j]
```
sktzwhj: Wh_2 -> Wh_j?

stellargraph/mapper/node_mappers.py
adocherty: Good catch .. yes we should set the flag to False.

adocherty:
> The changes look good to me. In terms of the changes we may have for the ongoing work, my understanding is that we will still only support the dense version for saliency maps due to the potential edge constraint. Is that correct?

Yes, we are keeping the dense versions of both GCN and GAT which we will use for saliency maps.

stellargraph/layer/graph_attention.py
adocherty: Good point, I'll add a comment.

stellargraph/layer/graph_attention.py
adocherty: Well spotted!

demos/node-classification/gat/gat-cora-example.py
adocherty: Fixed.

stellargraph/mapper/node_mappers.py
adocherty: I will note it's hard to test this in a unit test framework as we would have to change the Keras backend for that test.

PR #430, opened 2019-06-21
Hateful users Jupyter notebook

PantelisElinas:
Hi @adocherty and @annitrolla 

I have created a demo notebook for the hateful users on Twitter use-case. It can be used to build a predictive model using GraphSAGE, GCN, or GAT and compare with a Logistic Regression model trained on the same split of the data. 

I have created a new folder under demos named use-cases. I think notebooks that are not demonstrating a specific algorithm applied to a specific type of graph-ml problem e.g, node classification or edge prediction, but rather show how to use one or more methods to solve an interesting use-case should be placed in this new folder. I'm open for suggestions if you find this is not the best folder structure.

I have tested the notebook against the latest StellarGraph release, 0.7.0. Let me know of any issues.

Regards,

P.


adocherty:
* I agree with @annitrolla that the `use-cases` directory would be better without the intermediate `node-classification` directory.

* "Of the original **1037** node features, we are keeping only **204** that are based on a user's attributes and tweet lexicon."  Some justification for this would improve understanding of what we are doing.

* Having to add an axis back again for the targets is a little clunky:
```
annotated_user_features = annotated_users.drop(columns=['hate'])
annotated_user_targets = annotated_users['hate']
print(annotated_user_targets.value_counts())
...
train_targets = train_targets.values
train_targets = train_targets[...,np.newaxis]
test_targets = test_targets.values
test_targets = test_targets[...,np.newaxis]
```

  The reason this is needed is that `annotated_users['hate']` is a `pd.Series` not a `pd.DataFrame`. By having double brackets for the selection we get a `pd.DataFrame` and don't have to add the axis back:
```
annotated_user_targets = annotated_users[['hate']]
print(annotated_user_targets.hate.value_counts())
...
```
Note the change to printing the value counts.  Also, the train_targets can be kept as a DataFrame and used in the flow method, except for calculating the class weights does require using `train_targets.values`.

* "Train the model, keeping track of its loss and accuracy on the training set, and its performance on the validation set during the training (e.g., for early stopping), "  we aren't using early stopping here, so perhaps remove it.

* The table showing FPR and TPR at the end is hard to interpret. Giving approximate numbers would be easier, for example:
```
print("At 1% FPR, GraphSAGE TPR={:.3f}, LR TPR={:.3f}".format(np.interp(0.01, fpr, tpr), np.interp(0.01, fpr_lr, tpr_lr)))
```
  You could also plot these as labelled points on the ROC curve.

PantelisElinas:
Hi @adocherty and @annitrolla 

I have tackled all the issues you raised. Can you have another look?

Thank you!

annitrolla:
Looks good:
found two typos:
 the **totail** number of nodes 
 and the **reamining** 85%

PR #431, opened 2019-06-21
add shellcheck to pipeline + format code

TheDen:
PR:

* Adds [shellcheck](https://github.com/koalaman/shellcheck) to the CI
* Severity level is at default (`style`)
* PR also formats the code so the build passes the `shellcheck` step

Since it is trivial to add unsafe shell code, this enforces best practices when new shellcode is added.


TheDen:
Ah, it means I made sure to format the shellcode in this PR so it passes `shellcheck` in the CI 👍

PR #443, opened 2019-07-02
remove docker cloud image cleanup

TheDen:
PR:

* Removes the cleanup of auxiliary docker images
* This will be done outside the pipeline


TheDen:
Yep, it's more reliable to automatically clean it outside and it won't take up CI time

PR #446, opened 2019-07-09
Restricted tensorflow to less than version 2.0

geoffj-d61:
Tested that `pip install stellar graph`, `pip install stellar graph[demos]` and `python setup.py develop` all install tensor flow < 2.0 under a Python 3.7 environment.


adocherty:
This looks good to me! Thanks for fixing the bug 👍 

Just a reminder to also add reviewers as assignees to the PR, so it shows up in ZenHub when we view our own tasks.

Issue #425, opened 2019-06-14
GraphSAGE and GAT/GCN predictions have different shapes


adocherty:
**Describe the bug**

When using a GraphSAGE model the outputs have an extra dimension before the feature dimension that captures the number of head graphs there are (1 for node inference, 2 for link classification).

For example:
```
graphsage_model = GraphSAGE(
    layer_sizes=[32, 32],
    generator=train_gen,
)
>> x_inp, x_out = graphsage_model.build(flatten_output=False)

>> K.int_shape(x_out)
(None, 1, 32)
```

When the `flatten_output=True` argument is specified this extra dimension is flattened:
```
>> x_inp, x_out = graphsage_model.build(flatten_output=True)

>> K.int_shape(x_out)
(None, 32)
```

However, we cannot use this flag with link prediction.

When using a GAT/GCN model the outputs have a single batch dimension and do not have an extra head graph dimension:

For example:
```
gcn = GCN(
    layer_sizes=[16, 16],
    activations=['relu', 'relu'],
    generator=generator,
    )
>> x_inp, x_out = gcn.node_model()

>> K.int_shape(x_out)
(1, None, 16)
```

For single feature outputs we have the following situation:

GraphSAGE must have outputs divisible by 2 so this doesn't work.

For GCN/GAT we have an extra dimension:
```
>> gcn = GCN(
    layer_sizes=[16, 1],
    activations=['relu', 'relu'],
    generator=generator,
    )
>> x_inp, x_out = gcn.node_model()

>> K.int_shape(x_out)
(1, None, 1)
```

This means that the output needs to be flattened before feeding a vector of binary labels to the `.fit` method:
```
>> x_out_flat = layers.Reshape((-1,))(x_out)

>> K.int_shape(x_out)
(1, None)
```

**Expected behavior**
The output shapes from the GraphSAGE and GCN/GAT methods should be closer (apart from the required singleton batch dimension of the full-batch methods). 

Specifically, we should have sensible default behaviour for GraphSAGE and not have to specify the `flatten_output=True` argument:
```
>> x_inp, x_out = graphsage_model.build()

>> K.int_shape(x_out)
(None, 32)
``` 

We can introduce the `flatten_output=True` option to GCN to flatten binary prediction outputs:
```
>> gcn = GCN(
    layer_sizes=[16, 1],
    activations=['relu', 'relu'],
    generator=generator,
    )
>> x_inp, x_out = gcn.node_model(flatten_output=True)

[...]

PR #447, opened 2019-07-11
Bugfix/flatten output=true

youph:
This partially addresses issue #425 
- The "ip" edge embedding method for link inference is fixed to have the output shape `(None, 1)` regardless of the shape of the input tensors for (src, dst) nodes. This is done by adding a line `out = Reshape((1,))(out)`. So now both code snippets below work for link prediction, producing same shape for `prediction` tensor:
```
x_inp, x_out = graphsage.build(flatten_output=False)
prediction = link_classification(output_dim=1, output_act="relu", edge_embedding_method='ip')(x_out)
K.int_shape(x_out[0])    # gives (None, 1, 20)
K.int_shape(prediction)  # gives (None, 1)
```
and 
```
x_inp, x_out = graphsage.build(flatten_output=True)
prediction = link_classification(output_dim=1, output_act="relu", edge_embedding_method='ip')(x_out)
K.int_shape(x_out[0])    # gives (None, 20)
K.int_shape(prediction)  # gives (None, 1)
```
- The default value for `flatten_output` argument for `build` method is changed to `True`, in all model classes. This addresses the comment "Specifically, we should have sensible default behaviour for GraphSAGE and not have to specify the flatten_output=True argument" in #425


adocherty:
Hi,

I don't like the default being `flatten_output=True`. The reason for this is that any structure in the predictions is lost.  If users specify anything other than a vector of predictions for each node, this is flattened. 

For GraphSAGE/HinSAGE this flag is needed because there is an extra neighbourhood dimension used to enumerate the neighbours as the tensors are passes through the model.  I've changed the code to always remove this dimension in the output tensors, as long as this dimension is one.

I've also removed the `flatten_output` option altogether. I don't think there is any real justification for keeping it around.

Let me know what you think of these proposed changes!

A.

stellargraph/layer/graphsage.py
```diff
@@ -690,12 +690,12 @@ def node_model(self, flatten_output=False):
         # Output from GraphSAGE model
         x_out = self(x_inp)
 
-        if flatten_output:
-            x_out = Reshape((-1,))(x_out)
+        if K.int_shape(x_out)[1] == 1:
```
youph: please add a comment here explaining this "special case"

youph: Inconsistency with HinSAGE: in HinSAGE, removal of the "neighbourhood" dimension is done in `HinSAGE.__call__()` method (lines 382-385 in `hinsage.py`), while here it is done in `GraphSAGE.node_model()` method. Perhaps this should be moved to `GraphSAGE.__call__()` as well.

Also, a comment should be added to this operation, similar to that in HinSAGE

stellargraph/layer/hinsage.py
```diff
@@ -379,6 +379,14 @@ def apply_layer(x: List, layer: int):
             h_layer = apply_layer(h_layer, layer)
             self.layer_tensors.append(h_layer)
 
+        print(h_layer)
```
youph: I removed the print statements

stellargraph/layer/hinsage.py
```diff
@@ -423,16 +431,16 @@ def get_shape(stree, cnode, level=0):
 
         return [input_shapes[ii] for ii in range(len(self.subtree_schema))]
 
-    def build(self, flatten_output=False):
+    def build(self):
         """
         Builds a HinSAGE model for node or link/node pair prediction, depending on the generator used to construct
         the model (whether it is a node or link/node pair generator).
 
         Args:
```
youph: I removed the Args description, as there are no args anymore

PR #448, opened 2019-07-12
Add demo for hateful twitter user dataset.

sktzwhj:
Hi @annitrolla and @habiba-h .

This PR adds the interpretability demo for hateful twitter dataset. This shows that we are able to compute the saliency maps for relatively large datasets like hateful twitter users. We may refine it a bit. 

Comments and suggestions are appreciated. 

Cheers,
-Huijun


sktzwhj:
Hi @annitrolla @habiba-h and @nhungphnguyen , 

I will modify the notebook soon. Close this PR for now and will re-invite all of you to review later.

-Huijun

sktzwhj:
> The notebook runs smoothly without any problems and is a nice guide how to apply saliency maps. If this notebook for the internal use only, then it can be approved as it is. However, if the plan is to publish it as a demo or use-case in the library, then the notebook is under-commented and requires to build a story. Also it would be good to eliminate the duplication with the Pantelis's hateful-twitter use-case by just loading the model from previous notebook or give a link to his notebook and create a minimal code for one single model. Next, I list the comments cell-by-cell:
> 
> A suggestion: cell 6 in data_cleaning function replace several `feat.columns.str.contains` with single one: `feat.drop(feat.columns[feat.columns.str.contains("is_|_glove|c_|sentiment")], axis=1, inplace=True)`
> The new function, therefore, would be:
> 
> ```
> def data_cleaning(feat):
>     feat = feat.drop(columns=["hate_neigh", "normal_neigh"])
>     
>     # Convert target values in hate column from strings to integers (0,1,2)
>     feat['hate'] = np.where(feat['hate']=='hateful', 1, np.where(feat['hate']=='normal', 0, 2))
>     
>     # missing information
>     number_of_missing = feat.isnull().sum()
>     number_of_missing[number_of_missing!=0]
>     
>     # Replace NA with 0
>     feat.fillna(0, inplace=True)
> 
>     # droping info about suspension and deletion as it is should not be use din the predictive model
>     feat.drop(feat.columns[feat.columns.str.contains("is_|_glove|c_|sentiment")], axis=1, inplace=True)
> 
>     # drop hashtag feature
>     feat.drop(['hashtags'], axis=1, inplace=True)
> 
>     # Drop centrality based measures
>     feat.drop(columns=['betweenness', 'eigenvector', 'in_degree', 'out_degree'], inplace=True)
>     
>     feat.drop(columns=['created_at'], inplace=True)
>     
>     return feat
> ```
> 
> 1. The notebook is about interpretation. I think we can safely delete cells: 13,14,15.
Deleted. I have also removed several other cells which are not quite relevant.

> 2. Does this interpretability notebook will go to demos in our library as a standalone notebook? If yes, then I’d suggest to keep the description and cells minimal (just quickly train the model without explanations and the cells that can specify several models, e.g. 72, 74 etc) to avoid much of the duplication as we can refer to the use-case in our library for the details. Another option is to save the model in Pantelis’s use-case and load it here starting from cell 93 and link to his notebook.   If the goal of this notebook is different, then you can safely ignore this comment :).

I tend to do the first and have deleted the unnecessary cells.

> 3. cell 151: You comment that you take all  nodes classified as hateful, but then just specify idx=2. I think it would be better if you provide a bit more explanation before this cell .
Description added. 

> 4. before cell 152: “Our sparse support for saliency maps” - rephrase e.g. Our support for sparse matrices <…>
Thanks. Fixed.

> 5. It is also not clear in cell 152 you say that “We then calculate the importance of the features for each node in the graph” but in the  call `node_feature_importance = int_saliency.get_integrated_node_masks(target_idx, 0)` you specify the target_idx. May be better to explain more about that.

Explanation added.
> 6. Provide  comments/descriptions before the cells 153,154 , 156 to guide a reader.
> 7. cell 155: can we truncate the output limiting it to 5 or 6 lines? Also, comment what exactly this output is.
> 8. the same comment for the cell 157
> 9. the rest of the notebook lacks the descriptions of the steps as well as the conclusion.

To address the above comments regarding descriptions of cells, I have added more now to clarify.

> 10. Also, should we add a better explanation for the results?

I think this can go to future work as it requires non-trivial work to look into the original tweets and maybe we require good visualization to deal with such a large graph. For this demo, I made it mainly to show that we have the capability to efficiently compute the saliency maps for relatively large graphs.

annitrolla:
Looks much better. Make sure you delete the NOTE at the end when merging into library.

Issue #297, opened 2018-11-29
Change StellarGraph class to NOT inherit from networkx.Graph() class


youph:
### Description
Currently, `StellarGraph` class inherits from `networkx.Graph` class, inheriting all of its methods. While this allows us to use the networkx methods for Stellargraph objects, the problem is that many of these methods (e.g., `.copy()`) produce a `networkx.Graph` object as a result, i.e., applying these methods to `StellarGraph`-typed objects changes the type of the object to `networkx.Graph`, losing the information specific to the original `StellarGraph` object (e.g., node features). 

To avoid this confusing behaviour, we need to make `StellarGraph` a `class StellarGraph(object)`, with constructor taking two arguments: a networkx graph, and an optional pandas dataframe with node attributes. We also need to implement the methods that are used by explorers, etc., e.g., `.nodes()`, `.edges()`, `.neighbours()`, `.copy()`, etc. While this requires extra work to implement, the advantage is that the result of these methods would be of the same type, `StellarGraph`, with all the internal information retained.

### Done Checklist (Development)
- [x] Produced code for required functionality
- [x] Branch and Pull Request build on CI
- [x] Branch and Pull Request pass unit tests on CI
- [ ] Version number reflects new status
- [x] Peer Code Review Performed
- [x] Code well commented
- [ ] Documentation in repo
- [ ] Team demo

youph:
@adocherty We've discussed this a while ago, creating an issue to address in this release.

youph:
@adocherty @PantelisElinas this issue has been in the backlog for a while. Let's discuss to see if this is still relevant, and when we should address it.

geoffj-d61:
@youph @adocherty I figured I'd have a look at this as well whilst I'm thinking about the whole StellarGraph class implementation framework (unless anyone has objections). Subsequent integration of the proposed new classes into the system will require me to understand just what functionality is currently used, anyway.

geoffj-d61:
Hi @adocherty , I have a local branch (not yet pushed) that makes StellarGraph and StellarDiGraph pure interfaces (in the Java or C++ concepts), and makes StellarGraphBase (renamed to NetworkXStellarGraph) a functional but hidden implementation. This currently passes all the unit tests, but I have forgotten how to manually run the demo tests. Can you please remind me how?

[...]

PR #451, opened 2019-07-18
Try to ensure all edges and nodes get types

geoffj-d61:
Basic behaviour:
- define global default type labels, which can be overridden in `StellarGraphBase`
- define private getters that set the default type if not present
- replace all label fetches, e.g. `node[self._node_type_attr]`, with calls to getters

Warning: It seems to my uneducated eye that graph methods not directly defined on `StellarGraphBase` are delegated to `NetworkX`. Ideally, such delegated calls should mark the graph as being dirty, and there should be some explicit validation call that gives all nodes and edges types. Currently `create_graph_schema` plays this role, but we cannot ensure that it is always called.


adocherty:
Hi @geoffj-d61 ,

You are right that all calls to graph methods are delegated to NetworkX.  In fact we don't support  modifications to the graph that change nodes and they will fail.  We have intended to make the `StellarGraph` object immutable in some way, but have not gotten around to doing so. This is because we have been planning to switch from NetworkX to our own graph class for some time.  This is currently being planned, see issue #297 & https://github.com/stellargraph/stellar-research/issues/61.

I'm OK with these changes, but I'd prefer to prevent changes to the graph in the StellarGraph object. I probably should have made these points in the original issue!

Best regards,
Andrew

geoffj-d61:
@adocherty, no problem - it has been a useful learning exercise for me. From a design point of view, my personal feeling would be to wait for a redesign/rewrite of the architecture anyway. I can kill this pull request if you like.

adocherty:
Let's merge this PR: you've done the work and fixed the problem, so why not?   Either way we should move forward with investigating a new graph class.

Issue #464, opened 2019-07-29
Directed SampledBFS walks


PantelisElinas:
### Description

To support directed GraphSAGE, we need to first implement directed SampleBFS walks.

### Done Checklist (Development)
- [x] Produced code for required functionality
- [x] Branch and Pull Request build on CI
- [x] Branch and Pull Request pass unit tests on CI
- [x] Branch and Pull Request pass integration tests on CI
- [x] Peer Code Review Performed
- [x] Code well commented
- [ ] Documentation in repo
- [ ] CHANGELOG.md updated

PR #476, opened 2019-08-14
Directed BFS sampling

geoffj-d61:
The purpose of ticket #464 is to create a version of breadth-first graph sampling that can work on directed graphs, as a prelude to implementing directed GraphSAGE. There is a document which covers BFS and GraphSAGE on the shared drive under (approx.) IA > Team folders > ML > Graph ML library. This document explains the choice of data structures.


stellargraph/data/explorer.py
```diff
@@ -107,6 +81,114 @@ def run(self, **kwargs):
         """
         raise NotImplementedError
 
+    def _raise_error(self, msg):
+        full = "({}) " + msg
+        raise ValueError(full.format(type(self).__name__))
```
huonw: This would be better as:

```suggestion
        raise ValueError("({}) {}".format(type(self).__name__, msg))
```

As is, there will be weird errors if `msg` contains a `{` because the `format` will try to interpret that.

stellargraph/data/explorer.py
```diff
@@ -107,6 +81,114 @@ def run(self, **kwargs):
...
+            self._raise_error(
+                "The random number generator seed value, seed, should be non-negative integer or None."
+            )
+        # seed the random number generator
+        return random.Random(seed)
+
+    # For neighbourhood sampling
+    def _check_sizes(self, n_size):
+        err_msg = "The neighbourhood size must be a list of non-negative integers."
+        if n_size is None or type(n_size) != list:
```
huonw: Does this need to be a list? Can it be a subclass of a list? I think that should be accepted, e.g. https://docs.python.org/3/library/functions.html#type

> The `isinstance()` built-in function is recommended for testing the type of an object, because it takes subclasses into account.

```suggestion
        if n_size is None or not isinstance(n_size, list):
```

(This also applies to other `type(...)` calls in this code.)

---

In addition, the `None` check is redundant (`type(None)` is `<class 'NoneType'>` (`!= list`), and it isn't a subclass of `list` either):

```suggestion
        if not isinstance(n_size, list):
```

Finally, can this be a tuple? I would expect so? Do we need to check the type here, vs. just checking its length and then iterating?

stellargraph/data/explorer.py
```diff
@@ -107,6 +81,114 @@ def run(self, **kwargs):
...
+    Base class for exploring graphs with heterogeneous edge types.
+    """
+
+    def __init__(self, graph, graph_schema=None, seed=None):
+        super().__init__(graph, graph_schema, seed)
+
+        # Create a dict of adjacency lists per edge type, for faster neighbour sampling from graph in SampledHeteroBFS:
+        # TODO: this could be better placed inside StellarGraph class
+        edge_types = self.graph_schema.edge_types
+        self.adj = {et: defaultdict(lambda: [None]) for et in edge_types}
```
huonw: If this represents no neighbours, should this be

```suggestion
        self.adj = {et: defaultdict(lambda: []) for et in edge_types}
```

If not, why `[None]` exactly?

stellargraph/data/explorer.py
```diff
@@ -107,6 +81,114 @@ def run(self, **kwargs):
...
+
+        for n1, nbrdict in graph.adjacency():
+            for et in edge_types:
+                neigh_et = [
+                    n2
+                    for n2, nkeys in nbrdict.items()
+                    for k in iter(nkeys)
```
```suggestion
                    for k in nkeys
```

I think?

stellargraph/data/explorer.py
```diff
@@ -774,45 +599,42 @@ def run(self, nodes=None, n=1, n_size=None, seed=None):
...
+                    else:
+                        # sample with replacement
+                        neighbours = [
+                            rs.choice(neighbours) for _ in range(n_size[cur_depth])
+                        ]
 
-                        # add them to the back of the queue
-                        q.extend([(sampled_node, depth) for sampled_node in neighbours])
+                    # add them to the back of the queue
+                    q.extend([(sampled_node, depth) for sampled_node in neighbours])
```
huonw: The intermediate list isn't necessary, if you're refactoring this code:

```suggestion
                    q.extend((sampled_node, depth) for sampled_node in neighbours)
```

stellargraph/data/explorer.py
```diff
@@ -774,45 +599,42 @@ def run(self, nodes=None, n=1, n_size=None, seed=None):
...
+                        continue
+                    neighbours = (
+                        self.neighbors(cur_node) if cur_node is not None else []
+                    )
+                    if len(neighbours) == 0:
+                        # Either node is unconnected or is in directed graph with no out-nodes.
+                        neighbours = [None] * n_size[cur_depth]
+                    else:
+                        # sample with replacement
+                        neighbours = [
```
huonw: This can become a generator too:

```python
                        neighbours = (
                            rs.choice(neighbours) for _ in range(n_size[cur_depth])
                        )                         
```

stellargraph/data/explorer.py
```diff
@@ -774,45 +599,42 @@ def run(self, nodes=None, n=1, n_size=None, seed=None):
...
+
+                    # consider the subgraph up to and including max_hops from root node
+                    if depth > max_hops:
+                        continue
+                    neighbours = (
+                        self.neighbors(cur_node) if cur_node is not None else []
+                    )
+                    if len(neighbours) == 0:
+                        # Either node is unconnected or is in directed graph with no out-nodes.
+                        neighbours = [None] * n_size[cur_depth]
```
huonw: Why does this new code need to include all these `None`?

stellargraph/data/explorer.py
```diff
@@ -1021,93 +756,137 @@ def _check_parameter_values(self, nodes, n, n_size, graph_schema, seed):
...
-                )
-            )
+        self._check_sizes(n_size)
 
-        if type(n) != int:
-            raise ValueError(
-                "({}) The number of walks per root node, n, should be integer type.".format(
-                    type(self).__name__
-                )
+        if graph_schema is not None and type(graph_schema) is not GraphSchema:
```
huonw: I think this is already checked by `GraphWalk.__init__`?

stellargraph/data/explorer.py
```diff
@@ -774,45 +599,42 @@ def run(self, nodes=None, n=1, n_size=None, seed=None):
...
-                    if depth <= d:
-                        neighbours = self.neighbors(frontier[0])
-                        if len(neighbours) == 0:
-                            break
-                        else:
-                            # sample with replacement
-                            neighbours = [
-                                rs.choice(neighbours) for _ in range(n_size[depth - 1])
-                            ]
+                    cur_node, cur_depth = q.pop(0)
```
huonw: `pop(0)` on a list likely makes this loop quadratic in its maximum length (or something along those lines). https://docs.python.org/3/library/collections.html#collections.deque would be better, while you're touching this code.

stellargraph/data/explorer.py
```diff
@@ -1021,93 +756,137 @@ def _check_parameter_values(self, nodes, n, n_size, graph_schema, seed):
...
+
+                [node out_i  out_i.in_j  out_i.in_j.in_k ...]
+        """
+        rs = self._check_parameter_values(nodes, n, in_size, out_size, seed)
+        max_hops = len(in_size)
+        max_slots = 2 ** (max_hops + 1) - 1
+
+        samples = []
+
+        for node in nodes:  # iterate over root nodes
```
huonw: There's a lot of commonality here with `SampledBreadthFirstWalk`. Do you think it's reasonable to try to factor that out?

stellargraph/data/explorer.py
```diff
@@ -1021,93 +756,137 @@ def _check_parameter_values(self, nodes, n, n_size, graph_schema, seed):
...
+                 [in_1.in_1.in_1...in_n.in_p.in_r] [in_1.in_1.out_1...in_n.in_p.out_s] ...
+                 ...]
+
+            where a single, undirected walk might be, for example:
+
+                [node out_i  out_i.in_j  out_i.in_j.in_k ...]
+        """
+        rs = self._check_parameter_values(nodes, n, in_size, out_size, seed)
+        max_hops = len(in_size)
+        max_slots = 2 ** (max_hops + 1) - 1
```
huonw: What's the meaning of this? What are the slots representing?

geoffj-d61:
> My feedback is mostly in the form of questions, because I don't have much of a picture of what the overall goal/design is here.
> 
> I haven't looked at the tests at all.

Firstly, did you read the google doc I wrote? If not, that should give you a better idea of what slots are and why they are needed, and what data structures I've designed.

Also, all the tests that say, hey this must be a list - they are not my tests! I agree with you, most things should be iterables not concrete lists. We plan to rationalise the library, and I would like such things to form part of the design.

stellargraph/data/explorer.py
geoffj-d61: Yes and no - the output data structure is different and the sequence of sampling is different. But on the other hand, yes, my preference is to refactor code duplication/commonality into modular reuse. Sadly, the code base is a complete mess of copy-pasta and lack of design.

stellargraph/mapper/node_mappers.py
```diff
@@ -254,13 +254,6 @@ def sample_features(self, head_nodes, sampling_schema):
         # The number of samples for each head node (not including itself)
         num_full_samples = np.sum(np.cumprod(self.num_samples))
 
-        # Isolated nodes will return only themselves in the sample list
-        # let's correct for this by padding with None (the dummy node ID)
-        node_samples = [
```
youph: why has this been removed?

stellargraph/data/explorer.py
```diff
@@ -51,42 +49,18 @@ def __init__(self, graph, graph_schema=None, seed=None):
...
             self.graph_schema = self.graph.create_graph_schema(create_type_maps=True)
         else:
             self.graph_schema = graph_schema
 
-        if self.graph_schema is not None and type(self.graph_schema) is not GraphSchema:
-            raise ValueError(
-                "({}) The parameter graph_schema should be either None or of type GraphSchema.".format(
-                    type(self).__name__
-                )
+        if type(self.graph_schema) is not GraphSchema:
```
youph: can this ever break, e.g., if after line 55 self.graph_schema is still None? If yes, this should be reverted to what it was before.

stellargraph/data/explorer.py
```diff
@@ -51,42 +49,18 @@ def __init__(self, graph, graph_schema=None, seed=None):
...
+            self._raise_error(
+                "The parameter graph_schema should be either None or of type GraphSchema."
             )
 
-        # Create a dict of adjacency lists per edge type, for faster neighbour sampling from graph in SampledHeteroBFS:
```
youph: any reason for moving this?

stellargraph/data/explorer.py
```diff
@@ -107,6 +81,114 @@ def run(self, **kwargs):
         """
         raise NotImplementedError
 
+    def _raise_error(self, msg):
+        full = "({}) " + msg
+        raise ValueError(full.format(type(self).__name__))
+
+    def _check_parameter_values(self, nodes=None, n=0, length=0, seed=None):
```
youph: wouldn't it be nicer to define all the internal methods before defining methods like `run`?

stellargraph/data/explorer.py
```diff
@@ -107,6 +81,114 @@ def run(self, **kwargs):
...
+        if type(seed) != int:
+            self._raise_error(
+                "The random number generator seed value, seed, should be integer type or None."
+            )
+        if seed < 0:
+            self._raise_error(
+                "The random number generator seed value, seed, should be non-negative integer or None."
+            )
+        # seed the random number generator
+        return random.Random(seed)
```
youph: I don't think it's a good idea to have this return a random state: the purpose of this method is checking parameter values, so it should just do that. We can have another method for initialising the random state.

stellargraph/data/explorer.py
youph: Good idea to put common checks into the base class!

stellargraph/mapper/link_mappers.py
```diff
@@ -366,13 +366,6 @@ def get_levels(loc, lsize, samples_per_hop, walks):
         for hns in zip(*head_links):
             node_samples = self.sampler.run(nodes=hns, n=1, n_size=self.num_samples)
 
-            # Isolated nodes will return only themselves in the sample list
```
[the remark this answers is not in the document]
geoffj-d61: Partly because it breaks a test case where someone wraps a NetworkX DiGraph with StellarGraph. And partly because it is a bad idea to have a special representation in one class, but correct for it in a completely different class. Especially when the correction is simply to put the None's back that were left off.

stellargraph/mapper/node_mappers.py
geoffj-d61: See above.

stellargraph/data/explorer.py
geoffj-d61: Only if _create_graph_schema() can ever return None, which should be an error.

stellargraph/data/explorer.py
geoffj-d61: Moved it into its own base class, since it is only used in one place.

stellargraph/data/explorer.py
geoffj-d61: Styles vary - some people like to define all helper functions before first use (a la Pascal), and some like to define them after first use. I don't think we have a standard yet.

stellargraph/data/explorer.py
geoffj-d61: I agree, because it breaks the Single Responsibility principle. However, it also removes a lot of explicit code reuse. I could add one more method just for obtaining the random seed.

stellargraph/mapper/link_mappers.py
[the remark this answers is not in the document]
youph: Ok, let's talk more :) Are they not being left off anymore?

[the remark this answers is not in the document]
geoffj-d61: And yes, the sampler class in data/explorer.py explicitly inserts the None's.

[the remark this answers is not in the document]
geoffj-d61: At the point of use, if a neighbourhood is empty we still need it full of None's, in order to get the size right. We can therefore either fill it full of None at the point of sampling, or try to reduce space by leaving them off, but at the cost of logic having to reconstruct them later, at a different point in code. Seems easier to do the simple thing, so that the output is always guaranteed to be of the correct size and shape.

stellargraph/data/explorer.py
youph: Let's try and stick with the Single Responsibility principle.

stellargraph/data/explorer.py
youph: But this required creating a new base class for heterogeneous walks. I think this is not necessary, so I'd prefer to have only one base class here. Once the StellarGraph class is refactored, we'll either get rid of explicit use of adjacency lists per edge type here, or will use them in all descendent walk classes.

stellargraph/mapper/link_mappers.py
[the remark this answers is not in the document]
youph: Makes sense, but let me think about edge cases...

stellargraph/data/explorer.py
geoffj-d61: Good catch. Pointless constructing a temporary, new string anyway, since format will do that for us.

stellargraph/data/explorer.py
geoffj-d61: Another good catch (but actually not my code or my design - the basic deign assumes lists everywhere; erk).

stellargraph/data/explorer.py
geoffj-d61: You will need to ask the author; I understood it once but forget. I do recall it is shorthand for the process of inserting a list of None for a missing key (yes, the None does need to be there).

stellargraph/data/explorer.py
geoffj-d61: I'll try it your way, I think you are correct.

stellargraph/data/explorer.py
geoffj-d61: Didn't know that one!

stellargraph/data/explorer.py
geoffj-d61: E.g. sample exactly 10 nodes from the out-node neighbourhood; oh dear, that neighbourhood is empty. Q: How do we get exactly 10? A: Fill the result with 10 None values (which will later get mapped to special, zero-valued features).

stellargraph/data/explorer.py
youph: What's wrong with expecting lists? What else would you pass there?

stellargraph/data/explorer.py
geoffj-d61: Yes.

stellargraph/data/explorer.py
geoffj-d61: The general principle would be an iterator (if we didn't care about length). If you think in terms of java, Python's list is a concrete class, which breaks the D in SOLID (dependency inversion, i.e. depend on abstractions rather than concrete classes).

stellargraph/mapper/link_mappers.py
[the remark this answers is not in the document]
youph: BTW: looks like we can always use `SampledHeterogeneousBreadthFirstWalk` instead of `SampledBreadthFirstWalk`? It already does the right thing filling None's in place of missing nodes. In case of homogeneous graphs, there'll only be one node type to loop through.

stellargraph/data/explorer.py
youph: Yes, a comment would be good

geoffj-d61: Seriously!? I wrote a big comment! Ok, I will try to add more explanation (but it is explained in the design document).

stellargraph/data/explorer.py
```diff
@@ -1021,93 +755,137 @@ def _check_parameter_values(self, nodes, n, n_size, graph_schema, seed):
...
+                    q.extend(
+                        [(sampled_node, depth, slot) for sampled_node in neighbours]
+                    )
+
+                # finished multi-hop neighbourhood sampling
+                samples.append(sample)
+
+        return samples
+
+    def _sample_neighbours(self, rs, node, idx, size):
```
youph: This is a nice general method. Do you think it should replace `GraphWalk.neighbors`?
Also, could you please add a docstring explaining the args, especially `idx`? It's clear now, but I'm sure we can forget in the future.

stellargraph/data/explorer.py
youph: @geoffj-d61 Yes! Remember, this is an open-source code base, while the doc you wrote is only visible to us.

geoffj-d61: Ok, have added a comment and addressed some other review issues. See latest push.

stellargraph/mapper/link_mappers.py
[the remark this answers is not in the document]
geoffj-d61: We could - it was interesting to find, after I had included the explicit None's, that 'hetero' did the same thing. I guess the thing to think about (I have no opinion here) is whether we care about the extra overhead involved in repeatedly looking up a single type in a dictionary? If we do, then we keep two classes, one optimised for homogeneous and one not. If we don't, then we could just keep the more general class.

stellargraph/data/explorer.py
geoffj-d61: Yes, fair point about public vs private.

stellargraph/data/explorer.py
geoffj-d61: I'll look into this now; I agree in principle, if pop(0) on a list keeps creating a new list.

stellargraph/data/explorer.py
geoffj-d61: I think `in_edges` and `out_edges` methods only exist in DiGraph.

youph: We can either check if the graph is directed inside this method (the easier solution?), or make all graphs directed.

geoffj-d61: My preference is that all graphs are directed by default, and the `neighbors()` method gives back nodes from both in and out neighbourhoods. Alternatively, for an undirected graph we could add `in_edges()` and `out_edges()` methods.

youph:
Another question:
- `explorer.py`, line 165: why can't we have `n_size=0`? Setting `n_size=0` is our way to easily switch-off neighbour sampling in GraphSAGE and HinSAGE!

youph:
Another comment:
`explorer.py`, lines 673 and 777: you reuse `GraphWalk._check_parameter_values` method here, but pass `len(n_size)` as `length` argument. This is somewhat confusing, as in the declaration of `GraphWalk._check_parameter_values`, `length` is defined as "Maximum length of walk measured as the number of edges followed from root node" - which is completely irrelevant to what we are in fact passing to it here. One way to fix this would be making the docstring in `GraphWalk._check_parameter_values` more general, to cover all possible meanings of the arguments. Another way would be to revert to walker-specific methods for checking parameter values - less code reuse but easier to understand.

geoffj-d61:
@youph if I reworded "Maximum length of walk measured as the number of edges followed from root node" to be something like "Length of walk; the maximum number of hops from the root node", would that be suitable? Or "The path length n for n-hop walks"? Or give me your own wording.

I agree that "number of edges followed from root node" is misleading; my intent was that a single walk only follows one edge at a time, even if the sampler is allowed to amalgamate multiple walks into one structure.

stellargraph/data/explorer.py
huonw: > What's wrong with expecting lists? What else would you pass there?

Concretely, I could easily imagine passing a tuple here: `(10, 20)` instead of `[10, 20]`.

geoffj-d61:
The build is getting this error: ```E       AttributeError: module 'keras.backend' has no attribute 'tf'``` Runs fine on my tests, is this an issue with buildkite - the maligned test is `tests/layer/test_graph_attention.py`.

geoffj-d61:
Will merge as recommended by @PantelisElinas.

PR #477, opened 2019-08-15
added thread lock acquire and release to OnDemandLinkSequence, to allow multithreading

youph:
This patch, inspired by [this post](https://www.bogotobogo.com/python/Multithread/python_multithreading_Synchronization_Lock_Objects_Acquire_Release.php), appears to have fixed the multithreading issue. The way to use is
```
model.fit_generator(
    train_gen,
    epochs=epochs,
    verbose=1,
    use_multiprocessing=False,
    workers=4,
    shuffle=True,
)
```

@PantelisElinas the speed up is significant on CPU; let's try this on `fractal` using GPU.

### DIFF
--- a/demos/embeddings/embeddings-unsupervised-graphsage-cora.ipynb
+++ b/demos/embeddings/embeddings-unsupervised-graphsage-cora.ipynb
@@ -451,7 +451,7 @@
     "    epochs=epochs,\n",
     "    verbose=1,\n",
     "    use_multiprocessing=False,\n",
-    "    workers=0,\n",
+    "    workers=4,\n",
     "    shuffle=True,\n",
     ")"
    ]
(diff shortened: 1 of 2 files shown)


youph:
> Simple fix, looks good. Do you have the CPU timing results, with and without the fix?

Yes, the training time per epoch on `2.9 GHz Intel Core i7` CPU is:
with `workers=0` (executing the generator on the main thread): 83s/epoch
with `workers=1`: 63s/epoch
with `workers=2`: 38s/epoch
with `workers=4`: 32s/epoch

PantelisElinas:
Hi @youph 

the fix looks good and I have not come across the exception we used to get. Also, here are the timing results from our workstation.

**Times are per epoch** on CPU and GPU using multi-threading. The machine is Intel i9-7960X 2 2.80GHz with NVIDIA GeForce RTX 2080.

CPU:
workers=0: 37s
workers=2: 18s
workers=4: 14s
workers=6: 14s
workers=8: 14s

GPU:
workers=0: 42s
workers=2: 17s
workers=4: 19s
workers=8: 18s

GPU utilisation for the above experiments was 10-15% (very low) but we already know that with GraphSAGE we have a bandwidth problem sending the node features to the GPU. In other words, there is little benefit using the GPU, in fact, the extra overhead of having to send data to the GPU hurts performance.

Regards,

P.

PR #483, opened 2019-09-05
Encapsulate node data

geoffj-d61:
Implemented a module for encapsulating node data in a variety of formats into a common interface.


geoffj-d61:
@huonw Could you please have a look at this and see if I used more idiomatic Python this time. Since it started as a copy-and-paste of the edge data code, I expect some of your previous comments/complaints to still be valid here - please make them again, as necessary, because this is a cleaner version and I need to retrofit some changes to the edge data module. I'd rather fix this one to your satisfaction first!

adocherty:
As far as I understand the code you have the following principles:
• Node data can be passed to `to_node_data` in different formats such as as a Pandas DataFrame, np.ndarray, dictionary of dataframes etc.
• The node data used is node ID and node type.
• The data will not be copied but will be stored (in the appropriate EdgeData object) as passed and can be iterated over node (row) at a time.

I have some questions about the goal of this PR, and I've added a comment about this to the linked issue. Additionally how this NodeData encapsulation work with the EdgeData encapsulation of PR #482, in particular If we encapsulate a list of nodes without IDs, how would the EdgeData object use these node ids in the edgelist?  Additionally, how would we be able to assign features to the nodes and look up the feature for the corresponding node ID?

I am thinking that initially we should be more rigid in what we allow to specify the graph and the features of the nodes and edges. I've added my ideas for this to the issue for the higher-level graph ingestion here: 
https://github.com/stellargraph/stellargraph-ml-private-issues/issues/144

Issue #484, opened 2019-09-06
Implement PPNP and APPN graph ml algorithms


kieranricardo:
### Description

Feature request to implement the Personalised Propagation of Neural Predictions (PPNP) and Approximate PPNP (APPNP) algorithms described in: 
    https://openreview.net/pdf?id=H1gL-2A9Ym

### User Story
As a user I want to be able to use the PPNP and APPNP algorithms with the stellargaph library so that I can perform node classification on large graphs. 

### Done Checklist (Development)
- [X] Assumptions of the user story met
- [X] Produced code for required functionality
- [ ] Branch and Pull Request build on CI
- [ ] Branch and Pull Request pass unit tests on CI
- [ ] Branch and Pull Request pass integration tests on CI
- [ ] Version number reflects new status
- [ ] Peer Code Review Performed
- [ ] Code well commented
- [ ] Documentation on Google Docs
- [ ] Documentation in repo
- [ ] CHANGELOG.md updated
- [ ] Team demo
- [ ] Mini-meetup talk
- [ ] Stakeholder sign-off

PR #485, opened 2019-09-06
Added PPNP and APPNP algorithms + unit tests.

kieranricardo:
#484


tests/layer/test_appnp.py
```diff
@@ -0,0 +1,76 @@
...
+    adj = nx.to_numpy_array(G)[None, :, :]
+    n_nodes = features.shape[0]
+
+    nodes = G.nodes()
+    node_features = pd.DataFrame.from_dict(
+        {n: f for n, f in zip(nodes, features)}, orient="index"
+    )
+    G = StellarGraph(G, node_features=node_features)
+
+    generator = FullBatchNodeGenerator(G, sparse=False, method="none")
```
PantelisElinas: Shouldn't this be method="gcn"?

tests/layer/test_appnp.py
```diff
@@ -0,0 +1,76 @@
...
+    assert preds_1.shape == (1, 2, 2)
+
+    # Check fit_generator method
+    preds_2 = model.predict_generator(generator.flow(["a", "b"]))
+    assert preds_2.shape == (1, 2, 2)
+
+    assert preds_1 == pytest.approx(preds_2)
+
+
+def test_PPNP_apply_sparse():
```
PantelisElinas: Shouldn't this be called test_APPNP_apply_sparse?

tests/layer/test_appnp.py
```diff
@@ -0,0 +1,76 @@
+from stellargraph.layer.appnp import *
```
PantelisElinas: You have no tests for edge cases. What if the generator is of the wrong type? What if approx_iter is a negative number or a float? What if layer_sizes is of different length than activations. What if the last layer's size is not equal to the number of classes to predict,  and so on?

You need to add more tests!

tests/layer/test_ppnp.py
```diff
@@ -0,0 +1,76 @@
+from stellargraph.layer.ppnp import *
```
PantelisElinas: See my comment for the lack of tests for edge cases for APPNP. The same applies here for PPNP.

tests/layer/test_ppnp.py
```diff
@@ -0,0 +1,76 @@
...
+    adj = nx.to_numpy_array(G)[None, :, :]
+    n_nodes = features.shape[0]
+
+    nodes = G.nodes()
+    node_features = pd.DataFrame.from_dict(
+        {n: f for n, f in zip(nodes, features)}, orient="index"
+    )
+    G = StellarGraph(G, node_features=node_features)
+
+    generator = FullBatchNodeGenerator(G, sparse=False, method="none")
```
PantelisElinas: You never check the case passing method="gcn".

PantelisElinas (changes requested):
Hi @kieranricardo 

thank you for this implementation of PPNP and APPNP algorithms. Overall it looks very good. I have made some suggestions regarding some small improvements that I would like you to address before we can accept your pull request. 

Also, here are a few comments about your demo Jupyter notebook,

- At the start of the notebook, please add a brief description of the task, algorithms (PPNP, APPNP), and a reference to the paper. I suggest you use the sgc demo notebook as a guide.
- Cell 16, the last layer has to have the same size as the number of classes since you explicitly add a SoftMax layer during model creation following the last NN layer. However, there is nowhere any indication or check in the code to inform the user of this ( I can see it is mentioned in the doc string but there is a good chance most users won't see this at first). Same holds of the APPNP model in Cell 27. It might be a good idea to specify the model using something like 
 ```ppnp = PPNP([64, 64, train_targets.shape[-1]], ['relu', 'relu', 'relu'],  generator, transport_probability=0.1,  dropout=0.5, kernel_regularizer=keras.regularizers.l2(0.001))```
 - Cell 29, you are trying to load the wrong model file, you have 'logs/best_model.h5' whereas it should be 'logs/best_appnp_model.h5'
 - You should add some kind of a conclusion at the end of the notebook 

Regards,

P.

kieranricardo:
Hi @PantelisElinas,

Thanks for your feedback! I'll add in your suggested changes and create a more rigorous suite of unit tests.

stellargraph/layer/ppnp.py
```diff
@@ -0,0 +1,188 @@
...
+            A_indices_t = Input(batch_shape=(1, None, 2), dtype="int64")
+            A_values_t = Input(batch_shape=(1, None))
+            A_placeholders = [A_indices_t, A_values_t]
+
+        else:
+            # Placeholders for the dense adjacency matrix
+            A_m = Input(batch_shape=(1, N_nodes, N_nodes))
+            A_placeholders = [A_m]
+
+        # TODO: Support multiple matrices
```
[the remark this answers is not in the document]
adocherty: This is a left-over from GCN – and something I should get on to fixing there!

stellargraph/layer/appnp.py
```diff
@@ -0,0 +1,244 @@
...
+                    activation=a,
+                    use_bias=self.bias,
+                    kernel_regularizer=self.kernel_regularizer,
+                )
+            )
+
+        feature_dim = self.layer_sizes[-1]
+
+        for ii in range(approx_iter):
+            self._layers.append(Dropout(self.dropout))
```
adocherty: Should there be a dropout between the layers of graph convolution? This is approximating the matrix inverse applied to the output activations (corresponding to the final classes we wish to classify) so I’m not sure dropout is suitable here. That being said, NN design is very empirical so we can test if there are better results with or without dropout

stellargraph/layer/appnp.py
```diff
@@ -0,0 +1,244 @@
...
+            #use graph convolutions to perform the AH matrix multiplication (A = adjacency matrix,
+            # H = previous layer feature matrix)
+            #set the weights W = I (identity), bias=0, activation=linear, and make the layer non-trainable
+            # so the convolution becomes: activation(AHW + bias) = (AHI + 0) = AH
+            #add propogation step with lambda function in the __call__ function
+            self._layers.append(
+                GraphConvolution(
```
adocherty: Why are you using a graph convolutional layer here but not in PPNP?  It seems a bit heavyweight to use a convolutional layer instead of creating a new matrix multiplication layer, especially as you need to then append another layer to add the transport vector.  I would recommend creating a new Layer similar to the `GraphConvolutionalLayer` that does the computations you require for each APPNP iteration.

stellargraph/layer/appnp.py
```diff
@@ -0,0 +1,244 @@
...
+                    kernel_initializer='Identity',
+                    final_layer=False,
+                    trainable=False
+                )
+            )
+        #apply softmax
+        self._layers.append(Softmax())
+
+        #gather the nodes from the output indices - inputs = [all node predictions, output node indices]
+        self._layers.append(
```
adocherty: If you specify in the final graph convolutional layer `final_layer=True` it will gather the outputs according to the output indices without requiring this final Lambda layer. Or am I missing something?

stellargraph/layer/appnp.py
```diff
@@ -0,0 +1,244 @@
...
+            elif isinstance(layer, Dense):
+                h_layer = layer(h_layer)
+                # store feature layer for propogation
+                feature_layer = h_layer
+            else:
+                # For other (non-graph) layers only supply the input tensor
+                h_layer = layer(h_layer)
+
+        #apply gather layer
+        h_layer = self._layers[-1]([h_layer, out_indices])
```
adocherty: You could also just create and apply the gather layer here, not all layers used in the model need to be stored in `._layers`.

stellargraph/layer/ppnp.py
```diff
@@ -0,0 +1,188 @@
...
+                )
+            )
+
+        # Create propogation matrix and propogation layer
+        Aadj = generator.Aadj.todense()
+
+        # propagation matrix is non-sparse : create and store at initalization to save time
+        propogation_matrix = self.transport_probability * np.linalg.inv(np.eye(Aadj.shape[0]) -
+                                                                  (1 - self.transport_probability) * Aadj)
+        propogation_matrix = tf.convert_to_tensor(propogation_matrix, tf.float32)
```
adocherty: Here you take the matrix in the generator and create a static tensor representation of the matrix. This means you cannot use the same model on different graphs as the model contains the graph. This is an important caveat of this approach: PPNP embeds the graph in the model so using a generator with another graph will fail, whereas APPNP as it uses the `GraphConvolutionalLayer` the model can be transferred between graphs by simply providing a generator object with the new graph.

adocherty: Rather than doing this I would use a graph multiplication layer, as for the APPNP code. We can avoid calculating the inverse in the graph each time the model is run, by adding an option to the `FullBatchNodeGenerator` to do this pre-processing. The `FullBatchNodeGenerator` could then be called with `method='ppnp'` instead of `method='gcn'`.

adocherty: A note for future development: having the pre-processed adjacency matrix as a static tensor that can be stored in GPU memory rather than transferred at each iteration of training is a good idea.  However, to do this and allow applying or training the model on multiple graphs, we need to find a way to allow updating this adjacency matrix when we change the graph that is easy for the user.

adocherty (commented):
Hi, I've looked through the implementation of both algorithms and they look good.  There is an important caveat that the PPNP model will not be transferrable between graphs (for inductive learning for example). We can either change how we have implemented the PPNP class now or go ahead with the current implementation had add a follow-up issue: @PantelisElinas what do you think?

stellargraph/layer/appnp.py
kieranricardo: Hi @adocherty, I agree with you there! The authors use dropout in between each iteration and I tried to make my implementation true to the paper. Happy to test and report back.

stellargraph/layer/appnp.py
kieranricardo: I have to do one last addition of the feature layer after the final convolution - so the gathering operation has to be done after the final addition not the final convolution. I actually tried this earlier by excluding the final addition of the feature layer but this tanked the accuracy!

stellargraph/layer/appnp.py
kieranricardo: Good idea, I'll move the gather layer over.

stellargraph/layer/appnp.py
kieranricardo: Good idea, I'll create a new matrix multiplication layer.

stellargraph/layer/ppnp.py
kieranricardo: @adocherty good idea, I'll add this into the FullBatchNodeGenerator and either add a warning or an error if a user tries to use `sparse=True` while using `method="ppnp"` so they don't accidentally run things expecting a sparse matrix.

kieranricardo:
@PantelisElinas @adocherty I stumbled across a test_gcn.py bug that I copied over to my unit tests. \`test_GCN_apply_sparse()` is actually only testing on dense adjacency matrices. I've fixed this in my unit tests, should I copy the fix over for test_gcn.py in this pull request or create a new issue?

adocherty:
> @PantelisElinas @adocherty I stumbled across a test_gcn.py bug that I copied over to my unit tests. `test_GCN_apply_sparse()` is actually only testing on dense adjacency matrices. I've fixed this in my unit tests, should I copy the fix over for test_gcn.py in this pull request or create a new issue?

I think fixing it here should be fine.  Good catch!

stellargraph/layer/appnp.py
adocherty: I see, in that case a new matrix multiplication layer with the added teleport will make this neater.

stellargraph/layer/appnp.py
adocherty: Oh really? I missed that in the paper! Keep it how it's implemented then – we can test out other ways later 😄

kieranricardo: I did a couple of runs and the test set accuracy doesn't change so I'll remove the unnecessary dropout layers.

kieranricardo: I didn't see your comment above! Should I put the extra dropout layers back in?

adocherty: As you have done it now, I'd be inclined to leave them out. @PantelisElinas what do you think?

kieranricardo:
@PantelisElinas @adocherty I implemented the changes you suggested :)

I also added the test_gcn.py bug fix and I added functionality to APPN to allow the user to propagate any Keras model through the graph. Check out the last part of the demo for an example, its super fast and because it only uses node features during training the training stage can be distributed across as many clusters as the user wants!

stellargraph/layer/appnp.py
PantelisElinas: I think we should try to be as close as possible to the paper in case a researcher uses our implementation as a baseline for comparison of their own methods against PPNP.

PantelisElinas:
> @PantelisElinas @adocherty I stumbled across a test_gcn.py bug that I copied over to my unit tests. `test_GCN_apply_sparse()` is actually only testing on dense adjacency matrices. I've fixed this in my unit tests, should I copy the fix over for test_gcn.py in this pull request or create a new issue?

Yes, please update the code as necessary.

PantelisElinas:
@kieranricardo your branch conflicts with the latest in develop. Can you please resolve these conflicts in your branch and then push again so we can finish the review?

Regards,

P.

PantelisElinas:
@kieranricardo 

for you demo, I suggest we create a separate notebook for demonstrating scalability. You should demonstrate this on a larger than Cora dataset. You can use any of the large datasets in the PPNP paper. What do you think @youph and @adocherty? 

Why are you using 1000 nodes for training? All works use 140 samples for training, 500 for validation and the rest for testing; strictly speaking, research papers use only 1000 nodes for training but for all our demos using Cora, we have not limited ourselves to 1000 test nodes.

In my experiments using your notebook and 140 training examples, the more scalable model achieves an accuracy of ~0.74 vs ~0.81 for the end-to-end trained PPNP and APPNP models. That is, for smaller training sets, the former is better than the MLP on the node features (~0.60 in my experiments) but not as good as the end-to-end trained models. What I am trying to say is that your below comment about the relative performance of the 4 methods, should be revised accordingly,

> Our accuracy is almost as good as end-to-end trained PPNP and APPNP but now we can handle abitrarily large graphs!

Also, I suggest you scale back the language in the descriptions, e.g., drop terms such as "super" and "exascale", since, especially for the latter, there are probably other reasons why our implementation would not scale to anything that large. 

Please re-write the following for clarity, 

> But, we can get around this by JUST training the fully connected network and then propogate the trained fully connected network instead of training the whole network end-to-end. 

Where is the following demonstrated in this notebook?

> The advantage of this is that we can distribute the training across as many GPUs as we like and handle abritrarily large graphs.

Do not write words in all capital letters, e.g., JUST and ANY,  just for emphasis; if necessary, please use bold font.

Finally, add a reference to the BERT paper as many people may not be familiar with this work.

stellargraph/layer/appnp.py
```diff
@@ -1,17 +1,150 @@
...
+            raise ValueError(
+                "Currently full-batch methods only support a batch dimension of one"
+            )
+
+        # Remove singleton batch dimension
+        features = K.squeeze(features, 0)
+        propagated_features = K.squeeze(propagated_features, 0)
+        out_indices = K.squeeze(out_indices, 0)
+
+        # Calculate the layer operation of GCN
```
adocherty: This isn't GCN

stellargraph/layer/appnp.py
```diff
@@ -182,30 +301,23 @@ def __call__(self, x):
...
 
-        #apply gather layer
-        h_layer = self._layers[-1]([h_layer, out_indices])
         return h_layer
 
+
     def node_model(self):
         """
-        Builds a APPNP model for node prediction
+        Builds a APPNP model for node
```
adocherty: This sentence is unfinished, why not add

stellargraph/mapper/node_mappers.py
```diff
@@ -336,7 +337,7 @@ def flow_from_dataframe(self, node_targets, shuffle=False):
 
 
 class HinSAGENodeGenerator:
-    """Keras-compatible data mapper for Heterogeneous GraphSAGE (HinSAGE)
+    """Keras-compatible data mapper for Heterogeprop_generator.Aadjneous GraphSAGE (HinSAGE)
```
adocherty: This looks like a copy paste accident!

stellargraph/mapper/node_mappers.py
```diff
@@ -704,14 +705,15 @@ class FullBatchNodeGenerator:
 
     """
 
-    def __init__(self, G, name=None, method="gcn", k=1, sparse=True, transform=None):
+    def __init__(self, G, name=None, method="gcn", k=1, sparse=True, transform=None, a=0.1):
```
adocherty: Let's add `method=ppnp` as an option in the docstring of this class.

stellargraph/core/utils.py
```diff
@@ -123,6 +123,14 @@ def chebyshev_recurrence(T_k_minus_one, T_k_minus_two, X):
     return T_k
 
 
+def PPNP_Aadj_feats_op(features, A, k=1, a=0.1, method="gcn"):
```
adocherty: I think a brief docstring would be useful here. Also the current implementation causes a warning in SciPy:
```
python3.6/site-packages/scipy/sparse/linalg/dsolve/linsolve.py:203: SparseEfficiencyWarning: spsolve is more efficient when sparse b is in the CSC matrix format
  'is in the CSC matrix format', SparseEfficiencyWarning)
```

Perhaps convert to CSC before calculating the inverse?  It is converted to the required COO format in the `SparseFullBatchNodeSequence` anyway.

adocherty (changes requested):
The changes look good.  Creating an APPNP model from a pre-trained node feature Keras model is awesome!  I've got a few comments on the changes below:

I think you've checked in this file by accident which isn't related to your changes:
`demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb`
Restore this file from develop before merging (it currently contains error messages).

In 
`demos/node-classification/ppnp/ppnp-cora-node-classification-example.ipynb`

* We have the data located in `~/data/cora` in all other notebooks, let's use that here:
```
data_dir = os.path.join("~", "DataSets "cora")
```

* You explain `layer_sizes`, `activations` and `dropout` arguements, let's also explain `trainsport_probability` and `kernel_regularizer`

* Should the final activation here be 'linear' not 'relu' as you will add a softmax layer later?

* Use as many keyword arguments as possible to help the user, for example:
```
ppnp = PPNP(
    layer_sizes=[64, 64, train_targets.shape[-1]],
    activations=['relu', 'relu', 'relu'], 
    generator=generator,
    transport_probability=0.1, 
    dropout=0.5,
    kernel_regularizer=keras.regularizers.l2(0.001)
)
```
 * This will help as we are unify the API across multiple models and the order of the arguments may change, but keyword arguments will still work.  The same comment applies to the APPNP model.

* You have created two new layers `APPNPPropagationLayer` and `PPNPPropagationLayer` let's add them to the `custom_keras_layers` in `stellargraph/__init__.py` so that we can easily load saved models.

*

adocherty:
> @kieranricardo
> 
> for you demo, I suggest we create a separate notebook for demonstrating scalability. You should demonstrate this on a larger than Cora dataset. You can use any of the large datasets in the PPNP paper. What do you think @youph and @adocherty?

I think we can make this another issue, let's finish this implementation up with the current example Cora dataset.

> Why are you using 1000 nodes for training? All works use 140 samples for training, 500 for validation and the rest for testing; strictly speaking, research papers use only 1000 nodes for training but for all our demos using Cora, we have not limited ourselves to 1000 test nodes.

I agree, all our other demos use 140 nodes here.
 
> Also, I suggest you scale back the language in the descriptions, e.g., drop terms such as "super" and "exascale", since, especially for the latter, there are probably other reasons why our implementation would not scale to anything that large.

I agree with this, we can focus on scalable training in a separate issue with a larger dataset.

> Finally, add a reference to the BERT paper as many people may not be familiar with this work.

I would say "state-of-the-art NLP model (for example BERT)" and give the reference to the BERT paper.

adocherty:
Sorry for all the comments, but one more thing:  Could you look at the `docs/api.txt` file and add a section for the PPNP and APPNP models?  That will expose the documation in the class doctrings so it appears on our API documentation page: https://stellargraph.readthedocs.io/

stellargraph/mapper/node_mappers.py
kieranricardo: Nice pickup!

stellargraph/core/utils.py
kieranricardo: Good idea, I've chucked in a doc string. And I've converted to a numpy array instead because only `use_sparse = False` is supported for `method = 'ppnp'` now.

stellargraph/layer/appnp.py
kieranricardo: I've put the extra dropout layers back in.

kieranricardo:
@PantelisElinas @adocherty thanks for your comments!

I've toned down my demo, reduced the number of training nodes to 140, revised my comparison between the 4 models and tried to improve clarity wherever possible.

I also exposed my custom layers and I also added them to api.txt. I've never use doc strings before, is there a way to check that I've added them correctly?

@adocherty the activations specified are the activations for the fully connected layers before the graph propagation so no softmax is needed. I've also changed things so that the user manually adds the softmax layer as in GCN.

re. demoing with a larger dataset I'd be really keen to implement this! I keep hearing about a bitcoin and a twitter dataset in meetings; what do you think of using one of the real world datasets that the stellar team has?

youph:
Hi Kieran,
Demoing PPNP and APPNP on Twitter dataset would be great – in fact, there’s an issue for that https://github.com/stellargraph/stellargraph-ml-private-issues/issues/150
If you can’t see the private issues, here’s the description:

Description
Apply the PPNP and APPN algorithms, developed in #484, to Twitter use case, and compare the resulting metrics to GCN, GAT, and GraphSAGE.
This is to convince ourselves it works well, before talking to Platform about implementing these on Spark.
Done Checklist
•	 Demo notebook applying PPNP and APPN to Twitter
•	 Comprehensive comparison of PPNP, APPN, GAT, GCN, and GraphSAGE on Twitter use case, using Evaluation System

The “baseline” notebook with GraphSAGE, GCN, and GAT algorithms is in https://github.com/stellargraph/stellargraph/blob/develop/demos/use-cases/hateful-twitters.ipynb
It would be good to expand it with PPNP and APPNP.

adocherty:
> I've fixed the merge issues and reverted the gcn demo (really didn't mean to commit that...)

I've done that myself many times 😆 

> I also exposed my custom layers and I also added them to api.txt. I've never use doc strings before, is there a way to check that I've added them correctly?

Great, you can build the documentation using Sphinx. I should add some information about how to do this to the contributors file, in the mean time you can build the documentation it like this:
```
pip install sphinx sphinx-rtd-theme
cd docs
make html
open _build/html/index.html
```
 
> @adocherty the activations specified are the activations for the fully connected layers before the graph propagation so no softmax is needed. I've also changed things so that the user manually adds the softmax layer as in GCN.

Great!

adocherty (dismissed):

PantelisElinas (dismissed):
Looks good!

Regards,

P.

adocherty:
We should try and get this merged by Thursday, before the release!

PantelisElinas:
@kieranricardo have you finished making changes? We are happy with the current implementation, so if you don't want to change anything else, we can merge it this afternoon. Let us know.

Regards,

P.

PantelisElinas (dismissed):

kieranricardo:
hi @PantelisElinas,

Yep, I'm done making changes - if you're happy merge away!

PantelisElinas:
@kieranricardo just make a final update so that we can merge it.

Regards,

P.

kieranricardo:
@PantelisElinas done, the only change I made was adding myself to the contributors.

youph:
@kieranricardo 
there’s an error in the `ppnp-cora-node-classification-example` demo notebook, cell 30: the 2nd Dense layer should apply to `layer` instead of `in_layer`. Could you please fix that?

Issue #381, opened 2019-05-02
Add 'activations' to GraphSAGE and HinSAGE classes


adocherty:
### Description

Currently the `GCN` and `GAT` layers allow specifying the activations at each layer of the network. GraphSAGE and HinSAGE do not, but rather hard code the activations to "relu" for all layers except the final one which is "linear".

Let's change the GraphSAGE class to accept an "activations" argument to specify the activations at all levels. The default behaviour should match the current implementation.

### User Story
As a: user of stellargraph library
I want: have similar configuration options for each graph ML class
so that: I can switch between models easily

### Done Checklist (Development)
- [x] Assumptions of the user story met
- [x] Produced code for required functionality
- [x] Branch and Pull Request build on CI
- [x] Branch and Pull Request pass unit tests on CI
- [ ] Version number reflects new status
- [x] Peer Code Review Performed
- [x] Code well commented
- [ ] Documentation in repo
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #493, opened 2019-09-24
Unify activations for GraphSAGE, HinSAGE, GCN and GAT

geoffj-d61:
Issue #381 requires optional activations to be added to GraphSAGE and HinSAGE, with a default to match the original fixed behaviour of `['relu', ..., 'relu', 'linear']`.

In addition, I have made the required activations of GCN and GAT to also be optional with the standard default, so that all models have similar behaviours in regards to activations.

### DIFF
--- a/stellargraph/layer/gcn.py
+++ b/stellargraph/layer/gcn.py
@@ -273,28 +273,30 @@ class GCN:
             x_inp, predictions = gcn.node_model()
 
     Args:
-        layer_sizes (list of int): list of output sizes of GCN layers in the stack
-        activations (list of str): list of activations applied to each layer's output
-        generator (FullBatchNodeGenerator): an instance of FullBatchNodeGenerator class constructed on the graph of interest
-        bias (bool): toggles an optional bias in GCN layers
-        dropout (float): dropout rate applied to input features of each GCN layer
-        kernel_regularizer (str): normalization applied to the kernels of GCN layers
+        layer_sizes (list of int): Output sizes of GCN layers in the stack.
+        generator (FullBatchNodeGenerator): The generator instance.
+        bias (bool): If True, a bias vector is learnt for each layer in the GCN model.
+        dropout (float): Dropout rate applied to input features of each GCN layer.
+        kernel_regularizer (str): Normalization applied to the kernels of GCN layers.
+        activations (list of str): Activations applied to each layer's output;
+            defaults to ['relu', ..., 'relu', 'linear'].
     """
 
     def __init__(
         self,
         layer_sizes,
-        activations,
         generator,
         bias=True,
         dropout=0.0,
         kernel_regularizer=None,
+        activations=None,
     ):
         if not isinstance(generator, FullBatchNodeGenerator):
             raise TypeError("Generator should be a instance of FullBatchNodeGenerator")
 
         assert len(layer_sizes) == len(activations)
 
+        n_layers = len(layer_sizes)
         self.layer_sizes = layer_sizes
         self.activations = activations
         self.bias = bias
@@ -311,8 +313,16 @@ class GCN:
                 num_of_nodes=self.generator.Aadj.shape[0]
             )
 
+        # Activation function for each layer
+        if activations is None:
+            activations = ["relu"] * (n_layers - 1) + ["linear"]
+        elif len(activations) != n_layers:
+            raise ValueError(
+                "Invalid number of activations; require one function per layer"
+            )
+        self.activations = activations
+
         # Initialize a stack of GCN layers
-        n_layers = len(self.layer_sizes)
         self._layers = []
         for ii in range(n_layers):
             l = self.layer_sizes[ii]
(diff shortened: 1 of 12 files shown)


stellargraph/layer/gcn.py
```diff
@@ -311,8 +311,16 @@ def __init__(
                 num_of_nodes=self.generator.Aadj.shape[0]
             )
 
+        # Activation function for each layer
+        if activations is None:
+            activations = ["relu"] * (n_layers - 1) + ["linear"]
```
adocherty: Having a final activation of linear is appropriate GraphSAGE as, by default, it has L2 normalization applied afterwards, it is perhaps not appropriate for GCN.  I think having all layers default to "relu" would be better.

stellargraph/layer/graph_attention.py
```diff
@@ -720,17 +720,19 @@ def __init__(
 
         # Check activations (list of str):
         # check type:
+        if activations is None:
```
adocherty: A better default would be all "elu" (which is the configuration used in the original paper).

stellargraph/layer/graphsage.py
```diff
@@ -837,16 +849,15 @@ def apply_layer(x: List, num_hops: int):
...
         h_layer = xin
         for layer in range(0, self.max_hops):
             h_layer = apply_layer(h_layer, layer)
-            self.layer_tensors.append(h_layer)
 
         # Remove neighbourhood dimension from output tensors of the stack
         # note that at this point h_layer contains the output tensor of the top (last applied) layer of the stack
         h_layer = [
-            Reshape(K.int_shape(x)[2:])(x) for x in h_layer if K.int_shape(x)[1] == 1
+            Reshape(K.int_shape(x)[2:])(x) if K.int_shape(x)[1] == 1 else x
```
adocherty: Good catch!

stellargraph/layer/hinsage.py
```diff
@@ -310,13 +313,20 @@ def eval_neigh_tree_per_layer(input_tree):
             for layer, dim in enumerate([self.input_dims] + layer_sizes)
         ]
 
+        # Activation function for each layer
+        if activations is None:
+            activations = ["relu"] * (self.n_layers - 1) + ["linear"]
+        elif len(activations) != self.n_layers:
+            raise ValueError(
```
adocherty: I'd perhaps raise an IndexError here.
Also in GraphSAGE line 732.

geoffj-d61: This one I'm not sure about - various code raises ValueError, TypeError, etc. Sometimes I can see the rationale, e.g. a str was passed where an int was required - maybe TypeError is sensible. Other times, it doesn't seem so clear-cut, so I have opted generically for ValueError, sort of as a proxy to Java's IllegalArgumentException. I'm open to discussion on this point; maybe we can have (or already have) some clear guidelines.

stellargraph/layer/graphsage.py
geoffj-d61: Only, I have no idea what mess it might make if the first dimension is ever not one!

stellargraph/layer/hinsage.py
geoffj-d61: Just to capture my thinking, suppose the input requirement was for a list of exactly N integers between 0 and M, but suppose the actual input given was a list comprised of N-3 integers between 0 and M, one integer > M, and one string. Then:
* if length is checked first, we might get an IndexError;
* if type is checked first, we might get a TypeError;
* if value is checked first, we might get a ValueError.
Hence my reasoning that it might be easier (including for any try/catch wrapper) to just have a single exception that indicates that not all conditions have been satisfied.

geoffj-d61:
Hi @adocherty can you please review this (I made the changes you suggested about activation defaults, but in taking @youph off as a reviewer it no longer thinks you approved it!?).

stellargraph/layer/graphsage.py
adocherty: Well, it would have made a total mess before. Now, there's a chance of it working 😆

stellargraph/layer/hinsage.py
adocherty: Yes, it's rather ambiguous in Python what error to raise, especially when there are multiple things wrong in the input.  Nonetheless, regardless of which order things are checked as long as we state what the error is the user will get a more informative message.

stellargraph/layer/graphsage.py
youph: Good catch indeed. Why did it work before?

geoffj-d61: Because the code is designed to aggregate neighbourhoods down to a single feature vector per node, with leading dimension of 1. So I guess it always managed to do this - not sure in what cases it might possibly fail.

youph: Should this be generalised to
```
h_layer = [
            Reshape(K.int_shape(x)[2:])(x) if K.int_shape(x)[1] == head_shape else x
            for x in h_layer
        ]
```
?

geoffj-d61: Possibly. Personally, I'd like to first see a use case where node features are not vectors.

PR #494, opened 2019-09-25
Unify regularisation of GraphSAGE, HinSAGE, GCN and GAT

geoffj-d61:
Added or modified passing of optional regularisers, initialisers and constraints from GraphSAGE, HinSAGE, GAT and GCN to the underlying layers.


geoffj-d61:
Due to touching code in common, this branch feature/469-add-regularisation was branched from feature/381-graphsage-activations, which is being reviewed in PR #493. Here only the ability to specify regularisation of the underlying layers needs testing.

stellargraph/layer/gcn.py
```diff
@@ -56,47 +56,42 @@ class GraphConvolution(Layer):
...
-        kernel_initializer (str): name of layer bias f the initializer for kernel parameters (weights)
-        bias_initializer (str): name of the initializer for bias
-        attn_kernel_initializer (str): name of the initializer for attention kernel
-        kernel_regularizer (str): name of regularizer to be applied to layer kernel. Must be a Keras regularizer.
-        bias_regularizer (str): name of regularizer to be applied to layer bias. Must be a Keras regularizer.
-        activity_regularizer (str): not used in the current implementation
-        kernel_constraint (str): constraint applied to layer's kernel
-        bias_constraint (str): constraint applied to layer's bias
+
+    Optional Args:
```
adocherty: All arguments optional or otherwise should be under `Args:` to follow the Google docstring styles. You can check they are rendered correctly by building the api docs see the info here:
https://github.com/stellargraph/stellargraph-ml-private-issues/blob/master/README.md

stellargraph/layer/gcn.py
```diff
@@ -56,47 +56,42 @@ class GraphConvolution(Layer):
...
-        final_layer=False,
-        kernel_initializer="glorot_uniform",
-        bias_initializer="zeros",
-        kernel_regularizer=None,
-        bias_regularizer=None,
-        activity_regularizer=None,
-        kernel_constraint=None,
-        bias_constraint=None,
-        **kwargs
+        self, units, activation=None, use_bias=True, final_layer=False, **kwargs
```
adocherty: I'm not sure I like the use of `**kwargs` here, it makes it harder for users to know what the arguments can be. Why not list them all?

stellargraph/layer/graph_attention.py
```diff
@@ -615,6 +619,18 @@ class GAT:
         normalize (str or None): normalization applied to the final output features of the GAT layers stack. Default is None.
         activations (list of str): list of activations applied to each layer's output; defaults to ['relu', ..., 'relu', 'linear'].
         saliency_map_support (bool): XXX What is this?
```
adocherty: If calculating saliency maps using the tools in `stellargraph.utils.saliency_maps` this should be True. Otherwise this should be False.

stellargraph/layer/graph_attention.py
```diff
@@ -615,6 +619,18 @@ class GAT:
...
+
+
+    Optional Args:
+        kernel_initializer (str or func): The initialiser to use for the head weights.
```
adocherty: While in the research code they exposed the intializers and regularizers for all internal options, I think we should try and guide the user and have  fewer options at the top level class.  I think having `kernel_regularizer` and `attn_kernal_regularizer` is all we need for now. The other options I think are too much, although we can still support them in the underlying layers.

This also applies to:
gcn.py: line 61 (GCN)
graphsage.py: line 732 (GraphSAGE)
graphsage.py: line 1046 (DirectedGraphSAGE)
hinsage.py: line 46 (HinSAGE)

stellargraph/layer/graphsage.py
```diff
@@ -417,26 +457,34 @@ def _build_group_weights(self, in_shape, out_size, group_idx=0):
...
+                initializer=self.kernel_initializer,
+                regularizer=self.kernel_regularizer,
+                constraint=self.kernel_constraint,
                 trainable=True,
             )
             b_pool = self.add_weight(
                 name=f"b_pool_g{group_idx}",
                 shape=(self.hidden_dim,),
-                initializer=self._initializer,
+                initializer=self.kernel_initializer,
```
adocherty: This is a bias for the pooling function. Applying kernel initializers, constraints and regularization here is interesting.  We really don't know the effects of regularization or constraints in these more complicated aggregation classes.

This is just a comment, I don't know what to do about it. Maybe add a comment in the code.

stellargraph/layer/gcn.py
geoffj-d61: Partly because the auto code checker doesn't like too many arguments (for good reason), and partly for the same reason you mention above - we can have additional (non-typical) arguments for expert users that are not mentioned in the method docs, to avoid confusing the typical user.

stellargraph/layer/graphsage.py
geoffj-d61: In every case that I modified, there was a pre-existing use of `initializer=self._initializer` that I altered, so I am not in effect doing anything different, especially since the default case should have exactly the same behaviour. I take your point, however, that the intention might differ.

geoffj-d61: If it is really a bias term, should it take the bias regulariser, etc.? As I mentioned, the previous code in fact assumed 'glorot_uniform' initialisation by default. I can just hard wire it to that, if you prefer.

stellargraph/layer/gcn.py
adocherty: I'm happy with that 👍

stellargraph/layer/graphsage.py
adocherty: Actually good point, for bias we should really be initializing them to zeros. Perhaps you can use the `bias_initializer` here and set the default for that to be zero:
https://datascience.stackexchange.com/questions/17987/how-should-the-bias-be-initialized-and-regularized/18145#18145

geoffj-d61: @adocherty,  now that I have looked more closely at the code, you are right - it is definitely a bias term. Unless you think otherwise, I am going to apply the bias regularisers to it, including the default of 'zeros', which will be a change from previous behaviour. Do we have any demos that use the pooling aggregators?

geoffj-d61: Also, should the AttentionalAggregator of GraphSAGE allow the attn_kernel_regularizer?

geoffj-d61: Hi @adocherty , could you please give this PR another look over? I think I have resolved all your requests - the outstanding issue is my question above, re allowing attn_kernel_regularizer in GraphSAGE for the AttentionalAggregator case.

adocherty: > Unless you think otherwise, I am going to apply the bias regularisers to it, including the default of 'zeros', which will be a change from previous behaviour.

I agree.

> Do we have any demos that use the pooling aggregators?

We have no demos that use the pooling aggregators at the moment, so these changes won't be a problem.

> Also, should the AttentionalAggregator of GraphSAGE allow the attn_kernel_regularizer?

I don't think that we need to expose this at the GraphSAGE class level at the moment.  The AttentionalAggregator is still pretty experimental.

PR #499, opened 2019-10-09
Fixed GAT demo with fit method in gat-cora-example.py

adocherty:
Updated the demo to remove the batch size which prevented the `.fit` method from working with the new singleton batch sizes for GCN/GAT.


adocherty:
> LGTM.
> Did we describe the new singleton batch dimension in the corresponding release description?

Good point, I think we did in the Changelog & docstrings. I'll check.

Issue #164, opened 2018-08-13
Build up install procedure for stellar-ml


arc0:
### Description
We need to be able to install the library easily. This will involve writing and testing the:
- setup.py (build + install) procedure
- pip install procedure
- dependencies

### Done Checklist (Development)
- [ ] Produced code for required functionality
- [ ] Peer Install Review Performed
- [ ] Tested on clean machines + environments

youph:
@adocherty can this be moved to Done or Closed?

PR #502, opened 2019-10-16
Feature/rgcn

kieranricardo:
#164
Implementation of RGCN

#### Reviewer Checklist
- [x] Code correctly implements R-GCN algorithm
- [x] Unit tests
- [x] Demo notebook runs and clearly demonstrates/explains how to train R-GCN models
- [x] Documentation


stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,496 @@
...
+
+        Args:
+            input_shapes (list of int): shapes of the layer's inputs
+            (node features, node_indices, and adjacency matrices)
+
+        """
+        feat_shape = input_shapes[0]
+        input_dim = int(feat_shape[-1])
+
+        if self.num_bases != 0:
```
PantelisElinas: I guess negative values are good here!

Maybe change it to self.num_bases > 0 so that at least negative values are treated as zero. A check for a valid value and a unit test to go with it would be helpful.

kieranricardo: Nice pickup! I'll change it to `self.num_bases > 0` and add a unit test around this

stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,496 @@
...
+        features = K.squeeze(features, 0)
+        out_indices = K.squeeze(out_indices, 0)
+
+        # Calculate the layer operation of RGCN
+        output = K.dot(features, self.self_kernel)
+
+        for i in range(self.num_relationships):
+            h_graph = K.dot(As[i], features)
+            output += K.dot(h_graph, self.relational_kernels[i])
+
```
PantelisElinas: Either the above lines are incorrect or I have misunderstood the code or the paper. 

If I am not mistaken, according to equations 2 and 4 in the paper, there is a weight matrix, `W_r`, for each relationship type. In the above code, there is only one W, `self.self_kernel`. Isn't this wrong? In the paper's figure 2, they talk about parameter sharing but I don't think they mean sharing across relationships but rather across nodes. I may be wrong, but can you please check?

stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,496 @@
...
+        """
+        Apply a stack of RGCN layers to the inputs.
+        The input tensors are expected to be a list of the following:
+        [Node features shape (1, N, F), Output indices (1, O)] +
+        [Adjacency indices for each relationship (1, E, 2) for _ in range(R)]
+        [Adjacency values for each relationshiop (1, E) for _ in range(R)]
+
+
+        where N is the number of nodes, F the number of input features,
+              E is the number of edges, O the number of output nodes,
```
PantelisElinas: Can you please use another symbol instead of O? When I first saw it, I thought it was a zero. Maybe use Z?

stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,496 @@
...
+                # For an RGCN layer add the adjacency matrices and output indices
+                # Note that the output indices are only used if `final_layer=True`
+                h_layer = layer([h_layer, out_indices] + Ainput)
+            else:
+                # For other (non-graph) layers only supply the input tensor
+                h_layer = layer(h_layer)
+
+        return h_layer
+
+    def node_model(self):
```
PantelisElinas: We are trying to standardise on replacing `node_model()` with `build()`. The goal is to make `build()` return the correct model, node or link, based on the generator. Using `build()` will require the user to change fewer lines of code to re-use their scripts between tasks. Note that not all models have `build()` implemented yet, but we want to provide this method for all new model classes.

stellargraph/mapper/node_mappers.py
```diff
@@ -1128,3 +1128,275 @@ def flow_from_dataframe(self, node_targets, shuffle=False):
         return NodeSequence(
             self, node_targets.index, node_targets.values, shuffle=shuffle
         )
+
+
+class RelationalSparseFullBatchNodeSequence(Sequence):
+    """
+    Keras-compatible data generator for for node inference models
```
PantelisElinas: Spelling?

stellargraph/mapper/node_mappers.py
```diff
@@ -1128,3 +1128,275 @@ def flow_from_dataframe(self, node_targets, shuffle=False):
...
+            self.targets = None
+
+    def __len__(self):
+        return 1
+
+    def __getitem__(self, index):
+        return self.inputs, self.targets
+
+
+class RelationalDenseFullBatchNodeSequence(Sequence):
```
PantelisElinas: This class looks nearly identical to the Sparse version above with 3-4 lines different to handle the different types of adjacency matrices.

Is there are a reason why we cannot combine them into a single class with a `sparse` parameter with default value `True`?

stellargraph/mapper/node_mappers.py
```diff
@@ -1128,3 +1128,275 @@ def flow_from_dataframe(self, node_targets, shuffle=False):
...
+        transform (callable): an optional function to apply on features and adjacency matrix
+            the function takes (features, Aadj) as arguments.
+        sparse (bool): If True (default) a list of sparse adjacency matrices is used,
+            if False a list of dense adjacency matrices is used.
+
+    """
+
+    def __init__(self, G, name=None, sparse=True, transform=None):
+
+        if not isinstance(G, StellarGraphBase):
```
PantelisElinas: In the docstring above, it says that G should be StellarDiGraph, but here the check if for the more general StellarGraph. Shall we update this test to be a bit more strict?

stellargraph/mapper/node_mappers.py
```diff
@@ -1128,3 +1128,275 @@ def flow_from_dataframe(self, node_targets, shuffle=False):
...
+            if not is_real_iterable(targets):
+                raise TypeError("Targets must be an iterable or None")
+
+            # Check targets correct shape
+            if len(targets) != len(node_ids):
+                raise TypeError("Targets must be the same length as node_ids")
+
+        # The list of indices of the target nodes in self.node_list
+        # use dictionary for faster index look-up time
+        node_index = dict(zip(self.node_list, range(len(self.node_list))))
```
PantelisElinas: You perform the same operation in the `__init__` method. Can you do it once and store it with the object in `self.node_index`? Or are you trying to save memory by using a temporary object?

tests/layer/test_rgcn.py
```diff
@@ -0,0 +1,322 @@
...
+    conf = rgcn_layer.get_config()
+
+    assert conf["units"] == 16
+    assert conf["activation"] == "linear"
+    assert conf["num_bases"] == 0
+    assert conf["num_relationships"] == 5
+    assert conf["use_bias"] == True
+    assert conf["kernel_initializer"]["class_name"] == "GlorotUniform"
+    assert conf["bias_initializer"]["class_name"] == "Zeros"
+    assert conf["kernel_regularizer"] == None
```
PantelisElinas: I think comparing for equality with None should use `is` not `==`.

tests/layer/test_rgcn.py
```diff
@@ -0,0 +1,322 @@
...
+    assert conf["bias_constraint"] == None
+
+
+def test_RelationalGraphConvolution_init():
+    rgcn_layer = RelationalGraphConvolution(
+        units=16, num_relationships=5, num_bases=0, activation="relu"
+    )
+
+    assert rgcn_layer.units == 16
+    assert rgcn_layer.use_bias == True
```
PantelisElinas: Shall we be using `is` here instead of `==`?

tests/mapper/test_relational_node_mappers.py
```diff
@@ -0,0 +1,191 @@
...
+
+    n_feat = 2
+    target_dim = 5
+
+    gnx, features = create_graph_features()
+    nodes = list(gnx.nodes)
+    node_features = pd.DataFrame.from_dict(
+        {n: f for n, f in zip(nodes, features)}, orient="index"
+    )
+    G = StellarDiGraph(gnx, node_features=node_features)
```
PantelisElinas: Please see my previous comments about the correct type of the graph input to RelationalFullBatchNodeGenerator. Does the input graph have to be directed or can it also be undirected? You only have a test for directed graphs.

tests/mapper/test_relational_node_mappers.py
```diff
@@ -0,0 +1,191 @@
...
+
+        generator = RelationalFullBatchNodeGenerator(self.G)
+        assert len(generator.As) == self.num_relationships
+        assert all((A.shape == (self.N, self.N)) for A in generator.As)
+
+        assert generator.features.shape == (self.N, self.n_feat)
+
+    def test_generator_constructor_wrong_G_type(self):
+        with pytest.raises(TypeError):
+            generator = RelationalFullBatchNodeGenerator(nx.Graph(self.G))
```
PantelisElinas: Can we add tests for other graph types? Should nx.DiGraph or nx.MultidiGraph (the type of nx graph you start with in `create_graph_features` work?

stellargraph/mapper/node_mappers.py
kieranricardo: I was trying to save memory, but I don't think time/memory overhead will be very much either. What do you think?

stellargraph/mapper/node_mappers.py
kieranricardo: The paper only talks about using directed graphs (I think relational graphs are inherently directed?) but I don't see any reason why this wouldn't work for undirected graphs. I'll update the docstring

tests/mapper/test_relational_node_mappers.py
kieranricardo: It can be both, I've added in unit tests for both directed and undirected

tests/layer/test_rgcn.py
kieranricardo: You're absolutely right, fixed!

stellargraph/layer/rgcn.py
kieranricardo: Should I use `build() `to call `node_model()` is the generator is node generator or removing the `node_model()` function entirely?

stellargraph/mapper/node_mappers.py
kieranricardo: Great idea, I was copying the API for the `FullBatchNodeSequence` but this sounds much better. I'll implement this.

tests/mapper/test_relational_node_mappers.py
kieranricardo: This test is for making sure that passing any networkx graph to the generator will fail, only `StellarGraphBase` objects should be accepted.

stellargraph/layer/rgcn.py
PantelisElinas: I think it might be best to remove node_model all together by just making it a "private" method we can call from build().

stellargraph/mapper/node_mappers.py
PantelisElinas: Let's just do it in __init__. It will be easier to move it out of __init__ later, especially after we have re-worked the StellarGraph class and we have a better memory management solution.

stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,554 @@
...
+            input_shapes (list of int): shapes of the layer's inputs
+            (node features, node_indices, and adjacency matrices)
+
+        """
+        feat_shape = input_shapes[0]
+        input_dim = int(feat_shape[-1])
+
+        if self.num_bases > 0:
+
+            self.bases = self.add_weight(
```
adocherty: Add some more comments on what this code block is doing

stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,554 @@
...
+                constraint=self.kernel_constraint,
+            )
+
+            self.coefficients = [
+                self.add_weight(
+                    shape=(self.num_bases,),  # hyperparametr B
+                    initializer=self.kernel_initializer,
+                    name="coeff",
+                    regularizer=self.kernel_regularizer,
+                    constraint=self.kernel_constraint,
```
adocherty: I'm not sure that using the same regulariser for the coefficents as the kernel is what's required. I'd not include them in the coefficients.

stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,554 @@
...
+                "Generator should be a instance of RelationalFullBatchNodeGenerator"
+            )
+
+        n_layers = len(layer_sizes)
+        self.layer_sizes = layer_sizes
+        self.activations = activations
+        self.bias = bias
+        self.num_bases = num_bases
+        self.dropout = dropout
+        self.generator = generator
```
adocherty: I'm trying to move away from storing the generator in the model. It seems too messy.

I have also introduced a 'multiplicity' property to the generators which is 1 for a node inference and 2 for link inference. To match the new-style full batch generators you could look at that.

See this PR: https://github.com/stellargraph/stellargraph/pull/543

stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,554 @@
...
+        self.support = 1
```
adocherty: This is not used in the code, remove?

stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,554 @@
...
+        Builds a RGCN model for node prediction. Link/node pair prediction will added in the future.
+
+        Returns:
+            tuple: (x_inp, x_out), where ``x_inp`` is a list of Keras input tensors
+            for the specified RGCN model and ``x_out`` contains
+            model output tensor(s) of shape (batch_size, layer_sizes[-1])
+
+        """
+
+        if self.generator is not None:
```
adocherty: Look at https://github.com/stellargraph/stellargraph/pull/543 and change this to use the `multiplicity` flag to match the new-style full batch generators (don't implement link prediction yet though, but it would be a good idea to make a new issue to do this)

stellargraph/mapper/full_batch_generators.py
```diff
@@ -238,3 +238,209 @@ def flow(self, node_ids, targets=None):
             return FullBatchNodeSequence(
                 self.features, self.Aadj, targets, node_indices
             )
+
+
+class RelationalFullBatchNodeSequence(Sequence):
```
adocherty: Move this class to the `mapper/sequences.py` file.

stellargraph/mapper/full_batch_generators.py
```diff
@@ -238,3 +238,209 @@ def flow(self, node_ids, targets=None):
...
+
+    def __getitem__(self, index):
+        return self.inputs, self.targets
+
+
+class RelationalFullBatchNodeGenerator:
+    """
+    A data generator for use with full-batch models on relational graphs e.g. RGCN.
+
+    The supplied graph G should be a StellarGraphBase object that is ready for
```
adocherty: It should be a  `StellarGraph` or `StellarDiGraph` object. `StellarGraphBase` is the abstract base class for these objects and users can't create one.

stellargraph/mapper/full_batch_generators.py
```diff
@@ -238,3 +238,209 @@ def flow(self, node_ids, targets=None):
...
+                node_index[n2] for n1, n2, etype in G.edges if etype == edge_type
+            ]
+            data = np.ones(len(col_index), np.float64)
+
+            # note that A is the transpose of the standard adjacency matrix
+            # this is to aggregate features from incoming nodes
+            A = sps.coo_matrix(
+                (data, (row_index, col_index)),
+                shape=(len(self.node_list), len(self.node_list)),
+            )
```
adocherty: It looks like the new `StellarGraph` class (under development) should have a method to create adjacency matrices for different relation types.  This should be captured somewhere 📣

stellargraph/mapper/full_batch_generators.py
```diff
@@ -238,3 +238,209 @@ def flow(self, node_ids, targets=None):
...
+
+            if transform is None:
+                # normalize here and add 1e-9 row sum to avoid harmless divide by zero warnings
+                d = sps.diags(
+                    np.float_power(np.array(A.sum(1)) + 1e-9, -1).flatten(), 0
```
adocherty: I don't particularly like adding a small number to solve DBZ problems. This depends on the float type of A, and the weights: it is fragile. Try this:

```
d = sps.diags(
    np.float_power(np.ravel(np.maximum(A.sum(axis=1),1)), -1), 0
)
```

stellargraph/mapper/full_batch_generators.py
```diff
@@ -238,3 +238,209 @@ def flow(self, node_ids, targets=None):
...
+                )
+                A = d.dot(A).tocsr()
```
adocherty: Why are you converting to CSR then on line 408 to COO?  After the dot it will most likely be CSR anyway.

demos/node-classification/rgcn/rgcn-aifb-node-classification-example.ipynb
```diff
@@ -0,0 +1,414 @@
...
+   "outputs": [],
+   "source": [
+    "from rdflib.extras.external_graph_libs import *\n",
+    "from rdflib import Graph, URIRef, Literal\n",
+    "\n",
+    "import networkx as nx\n",
+    "from networkx.classes.function import info\n",
+    "\n",
+    "import stellargraph as sg\n",
+    "from stellargraph.mapper.node_mappers import RelationalFullBatchNodeGenerator\n",
```
adocherty: Don't import from `node_mappers` : this file no longer exists. The imports should be:
```
from stellargraph.mapper import RelationalFullBatchNodeGenerator
```

To do this you'll need to add `RelationalFullBatchNodeGenerator` to the `__all__` variable at the top of the `mappers/full_batch_generators.py` file so they are inserted into the `mappers` namespace.

stellargraph/layer/rgcn.py
kieranricardo: Done!

stellargraph/mapper/full_batch_generators.py
kieranricardo: good point, G can be either `StellarGraph` or `StellarDiGraph` so I'll change the docstring to say this

stellargraph/layer/rgcn.py
kieranricardo: yeah agreed, your new method is way nicer - done!

stellargraph/layer/rgcn.py
kieranricardo: done!

stellargraph/layer/rgcn.py
kieranricardo: done!

stellargraph/layer/rgcn.py
kieranricardo: good point: I'll create different regularisers, initialisers, and constraints for the kernels, coefficients, and bases

demos/node-classification/rgcn/rgcn-aifb-node-classification-example.ipynb
kieranricardo: done!

stellargraph/mapper/full_batch_generators.py
kieranricardo: nice catch, was a copy paste error. fixed!

stellargraph/mapper/full_batch_generators.py
kieranricardo: much better solution than my `+1e-9`, fixed!

stellargraph/mapper/full_batch_generators.py
kieranricardo: done!

stellargraph/mapper/node_mappers.py
kieranricardo: done!

stellargraph/layer/rgcn.py
kieranricardo: done!

kieranricardo:
@PantelisElinas @adocherty thanks for your reviews! I think I've addressed all your comments + I've made things compatible with the new stellargaph class. All the units tests are passing but the build is failing (https://github.com/stellargraph/stellargraph/pull/562) so I think its ready for re-review

stellargraph/layer/rgcn.py
```diff
@@ -0,0 +1,598 @@
...
+    The generator object should be instantiated as follows::
+
+        generator = RelationalFullBatchNodeGenerator(G)
+
+    Note that currently the RGCN class is compatible with both sparse and dense adjacency
+    matrices and the :class:`RelationalFullBatchNodeGenerator` will default to sparse.
+
+    For more details, please see the RGCN demo notebook:
+
+    emos/node-classification/rgcn/rgcn-aifb-node-classification-example.ipynb
```
PantelisElinas: I think you mean "demos/...", missing the letter d.

kieranricardo: nice catch, I'll quickly fix that

stellargraph/layer/rgcn.py
```diff
@@ -480,7 +533,7 @@ def __call__(self, x):
 
         return h_layer
 
-    def node_model(self):
+    def _node_model(self):
```
adocherty: Currently all our methods have `node_model` and `link_model` so it would be good to talk in the new team about if these methods in the other classes should be deprecated .. for now just be aware that this makes RGCN only have the new `build` API.

stellargraph/mapper/full_batch_generators.py
```diff
@@ -379,10 +309,14 @@ def __init__(self, G, name=None, sparse=True, transform=None):
...
+                self.node_index[n1]
+                for n1, n2, etype in G.edges(triple=True)
+                if etype == edge_type
             ]
             row_index = [
-                node_index[n2] for n1, n2, etype in G.edges if etype == edge_type
+                self.node_index[n2]
+                for n1, n2, etype in G.edges(triple=True)
+                if etype == edge_type
             ]
```
adocherty: Perhaps this functionality could be something for the new StellarGraph class?  Also have a look at HinSAGE and the `GraphSchema` class which was supposed to be a place for this functionality.  *There is no need to change this PR* but perhaps in the  future these operations can be done in the StellarGraph class so they can be optimized.

Issue #503, opened 2019-10-17
New NetworkX 2.4 release removes deprecated Graph.node


huonw:
**Describe the bug**

StellarGraph indexes graphs via `.node` in many locations.

This was supported in [NetworkX 2.3](https://github.com/networkx/networkx/blob/networkx-2.3/networkx/classes/graph.py#L761), but was removed in [NetworkX 2.4](https://github.com/networkx/networkx/blob/networkx-2.4/networkx/classes/graph.py) (does not mention `def node` or `node =`) which was released ~2 hours ago.

The [release notes](https://networkx.github.io/documentation/stable/release/release_2.4.html) say:

> ## Deprecations
>
> ...
>
> Graph Object methods removed (marked as deprecated 2.1)
> 
> ...
> G.node –> use G.nodes

This causes problems like

```
AttributeError: 'StellarGraph' object has no attribute 'node'
```

These's many examples of this call:

```
$ git grep \\.node\\b
```
```
demos/calibration/calibration-pubmed-link-prediction.ipynb:    "    G.node[str(nid)][globalvar.TYPE_ATTR_NAME] = \"paper\"  # specify node type\n",
demos/calibration/calibration-pubmed-link-prediction.ipynb:    "    G.node[str(nid)][\"feature\"] = f"
demos/community_detection/attacks_clustering_analysis.ipynb:    "com_G = G.subgraph([n for n, attrdict in G.node.items() if attrdict['c_infomap'] == com_id])\n",
demos/embeddings/stellargraph-metapath2vec.ipynb:    "node_targets = [ g_nx.node[node_id]['label'] for node_id in node_ids]"
demos/embeddings/stellargraph-node2vec.ipynb:    "node_targets = [ g_nx.node[node_id]['subject'] for node_id in node_ids ]"
demos/ensembles/ensemble-link-prediction-example.ipynb:    "    G.node[nid][globalvar.TYPE_ATTR_NAME] = \"paper\"  # specify node type\n",
demos/ensembles/ensemble-link-prediction-example.ipynb:    "    G.node[nid][\"feature\"] = f"
demos/ensembles/ensemble-node-classification-example.ipynb:    "        g_nx.node[nid][globalvar.TYPE_ATTR_NAME] = \"paper\"  # specify node type\n",
demos/ensembles/ensemble-node-classification-example.ipynb:    "        g_nx.node[nid][globalvar.TYPE_ATTR_NAME] = \"paper\"  # specify node type\n",
[...]

PR #504, opened 2019-10-17
Replace `.node\b` with `.nodes` for NetworkX 2.4

huonw:
This was done with textual search and replace:

    git grep --files-with-matches '\.node\b' | xargs sed -i '' 's/\.node\[/.nodes[/g'
    git grep --files-with-matches '\.node\b' | xargs sed -i '' 's/\.node(/.nodes(/g'
    git grep --files-with-matches '\.node\b' | xargs sed -i '' 's/\.node\./.nodes./g'

See: #503


huonw:
I've merged this with `develop` which incorporates the hotfix #505, and thus made this PR also remove the version restriction on NetworkX.

adocherty:
@youph @PantelisElinas This is a straighforward fix for the NetworkX 2.4+ compatibility issues (removal of the `graph.node` interface). 

@huonw hasn't assigned reviewers but I think this can be reviewed and merged nonetheless?

youph:
@adocherty yes, let's review and merge. Does this change of method names affect the new abstract SG class?

geoffj-d61:
@youph , anything to do with NetworkX is likely to affect my new design - it depends on whether this is code purely internal to the operation of the NetworkXStellarGraph implementation or not. I suspect that it is, because the new design abstracts away from relying on internals, meaning that merging shouldn't be too bad.

adocherty:
In addition the function `nx.connected_component_subgraphs` is now deprecated and was being used in **./stellargraph-node2vec-weighted-random-walks.ipynb** – I've committed a change to fix this.

The original error was:
```
AttributeError                            Traceback (most recent call last)
<ipython-input-3-bbe10817d187> in <module>
     15 # nodes and subgraphs; having these in the data does not prevent the
     16 # algorithm from running and producing valid results.
---> 17 g_nx_wt = max(nx.connected_component_subgraphs(g_nx_wt, copy=True), key=len)
     18 print("Largest subgraph statistics: {} nodes, {} edges".format(
     19     g_nx_wt.number_of_nodes(), g_nx_wt.number_of_edges()))

AttributeError: module 'networkx' has no attribute 'connected_component_subgraphs'
```

youph:
Merge?

PR #505, opened 2019-10-17
Restrict NetworkX to < 2.4 and Tensorflow to <1.15.0

huonw:
There's been two package releases that have broken StellarGraph:

- NetworkX 2.4 removed `.node` (#503)
- Tensorflow 1.15.0 changed Keras in some as-yet-unknown way (#507)

This is the minimal change required to get StellarGraph working again, and so
I think should be considered for a 0.7.3 hotfix.

#504 attempts a real fix for the NetworkX change, but is significantly riskier for a hotfix.

See #503 and #507


youph:
Hotfix 0.7.3 released

PR #514, opened 2019-10-30
Restricted tf version to 1.14

kieranricardo:
Restricted tf version to 1.14

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 networkx>=2.2,<2.4
 setuptools>=39.1.0
-tensorflow>=1.12.1,<1.15
+tensorflow>=1.14,<1.15
 numba>=0.39.0
 numpy>=1.14
 scipy>=1.1.0
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ URL = "https://github.com/stellargraph/stellargraph"
 
 # Required packages
 REQUIRES = [
-    "tensorflow>=1.12,<1.15",
+    "tensorflow>=1.14,<1.15",
     "numpy>=1.14",
     "scipy>=1.1.0",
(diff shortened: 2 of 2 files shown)


youph:
Trying to reproduce...

youph:
with `tf==1.13.1` the tests produce lots of errors like this:
```TypeError: Layer graph_attention_sparse_4 does not support masking, but was passed an input_mask: [None, None, [None, None]]```
@adocherty We aren't using the masking, are we? Is it due to specifying node and feature dimensions of the sparse feature matrices as `None`s?

adocherty:
> with `tf==1.13.1` the tests produce lots of errors like this:
> `TypeError: Layer graph_attention_sparse_4 does not support masking, but was passed an input_mask: [None, None, [None, None]]`
> @adocherty We aren't using the masking, are we? Is it due to specifying node and feature dimensions of the sparse feature matrices as `None`s?

@youph  No we aren't using masking, I don't know what this message is coming from.  It could be due to the dimensions being `None`, but why does it only occur in version 1.13?

youph:
> > with `tf==1.13.1` the tests produce lots of errors like this:
> > `TypeError: Layer graph_attention_sparse_4 does not support masking, but was passed an input_mask: [None, None, [None, None]]`
> > @adocherty We aren't using the masking, are we? Is it due to specifying node and feature dimensions of the sparse feature matrices as `None`s?
> 
> @youph No we aren't using masking, I don't know what this message is coming from. It could be due to the dimensions being `None`, but why does it only occur in version 1.13?

It also occurs in 1.12

huonw:
For ease of future reference (particularly for those that don't have access to the zenhub board and it's extra layer of PR<->issue linking above Github's); the associated issue was https://github.com/stellargraph/stellargraph/issues/519.

Issue #144, opened 2018-07-31
Use enums for flag variables


arc0:
https://github.com/stellargraph/stellar-ml/blob/acfbb255bcb9f8a11c92a860ccf714caa16d7ece/stellar/data/edge_splitter.py#L213

Hi guys, usually string variables shouldn't be used for default arguments in Python as they are mutable.

Moreover it is best practice to use enums for flag arguments
```
from enum import Enum
class NegativeSamples(Enum):
    GLOBAL = 1
    LOCAL = 2

def train_test_split(self, something=NegativeSamples.GLOBAL)
```
Then the enums would float about in a 'utils.py' or 'const.py' file that people import. If the flags are common they can be promoted through __init__.py.

huonw:
Unfortunately, I'm not sure this is a good idea.

- most libraries use strings (and potentially other options too) e.g.:
  - `df.join(..., how="right")` in Pandas https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.join.html ,
  - `numpy.array(..., dtype="int32")`,
  - `Dense(kernel_initializer="uniform")`
- strings aren't mutable in Python (unlike `list`s and `dict`s) so the dangers are more controlled
- using enums is a bit clunky and annoying in Python, since it requires extra imports

We could potentially support both enums (good for "high assurance" cases) and strings (good for ease-of-use), but I'm not sure that's valuable. As such, I'm going to move this to Review/QA for discussion.

kjun9:
Yeah I also don't think we need enums. I'm not sure if it's python libraries in general or just the libraries that are geared towards data-science/notebook workflows, but using strings instead of enums seems to be the preferred approach.

PR #520, opened 2019-11-11
Feature/144 new stellargraph

geoffj-d61:
Notes:

- The code in this branch does not yet comprise a complete replacement for all uses of StellarGraph.
- The particular aim was to sufficiently capture the efficient handling of node and edge data in the form of Pandas data-frames and/or type-specific maps.
- The short-term goal was to document and implement enough of the StellarGraph interface to permit the node classification of the Cora dataset using GraphSAGE.
- Further work is required to finish capturing and implementing the full StellarGraph interface.
- However, the current work should serve as an example for review of the techniques used, in order to see if this approach is viable in regards to memory usage and speed of access.


stellargraph/core/graph_standard.py
```diff
@@ -0,0 +1,213 @@
...
+class StandardStellarGraph(StellarGraph):
+    """
+    Implementation based on encapsulating user-supplied edge and node data.
+    """
+
+    def __init__(self, is_directed=False, edges=None, nodes=None, **kwargs):
+        self._is_directed = is_directed
+        self._node_data = nd = self._get_node_data(nodes, kwargs)
+        self._edge_data = ed = self._get_edge_data(edges, kwargs)
+        self._edge_cache = EdgeCache(kwargs.get("edge_cache"), ed, nd)
```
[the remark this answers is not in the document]
geoffj-d61: I thought about passing in the edge cache type, but then this opens up the possibility of a user passing in their own implementation. At the time, I had specific reservations about why this could be bad, but I can't remember what they were (it might have been about exposing the interface or about testing, not sure).

stellargraph/data/edge_cache.py
```diff
@@ -0,0 +1,206 @@
...
+            if cache_type == EdgeCacheType.NONE:
+                return NoEdgeCache(*args, **kwargs)
+            if cache_type == EdgeCacheType.UNTYPED:
+                return UntypedEdgeCache(*args, **kwargs)
+            raise ValueError("Unknown cache type: {}".format(cache_type))
+        else:
+            return type.__call__(cls, *args, **kwargs)
+
+
+class EdgeCache(metaclass=EdgeCacheFactory):
```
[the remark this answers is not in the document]
geoffj-d61: The two specific data-types that @adocherty wanted to support are Pandas data-frames and a dictionary mapping types to blocks of data. My previous experiments also allowed NumPy arrays and lists of data, but I took these out for simplicity.

My basic goal was to have a unified interface, e.g. `NodeData` or `EdgeData`, that could efficiently reuse the data in the underlying data structure wherever possible, to avoid memory copying. The explosion of tests you mentioned is a valid concern, but in practice one only needs to test the (currently two) `NodeData` implementations and the (currently two) `EdgeData` implementations (some of that testing exists). I don't think it is useful beyond a basic test to combine the cross-product of all of these tests in `StellarGraph`.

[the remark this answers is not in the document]
geoffj-d61: The `EdgeCache` was an add-on that was required to speed up neighbourhood queries. The Single Responsibility principle, coupled with some deep thinking, suggests that `EdgeData` should not have to support caching, so it was designed as a wrapper around `EdgeData`.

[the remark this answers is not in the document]
geoffj-d61: In a very real sense, the `NodeData` implementation is exactly a mechanism that takes one of a variety of input data types and provides one output data type (of currently two). Previous experiments used an explicit method to do this, with explicitly documented arguments - I kept changing my mind about whether to expose this method or not.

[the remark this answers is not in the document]
geoffj-d61: I agree I have probably gone overboard with the factory instantiation approach. The benefit in the case of `StandardStellarGraph` versus `NetworkXStellarGraph` lies precisely in the fact that they currently have two very different argument lists, and this problem is hidden from the user. I'm not sure if there is a cleaner way to handle polymorphic arguments.

adocherty:
> The two specific data-types that @adocherty wanted to support are Pandas data-frames and a dictionary mapping types to blocks of data. My previous experiments also allowed NumPy arrays and lists of data, but I took these out for simplicity.

I think we need to come together as a group and discuss this more as I think there is miscommunication going on here.  What I wanted was a single underlying datatype that would be  sensible for the user to supply, performant for the algorithmic operations that we want, and be usable in-place without having to copy the data again.  

I think that a format very similar to the current Platform data format for it's Python library is a good place to start. This format would be a dictionary keyed by node type and with values as the node features and similarly for edges as I describe in this issue:
https://github.com/stellargraph/stellargraph-ml-private-issues/issues/144

Also, if you have not already done so I would take a look at the platform metadata documentation: https://github.com/stellargraph/neon/blob/develop/platform/docs/design/metadata.md

I also think Huon is exactly right that, at least for now, we should have all other graph formats converted to this internal data format.

We can discuss this further at the meeting on Monday, perhaps it would be useful to get a better feeling for the data format and the operations we can do some quick tests on performance using this format outside the StellarGraph library for now.

geoffj-d61:
Early experiments showed that a data-structure optimised for heterogeneous, type-specific access performed poorly for homogenous data. Possibly I implemented it badly, possibly there is scope for subsequent optimisations for the specifically homogenous case, or possibly the two are contra-optimal in efficiency. I suspect the latter, so in this instance I went for a common interface rather than a common implementation.

Additionally, the goalposts have drifted around somewhat over time. We have talked about having a type-specific map as the standard data-structure, but then also looked at a simple Pandas data-frame composite. Previously, we thought it was a good idea to also allow efficiency in extremely simple cases, such as the user specifying a list of source-target edge tuples and a list of node identifiers. I have been trying to satisfy the accumulated requirements.

stellargraph/data/edge_data.py
```diff
@@ -0,0 +1,937 @@
...
+                self.edge_weights(),
+            )
+        return zip(self.source_ids(), self.target_ids())
+
+
+#############################################
+# Pandas data-frame of edge data:
+
+
+class PandasEdgeData(MappedEdgeData):
```
adocherty: I think that while we should have the Pandas edgelist to store edge features and edge types, we should have a typed adjacency list (MDAL*) that will be used for fast access to in and out neighbours of nodes.

As far as I can see this is done implictly by using the `EdgeCache`, but in my opinion this adds a layer of unnecessary complexity.  I like the fact that the edge cache will avoid the start-up time of building the complete adjacency list and allows to build only the subset of the adjacency list that is actually required. However, I feel strongly that having this general `EdgeData` class is too abstract.

stellargraph/data/edge_cache.py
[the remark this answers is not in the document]
huonw: I'm going to have to keep thinking about the rest of your comments, but for now:

> The benefit in the case of StandardStellarGraph versus NetworkXStellarGraph lies precisely in the fact that they currently have two very different argument lists, and this problem is hidden from the user

I don't think the problem can be hidden from the user: the possible valid argument lists is somewhat fundamentally relevant to the user, since passing in a valid argument list is the only way to create a `StellarGraph` and use the rest of the library. If they can't work out how to construct this (i.e. what's valid to pass to a constructor of `StellarGraph`), they cannot use the library.

stellargraph/data/edge_data.py
geoffj-d61: I'm a bit confused again (if ever I wasn't before), because #144 also talks about edges being a type-map. And if heterogeneous edge data is flattened to a Pandas data-frame (not clear if we would do this or the user), then we need a systematic way of determining which columns form the features for which type. This is what EdgeData is for.
Also, the caching of edges for efficiency, e.g. MDAL*, turns out to be separate from the specification of edge data. EdgeCache was designed to allow different implementations for different usages, e.g. heterogenous versus homogeneous. I'm still worried that a one-size-fits-all is going to be inefficient for simple data.

stellargraph/data/edge_cache.py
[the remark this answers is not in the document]
geoffj-d61: True, but you're saying that means we must always have a well-defined argument list so IDEs can help us with parameters. That aim is helpful and desirable, but not even standard libraries like Keras always adhere to that.

My argument is not that we don't care or shouldn't aim for a consistent, explicit interface, but more that it is still too early to know what that looks like. For example, the `NetworkXStellarGraph` allows a `node_type_default` argument, but `StandardStellarGraph` calls it `default_node_type`. Must we explicitly allow all variants up front in the argument list, just because we haven't rationalised the nomenclature? I just want to defer that decision until we know more.

stellargraph/core/graph_standard.py
```diff
@@ -0,0 +1,213 @@
...
+    ##################################################################
+    # Computationally intensive methods:
+
+    def info(self, show_attributes: bool = True, sample: Optional[int] = None):
+        directed_str = "Directed" if self.is_directed() else "Undirected"
+        s = "{}: {} multigraph\n".format(type(self).__name__, directed_str)
+        s += " Nodes: {}, Edges: {}\n".format(
+            self.number_of_nodes(), self.number_of_edges()
+        )
+        # TODO Mimic NetworkXStellrGraph output
```
adocherty: BTW I have issues with the current info method as it parses the graph to identify the schema. I think in the future we should just report graph info that doesn't require parsing the graph, it's really annoying for the user to type `G.info()` and have to wait for 10 mins!

Also, in the future I think we should move to a pre-calculated schema in the graph class so we can display the schema info.

stellargraph/core/graph_standard.py
```diff
@@ -0,0 +1,213 @@
...
+        raise NotImplementedError
+
+    ##################################################################
+    # Private methods:
+
+    def check_graph_for_ml(self):
+        # TODO Actually do something!!
+        pass
+
+    def create_graph_schema(
```
adocherty: In the future I think we should move to a pre-calculated schema in the graph class.  How we can do this efficiently is an open question.

geoffj-d61:
Here is my current distillation of our discussions:

- `StellarGraph` and `StellarDiGraph` share a common interface. The user never sees this interface.
- `StellarGraph` and `StellarDiGraph` share a common base class, which delegates all methods to an underlying `graph` object. The user never sees the base class, nor any implementation details, other than what is visible in `StellarGraph` and `StellarDiGraph`.
- `StellarGraph` and `StellarDiGraph` have `__init__` methods that fully and explicitly list and document all parameters, for both the NetworkX and non-NetworkX implementations.
- The non-NetworkX implementation takes exactly one data-type for both its edges and nodes. This is a dictionary with the keys being strings representing edge/node types, and the values being Pandas data-frames.
- The node data-frame has an `id` column. All other columns are taken to be features.
- The edge data-frame has `id`, `src` and `dst` columns. All other columns are taken to be features.
- The non-NetworkX implementation will build an MDAL edge cache based on node types.

Notes and questions:
- If there are edge weights, these are part of the edge features. The user gets to specify the name of the edge weight column for any method that needs edge weights.
- Not sure if the user gets to specify what features they want. E.g. if we support edge features, then how do we distinguish those from the edge weights? Are the features fixed at class creation or specifiable upon demand like edge weights? If fixed, why aren't the edge weights also fixed at class creation?

stellargraph/core/graph_standard.py
adocherty: For example, with edges stored as a dictionary of edge types it will not require parsing all edges to extract the edge types.  Building up an inverse index of edge ID to edge type will also be straightforward.  Similarly for node types.

stellargraph/data/edge_cache.py
```diff
@@ -0,0 +1,206 @@
...
+            iterable: The identifiers of the neighbouring out-nodes.
+        """
+        raise NotImplementedError
+
+
+#############################################
+# Class for direct access to the edge data:
+
+
+class NoEdgeCache(EdgeCache):
```
adocherty: I'm not sure what circumstances we would not want to store the edge data as an adjacency list. What were your thoughts here?

geoffj-d61: I was thinking that in the situation where memory is very tight, we might want to wear the increased time of access to the underlying data by turning off the additional caching.

stellargraph/data/edge_cache.py
```diff
@@ -0,0 +1,206 @@
...
+    def out_nodes(self, node_id: Any) -> Iterable[Any]:
+        return self._edge_data.out_nodes(node_id)
+
+
+#############################################
+# Class for cached access to the edge data, ignoring
+# node and edge type information.
+
+
+class UntypedEdgeCache(EdgeCache):
```
adocherty: I think this is essentially SDAL?

stellargraph/data/edge_data.py
adocherty: > #144 also talks about edges being a type-map.

Yes, there is confusion here and I haven't been explaining my point of view well!  I agree with you that there needs to be three data structures: node features, edge features, and graph structure.

> and if heterogeneous edge data is flattened to a Pandas data-frame (not clear if we would do this or the user), then we need a systematic way of determining which columns form the features for which type.

I think I was wrong about this in https://github.com/stellargraph/stellargraph-ml-private-issues/issues/144,  I don't think we should flatten the array in our internal storage.  This was a potential way I though that users could want to specify the edge features, but let's focus on supporting the single unflattened format.

> Also, the caching of edges for efficiency, e.g. MDAL*, turns out to be separate from the specification of edge data

Agreed. The edge cache is the adjacency list for graph structure. 

> EdgeCache was designed to allow different implementations for different usages, e.g. heterogenous versus homogeneous. I'm still worried that a one-size-fits-all is going to be inefficient for simple data.

I did consider this, but I convinced myself that the overhead of looking up the type in the MDAL for a homogeneous graph would not be high.  However, I am not suggesting we have a one-size-fits all approach, but I'm thinking we can simplify things and for now have a single implementation and drop support for caching.

stellargraph/data/edge_cache.py
geoffj-d61: Yes, I started with the simplest case of homogeneous graphs, mainly because that is where I have spent most of my time (e.g. GraphSAGE). I have really only just started looking seriously at how heterogeneous graphs are being used (e.g. HinSAGE).

stellargraph/core/graph_standard.py
geoffj-d61: Agreed, but at the time I was being careful not to just remove any functionality that I didn't understand. I vaguely think `info()` exists purely because NetworkX has one?

geoffj-d61: And everywhere you look in the code you will find `create_graph_schema()`, so it might not be trivial to get rid of this.

adocherty:
> Here is my current distillation of our discussions:
> 
>     * `StellarGraph` and `StellarDiGraph` share a common interface. The user never sees this interface.

I'm not sure what you mean by "the user never sees this interface"?  We want a public API for the StellarGraph class so which interface are we talking about?

>     * `StellarGraph` and `StellarDiGraph` share a common base class, which delegates all methods to an underlying `graph` object. The user never sees the base class, nor any implementation details, other than what is visible in `StellarGraph` and `StellarDiGraph`.

Is this proposing we move from the current subclassing to e.g. `NetworkXStellarGraph` to having a `StellarGraph` object that has a `._graph` member that contains the actual graph information?

>     * `StellarGraph` and `StellarDiGraph` have `__init__` methods that fully and explicitly list and document all parameters, for both the NetworkX and non-NetworkX implementations.

I always like explicit parameter listing, but as you point out it's not always done in many popular libraries. I think it's a nice-to-have rather than something to drive our design.

>     * The non-NetworkX implementation takes exactly one data-type for both its edges and nodes. This is a dictionary with the keys being strings representing edge/node types, and the values being Pandas data-frames.

Agreed!

In fact, I'm actually thinking that we could move forward by dropping the NetworkX implementation entirely. We can have a converter that copies a NetworkX object to our internal format (EdgeData & adjacency list). This wouldn't be as nice as what you are proposing but will reduce the code complexity and allow us to focus on what is important for us: making our default implementation fast and memory efficient.  We can relegate NetworkX compatibility to small graphs where we don't have memory issues.

>     * The node data-frame has an `id` column. All other columns are taken to be features.

Agreed. We might want to make the `id` column the index – I've been wondering about the difference in speed between arbitrary indices using `.loc` and integer indices using `.iloc`  in Pandas DataFrames – do you have any ideas?

>     * The edge data-frame has `id`, `src` and `dst` columns. All other columns are taken to be features.

Agreed. Similarly to nodes ee might want to make the `id` column the index here.

>     * The non-NetworkX implementation will build an MDAL edge cache based on node types.

Agreed.  I think your current build-on-the-fly adjacency list is pretty cool, and I'd like to keep it.

> Notes and questions:
> 
>     * If there are edge weights, these are part of the edge features. The user gets to specify the name of the edge weight column for any method that needs edge weights.

Agreed – for methods that need edge weights they can access them in the same way as any edge feature, the graph class should not need to differentiate.
 
>     * Not sure if the user gets to specify what features they want. E.g. if we support edge features, then how do we distinguish those from the edge weights? Are the features fixed at class creation or specifiable upon demand like edge weights? If fixed, why aren't the edge weights also fixed at class creation?

I thought that edge features and edge weights should be fixed at class creation.  Why do you think the edge weights aren't fixed at class creation?

We also need to think about the `GraphSchema` and type maps, namely how can we look up the type of an edge/node quickly.

Thanks for this summary @geoffj-d61 – I think we are getting a good consensus.  Could you  write this up a bit more cleanly for Monday's meeting and we'll use it as a basis for the discussion.

stellargraph/data/edge_cache.py
adocherty: I understand your point – although this highlights a problem with the current design: For graphs with no edge features we store an edge list that has to be copied to an adjacency list for speed. To save memory we drop the adjacency list (the data structure we want for speed) leaving the edge list (the data structure we don't want). It would be better in this case to convert the edge list to an adjacency list and delete the edge list or directly ingest the adjacency list. 

Therefore I think it will be better to reduce memory and retain speed by looking at having an adjacency list but not the edge data. However, **I think we should leave these worries for the future and implement the minimum we require now**.

geoffj-d61:
My understanding from what people said is that the user should never see a `NetworkXStellarGraph` or other implementation, they should only ever see a `StellarGraph` object. This is what I meant. Similarly, the user shouldn't see `StellarGraphInterface` or `StellarGraphBase`, although the method signatures should flow though from the interface.

geoffj-d61:
In regards to Pandas `.loc` and `.iloc`, have a look at EdgeData and NodeData, I think one uses `.iloc` and the other `.loc`. The main difference is whether columns are addressed by index or name.

geoffj-d61:
Also, in regards to `id` maybe being the data-frame index, this is why I created NodeData and EdgeData, so that some of this complexity could be specified by the user and handled orthogonally to the rest of the code. I'm not sure why people are resistant to having a NodeData encapsulation (forgetting about implementations issues).

adocherty:
> In regards to Pandas `.loc` and `.iloc`, have a look at EdgeData and NodeData, I think one uses `.iloc` and the other `.loc`. The main difference is whether columns are addressed by index or name. Lookup by column name was much slower.

Then we should ensure that we do lookup by index for node node and edge data. Namely we should convert any user-specified node names to indices up front so that we can use indices from then on in the actual calculations. Perhaps in order to move towards this we can add, say `get_node_features_by_index` and `get_node_features_by_name` and the same for edges for interim compatibility and then move our algorithms to support the by index style. What are your thoughts?

geoffj-d61:
Part of the function of `NodeData` and `EdgeData` was to automatically build (where necessary) a mapping from identifier to index. This is also part of the reason why I figured it was necessary to specify columns in advance (at least for features, I guess - others could be dynamic).

Are you thinking that  `get_node_features_by_name` would be passed the column names on each call?

adocherty:
> Also, in regards to `id` maybe being the data-frame index, this is why I created NodeData and EdgeData, so that some of this complexity could be specified by the user and handled orthogonally to the rest of the code. I'm not sure why people are resistant to having a NodeData encapsulation (forgetting about implementations issues).

I think my main worry is that too much encapsulation will slow down us creating a fast implementation for a single data format.  I think we can be won over, but I'm still skeptical 😄 

> Part of the function of `NodeData` and `EdgeData` was to automatically build (where necessary) a mapping from identifier to index. This is also part of the reason why I figured it was necessary to specify columns in advance (at least for features, I guess - others could be dynamic).
> 
> Are you thinking that `get_node_features_by_name` would be passed the column names on each call?

Currently our algorithms just pass the node names to the `get_features_for_nodes` function.  I think we should move to mapping from name to index in the algorithms then passing the index of the nodes/edges. However, I think it would be good to start with backward compatibility using node names but with a view to moving to indices.

I understand part of the function of `NodeData` was to incorporate the mapping from identifier to index, but I think that this mapping could also be in the StellarGraph class and it is not really an argument for more abstraction. I'd like to hear more from you about how this is going to be applied at an algorithmic level – how can we use this mapping so that it's fast to access node and edge features for the algorithms using StellarGraph class?

stellargraph/core/graph_standard.py
adocherty: You are right, the main reason `.info` exists is it is extending the `.info` of the NetworkX graph.

No, it won't be trivial but this is an opportunity to relieve us of some technical debt so I wouldn't be constrained by us doing it another way.

geoffj-d61:
It seems to me that there are two possible implementation paths (not necessarily of equal value or feasibility):
1. The user calls initially a function that maps *every* node id to an index, and from then on only uses these indices; or
2. We have to do the id -> index lookup each time, as efficiently as possible.

If we have to do it the second way (and this is less of a constraint on the user, and minimises having them store the whole node mapping), then that lookup code has to live somewhere; whether it lives in a specific `StellarGraph` implementation or a specific `NodeData` implementation is moot. The benefit of it living in `NodeData` for me is simply the tighter coupling to the specific, underlying data-structure; the performance shouldn't vary much either way (apart from the call chain, which would be slightly faster if the functionality is in `StellarGraph`).

Apart from this `NodeData/StellarGraph` dichotomy, I'm not sure how I could make my existing implementation any faster (not saying it can't be, just that I don't know how). The current way node features are extracted is to map columns to indices (once) and pull out a single NumPy view of the features. Then every lookup first maps node identifiers to indices, then indexes the NumPy array. This was the fastest solution that I could find.

stellargraph/core/graph_standard.py
geoffj-d61: I'm certainly not arguing against doing it! Just that a lot of code will initially be broken. Some will be trivial (find and delete `info()` calls, but I think the graph schema stuff is going to be tricker.

adocherty:
> It seems to me that there are two possible implementation paths (not necessarily of equal value or feasibility):
> 
>     1. The user calls initially a function that maps _every_ node id to an index, and from then on only uses these indices; or
> 
>     2. We have to do the id -> index lookup each time, as efficiently as possible.
>  
> If we have to do it the second way (and this is less of a constraint on the user, and minimises having them store the whole node mapping), then that lookup code has to live somewhere; whether it lives in a specific `StellarGraph` implementation or a specific `NodeData` implementation is moot. The benefit of it living in `NodeData` for me is simply the tighter coupling to the specific, underlying data-structure; the performance shouldn't vary much either way (apart from the call chain, which would be slightly faster if the functionality is in `StellarGraph`).

We can use the look-up function in Pandas – using `.loc` allows us to use the named indices.

For performance it would be idea to have the ability to use integer indices both for looking up the node/edge data and in the adjacency list (edge cache). The goal here would be that user-facing methods would use node indices (for example the generators, the random walkers etc) and would convert them to integer indices under the hood which would then be used with the StellarGraph class. However, that would essentially assume case 1 when the user interacted with the StellarGraph class which would be confusing for the user (and likely for us).

> Apart from this `NodeData/StellarGraph` dichotomy, I'm not sure how I could make my existing implementation any faster (not saying it can't be, just that I don't know how). The current way node features are extracted is to map columns to indices (once) and pull out a single NumPy view of the features. Then every lookup first maps node identifiers to indices, then indexes the NumPy array. This was the fastest solution that I could find.

I have some timing of different approaches using Pandas and numpy here:
https://github.com/stellargraph/stellar-research/blob/develop/stellargraph_class/feature_sampling_timing_tests.ipynb

geoffj-d61:
Hi @kjun9, when the new teams start next year, I wouldn't mind if you or someone could take a look at this issue/PR, just so people know what work I did toward a new, non-NetworkX implementation of StellarGraph.

The design and implementation consists of:
- NodeData: an interface to node data, including implicit or explicit node identifiers, explicit or default node types, and optional node features. Two implementations exist for Pandas data-frames and dictionaries of type-to-data-blocks.

- EdgeData: like NodeData, but for edges (and not including edge features).

- EdgeCache: An interface to speed up access to edge neighbourhood queries. Two implementations include a null-cache (that uses the underlying EdgeData), and a dictionary of lists (DL) cache. The cache used is specifiable.

- StandardStellarGraph: An implementation of enough of the interface to permit node-classification of the Cora data-set using GraphSAGE.

The whole design/implementation was deemed to be overly complex, but from a software engineering perspective, every decision point was made after careful consideration of the various edge cases.

It would be nice if at least some of the design was salvageable.

huonw:
I've picked this work up; I've done the merge in #597 (a separate PR while it's a draft to try to minimise the notifications going to anyone involved with this particular one), and have split out some small pieces (#598, #599, #600) since there's a relatively clear plan for those chunks as set out by this work. I'll leave this PR open for now, as a reminder to go through all the discussion here to make sure it's reflected, and may even switch back to this one for the final review (for continuity).

huonw:
I didn't end up continuing with #597, but rather took the ideas here and in the other research done by @geoffj-d61 and @adocherty to implement a more distilled version of this in #668. That has now landed and I'm continuing with various bits of follow up work to perfect it. As such, I'm closing this.

PR #521, opened 2019-11-14
Feature/clustergcn

PantelisElinas:
Implementation of ClusterGCN with accompanying demo for node classification.

[1], Cluster-GCN: An Efficient Algorithm for Training Deep and Large Graph Convolutional Networks, W. Chiang, X. Liu, S. Si, Y. Li, S. Bengio, C. Hsieh, KDD, 2019.

#### Reviewer Checklist
- [x] Code correctly implements Cluster-GCN algorithm
- [x] Unit tests
- [x] Demo notebook runs and clearly demonstrates/explains how to train Cluster-GCN models
- [x] Documentation


demos/README.md
```diff
@@ -7,7 +7,7 @@ examples demonstrate using the `StellarGraph` library to build machine learning
 workflows on both homogeneous and heterogeneous networks.
 
 Each folder contains one or more examples of using the StellarGraph implementations of the
-state-of-the-art algorithms, attri2vec [4], GraphSAGE [3], HinSAGE, GCN [6], GAT [7], PPNP/APPNP [9], 
+state-of-the-art algorithms, attri2vec [4], GraphSAGE [3], HinSAGE, GCN [6], GAT [7], Cluster-GCN [10], PPNP/APPNP [9],
```
geoffj-d61: What does the ordering of this list imply?

docs/api.txt
```diff
@@ -19,7 +19,7 @@ Generators
 -----------
 
 .. automodule:: stellargraph.mapper
-  :members: FullBatchNodeGenerator, GraphSAGENodeGenerator, GraphSAGELinkGenerator, HinSAGENodeGenerator, HinSAGELinkGenerator, Attri2VecNodeGenerator, Attri2VecLinkGenerator
+  :members: FullBatchNodeGenerator, GraphSAGENodeGenerator, ClusterNodeGenerator, GraphSAGELinkGenerator, HinSAGENodeGenerator, HinSAGELinkGenerator, Attri2VecNodeGenerator, Attri2VecLinkGenerator
```
geoffj-d61: Again, is there a particular reason this was not just placed at the end of this list? I almost thought you were grouping node generators together, except that `HinSAGENodeGenerator` still comes after `GraphSAGELinkGenerator`.

tests/layer/test_cluster_gcn.py
```diff
@@ -0,0 +1,241 @@
...
+    # Test with final_layer=False
+    out = ClusterGraphConvolution(2, final_layer=False)([x_t, output_indices_t, A_mat])
+    model = keras.Model(inputs=[x_t, A_t, output_indices_t], outputs=out)
+    preds = model.predict([x, adj, out_indices], batch_size=1)
+    assert preds.shape == (1, 3, 2)
+
+    # Now try with final_layer=True
+    out = ClusterGraphConvolution(2, final_layer=True)([x_t, output_indices_t, A_mat])
+    # The final layer removes the batch dimension and causes the call to predict to fail.
+    # We are going to manually added the batch dimension before calling predict.
```
geoffj-d61: `add` not `added`.

tests/layer/test_cluster_gcn.py
```diff
@@ -0,0 +1,241 @@
...
+    out = K.expand_dims(out, 0)
+    model = keras.Model(inputs=[x_t, A_t, output_indices_t], outputs=out)
```
geoffj-d61: Why does the `Model` and `ClusterGraphConvolution.__call__` have the arguments in different orders?

tests/layer/test_cluster_gcn.py
```diff
@@ -0,0 +1,241 @@
...
+        layer_sizes=[2], generator=generator, activations=["relu"], dropout=0.5
+    )
+
+    assert cluster_gcn_model.layer_sizes == [2]
+    assert cluster_gcn_model.activations == ["relu"]
+    assert cluster_gcn_model.dropout == 0.5
+
+
+def test_ClusterGCN_apply():
+    G, features = create_graph_features()
```
geoffj-d61: Given that this block of code is (mostly) reused multiple times, I would encapsulate it along with a switch for computing `adj`, e.g.: ```def create_stellargraph(with_adj=False):
    Gnx, features = create_graph_features()
    nodes = Gnx.nodes()
    node_features = pd.DataFrame.from_dict(
        {n: f for n, f in zip(nodes, features)}, orient="index"
    )
    G = StellarGraph(Gnx, node_features=node_features)
    if with_adj:
        adj = nx.to_numpy_array(Gnx)[None, :, :]
        return G, adj
    return G```

tests/mapper/test_cluster_gcn_node_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+import pandas as pd
+import numpy as np
+import pytest
+
+
+def create_graph_features():
+    G = nx.Graph()
+    G.add_nodes_from(["a", "b", "c", "d"])
+    G.add_edges_from([("a", "b"), ("b", "c"), ("a", "c"), ("b", "d")])
+    G = G.to_undirected()
```
geoffj-d61: My understanding is that `nx.Graph` is already undirected.

tests/mapper/test_cluster_gcn_node_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+    return G, np.array([[1, 1], [1, 0], [0, 1], [0.5, 1]])
+
+
+def test_ClusterNodeSequence_init():
+    G, features = create_graph_features()
```
geoffj-d61: See my earlier comment about encapsulating the creation of a `StellarGraph`.

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -0,0 +1,361 @@
...
+    Args:
+        G (StellarGraphBase): a machine-learning StellarGraph-type graph
+        k (int): The number of clusters if parameter `clusters` is None. Otherwise it is ignored.
+        lam (float): The mixture coefficient for adjacency matrix normalisation.
+        clusters (list): a list of lists of node IDs such that each list corresponds to a cluster of nodes
+            in G. The clusters should be non-overlapping. If None, the G is clustered into k clusters.
+        name (str): an optional name of the generator
+    """
+
+    def __init__(self, G, k=1, q=1, lam=0.1, clusters=None, name=None):
```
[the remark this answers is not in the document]
geoffj-d61: @adocherty , what is our position on the use of cryptic parameter names? Do we not care because they are documented, or do we want the names to be more mnemonically aligned to their purpose?

geoffj-d61: Since `k` is optional if `clusters` is not `None`, I would put it after `clusters` in the parameters. Actually, I would put `clusters` next after `G` as the second most important parameter. Also, you could consider having `clusters` be polymorphic, i.e. a list if you are actually specifying clusters, or an integer if you are specifying the number of clusters.

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -0,0 +1,361 @@
...
+        self.k = k
+        self.q = q  # The number of clusters to sample per mini-batch
+        self.lam = lam
+        self.clusters = clusters
+
+        if clusters:
+            self.k = len(clusters)
+
+        # Some error checking on the given parameter values
+        if lam < 0 or lam > 1 or not isinstance(lam, float):
```
geoffj-d61: Type test should go before value tests.

geoffj-d61: And technically, our newest convention is that failure of the type test should result in a `TypeError`.

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -0,0 +1,361 @@
...
+        G.check_graph_for_ml()
+
+        self.node_list = list(G.nodes())
+
+        # We need a schema to check compatibility with ClusterGCN
+        self.schema = G.create_graph_schema(create_type_maps=True)
+
+        # Check that there is only a single node type
+        if len(self.schema.node_types) > 1:
+            raise TypeError(
```
geoffj-d61: `ValueError` here.

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -0,0 +1,361 @@
...
+            if not is_real_iterable(targets):
+                raise TypeError(
+                    "{}: Targets must be an iterable or None".format(
+                        type(self).__name__
+                    )
+                )
+
+            # Check targets correct shape
+            if len(targets) != len(node_ids):
+                raise TypeError(
```
geoffj-d61: `ValueError` here.

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -0,0 +1,361 @@
...
+                )
+
+        return ClusterNodeSequence(
+            self.graph,
+            self.clusters,
+            targets=targets,
+            node_ids=node_ids,
+            q=self.q,
+            name=name,
+        )
```
geoffj-d61: Do we not pass `self.lam` into `ClusterNodeSequence`?

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -0,0 +1,361 @@
...
+        self.normalize_adj = normalize_adj
+        self.q = q
+        self.lam = lam
+        self.node_order = list()
+        self._node_order_in_progress = list()
+        self.__node_buffer = dict()
+
+        if len(clusters) % self.q != 0:
+            raise ValueError(
+                "The number of clusters should be exactly divisible by q. However {} is not exactly divisible by {}".format(
```
geoffj-d61: Maybe
```The number {} of clusters is not exactly divisible by q={}```
?

File: stellargraph/mapper/mini_batch_node_generators.py [hunk 14]
```diff
@@ -0,0 +1,361 @@
...
+                raise ValueError(
+                    "When passed together targets and indices should be the same length."
+                )
+
+            self.target_ids = list(node_ids)
+            self.targets = np.asanyarray(targets)
+            self.target_node_lookup = dict(
+                zip(self.target_ids, range(len(self.target_ids)))
+            )
+            self.target_ids = list(node_ids)
```
geoffj-d61: I count three times this assignment is (potentially) made. Some code rationalisation is in order.

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def __len__(self):
+        num_batches = len(self.clusters_original) // self.q
+        return num_batches
+
+    def __getitem__(self, index):
+        # The next batch should be the adjacency matrix for the cluster and the corresponding feature vectors
+        # and targets if available.
+        cluster = self.clusters[index]
+        g_cluster = self.graph.subgraph(
```
geoffj-d61: This is going to cause me some fun in the branch where `graph` is no long a `NetworkX` object!

demos/node-classification/cluster-gcn/cluster-gcn-node-classification.ipynb
```diff
@@ -0,0 +1,1657 @@
...
+   "metadata": {},
+   "source": [
+    "### Train the model"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We are not ready to train the `ClusterGCN` model, keeping track of its loss and accuracy on the training set, and its generalisation performance on a validation set."
```
geoffj-d61: Should this be `We are now`?

stellargraph/layer/cluster_gcn.py
```diff
@@ -0,0 +1,387 @@
...
+            x (Tensor): input tensors
+
+        Returns:
+            Output tensor
+        """
+        x_in, out_indices, *As = x
+
+        Ainput = [Lambda(lambda A: K.squeeze(A, 0))(A) for A in As]
+
+        if len(Ainput) != 1:
```
geoffj-d61: Are there future plans to allow multiple adjacency matrices to be passed in? If not, then this all seems overly elaborate.

demos/README.md
PantelisElinas: Nothing, it is just the list of algorithms we have implemented.

stellargraph/layer/cluster_gcn.py
PantelisElinas: I've removed it.

stellargraph/mapper/mini_batch_node_generators.py
PantelisElinas: I'll update accordingly, but this newest convention is news to me. Where have we documented these conventions for testing?

stellargraph/mapper/mini_batch_node_generators.py
PantelisElinas: This is probably up to debate as the problem is the the graph has more than one node types. Our FullBatchNodeGenerator that we use for our GCN implementation return TypeError for the same check, so for consistency, I will not be changing it.

stellargraph/mapper/mini_batch_node_generators.py
PantelisElinas: Fixed!

stellargraph/mapper/mini_batch_node_generators.py
geoffj-d61: Not sure, but it was the end result of conversations between @adocherty and myself, which I gather also come from the Python convention regarding errors.

stellargraph/mapper/mini_batch_node_generators.py
geoffj-d61: Yes, but the type of `self.schema.node_types` is not the problem here (it is presumably a list), it is the value of that variable that is wrong, hence `ValueError`. The use of `TypeError` has nothing to do with node types. 

And I have to say that it is a facile argument to not make a suggested correction just because the same problem occurs in other places! We should be making our (collective `our`) code cleaner whenever we touch it, not leave the code smell intact.

stellargraph/mapper/mini_batch_node_generators.py
PantelisElinas: I've corrected this. Certainly not necessary to assign it 3 times!

tests/layer/test_cluster_gcn.py
PantelisElinas: Because the TensorFlow team won't follow our lead!

geoffj-d61: So doing the following wouldn't work?
```    model = keras.Model(inputs=[x_t, output_indices_t, A_t], outputs=out)
    preds = model.predict([x, out_indices, adj], batch_size=1)
```

PantelisElinas: On first reading I misunderstood what parameter you had an issue with; I should have had a closer look.

At any rate, the above would work. But it would not be consistent with our GCN implementation this is based on. I know you don't like the justification but this is why it is the way it is. Basically, I like to be consistent in our inconsistency to prevent tripping users. If we think this is an important issue, then we should create another ticket and update across the board.

tests/layer/test_cluster_gcn.py
PantelisElinas: I've simplified it. Thank you!

tests/mapper/test_cluster_gcn_node_mapper.py
PantelisElinas: You are correct. I've removed the extra call.

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -251,12 +260,10 @@ def __init__(
                     "When passed together targets and indices should be the same length."
```
geoffj-d61: `targets is not None` (above) must already be true, since it was tested previously.

tests/layer/test_cluster_gcn.py
geoffj-d61: Oh ok, it is done that way because our GCN is one that way. I understand now.

geoffj-d61:
In `stellargraph/layer/cluster_gcn.py`, copyright should just be 2019 not 2018-2019?

stellargraph/mapper/mini_batch_node_generators.py
PantelisElinas: Good catch!

stellargraph/mapper/mini_batch_node_generators.py
PantelisElinas: I've changed it to ValueError. I don't think this is issue is important enough to spend too much of our time arguing the correct exception to return.

stellargraph/mapper/mini_batch_node_generators.py
PantelisElinas: I'll update to remove k and accept clusters as either int or list. It's a good suggestion. One less parameter to worry about and less complicated logic to figure out what the user wants to do and less likely for her to call the method incorrectly.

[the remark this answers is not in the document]
PantelisElinas: I've removed k so it is not an issue anymore. The parameters q and lam are the same as the symbols used in the paper. For the latter, I could not use lambda for the obvious reason so I thought lam is close enough.

geoffj-d61:
Unless I've lost track of some issues, this is looking pretty good. Still outstanding:
- I made a comment about some text in the Jupyter notebook being possibly incorrect (appears above, somewhere).
- I made a comment somewhere on a commit to the effect that if `cluster_gcn.py` is a new file, then the copyright should just be for 2019.

geoffj-d61:
@PantelisElinas , I think you need to run `black` on tests and stellargraph.

PantelisElinas:
@geoffj-d61 

I think the code is ready for your review now.

Regards,

P.

PR #523, opened 2019-11-14
Added script to format and remove warnings from demo notebooks

adocherty:
## Description

The script `format_notebooks.py` has been added which formats and prepares Notebooks for uploading to GitHub by doing the following:
* Removes Tensorflow warnings
* Removes `stderr` outputs
* Formats all code cells using Black
* Renumbers all code cells starting from 1
* Sets the default kernelspec to "Python 3"

These are all optional. Currently they are set as options in the script, TODO is to make these cmd line arguments.

## Review Checklist

I would suggest trying modifying one of the demo notebooks by changing kernel, running to get Tensorflow warning messages, and making the code ugly (e.g. code in a few cells is all one line). Run the format script and check that the notebook is cleaned, has the default kernel again, and the code is reformatted.

- [ ] Check `scripts/README.md` is well written and contains required info for runnning scripts.
- [ ] Check `format_notebooks.py` script works as expected and produces check-in ready notebooks.


huonw:
In neon, we require notebooks have no output (https://github.com/stellargraph/neon/issues/1894, https://github.com/stellargraph/neon/pull/1903), which we did because we decided the pros outweighed the cons:

> Pros:
> 
> - the diffs are more focused (especially for nondeterministic code, where the output may change for consecutive reruns even without the code changing, so every checkin will be changing that code)
> - users don't have to remember to update the output before committing
> - the notebooks are smaller on disk and more focused (especially for notebooks that contain plots/images and other large output)
> 
> Cons:
> 
> - someone cannot view the notebook on GitHub and see an example of the output

I've found that the third point in particular can result in huge repository sizes very quickly, e.g. https://github.com/stellargraph/neon/pull/1615 noted that the scenario-2 repo that contained 28 commits of a few notebooks was ~7MB compressed (with ~3MB of images). That could be cut by 70% compressing the images to 400k and stripping out all the notebook outputs.

Is the pros/cons balance different for this repo?

adocherty:
Hi @huonw I think the pro/cons balance is different for stellargraph library.  Here we really want people to see the demos easily in the repo so they can more easily see what the library does (we want to increase adoption).

However, I was thinking that another way (perhaps better) to do this would be to use this script with the `--html` option to convert the notebooks with outputs to HTML files to be stored somewhere, then the notebooks themselves could have the output removed.

However, for now I think outputs are useful to potential users. Although I have no data to back that up...

scripts/README.md
```diff
@@ -0,0 +1,66 @@
+## Development Scripts Readme
+
+This directory contains scripts to perform testing and maintainence tasks for developers of the StellarGraph  library. These scripts should not be needed for users of the library.
+
+Currently there are two scripts, one to formats and clean the specified Juptyer notebooks, the other to run specific demo notebooks and check if there are any errors.  In the future the functionality of these two scripts may be combined.
+
+
+### Format and clean up demo Jupyter notebooks
+
+The script `format_notebooks.py` uses `nbconvert` to process all code and output cells in the specified locations optionally remove Tensorflow warnings and stderr outputs, format and number the code cells, and set the kernel to the default of "Python 3".
```
PantelisElinas: @adocherty Do we need a requirements.txt for the scripts?

When I try to run the following command,

> python scripts/format_notebooks.py -h

I get the following error,

```
  File "scripts/format_notebooks.py", line 32, in <module>
    from black import format_str, FileMode, InvalidInput
ModuleNotFoundError: No module named 'black'
```

Clearly, I can install black to get around this problem, but what about the version. Also, do we need to install nbconvert (is this another python library?)

I had a look at the StellarGraph requirements.txt and the above are not included.

adocherty: I've added a requirements.txt file containing black, nbconvert, (part of and installed by Jupyter so should be covered, but just in case) and treon (for the other demo script).  Added a requirements section in the readme.

PR #525, opened 2019-11-17
Fixed appnp to stop excess dropout propagation and added change to pp…

kieranricardo:
This fixes the issue in the APPNP class that causes appnp to propagate excessive dropout layers.  This also adds a fix into the demo so that the softmax layer is no longer propagated. The demo scaleable and regular appnp structures should now be the same.


kieranricardo:
@youph 

- I combined all layers into self._layers
- great idea, I think there's a fairly easy way to refactor this but it'll involve an API change. Is this still suitable for a hotfix? It might be better suited for the next release
- I think you're right that the dropout after each propagation step is unnecessary (I only had them because the paper did). I removed these

youph:
I reverted to thinking that dropout layers between the propagation layers should be there, for the following reasons:
- having them removes the gap between train and validation loss for PPNP and APPNP models, i.e., prevents overfitting, albeit without improving the final validation loss (i.e., without affecting model generalisation). However, it is possible that in some cases preventing overfitting will actually be beneficial for generalisation.
- the paper applies dropout: they ```use dropout with dropout rate d = 0.5 on both layers and the adjacency matrix```

youph:
Adding the propagation dropout parameter changes the API, and is not suitable for the hotfix. But this is a good idea nevertheless, so let's create a new issue and add it in a separate PR into `develop`.

Issue #528, opened 2019-11-25
Write benchmarks for core stellargraph operations


huonw:
### Description

Per the stellargraph planning meeting, we should have some benchmarks to validate and track the performance of any changes to the `StellarGraph` class as part of #325.

Create benchmarks for important measures:
- Memory usage of StellarGraph creation
- Performance of sampling neighbours
- Performance of random access of node and edge features

### Done Checklist (Development)
- [ ] Assumptions of the user story met
- [ ] Produced code for required functionality
- [ ] Branch and Pull Request build on CI
- [ ] Branch and Pull Request pass unit tests on CI
- [ ] Branch and Pull Request pass integration tests on CI
- [ ] Version number reflects new status
- [ ] Peer Code Review Performed
- [ ] Code well commented
- [ ] Documentation on Google Docs
- [ ] Documentation in repo
- [ ] CHANGELOG.md updated
- [ ] Team demo
- [ ] Mini-meetup talk
- [ ] Stakeholder sign-off

PR #529, opened 2019-11-25
Add benchmarks measuring the time of neighbour and feature queries

huonw:
This adds two benchmarks that construct a small graph (100 nodes, 200 edges)
and do:

- "get neighbours" on every node in the graph
- "get features" on a random subset of the nodes in the graph, with 4 different configurations, along 2 independent axes:
   - either homogeneous or heterogenous (with 4 node types)
   - "specifying" the node type (via the `node_type` argument) or "inferring" it

Current output:

```
$ py.test tests -k benchmark_get
==================================================================================================================================================== test session starts =====================================================================================================================================================
platform darwin -- Python 3.6.9, pytest-5.3.0, py-1.8.0, pluggy-0.13.1
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/wil9dw/projects/stellargraph/stellargraph
plugins: benchmark-3.2.2
collected 302 items / 293 deselected / 9 selected                                                                                                                                                                                                                                                                            

tests/core/test_stellargraph.py .........                                                                                                                                                                                                                                                                              [100%]
...
------------------------------------------ benchmark 'StellarGraph neighbours': 1 tests ------------------------------------------
Name (time in us)                     Min       Max     Mean  StdDev   Median     IQR   Outliers  OPS (Kops/s)  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------
test_benchmark_get_neighbours     21.1061  133.4969  23.1896  6.8984  21.4381  1.0589  1088;1128       43.1227   21868           1
----------------------------------------------------------------------------------------------------------------------------------

-------------------------------------------------------------------------------- benchmark 'StellarGraph node features': 4 tests ---------------------------------------------------------------------------------
Name (time in us)                              Min                 Max               Mean             StdDev             Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_get_features[specify-4]     25.8561 (1.0)      205.9911 (1.03)     29.6558 (1.0)       8.6885 (1.0)      27.5809 (1.0)       1.1480 (1.06)    1545;2098       33.7202 (1.0)       24575           1
[...]

PR #530, opened 2019-11-26
Add benchmark for the memory use of creating a StellarGraph object

huonw:
This (ab)uses pytest-benchmark's ability to add a custom timer, by
creating a "timer" based on allocations. This timer gives the delta of
the bytes allocated and freed between the two "now" instants
(`snapshot` calls), and so can be used to track the size of an object
created after the first one, and held alive until after the second
one. For instance:

```python
start = snapshot()
keep_it_alive = create_some_object()
end = snapshot()

# doesn't need to be deleted, but can be, as long as the variable outlives the
# call for `end`
del keep_it_alive

print(end - start, "bytes")
```

The allocations are measured using
[`tracemalloc`](https://docs.python.org/3/library/tracemalloc.html),
available in Python 3.4 and later (and, since Python 3.6, has numpy
integration:
https://github.com/numpy/numpy/tree/master/tools/allocation_tracking).

The current results are:

```
$ pytest tests/ -k test_allocation
...

------------------------------------------------------------------------------------------------------------ benchmark 'StellarGraph creation': 6 tests ------------------------------------------------------------------------------------------------------------
Name (time in s)                                                                Min                       Max                      Mean                StdDev                    Median                IQR            Outliers     OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_allocation_benchmark_creation_from_networkx[None-0-0]               1,512.0000 (1.0)          2,080.0000 (1.0)          1,578.8000 (1.0)        121.9429 (1.08)         1,576.0000 (1.0)      64.0000 (inf)           1;1  0.0006 (1.0)          20           1
test_allocation_benchmark_creation_from_networkx[100-0-0]                1,576.0000 (1.04)         2,080.0000 (1.0)          1,601.2000 (1.01)       112.6978 (1.0)          1,576.0000 (1.0)       0.0000 (1.0)           1;1  0.0006 (0.99)         20           1
test_allocation_benchmark_creation_from_networkx[None-100-200]         143,752.0000 (95.07)      157,552.0000 (75.75)      144,578.0000 (91.57)    3,054.0825 (27.10)      143,896.0000 (91.30)     0.0000 (1.0)           1;3  0.0000 (0.01)         20           1
test_allocation_benchmark_creation_from_networkx[100-100-200]          232,336.0000 (153.66)     246,184.0000 (118.36)     233,034.8000 (147.60)   3,095.1325 (27.46)      232,336.0000 (147.42)    0.0000 (1.0)           1;2  0.0000 (0.01)         20           1
test_allocation_benchmark_creation_from_networkx[None-1000-5000]     3,170,936.0000 (>1000.0)  3,184,720.0000 (>1000.0)  3,172,150.0000 (>1000.0)  2,961.5994 (26.28)    3,171,512.0000 (>1000.0)   0.0000 (1.0)           1;4  0.0000 (0.00)         20           1
test_allocation_benchmark_creation_from_networkx[100-1000-5000]      4,006,040.0000 (>1000.0)  4,019,984.0000 (>1000.0)  4,007,377.2000 (>1000.0)  2,971.6428 (26.37)    4,006,744.0000 (>1000.0)   0.0000 (1.0)           1;4  0.0000 (0.00)         20           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
```

Why use pytest-benchmark?

Pros:

- built-in outlier detection, summarisation

- built-in saving and comparison
  (https://pytest-benchmark.readthedocs.io/en/latest/comparing.html)

Cons:

- the units are seconds instead of bytes

- requires a few moving parts, although the current implementation has
  checking/exceptions for the failure cases (not specifying the custom
  timer, and not returning an appropriate object)

https://github.com/ionelmc/pytest-benchmark/issues/28 is related.

See: #528


tests/test_utils/alloc.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2017-2018 Data61, CSIRO
```
geoffj-d61: If this is a new file, then copyright should be 2019.

tests/core/test_stellargraph.py
```diff
@@ -427,3 +428,44 @@ def test_feature_conversion_from_iterator():
...
+
+@pytest.mark.benchmark(group="StellarGraph creation", timer=snapshot)
+# various element counts, to give an indication of the relationship
+# between those and memory use (0,0 gives the overhead of the
+# StellarGraph object itself, without any data)
+@pytest.mark.parametrize("num_nodes,num_edges", [(0, 0), (100, 0), (100, 200)])
+# various feature sizes (including no features) to capture that cost
+@pytest.mark.parametrize("feature_size", [None, 1, 100])
+def test_benchmark_creation(allocation_benchmark, feature_size, num_nodes, num_edges):
+    g = example_benchmark_graph(feature_size, num_nodes, num_edges)
```
geoffj-d61: Isn't `g` here already a StellarGraph? Why does `f()` wrap it again? Shouldn't `f()` just return `g`?

tests/test_utils/alloc.py
huonw:
```suggestion
# Copyright 2017-2019 Data61, CSIRO
```

tests/core/test_stellargraph.py
huonw: No, the benchmarked function `f` needs to do any allocations inside it for them to be measured correctly. This is written a bit strangely while `StellarGraph` inherits from `nx.Graph`, because there's no way to refer to a `StellarGraph` and its data as purely a networkx object. Once #501 lands, and, supposing we had a `.to_networkx` method, I'd rewrite this as is as:

```python
g = example_benchmark_graph(...)
nx_g = g.to_networkx()

def f():
    if feature_size is None:
        return StellarGraph(nx_g)
    else:
        ...
```

Your comment makes me realise:

- the name of the function could be improved: I'll rename it to `test_benchmark_creation_from_networkx`. 
- maybe `example_benchmark_graph` should be changed to return the networkx graphs directly

I'll also try to add a benchmark that uses the new style `node_features` method to specify features, but this benchmark is enough to validate the `alloc.py` approach I've gone with.

tests/test_utils/alloc.py
geoffj-d61: The paradigm we've been using is that any file created in 2019 is copyrighted just for 2019, not from the start of the project. Any modifications in 2020 would then change the copyright to 2019-2020.

huonw:
One thing to be clear here is: this is (trying to) measure the total size of _new allocations that continue living_ after the benchmarked function exits: it is _not_ measuring the total amount of allocations in that period, or the peak memory usage. This is because:

- I didn't think of doing it another way, at the time
- it is a better measure of the long-term impact on memory of creating the object in question (that is, if you create a `StellarGraph` object, how much memory does exactly that object take up, once you start doing other things)

---

My impression is that any temporary objects created during construction are typically going to be small, and most large values created are going to be long-lived. If it's likely to be useful, we could have separate benchmarks that use [`get_traced_memory`](https://docs.python.org/3/library/tracemalloc.html#tracemalloc.get_traced_memory) to measure the peak memory use as well.

I experimented with the above and found the following very approximate/preliminary results:

|features in nx|feature size|# nodes|# edges|measurement|bytes|increase (&times;)|
|---|---|---|---|---|---|---|
|False|1|0|0 |long living|4504||
|False|1|0|0 |peak|5072 |1.13|
|False|1|100|0 |long living|67768||
|False|1|100|0 |peak|73512 |1.08|
|False|1|100|200 |long living|189504||
|False|1|100|200 |peak|221688 |1.17|
|False|100|0|0 |long living|4504||
|False|100|0|0 |peak|5072 |1.13|
|False|100|100|0 |long living|150968||
|False|100|100|0 |peak|166192 |1.10|
|False|100|100|200 |long living|272128||
|False|100|100|200 |peak|287352 |1.06|
|False|None|0|0 |long living|4504||
|False|None|0|0 |peak|5072 |1.13|
|False|None|100|0 |long living|61856||
|False|None|100|0 |peak|73512 |1.19|
|False|None|100|200 |long living|184960||
|False|None|100|200 |peak|220632 |1.19|
|True|1|0|0 |long living|4504||
|True|1|0|0 |peak|5072 |1.13|
|True|1|100|0 |long living|67648||
|True|1|100|0 |peak|74864 |1.11|
|True|1|100|200 |long living|190816||
|True|1|100|200 |peak|220696 |1.16|
|True|100|0|0 |long living|4504||
|True|100|0|0 |peak|5072 |1.13|
|True|100|100|0 |long living|150848||
|True|100|100|0 |peak|157936 |1.05|
|True|100|100|200 |long living|272352||
|True|100|100|200 |peak|279440 |1.03|
|True|None|0|0 |long living|4504||
|True|None|0|0 |peak|5072 |1.13|
|True|None|100|0 |long living|61856||
|True|None|100|0 |peak|73512 |1.19|
|True|None|100|200 |long living|182176||
|True|None|100|200 |peak|220344 |1.21|

That is, the peak memory usage of creating a `StellarGraph` is currently somewhere between 3% and 21% larger than the memory use of a long-lived `StellarGraph`.

tests/test_utils/alloc.py
huonw: 👍 

```suggestion
# Copyright 2019 Data61, CSIRO
```

FYI, the "suggestions" button on GitHub (the little page icon contain a +/-) makes it easy to be clear exactly what's meant for single-line changes like this and easy for the PR author to apply.

geoffj-d61: Thanks, I got Terry to show me where that *tiny* little button lives!!

tests/core/test_stellargraph.py
```diff
@@ -19,6 +19,7 @@
 import networkx as nx
 from stellargraph.core.graph import *
 from stellargraph.data.converter import *
+from ..test_utils.alloc import snapshot, allocation_benchmark
```
geoffj-d61: Copyright (earlier in file - can't seem to make suggestions for unmodified lines) should become `2017-2019` as this file has been touched.

huonw: I've updated it, but this seems like somewhat distracting busy-work for something of limited value, especially since these aren't the first edits this year (although they are the first non-trivial ones):

image

geoffj-d61: I agree, others clearly failed to follow the procedure. But since it is a publicly visible library, it is part of the standard checks, like blacking the code and running the unit tests. Might as well fix these things when we notice them, trivial though they are.

tests/core/test_stellargraph.py
```diff
@@ -427,3 +429,59 @@ def test_feature_conversion_from_iterator():
...
+    else:
+        node_features = {}
+        for ty in range(n_types):
+            type_nodes = range(ty, n_nodes, n_types)
+            if len(type_nodes) > 0:
+                node_features[ty] = pd.DataFrame(
+                    [np.ones(feature_size)] * len(type_nodes), index=type_nodes
+                )
+
+    return G, node_features
```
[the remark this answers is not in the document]
huonw: I agree about consolidating.

> when we merge these it seems we will have some conflicts with two functions example_benchmark_graph that take different args and return different objects.

I was going to merge #529 first and then resolve the conflicts in this PR.

tests/core/test_stellargraph.py
```diff
@@ -427,3 +429,59 @@ def test_feature_conversion_from_iterator():
...
+
+@pytest.mark.benchmark(group="StellarGraph creation", timer=snapshot)
+# various element counts, to give an indication of the relationship
+# between those and memory use (0,0 gives the overhead of the
+# StellarGraph object itself, without any data)
+@pytest.mark.parametrize("num_nodes,num_edges", [(0, 0), (100, 0), (100, 200)])
+# various feature sizes (including no features) to capture that cost
+@pytest.mark.parametrize("feature_size", [None, 1, 100])
+# test both features
+@pytest.mark.parametrize("features_in_nodes", [False, True])
```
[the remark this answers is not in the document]
huonw: Do you have an alternative? I suspect that the `features_in_node` option will become irrelevant and thus totally disappear with a non-networkx implementation, but for now it's useful as a baseline? But... maybe that's irrelevant anyway, and we can still remove?

That is, remove this and add homogeneity

```suggestion
@pytest.mark.parametrize("num_types", [1, 4])
```

or, keep it and add homogeneity?

```suggestion
@pytest.mark.parametrize("features_in_nodes", [False, True])
@pytest.mark.parametrize("num_types", [1, 4])
```

tests/test_utils/alloc.py
```diff
@@ -0,0 +1,88 @@
...
+    marker = request.node.get_closest_marker("benchmark")
+    options = dict(marker.kwargs) if marker else {}
+    correct_timer = "timer" in options and options["timer"] is snapshot
+    if not correct_timer:
+        raise ValueError(
+            "allocation_benchmark fixture can only be used in functions with @pytest.mark.benchmark(..., timer=%s.%s, ...)"
+            % (__name__, snapshot.__name__)
+        )
+
+    benchmark.extra_info["allocation_benchmark"] = True
```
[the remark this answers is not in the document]
huonw: It's not required, but I thought it would be nice to note that the "timing" isn't actually times, if we ever want to do analysis on the JSON files in future. I'll add a comment.

tests/test_utils/alloc.py
```diff
@@ -0,0 +1,88 @@
...
+            raise ValueError(
+                "benchmark function returned None: allocation benchmarking is only reliable if the object(s) of interest is created inside and returned from the function being benchmarked"
+            )
+
+    # Running without a GC for an memory benchmark? This ensures that all objects created in the
+    # benchmarked function get placed into the young generation (0), and so cleaning those up to
+    # leave only the long-lived objects from each function execution is just a `gc.collect(0)` which
+    # is very fast (much faster than `gc.collect()`). For measuring peak memory use this gives the
+    # worst-case peak: the case when no GC runs at all during the benchmarked-function and so only
+    # obvious deallocations occur.
```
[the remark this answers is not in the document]
huonw: > You are disabling the GC to run the tests, is this the same as "running without a GC"?

Yes, I was taking "disabled GC" = "no GC". I'll rephrase because I guess it's a bit sloppy.

> Are you saying if you perform a gc.collect(0) we could measure the worst case peak?

Ah, no, that's meant to be read as:

> Disabling GC is beneficial for measuring the peak memory use too, if that was implemented (e.g. via `tracemalloc.get_traced_memory`), as it gives us the worst-case peak: the case when no GC runs at all during the benchmarked-function and so only obvious deallocations occur.

Does that make more sense? I'll rephrase (e.g. copy that in), or even remove, if we're not going to do anything with the peak memory use.

tests/core/test_stellargraph.py
huonw: I've filed an issue https://github.com/stellargraph/stellargraph/issues/533 and opened a pull request #534, so that there's no need for a human procedure here and it's enforced automatically 😄

geoffj-d61: > no need for a human procedure here

Correct me if I'm wrong, but I thought CI could check but not fix this automatically (due to `mtime` issues), so people still need to manually adjust the year?

huonw: Yes, but no one will have to remember to think about it: it'll be automatically flagged and so remind PR authors to fix it, and reviewers can just rely on that CI action instead of having to catch it themselves.

geoffj-d61: Yeah, but for me it is on the order of forgetting to `black` the code and thus having to do a second commit - yes, great that CI catches it, but should have been on a mental checklist before pushing the commits. And I'm not trying to disparage your effort, I think it is a great new check.

tests/test_utils/alloc.py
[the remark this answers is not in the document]
adocherty: Good to know 👍

tests/test_utils/alloc.py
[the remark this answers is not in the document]
adocherty: Why do you mention gc.collect(0) s faster when GC is disabled, but then don't use this?  The second paragraph is great 👍

tests/core/test_stellargraph.py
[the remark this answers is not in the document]
adocherty: I would tend to remove it: as you say, it will be irrelevant soon (hopefully!)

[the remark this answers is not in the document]
huonw: Sounds good!

tests/test_utils/alloc.py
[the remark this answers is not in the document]
huonw: > Why do you mention gc.collect(0) s faster when GC is disabled, but then don't use this?

Ah, the key here is: disabling GC allows us to use `gc.collect(0)`.

Collecting just the young generation with `gc.collect(0)` is always faster than collecting all generations with `gc.collect()`, whether or not GC is stopped from running automatically (which is what `gc.disable()` is doing). By doing the latter, GC never runs without us asking for it, and so objects never move from the young generation (gen 0) where they are first allocated to gen 1 (or from gen 1 to gen 2).

In addition, this code does use `gc.collect(0)` (in `snapshot`).

Does that clarify the motivation better? I'll think how I can rephrase the comment.

tests/core/test_stellargraph.py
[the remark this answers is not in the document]
huonw: I've switched to:

- sizes: (0, 0), (100, 200), (1000, 5000)
- feature size: `None`, `100`

I've retained the (0, 0) because I think it's helpful to know the "base" level for these sort of benchmarks, to know the lower bound for improvements, and to know how much an improvement actually changes the thing being edited (e.g. a change to vertices that takes the overall size from 100k to 90k might still be a 6x improvement in the size of vertices if they went from 12k to 2k). This is also why I had benchmarks with no edges.

I could be convinced otherwise, if we decide that's unlikely to be interesting.

tests/test_utils/alloc.py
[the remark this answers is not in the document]
huonw: I've rewritten the comment. Maybe that's clearer?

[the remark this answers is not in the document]
adocherty: Yes, the description is super clear now, but still in the final line you are using `gc.collect()` not `gc.collect(0)` so this comment is informative about gc in Python but doesn't seem to match what you actually do, or am I missing something?

tests/core/test_stellargraph.py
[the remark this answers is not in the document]
adocherty: True, there will always be a base level to consider.  I am happy with these 👍

tests/test_utils/alloc.py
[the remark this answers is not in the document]
huonw: Ah, no, that is just cleaning up everything after the benchmark has completely run; this comment ties into `snapshot`'s `gc.collect` call.

huonw:
> Drop the 'if' from "Thus... if the benchmarks" (line 82) - git won't let me make this a suggestion.

Done (how peculiar that GitHub wasn't letting it happen 🤔 ).

> I know you expanded this explanation for Andrew, but I find it very verbose. I'm not suggesting you change it again, but you could get away with just saying that short-term objects are found in gen 0 and long-term objects are found in gen 1 or 2, and thus gc.collect(0) will remove the short-term objects prior to memory measurement of the long-term objects.

Sure, but this doesn't explain why short-term objects are guaranteed to be in gen 0, which is only true with `gc.disable()`. I was trying to make it complete so that it makes sense even for people unfamiliar with garbage collection and how it works, as I imagine most engineers working on this library will be. 😄

tests/test_utils/alloc.py
```diff
@@ -100,5 +103,7 @@ def run_it(f):
 
     if gc_was_enabled:
         gc.enable()
-        # forcibly clean up anything left over in this benchmark
-        gc.collect()
+
+    # This benchmark is finished, and now forcibly clean up any garbage left over from the
+    # small/partial collections that were being during the benchmark.
```
geoffj-d61: My impression was that the final `gc.collect()` now cleans up the long-term objects created during the benchmark - or have I got that wrong?

huonw: Yes, that's right; the benchmark does a few small collections, so the long term objects crated by each call to the benchmarking function may not have been collected. This call then collects those (if they're collectable at all). I guess the comment is a bit unclear, potentially because it's riddled with typos... let me fix those.

tests/test_utils/alloc.py
```diff
@@ -104,6 +104,7 @@ def run_it(f):
     if gc_was_enabled:
         gc.enable()
 
-    # This benchmark is finished, and now forcibly clean up any garbage left over from the
-    # small/partial collections that were being during the benchmark.
+    # This benchmark is finished, so now clean up any longer-lived garbage left over from the
+    # small/partial collections of short-term objects that happened during the benchmark (in
```
geoffj-d61: (Git says 'Suggestions cannot be applied while viewing a subset of changes'. Dunno what it is on about!)

I'm confused again - is it 'short-term' or 'long-term' here? Our conversation suggested the final `gc.collect()` cleans up the *long-term* objects (if possible), since the short-term objects were already cleaned up by `gc.collect(0)`.

geoffj-d61: Ok, I think you mean that the 'collections of short-term objects', i.e. `gc.collect(0)`, will have left over some garbage. I think I keep interpreting 'collections' like 'lists' instead of 'garbage collections'! All good.

geoffj-d61:
And as for git not allowing suggestions, somehow there was a filter being applied (no idea why or how). Once I closed all filters, it works properly again.

Issue #255, opened 2018-10-08
Word2Vec in StellarGraph


PantelisElinas:
### Description

Random walk-based representation learning methods in StellarGraph, e.g., node2vec, metapath2vec, deepwalk, make use of Gensim's implementation of word2vec. This issue is about implementing word2vec as part of StellarGraph. 

Furthermore, we would like to improve the implementation so that we can extend word2vec such that it learns a node representation in conjunction with the downstream task; the latter should be possible but optional.

### User Story
As a: Research Engineer and StellarGraph developer
I want: an implementation of the Word2Vec algorithm consistent with the StellarGraph API as well as enhanced such that node representation can be learned together with the downstream task
so that: I can remove dependence of StellarGraph on Gensim and develop better graph ML models (with the extended Word2Vec algorithm).

### Done Checklist (Development)
- [ ] Produced code for required functionality
- [ ] Unit tests
- [ ] Branch and Pull Request build on CI
- [ ] Peer Code Review Performed
- [ ] Code well commented

youph:
This is a duplicate of

yvonnegh:
Daokun to finish before going away on leave 14 Jan (return Feb).  Yuriy to hopefully review if he has capacity.

PR #536, opened 2019-11-29
Feature/node2vec for issue Word2Vec in StellarGraph #255

daokunzhang:
This is the pull request for adding Keras Node2Vec layer to StellarGraph library.

I mainly made the following changes:

1. Add the Node2Vec layer.
2. Add the Node2VecNodeGenerator and Node2VecLinkGenerator mapper.
3. Change the interface for the `__init__` and  `run` function of BiasedRandomWalk to make UnSupervisedSampler support BiasedRandomWalk
4.  Write the Keras-node2vec notebook examples for embedding learning and node classification
5. Add the unit test for the added node2vec layer and mappers


daokunzhang:
Hi @youph ,

The conflicts of the node2vec pull request have been processed. You can review now.

Best regards!
Daokun

stellargraph/data/explorer.py
```diff
@@ -256,36 +256,47 @@ def naive_weighted_choices(rs, weights):
...
 class BiasedRandomWalk(GraphWalk):
+
     """
     Performs biased second order random walks (like those used in Node2Vec algorithm
     https://snap.stanford.edu/node2vec/) controlled by the values of two parameters p and q.
+        
+    Args:
+        graph: The Stellargraph to perform uniform random walk on
+        graph_schema: The schema of the given graph
+        p: <float> Defines probability, 1/p, of returning to source node
```
youph: I'm not sure about providing p, q at `__init__` rather than at `run` the way it's done here:
- it's better if all parameters of the walk (p, q, n, length, etc.) are passed in one place: either at `__init__` or at `run`. Here p and q are passed to `__init__`, and n and length are passed to `run`. Might be confusing.
- our initial choice was to only pass the graph to `__init__`, so that a user could reuse the same walker, defined on a graph, by running it with different parameters. 
- we could switch to an alternative way of passing all the parameters to `__init__` instead, but then users will have to create multiple walker instances if they wanted to do something like parameter search for their problem, one walker instance per parameter set.
@adocherty @PantelisElinas do you have an opinion on this?

stellargraph/data/unsupervised_sampler.py
```diff
@@ -141,49 +160,62 @@ def generator(self, batch_size):
...
                 target = walk[0][0]
                 context_window = walk[0][1:]
                 for context in context_window:
                     # Don't add self pairs
                     if context == target:
                         continue
 
-                    positive_pairs.append((target, context))
-                    sample_counter += 1
+                    for _ in range(2):
```
youph: why is this loop needed now?

daokunzhang: Here the bidirectinoal=True can increase the diversity of the node context pairs in a same batch. If bidirectional is set to false, the node context pairs generated in a same batch tend to share a comment target nodes. When the option is set to Ture, we will have node context pairs with different target nodes. This is helpful in my previous tests.

stellargraph/layer/node2vec.py
```diff
@@ -0,0 +1,195 @@
...
+    Args:
+        emb_size (int): The dimension of node embeddings.
+        generator (Sequence): A NodeSequence or LinkSequence.
+        node_num (int): The number of nodes in the given graph.
+    """
+
+    def __init__(self, emb_size, generator=None, **kwargs):
+
+        # Get the node_num from the generator if it is given
+        self.generator = generator
```
youph: Do we need to check that the generator type is compatible?

stellargraph/layer/node2vec.py
```diff
@@ -0,0 +1,195 @@
...
+        Returns:
+            tuple: (x_inp, x_out) where ``x_inp`` is a list of Keras input tensors for (src, dst) nodes in the node pairs
+            and ``x_out`` is a list of output tensors for (src, dst) nodes in the node pairs.
+
+        """
+        # Expose input and output sockets of the model, for source node:
+        x_inp_src, x_out_src = self.node_model()
+
+        # Expose input and out sockets of the model, for target node:
+        x_inp_dst = Input(shape=(1,))
```
youph: Why is this different from line 146? Can we just do 
```
x_inp_dst, x_out_dst = self.node_model()
```
?

stellargraph/layer/node2vec.py
daokunzhang: the check is done in the function of _get_sizes_from_generator

stellargraph/layer/node2vec.py
```diff
@@ -0,0 +1,195 @@
...
+        Construct node representations from node ids through a look-up table.
+
+        Args:
+            xin (Keras Tensor): Batch input node ids.
+
+        Returns:
+            Output tensor.
+        """
+        input_initializer = keras.initializers.RandomUniform(minval=-1.0, maxval=1.0)
+        input_embedding = Embedding(
```
youph: I think layers should be initialized in the `__init__` method, and applied in `__call__` - see, e.g., the GAT class. 
This would allow you to simplify the `link_model` method below to 
```
x_inp_src, x_out_src = self.node_model()
x_inp_dst, x_out_dst = self.node_model()
...
```

stellargraph/layer/node2vec.py
daokunzhang: We use different embedding tables. In the node_model, we use the input_embedding table which is used to generate node embedding vectors, here we use the output_embedding table   for context node look-up to predict its existence.

stellargraph/layer/node2vec.py
daokunzhang: Yeah, its better to place the initialisation to the __init__ method. But we should make a difference about the definition of (x_inp_src, x_out_src) and (x_inp_dst, x_out_dst), as they use different embedding look-up tables according to the word2vec paper and its original C implementation.

stellargraph/layer/node2vec.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        # Expose input and output sockets of the model, for source node:
+        x_inp_src, x_out_src = self.node_model()
+
+        # Expose input and out sockets of the model, for target node:
+        x_inp_dst = Input(shape=(1,))
+        output_initializer = keras.initializers.TruncatedNormal(
+            stddev=1.0 / math.sqrt(self.emb_size * 1.0)
+        )
+        output_embedding = Embedding(
```
youph: This is wrong, as this creates a different model for dst nodes in (src, dst) node pairs, whereas src and dst nodes should be embedded by the **same** model! See above comments on how to fix this.

stellargraph/layer/node2vec.py
youph: Fixed

stellargraph/layer/node2vec.py
youph: Fixed

stellargraph/layer/node2vec.py
youph: I take it back - it follows the original word2vec implementation with 2 different embedding layers

daokunzhang:
@youph,

I tend to keep the api of BiasedRandomWalk as before. It is not necessary to unify the API for all walkers. Among them, UniformRandomWalk, BiasedRandomWalk and UniformRandomMetaPathWalk are used to generate (`target`, `context`) pairs, while SampledBreadthFirstWalk, SampledHeterogeneousBreadthFirstWalk and DirectedBreadthFirstNeighbors are used to sample features for GraphSage algorithms. For UnsupervisedSampler, only UniformRandomWalk, BiasedRandomWalk and UniformRandomMetaPathWalk are required to perform run function to sample target context node pairs. Currently we haven't realised keras based metapath2vec. So we only require UnsupervisedSampler to support UniformRandomWalk and BiasedRandomWalk. In this implementation, I just use if ... else ... to control which walker is used.

As the tensorflow 2.0 does not well support multiprocess for keras.fit_generator. I run the demo with previous lower version. In the future, we should investigate the multiprocess issue caused by tensorflow 2.0.

Regards!
Daokun

daokunzhang:
Hi @timpitman, 

By handling the problem raised by Yuriy and merging the latest develop branch, I have updated the keras word2vec pull request. Though the buildkite dose not succeed, the problem is occurred at test notebooks, which is caused by two new added notebooks.  According to the instruction given by buildkite, I should change the `notebook testing parallelism` to 38. However, after I changed it, it still reported failing check.

Could you please arrange someone to help he handling this buildkite check problem and review my new pull request?

Best regards!
Daokun

timpitman:
Hi @daokunzhang thanks for the updates!I'm trying to re-run the Buildkite pipeline now to check - your change to parallelism 38 was correct (that tells Buildkite that it's testing 38 notebooks now)

daokunzhang:
Hi @timpitman,

According to your instruction, I have made necessary changes on my new added notebooks to make parallelism able to override the parameter `work_number` and `walk_length` with much smaller values in CI.  The updated pull request has passed the buildkite check. Thank you very much!

However, it failed in the codecov/patch check. Does it mean my pull request make excessive changes? Though it is shown that there were in total 40 files changes, 25 dummy records with 0 changes are among them.

Daokun

.buildkite/notebook-parameters.yml
```diff
@@ -5,6 +5,8 @@ epochs: 2
 batch_size: 5
 # short random walks
 walk_length: 5
+# a small number of random walks
+walk_number: 2
```
timpitman: it might be best to move this to the bottom, with a comment about which notebooks use it, since it's not widespread.

timpitman:
> However, it failed in the codecov/patch check. Does it mean my pull request make excessive changes? Though it is shown that there were in total 40 files changes, 25 dummy records with 0 changes are among them.

I checked and you're right - the absolute coverage is good, I think it just got confused with the diff.

timpitman:
Some other basic questions about this @daokunzhang: what are the advantages and disadvantages of using the keras layer instead of gensim (eg accuracy, speed)?

daokunzhang:
Hi @timpitman,

The keras node2vec works slower than gensim word2vec. As gensim implementation is based on C programming with the hashing tricks for speeding up sigmoid calculation and negative sampling. The advantage of keras node2vec implementation over genism is that it provides flexible interfaces to support end-to-end learning. In other words, it can be integrated with a downstream task, such as node classification and link prediction to learn task-dependent node embeddings, which in contrary serves the downstream task better.

In following, I will try to make keras node2vec works faster. Traditional node2vec uses one-hot encoding scheme to represent each node, i.e., attaching each node with a d-dimensional vector, which will import too many parameters when the number of nodes is large. I will try to use the K-way D-dimensional discrete encoding scheme to represent each node, which is proposed in paper `Learning K-way D-dimensional Discrete Code For Compact Embedding Representations`(https://arxiv.org/abs/1711.03067). With the new node representing scheme, the number of parameters can be significantly reduced, and fewer iterations are required to obtain converged parameters. I will conduct some experiments to validate this.

Daokun

stellargraph/data/unsupervised_sampler.py
```diff
@@ -55,24 +69,25 @@ def __init__(self, G, nodes=None, length=2, number_of_walks=1, seed=None):
...
+                    "({}) Only the UniformRandomWalks and BiasedRandomWalks are possible".format(
+                        type(self).__name__
+                    )
+                )
+            else:
+                self.walker = walker
+        else:
+            self.walker = UniformRandomWalk(G, seed=seed)
+
+        if isinstance(self.walker, BiasedRandomWalk):
```
kjun9: I might look into refactoring `BiasedRandomWalk` (and other walker classes) to optionally take these parameters upon construction  rather than in `run`, so that we can avoid doing this - I think this make more sense than having to pass in the same set of parameters for each call to `run` and it seems generally useful to be able to instantiate and pass around a walker object that already has its hyper-parameters filled in?

stellargraph/layer/node2vec.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        # Expose input and output sockets of the model, for source node:
+        x_inp_src, x_out_src = self.node_model("input")
+        x_inp_dst, x_out_dst = self.node_model("output")
+
+        x_inp = [x_inp_src, x_inp_dst]
+        x_out = [x_out_src, x_out_dst]
+        return x_inp, x_out
+
+    def build(self):
```
kjun9: We ended up deprecating this method in https://github.com/stellargraph/stellargraph/pull/1140 - given that this is an entirely new model, maybe we don't need old methods with deprecation warnings either, such as the `default_model` method below? So in summary:

* `node_model` -> `_node_model` (i.e. should be private)
* `link_model` -> `_link_model` (i.e. should be private)
* `build` -> `in_out_tensors`
* remove `default_model`

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -223,6 +221,9 @@ def __init__(self, G, batch_size, num_samples, seed=None, name=None):
         self.num_samples = num_samples
         self.name = name
 
+        # Check if the graph has features
+        G.check_graph_for_ml()
```
kjun9: Hmm, maybe instead of having to add this line everywhere, we can just update `BatchedLinkGenerator.__init__` to be:
```python
class BatchedLinkGenerator(abc.ABC):
    def __init__(self, G, batch_size, schema=None, use_node_features=True):
        ...
        if use_node_features:
            G.check_graph_for_ml()

class Node2VecLinkGenerator(BatchedLinkGenerator):
    def __init__(self, G, batch_size, name=None):
        super().__init__(G, batch_size, use_node_features=False)
        ...
```
Would that work?

And similarly for the node generators

stellargraph/data/unsupervised_sampler.py
daokunzhang: Hi Kevin,

In my earlier version,  I ever tried to pass the hyperparameter (p, q) of `BiasedRandomWalk` into its initiate function rather than pass them into the `run` function. I will make changes to be consistent with the new interface.

kjun9: @daokunzhang This should be landing soon: https://github.com/stellargraph/stellargraph/pull/1187

kjun9: And this is related: https://github.com/stellargraph/stellargraph/pull/1179

daokunzhang: OK, I will merge it.

stellargraph/mapper/sampled_link_generators.py
daokunzhang: Good idea!

daokunzhang:
Hi @kjun9,

I have made the changes as per your comments. However, I do not know how to delete the dummy file changes record with no changes. Does it affect the merging? 

To address this problem, I can build a new pull request and transfer the changes into the new PR. Do you think it is necessary?

Best regards!
Daokun

kjun9:
> I have made the changes as per your comments. However, I do not know how to delete the dummy file changes record with no changes. Does it affect the merging?

> To address this problem, I can build a new pull request and transfer the changes into the new PR. Do you think it is necessary?

Hi @daokunzhang - I tried reverting the mode changes here https://github.com/stellargraph/stellargraph/pull/1203

You can either just do a `git merge kevin/node2vec` from this branch, or you can run the shell command yourself and commit it
```
ls -d $(find stellargraph demos tests -not -type d -perm 755) | xargs chmod 644
```
Hope that's clear!

daokunzhang:
Thanks @kjun9 ,

The mode change problem has been resolved.

demos/embeddings/stellargraph-keras-node2vec.ipynb
```diff
@@ -0,0 +1,464 @@
+{
```
kjun9, line 1: <blockquote>An example of implementing the keras Node2Vec representation learning algorithm using components from the stellargraph library</blockquote><p>How about something like : "This example demonstrates how to apply components from the stellargraph library to perform representation learning via Node2Vec. This uses a keras implementation of Node2Vec available in stellargraph instead of the reference implementation provided by <code>gensim</code>."? Or I think we could at least replace "implementing" with "demonstrating".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/536/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

kjun9, line 1: <p>This somehow ended up being added twice - could you remove one of them?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/536/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

kjun9, line 1: <p>Same as above, can remove one of these</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/536/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

kjun9, line 1: <blockquote>which actually performs the dot product of the 'input embedding' of the target node and the 'output embedding' of the context node followed by a sigmoid activation</blockquote><p>Doesn't it perform the dot product on the "output embedding" of both the target and context nodes? Also I think using the term "inner product" here might avoid a bit of confusion since we're passing in <code>ip</code>  instead of <code>dot</code>  although they do the same thing.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/536/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

kjun9, line 1: <p>Maybe a little bit more explanation for the loss function would be helpful for guiding users who are not as familiar with Node2Vec. Something like: "Our generator will produce batches of positive and negative context pairs as inputs to the model, so we use binary crossentropy much like a regular binary classification task."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/536/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

kjun9, line 1: <p>This seems like a very few number of epochs - is this usually the case for Node2Vec? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/536/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

kjun9, line 1: <p>Same as above, can remove one of these</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/536/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='39'/>

demos/node-classification/node2vec/stellargraph-keras-node2vec-node-classification.ipynb
```diff
@@ -0,0 +1,594 @@
+{
```
kjun9, line 1: <p>Similar to the other notebook, this cell is duplicated</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/536/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

kjun9, line 1: <p>I was under the impression that having <code>Node2Vec</code>  implemented with keras meant that we could do node classification via <code>Node2Vec</code>  in a single keras pipeline, instead of doing it separately after doing representation learning - is this not the case?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/536/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='37'/>

kjun9:
@daokunzhang I've reviewed the notebooks first, I'll go through the rest of the code now

stellargraph/layer/node2vec.py
```diff
@@ -0,0 +1,209 @@
...
+        input_initializer = keras.initializers.RandomUniform(minval=-1.0, maxval=1.0)
+        self.input_embedding = Embedding(
+            self.input_node_num,
+            self.emb_size,
+            input_length=1,
+            name="input_embedding",
+            embeddings_initializer=input_initializer,
+        )
+
+        output_initializer = keras.initializers.TruncatedNormal(
```
kjun9, line 68: should this initializer be exposed as a parameter, similar to our other models?

(similarly for the `input_initializer`)

stellargraph/layer/node2vec.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        if multiplicity is None:
+            multiplicity = self.multiplicity
+        if self.multiplicity == 1:
+            return self._node_model()
+        elif self.multiplicity == 2:
+            return self._link_model()
+        else:
+            raise RuntimeError(
+                "Currently only multiplicities of 1 and 2 are supported. Consider using node_model or "
```
kjun9, line 195: This error message should be updated to remove the mention of `node_model` and `link_model`

stellargraph/layer/node2vec.py
```diff
@@ -0,0 +1,209 @@
...
+            self._get_sizes_from_generator(generator)
+        else:
+            self._get_sizes_from_keywords(kwargs)
+
+        # Model parameters
+        self.emb_size = emb_size
+
+        # Initialise the embedding layers: input-to-hidden and hidden-to-output
+        input_initializer = keras.initializers.RandomUniform(minval=-1.0, maxval=1.0)
+        self.input_embedding = Embedding(
```
kjun9, line 60: If I understood correctly, you're referring to the "target node" as the `input` and the "context node" as `output`... Is this terminology standard for Node2Vec? 

I find it a bit confusing in the context of a keras model, since "input" and "output" already have a meaning (i.e. "input" tensors to the model vs the "output" tensors). Could we just name them as "target" and "context", similar to the way it's named in the generator?

kjun9: Also, I would've thought we actually want to only train a single embedding layer that's shared for both the target and context nodes? Otherwise we're saying that the embedding functions for the target node vs context node could be different.

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -501,6 +503,57 @@ def sample_features(self, head_links, batch_num):
...
+            A list of feaure arrays, with each element being the ids of
+            the sampled target and context node.
+        """
+
+        target_ids = [head_link[0] for head_link in head_links]
+        context_ids = [head_link[1] for head_link in head_links]
+        target_feats = self.graph._get_index_for_nodes(target_ids)
+        context_feats = self.graph._get_index_for_nodes(context_ids)
+        batch_feats = [np.array(target_feats), np.array(context_feats)]
+        return batch_feats
```
kjun9, line 554: I think `Graph._get_index_for_nodes` is already supposed to return a numpy array? In which case this whole method could be simplified:
```python
return [self.graph._get_index_for_nodes(ids) for ids in zip(*head_links)]
```

stellargraph/mapper/sampled_node_generators.py
```diff
@@ -563,3 +564,85 @@ def flow_from_dataframe(self, node_ids):
...
+            head_nodes: An iterable of head nodes to perform sampling on.
+
+        Returns:
+            A list of feature arrays, with each element being the id of each
+            head node.
+        """
+
+        head_node_ids = self.graph._get_index_for_nodes(head_nodes)
+        batch_feats = np.array(head_node_ids)
+        return batch_feats
```
kjun9, line 612: similar to the same method in the link generator, this could be:
```suggestion
        return self.graph._get_index_for_nodes(head_nodes)
```
or if you did want to assign it to a named variable first, I think `iloc` would be a more accurate term than `ids` here.
```suggestion
        head_node_ilocs = self.graph._get_index_for_nodes(head_nodes)
        return head_node_ilocs
```

stellargraph/mapper/sampled_node_generators.py
```diff
@@ -563,3 +564,85 @@ def flow_from_dataframe(self, node_ids):
...
+            self.sample_features, self.batch_size, node_ids, shuffle=False
+        )
+
+    def flow_from_dataframe(self, node_ids):
+        """
+        Creates a generator/sequence object for node representation prediction
+        with the supplied node ids.
+
+        Args:
+            node_ids: a Pandas DataFrame of node_ids.
```
kjun9, line 641: I think this could be a bit misleading, since we're actually just grabbing the index of the dataframe as the node IDs, rather than any of the contents of the dataframe. Probably would be an easy mistake to supply a dataframe with an "id" column instead.
```suggestion
        Creates a generator/sequence object for node representation prediction
        by using the index of the supplied dataframe as the node ids.

        Args:
            node_ids: a Pandas DataFrame indexed by node_ids.
```

stellargraph/layer/node2vec.py
daokunzhang: In the original word2vec implementation, two sets of embeddings i.e., `input_embedding` and `output_embedding` are learned for `target node` and `context node` respectively. Here, the `input_embedding` is actually the weight matrix from the input layer to the hidden representation layer, and the `output_embedding` is the weight matrix from hidden representation layer to the output layer. To be consistent with the word2vec algorithm, we should learn two different sets of embeddings for `target node` and `context node` with the purpose of making nodes sharing similar context nodes represented closely.

stellargraph/layer/node2vec.py
daokunzhang: I set the initializers by referring to the word2vec C and tensorflow implementations, which guarantees the good experimental results. Word2vec is sensitive to the initialization, other settings might fail to perform well.

stellargraph/layer/node2vec.py
daokunzhang: I will make the update.

demos/node-classification/node2vec/stellargraph-keras-node2vec-node-classification.ipynb
daokunzhang: Yeah! Actually, we can couple node2vec with a classification task within a single keras pipeline. Here, I just use the node classification to evaluate the quality of node embeddings learned by node2vec, with the expectation that keras node2vec achieves the similar classification performance to gensim node2vec. In the future pull request, I will try do node classification with node2vec in a single keras pipeline.

demos/embeddings/stellargraph-keras-node2vec.ipynb
daokunzhang: Thanks! Because I run format_notebooks.py the extra one is added.

daokunzhang: With enough random walks, Node2Vec will perform good enough. Accurally, in Word2Vec C implementation, only one epoch is required.

stellargraph/layer/node2vec.py
kjun9: Hmm okay, thanks for clarifying that. Do you think it's important that we use the terms `input_embedding` and `output_embedding`? Could it just be `target_embedding` and `context_embedding` instead? I'm just a bit concerned that it's confusing with the model's actual input and output tensors (i.e. `in_out_tensors`) - what do you think?

demos/embeddings/stellargraph-keras-node2vec.ipynb
kjun9: I see now that these are referring to the `input_embedding` and `output_embedding` defined in the model, rather than the "input" vs "output" I was thinking of... which still makes me think https://github.com/stellargraph/stellargraph/pull/536#discussion_r404464928 is relevant

kjun9: Ah yep that makes sense, since you can just run more walks to have a larger epoch 👍

daokunzhang: Thanks Kevin! I have change the introduction according to your suggestion and add an extra description "This implementation provides flexible interfaces to downstream tasks for end-to-end learning."

daokunzhang: I have deleted one of them.

daokunzhang: I have deleted one of them.

daokunzhang: Thanks for your feedback, Kevin. Here, it is indeed a bit confused to use the item ``input embedding`` and ``output embedding`` without any introduction. I plan to give an graphic illustration to demonstrate the architecture of word2vec, and referring the ``input embedding`` and ``output embedding`` to the input-to-hidden and hidden-to-output weight matrices respectively.

daokunzhang: Thanks for your suggestion. I will add the explanation.

daokunzhang: Thanks! I will add the explanation.

stellargraph/layer/node2vec.py
daokunzhang: That's a good idea! I will change the names to avoid confusing users.

stellargraph/mapper/sampled_link_generators.py
daokunzhang: In the old version ``Graph._get_index_for_nodes`` only return a list of node ids. It has been upgraded to return a numpy array. Obviously, your script is much more concise.

stellargraph/mapper/sampled_node_generators.py
daokunzhang: Thanks Kevin! I have changed it to ``return self.graph._get_index_for_nodes(head_nodes)``

stellargraph/mapper/sampled_node_generators.py
daokunzhang: Thanks Kevin. I have made your proposed changes.

demos/node-classification/node2vec/stellargraph-keras-node2vec-node-classification.ipynb
daokunzhang: I have deleted the extra cell.

daokunzhang:
Hi @kjun9,

I have handled your comments. In the two node2vec nodebooks, I add an graphic illustration as an brief introduction to node2vec and provide the reference to ``input_embedding`` and ``output_embedding``. Wish this updated version is more clear to understand.

Best regards!
Daokun

daokunzhang:
Hi @kjun9 ，

Could you please review my updated keras-node2vec pull request when you are free?

Thanks and best regards!
Daokun

daokunzhang:
> Thanks for making the updates @daokunzhang !
> 
> I did end up reading the reference paper again, but I struggled to find where it talks about the "input" vs "output" embeddings... equation 2 from [here](https://arxiv.org/pdf/1607.00653.pdf) just seems to refer to the same embedding function `f` for both the context and target nodes? I've pulled @PantelisElinas in to review as well since he might have better context around that than I do.

Hi @kjun9 ,

Thanks for your feedback. The equation 2 in the node2vec paper is confusing. Though it is claimed that the same parameter is used to represent ``input`` and ``output`` embeddings, the [implementation](https://github.com/aditya-grover/node2vec/blob/master/src/main.py) provided by authors directly deploys the gensim Word2Vec function, which is consistent with our [gensim node2vec notebook](https://github.com/stellargraph/stellargraph/blob/develop/demos/embeddings/stellargraph-node2vec.ipynb). Here, I try to make a consistency with the gensim Word2Vec implementation, which uses different parameters to represent the so-called ``input`` and ``output`` embeddings.

Regards,
Daokun

daokunzhang:
Hi @kjun9 ，

Thanks for your careful review! I have made necessary changes as per your suggestions. You can start the next round review now.

Best regards!
Daokun

kjun9:
@daokunzhang 

> Though it is claimed that the same parameter is used to represent input and output embeddings, the implementation provided by authors directly deploys the gensim Word2Vec function, which is consistent with our gensim node2vec notebook. Here, I try to make a consistency with the gensim Word2Vec implementation, which uses different parameters to represent the so-called input and output embeddings.

Thanks for the explanation, that makes sense 👍 This looks good to me now, just one last thing to update that I can see is that in the demo notebooks, we're trying to keep the first markdown cell only contain the title (for example https://github.com/stellargraph/stellargraph/blob/develop/demos/node-classification/graphsage/graphsage-cora-node-classification-example.ipynb) since this lets us render the colab/binder buttons right underneath the title. So I'd suggest just moving the rest of the markdown in that first cell to a second markdown cell!

daokunzhang:
Hi @kjun9 ,

The notebooks have been formatted as you requested.

Thanks and best regards!
Daokun

daokunzhang:
@kjun9 ,

The code has been updated to use node ilocs in sampled node and link generaters #1267! Could you please help me review it again and merge it ?

Thanks,
Daokun

kjun9:
@daokunzhang Hi daokun, sorry for the delay I'll take a look now!

daokunzhang:
Thanks @kjun9 ,

I have revised the **scripts/notebook_text_checker.py** file mode change problem. I have create the nblink files for the keras-node2vec-node-classification and keras-node2vec-embeddings demo files in the docs folder and added the word2vec_illustration image file to the docs folder. After that I added the component about keras node2vec to the  ``scripts/demo_indexing.py`` file and rerun it to overwrite the ReadME files. You can check it.

For my part, I cannot merge this pull request. Maybe it is caused that Yuriy ever requested changes but he haven't approved yet. Could you please help me merge it on your part if you are convenient?

Best regards!
Daokun

kjun9:
Thanks @daokunzhang I can see the notebook in the API documentation 👍  I'll merge this now. 

I think the demo-indexing could still be tweaked a bit, so I might still file an issue about that - I think this is the first time where we have two versions of an algorithm presented as demos (`gensim` vs stellargraph components) so I think it'd be nice for the table to communicate this more clearly, rather than showing them as two separate algorithms.

daokunzhang:
Thanks @kjun9 !

kjun9:
https://github.com/stellargraph/stellargraph/issues/1534

PR #539, opened 2019-12-03
Build the docker image using the buildkite docker-compose plugin

huonw:
This uses https://github.com/buildkite-plugins/docker-compose-buildkite-plugin to automate the building of the docker image. It doesn't work great for now, due to the restrictions around being a public repository.

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+echo "~~~ setting docker image tag"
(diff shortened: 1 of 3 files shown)


huonw:
This fails with:

```
 Error: Refusing to run /var/lib/buildkite-agent/builds/buildkite-linux-t2micro-i-0e9f3783173d401f3-1/stellar/stellargraph/.buildkite/hooks/pre-command, local hooks are disabled
```
  
At the moment and I'm not going to get back to it for now. I've opened this just to note the work for future reference.

PR #545, opened 2019-12-06
remove pushlogs

TheDen:
Since we have the public pipeline, this is no longer needed


huonw:
This can delete the checkstatus.sh and the continue on failure of the earlier wait, because all that only exists for this log pushing step.

PR #547, opened 2019-12-09
Bug-fix: Naive ensembling does not work with unsupervised graphsage

kiarie-ndegwa:
## Prob def:
Naive ensembling using unsupervised GraphSage does not work, as the `OnDemandLinkGenerator` was not included as an acceptable training generator type.

## Fix:
Added `OnDemandLinkGenerator` to train generator type onto `ensemble.py` class.

## Tests:
- [x] Unit test written and working
- [x] All SG tests passed: Run `py.test tests/`
- [x] Naive ensembling on unsupervised graphsage: training and inference found to be working on local environment post fix

### DIFF
--- a/tests/utils/test_ensemble.py
+++ b/tests/utils/test_ensemble.py
@@ -476,9 +476,7 @@ def test_unsupervised_Ensemble_fit_generator():
     graph = example_graph_1(feature_size=10)
 
     # base_model, keras_model, generator, train_gen
-    gnn_models = [
-        create_graphSAGE_model(graph),
-    ]
+    gnn_models = [create_graphSAGE_model(graph)]
 
     for gnn_model in gnn_models:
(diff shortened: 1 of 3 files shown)


tests/utils/test_ensemble.py
```diff
@@ -340,7 +365,7 @@ def test_Ensemble_fit_generator():
         ens = Ensemble(keras_model, n_estimators=2, n_predictions=1)
 
         ens.compile(
-            optimizer=Adam(), loss=categorical_crossentropy, weighted_metrics=["acc"]
```
kiarie-ndegwa: @adocherty : flaky about changing the loss function, here, I hope it's cool to leave this as is

kiarie-ndegwa: Reverted change and added unsupervised bagging unit tests.

PR #551, opened 2019-12-11
use the snapshot black tag + fmt everything

TheDen:
PR:

* Updates the pipeline to use `stellargraph/black:snapshot` which is currently version `19.10b0`
* The image's entrypoint will fomat everything in the repo by default
* This PR formats the remaining python files so that CI passes
* The docker-plugin is also updates for the `black` step


huonw:
Was there a reason that the `demos` folder wasn't being formatted with `black` before?

TheDen:
Not sure, perhaps @youph can clarify? (I think they should)

adocherty:
I think formatting the demos would be a good idea. There is also a script that formats the IPython notebooks in the demos folder as well in PR #523 .

youph:
> Not sure, perhaps @youph can clarify? (I think they should)
I think it never occurred to us to do that, but does black recognise .ipynb files? (notebooks)

TheDen:
Black does run again `ipynb` files, e.g.,

```$ black ./demos/node-classification/node2vec/stellargraph-node2vec-weighted-random-walks.ipynb
reformatted demos/node-classification/node2vec/stellargraph-node2vec-weighted-random-walks.ipynb
All done! ✨ 🍰 ✨  
```

TheDen:
I was mistaken, when directory running `black` against `ipynb` files it does format them but renders them unreadable, which I assume it treats them as python code. However, on the CI It won't happen because running `black .` or `black demos` it only formats `.py` so the `ipynb` will be left alone

PR #552, opened 2019-12-12
Temporal Random Walks

habiba-h:
#### Reviewer Checklist
- [ ] Code has no errors
- [ ] Unit tests are appropriate
- [ ] Demo notebook runs
- [ ] Code is well commented


habiba-h:
@kjun9 Temporal Random Walks is ready for review. Thanks.

stellargraph/data/explorer.py
```diff
@@ -849,3 +854,342 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+                        current_backwards_node = node
+                        current_backwards_time = start_edge[0][2]
+                    else:
+                        move_backwards = False
+
+                    walk = list()
+                    walk.append(node)  # start a walk
+                    walk.append(current_forward_node)
+
+                    while len(walk) < (length):
```
kjun9: I think we can reduce code duplication here by taking the "stepping"-logic out into its own function, since moving forwards vs backwards seems roughly the same. Would also make it a bit easier to read and make the codeclimate bot happier.

Here's my go at writing one:
```python
def _step(self, node: Any, time: int, is_forward: bool, edge_time_label) -> Optional[Tuple[Any, int]]:
    """Perform 1 temporal step from a node. Returns None if a dead-end is reached."""
    def check_time(t: int) -> bool:
        return (is_forward and t > time) or (not is_forward and t < time)

    edges = [
        (neighbour, t) for _, neighbour, t in self.graph.edges(node, data=edge_time_label)
        if check_time(t)
    ]

    if len(edges) > 0:
        next_edge = rs.sample(edges, 1)
        return next_edge[0][0], next_edge[0][1]  # (node, time)
    else:
        return None
```

And maybe each "walk" could also be a separate function, so that we have something that reads like
```python
def run(...):
    # do n walks for each starting node
    walks = []
    for node in nodes:
        if self.graph.node_degrees(node) > 0:
            walks += [self._walk(node, ...) for _ in range(n)]
    return walks

def _walk(...):
    # take steps until walk is of correct length or reached dead ends

def _step(...):
    # one temporal step forward or backward
```
What do you think? Happy to discuss more or flesh this out in more detail

stellargraph/data/explorer.py
```diff
@@ -849,3 +854,342 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+                            )
+                    else:
+                        self._raise_error(
+                            "Invalid time label ({}) between ({}) and ({}).".format(
+                                edge_time_label, k, v
+                            )
+                        )
+
+
+class TemporalBiasedRandomWalk(GraphWalk):
```
kjun9: I think once the `Uniform` version above has its main chunks of logic split out into smaller functions, we could probably think more about how to share more code between these two classes, unless you had explicit reasons for not wanting to do that and keep them entirely separate?

habiba-h:
wow! Thanks @kjun9 for the prompt review! The suggestions for the code fixes seem right. I will make the changes and let you know. I guess i'll get to these fixes next sprint(year?!).

stellargraph/data/explorer.py
```diff
@@ -849,3 +854,342 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+            <list> List of lists of nodes ids for each of the random walks
+
+        """
+        self._check_common_parameters(nodes, n, length, seed)
+        self._check_temporal_parameters(bidirectional, edge_time_label)
+        rs = self._get_random_state(seed)
+
+        walks = []
+        for node in nodes:  # iterate over root nodes
+            if not nx.is_isolate(self.graph, node):
```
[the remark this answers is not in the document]
habiba-h: Stellargraph class doesn't have the `node_degrees` method.

stellargraph/data/explorer.py
```diff
@@ -849,3 +854,342 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+
+                for walk_number in range(n):  # generate n walks per root node
+                    start_edge = rs.sample(
+                        list(self.graph.edges(node, data=edge_time_label)), 1
```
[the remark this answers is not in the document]
habiba-h: I think this should be a separate PR.

stellargraph/data/explorer.py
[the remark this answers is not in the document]
kjun9: If you merge `develop` into your branch you should get the latest updates on the `StellarGraph` class which now has a `node_degrees` method - let me know if you need help with `git` then I can try to explain further!

stellargraph/data/explorer.py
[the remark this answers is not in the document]
kjun9: Do you mind filing an issue to add/update methods in `StellarGraph` and `NetworkXStellarGraph` to be able to obtain neighbours of nodes as triples of `(src, dst, data)`?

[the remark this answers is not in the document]
kjun9: And yes I agree it should be a separate PR!

stellargraph/data/explorer.py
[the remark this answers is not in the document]
habiba-h: `node_degrees` method doesn't work with  `MultiDiGraph` and `MultiGraph`. Temporal graph is at least a multigraph.

stellargraph/data/explorer.py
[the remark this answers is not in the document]
habiba-h: Sure will add the issue to update the `StellarGraph` and `NetworkXStellarGraph` class methods dor neighbors, in_nodes, out_nodes etc.

stellargraph/data/explorer.py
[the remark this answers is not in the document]
kjun9: It seems to work for me:
```
(stellargraph-dev) [jun021@rain-ev stellargraph]$ python
Python 3.7.5 (default, Oct 25 2019, 10:52:18)
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import networkx as nx
>>> import stellargraph as sg
>>> nxg = nx.MultiDiGraph()
>>> nxg.add_nodes_from([1,2,3])
>>> nxg.add_edges_from([(1,2), (1,3)])
[0, 0]
>>> g = sg.StellarGraph(nxg)
>>> g.node_degrees()
DiMultiDegreeView({1: 2, 2: 1, 3: 1})
>>> g.node_degrees()[1]
2
```
Not sure what I'm missing - what did you try that didn't work for you?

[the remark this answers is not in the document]
habiba-h: This has been updated to `node_degrees`

stellargraph/data/explorer.py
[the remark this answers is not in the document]
habiba-h: done.

stellargraph/data/explorer.py
```diff
@@ -849,3 +854,342 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+                            move_forwards
+                        ):  # check to stop incase a dead end is reached moving forward
+                            forward_edges = []
+                            for _, neighbor, time in self.graph.edges(
+                                current_forward_node, data=edge_time_label
+                            ):
+                                if (
+                                    time > current_forward_time
+                                ):  # strictly look ahead of time to avoid being stuck in the current time
+                                    forward_edges.append((neighbor, time))
```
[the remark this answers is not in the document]
habiba-h: Fixed this part after the #646 updates.

stellargraph/data/explorer.py
habiba-h: Done. But would need thorough testing and unit tests.

stellargraph/data/explorer.py
habiba-h: Done. Now there is only `TemporalRandomWalk` class.  The run method takes a `step_type` value to choose between `uniform_random` or `biased_random` step (forward or backward) in time. Biased walk here means exponential random but could be any other distribution. The authors of this work proposed this exponential and linear type of forward step. Linear didn't make much sense so left it for now.

habiba-h:
@kjun9 I have updated the `TemporalRandomWalk` class based on the updates in the `neighbours` method in `StellarGraph`. Also, merged the two classes into one and cleaned up the implementation as you suggested. 
I have not fixed the unit test and the demo notebook. Also, haven't throughly  tested the new implementation.

stellargraph/data/explorer.py
```diff
@@ -796,3 +800,202 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+        walks = []
+        for node in nodes:  # iterate over root nodes
+            for walk_number in range(n):  # generate n walks per root node
+                walk = self._walk(
+                    node, length, bidirectional, edge_time_label, step_type, rs
+                )
+                walks.append(walk)
+        return walks
+
+    def _check_temporal_parameters(self, bidirectional, edge_time_label):
```
kjun9: This method seems to iterate through all edges in the graph to make sure that the edge weights are all positive numerical values - I think this is expensive enough that we don't want to do this up front, but instead just raise when we encounter a bad value during the random walk. I'm going to remove this.

stellargraph/data/explorer.py
```diff
@@ -796,3 +800,202 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+                            )
+                        )
+                    elif t < 0:  # check if edge has a negative timestamp
+                        self._raise_error(
+                            "Timestamp between nodes ({}) and ({}) is negative ({}).".format(
+                                node, neighbor, t
+                            )
+                        )
+
+    def get_exp_distribution(self, times, current_time, orientation="forward"):
```
kjun9: `orientation` is unused. I might rename this to `is_forward` for consistency, and try figure out how I should use it (I'm guessing all the `t_m - t_n` get reversed)

stellargraph/data/explorer.py
```diff
@@ -796,3 +800,202 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+    Performs uniform temporal random walks on the given graph
+    """
+
+    def run(
+        self,
+        nodes=None,
+        n=None,
+        length=None,
+        bidirectional=False,
+        edge_time_label="weight",
```
kjun9: We can remove `edge_time_label` and all its usage in this class entirely. This is a parameter that needs to be used when creating the graph i.e. `StellarGraph(nxg, edge_weight_label="time")` or something, not in this class.

stellargraph/data/explorer.py
kjun9: actually there's no point in reversing since all time distances are wrapped in`np.abs`, so I'm guessing it's the same regardless of direction

kjun9: actually^2 maybe I'm supposed to use `max(times)` instead of `min(times)` for going backwards... will just double check the paper

stellargraph/data/explorer.py
```diff
@@ -796,3 +800,202 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+            if step_type == "uniform_random":
+                next_edge = rs.sample(edges, 1)
+                return next_edge[0][0], next_edge[0][1]
+            else:
+                neighbors, times = zip(*edges)
+                dist = self.get_exp_distribution(
+                    times, time, "forward"
+                )  # exponential distribution
+                chosen_time = choice(times, 1, p=dist)[0]
+                next_edge = edges[times.index(chosen_time)]
```
kjun9: I think this is not very efficient `O(len(times))` and also problematic if the same time value occurs multiple times - will think of another way

stellargraph/data/explorer.py
```diff
@@ -796,3 +800,202 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+        Perform a time respecting random walk starting from the root nodes.
+
+        Args:
+            nodes: <list> The root nodes as a list of node IDs
+            n: <int> Total number of random walks per root node
+            length: <int> Maximum length of each random walk
+            bidirectional: <bool> Whether the walk extends on both direction.
+            seed: <int> Random number generator seed; default is None
+            step_type <string>: whether the next step taken in the walk is chosen uniformly at random or is biased (such as, exponential). 
+            The default is 'uniform_random', any other string value assumes a biased time respecting walk.
```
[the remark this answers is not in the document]
kjun9: Also going to update the docstring style to match the intended style in the rest of the codebase

stellargraph/data/explorer.py
```diff
@@ -796,3 +800,202 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+            if bidirectional:
+                move_backwards = True
+                current_backwards_node = node
+                current_backwards_time = current_forward_time
+            else:
+                move_backwards = False
+
+            walk.append(current_forward_node)
+
+            while len(walk) < (length):
```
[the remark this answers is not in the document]
kjun9: I decided to keep the while loop, but tried to simplify it a bit

kjun9:
The demo notebook links to this paper: http://ryanrossi.com/pubs/nguyen-et-al-WWW18-BigNet.pdf

After reading through that, I'm left a little bit confused:
* the exponential bias function doesn't seem like it matches the paper and there should be two different ones:
  * one for picking the initial edge which should favour edges that occur later in time, intuition being that the "older" parts of the graph are considered less informative than what's more recent.
  * one for picking a temporal neighbour which should favour edges that occur closer in time w.r.t the "current" time during the random walk (i.e. time of previously traversed edge).
* I couldn't find any mention of doing a `bidirectional` walk in the paper, so I'm either not looking in the right place, or there was a decision made elsewhere to include this idea in our implementation? I think this basically has the effect: the set of context pairs also including some neighbours that are connected earlier in time compared to the initial edge sampled. It's not clear to me whether this is what we want/don't want.

The exponential bias function implementation looks more like a mistake so I might just update that, but the bidirectional thing looks intentional so I could just leave it for now.

Also I think the demo notebook for this should really be use the same dataset as the paper so we can aim to produce the same benchmark given. I had trouble running the demo notebook with the spammer dataset since it looks like the edge list is over 20GB in size and so I don't think it's all that suitable as a demo.

kjun9:
I think equation 6 in that paper has confused its variables a little bit: it's supposed to be describing exponential decay, so the exponent should be a negative number so I think it should be `t(v) - t(w)` instead of `t(w) - t(v)` unless I'm misunderstanding what `v` and `w` are supposed to be? It also refers to `e = (u, v, t)` so I'm guessing it's gotten confused between using `u, v` vs `v, w` to referring to the target node and its neighbour.

kjun9:
@PantelisElinas @huonw I've added you as the new reviewers

As well as a review of the code itself, I'm interested to hear whether we think this is in a reasonable form to land as `@experimental` or not. Currently I think the demo especially needs some reworking (see [previous comment](https://github.com/stellargraph/stellargraph/pull/552#issuecomment-575460526)), and the tests also need a bit more fleshing out in terms of graduating from experimental entirely.

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+        def step(n, t, is_forward, bias_type=walk_bias):
+            return self._step(
+                n, time=t, is_forward=is_forward, bias_type=bias_type, np_rs=np_rs
+            )
+
+        # sample 1 edge to obtain starting time - any biases are applied as though we are
+        # looking backwards in time.
+        forward = step(node, None, is_forward=False, bias_type=initial_edge_bias)
+        if forward is None:
+            return [node]
```
huonw: If it's a bidirectional walk, couldn't we still have something to do here (i.e. couldn't the forward step fail, but the backward step succeed)?

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+
+        walk = []
+
+        # take steps until walk is of correct length or reached dead ends
+        while len(walk) < length:
+            if forward is None and backward is None:
+                break
+            if forward is not None:
+                n, t = forward
+                walk.append(n)
```
huonw: I think, if `bidirectional` is False, `node` never ends up in the walk: the only modification is this, which is appending the next node chosen by `step`, not `node` itself.

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+
+                # time is None indicates we should obtain the minimum available time for t_0
+                t_0 = time if time is not None else min(times)
+                biases = self._biases(
+                    times, t_0, bias_type=bias_type, is_forward=is_forward
+                )
+            else:
+                # default to uniform random sampling
+                biases = None
+            chosen_index = np_rs.choice(len(neighbours), p=biases)
```
huonw: I think this will be faster using `naive_weighted_choices` (in this file).

```suggestion
            chosen_index = naive_weighted_choices(np_rs, biases)
```

although that doesn't handle `biases == None` (at the moment).

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+
+        Returns:
+            List of lists of nodes ids for each of the random walks
+
+        """
+        self._check_common_parameters(nodes, n, length, seed)
+        np_rs = self._np_random_state if seed is None else np.random.RandomState(seed)
+
+        walks = []
+        for node in nodes:  # iterate over root nodes
```
huonw: This could be a list comprehension:

```python
return [
    self._walk(node, length, bidirectional, initial_edge_bias, walk_bias, np_rs)
    for node in nodes
    for _ in range(n)
]
```

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+            )
+
+        neighbours = [
+            (neighbour, t)
+            for neighbour, t in self.graph.neighbors(node, include_edge_weight=True)
+            if check_time(t)
+        ]
+
+        if len(neighbours) > 0:
+            if bias_type is not None:
```
huonw: I think it might be good to have a more unified treatment of `bias_type`, e.g. instead of just `_biases` have:

```python
def _edge_biases(self, neighbours, time, bias_type, is_forward):
    if bias_type is None:
        return None

    times = [t for _, t in neighbours]
    t_0 = time if time is not None else min(times)
    if bias_type == "exponential":
        return self._exp_biases(times, t_0, decay=is_forward)
    else:
        ...
```

and then this code looks like:

```python
if neighbours:
    biases = self._edge_biases(neighbours, bias_type=bias_type, is_forward=is_forward)
    chosen_index = ...
```

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+                or (not is_forward and t < time)
+            )
+
+        neighbours = [
+            (neighbour, t)
+            for neighbour, t in self.graph.neighbors(node, include_edge_weight=True)
+            if check_time(t)
+        ]
+
+        if len(neighbours) > 0:
```
```suggestion
        if neighbours:
```

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+        # sample 1 edge to obtain starting time - any biases are applied as though we are
+        # looking backwards in time.
+        forward = step(node, None, is_forward=False, bias_type=initial_edge_bias)
+        if forward is None:
+            return [node]
+        backward = node, forward[1]
+
+        walk = []
+
+        # take steps until walk is of correct length or reached dead ends
```
huonw: Per the discussion in section 2.3.3 of http://ryanrossi.com/pubs/nguyen-et-al-WWW18-BigNet.pdf , this is more of a maximum than a "correct" length:

```suggestion
        # take steps until walk is of maximum length or reached dead ends
```

In addition, that section talks about imposing a minimum length on the walks (`ω`); is that something we should do too?

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+        seed=None,
+    ):
+        """
+        Perform a time respecting random walk starting from the root nodes.
+
+        Args:
+            nodes (list): The root nodes as a list of node IDs
+            n (int): Total number of random walks per root node
+            length (int): Maximum length of each random walk
+            bidirectional (bool, default False): Whether the walk extends on both direction.
```
huonw: Does the paper talk about these sort of bidirectional walks?

tests/data/test_temporal_random_walker.py
```diff
@@ -0,0 +1,239 @@
...
+            ("m", "d", 4),
+        ]
+    )
+    g = StellarGraph(g)
+
+    return g
+
+
+class TestTemporalRandomWalk(object):
+    @pytest.mark.skip(reason="times aren't checked at the moment")
```
huonw: What does this mean?

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+                * "exponential" - heavily biased towards more recent edges.
+            walk_bias (str, optional): distribution to use when choosing a random
+                neighbour to walk through. Available options are:
+                * None (default) - neighbours are picked from a uniform distribution
+                * "exponential" - exponentially decaying probability, resulting in a bias
+                    towards shorter time gaps
+            seed (int, optional): Random number generator seed; default is None
+
+        Returns:
+            List of lists of nodes ids for each of the random walks
```
```suggestion
            List of lists of node ids for each of the random walks
```

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
             self._raise_error(
                 "The number of hops for the in and out neighbourhoods must be the same."
             )
+
+
+@experimental(reason="requires more thorough testing and documentation")
+class TemporalRandomWalk(GraphWalk):
```
huonw: This doesn't appear in the documentation. I think you'll need to add it to `docs/api.txt`:

https://github.com/stellargraph/stellargraph/blob/df198b8218d7d6ba06dbd85a12edafb71f6af591/docs/api.txt#L11-L15

stellargraph/data/explorer.py
huonw: Also, I think this length can be exceeded in a bidirectional walk, since it might go from `length - 1` and add both a forward and backward node to have length `length + 1`.

stellargraph/data/explorer.py
```diff
@@ -20,18 +20,22 @@
...
     "SampledHeterogeneousBreadthFirstWalk",
+    "TemporalRandomWalk",
     "DirectedBreadthFirstNeighbours",
 ]
 
 
 import numpy as np
+import math
 import random
+import scipy
```
huonw: I think neither `math` nor `scipy` are used in the current version of the code.

```suggestion
```

stellargraph/data/explorer.py
kjun9: If `bidirectional` is false, then the first iteration of the while loop has `backward = node` (line 911), so node does get added, then `backward` is set to `None`. I'd be interested in suggestions to achieve the same thing with more clarity.. I found the whole bidirectional logic in the while loop a bit clunky to write, I ended up going with something shorter rather than spelling out the 0th iteration first then doing a loop if that makes sense.

tests/data/test_temporal_random_walker.py
kjun9: It was a test that habiba had that I've "skipped" for now since I removed the corresponding check - I thought it was too expensive to check across all edges that the time values are in the correct format. (https://github.com/stellargraph/stellargraph/pull/552#discussion_r367211390)

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+            if forward is None and backward is None:
+                break
+            if forward is not None:
+                n, t = forward
+                walk.append(n)
+                forward = step(n, t, is_forward=True)
+            if backward is not None:
+                n, t = backward
+                walk = [n] + walk
+                backward = step(n, t, is_forward=False) if bidirectional else None
```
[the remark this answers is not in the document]
kjun9: See https://github.com/stellargraph/stellargraph/pull/552#discussion_r368771897

stellargraph/data/explorer.py
kjun9: Hmm I guess I can do
```suggestion
            chosen_index = naive_weighted_choices(np_rs, biases) if biases is not None else np_rs.choice(len(neighbours))
```
I think it's a bit awkward to update `naive_weighted_choices` to handle `biases==None` since then you need another parameter to specify the length

stellargraph/data/explorer.py
```diff
@@ -789,3 +793,136 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+        while len(walk) < length:
+            if forward is None and backward is None:
+                break
+            if forward is not None:
+                n, t = forward
+                walk.append(n)
+                forward = step(n, t, is_forward=True)
+            if backward is not None:
+                n, t = backward
+                walk = [n] + walk
```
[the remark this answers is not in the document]
kjun9: I've updated it to use `deque`

stellargraph/data/explorer.py
huonw: I suspect it may be discarded because it's too short to have information from the full context window being used for word2vec (or whatever), so may break things? I don't know.

stellargraph/data/explorer.py
huonw: Huh, that's subtle and I can definitely see where you're coming from with the clunkiness.

I don't love it, but how about something like:

```python
walk = deque([node])

forward_time = None
backward_time = None

while len(walk) < length:
    if len(walk) == 1:
        # first step is special
        forward, forward_time = step(walk[0], None, is_forward=False, bias_type=initial_edge_bias)
        walk.append(forward)
        if is_bidirectional:
            backward_time = forward_time
    elif forward_time is not None:
        forward, forward_time = step(walk[-1], forward_time, is_forward=True)
        walk.append(forward)

    if len(walk) == length:
        break

    if backward_time is not None:
        backward, backward_time = step(walk[0], backward_time, is_forward=False)
        walk.appendleft(backward)

    if backward_time is None and forward_time is None:
        break
```

With this model, the `walk` indexing and mutation logic could potentially be folded into `step`:

```python
walk = collections.deque([node])

def step(t, is_forward, is_first=False):
    n = walk[-1] if is_forward else walk[0]
    bias = initial_bias_type if is_first else bias_type
    result = self._step(n, t, ...)

    if result is None:
        return None

    next_node, next_time = result
    # the first step treats times as backwards, but still appends as if forward
    if is_forward or is_first: 
        walk.append(next_node)
    else:
        walk.prepend(next_node)

    return next_time
```

and the above becomes:

```python
forward_time = None
backward_time = None

while len(walk) < length:
    if len(walk) == 1:
        # first step is special
       forward_time = step(None, is_forward=False, is_first=True)
       if is_bidirectional:
           backward_time = forward_time
    elif forward_time is not None:
        forward_time = step(forward_time, is_forward=True)

    if len(walk) == length:
        break

    if backward_time is not None:
        backward_time = step(backward_time, is_forward=False)

    if backward_time is None and forward_time is None:
        # no eligible neighbours
        break
```

kjun9: Thanks I'll use that 👍  - I think using `if` instead of `elif` would maintain original behaviour of doing a forward step first after the initial edge selection

stellargraph/data/explorer.py
kjun9: I guess it's easy enough to add, but IMO I don't really see how it applies to temporal walks specifically, rather than random walks in general 🤷‍♂  and I would prefer to avoid adding more parameters that don't get used... but I guess following the paper "correctly" is also important. Do you prefer we add it?

stellargraph/data/explorer.py
```diff
@@ -789,3 +791,152 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+        backward_time = None
+
+        while len(walk) < length:
+            if len(walk) == 1:
+                # initial edge selection
+                forward_time = step(None, is_forward=False, is_first=True)
+                if bidirectional:
+                    backward_time = forward_time
+
+            if forward_time is not None:
```
huonw: Good catch with the `elif` vs. `if` behaviour.

Do we care about `length = 2`? If so, we'll need to add another check for `if len(walk) == length` here, otherwise the second step could add a third element.

Alternatively, and better... this `if` can just move out of the loop (rather silly of me to not do it like this to begin with...):


```python
forward_time = step(None, is_forward=False, is_first=True)
backward_time = forward_time if bidirectional else None

while len(walk) < length:
    if forward_time is not None:
        forward_time = step(forward_time, is_forward=True)

    ...
```

We may need a check for `length == 1`, but I think it's probably ok to add that check to `run`:

```python
def run(...):
    if length <= 1:
        raise ValueError("length: expected at least 2, found {length}")
```

because a walk of length `0` or `1` (or negative) is probably not very useful.

stellargraph/data/explorer.py
```diff
@@ -789,3 +791,152 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+
+@experimental(reason="requires more thorough testing and documentation")
+class TemporalRandomWalk(GraphWalk):
+    """
+    Performs uniform temporal random walks on the given graph
+    """
+
+    def run(
+        self,
+        nodes=None,
```
huonw: Given #704, I imagine some of these should become required? (Just a reminder for this to be updated, depending on the merge order.)

kjun9:
Currently this has been implemented in a way to allow `n` number of temporal walks of length `L` to be run from each node in the graph, but @PantelisElinas pointed out that the paper in fact outlines a method that selects the starting point of each temporal walk from a distribution (uniform, exponential, etc) over all edges, and repeats until there are enough context pairs.

I'm going to fix this up to match the paper, but...

* I'm a bit confused that the paper doesn't seem to address (at least from what I could understand) the fact that starting from edges means there's no guarantee for any particular node that temporal walks will produce any context pairs that include that node. In node2vec, you cannot learn representations for an isolated node, so similarly in this case if a node never ends up getting traversed, we won't be able to produce a representation for it. To me this seems kind of bad 🤷‍♂ 
* Also being edge-centric means this won't fit in as an option in the existing unsupervised sampler class very easily

stellargraph/data/explorer.py
```diff
@@ -780,3 +782,129 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+
+        Returns:
+            List of lists of node ids for each of the random walks
+
+        """
+        np_rs = self._np_random_state if seed is None else np.random.RandomState(seed)
+        walks = []
+        num_cw_curr = 0
+
+        edges = self.graph.edges(include_edge_weight=True)
```
kjun9: This currently doesn't work

stellargraph/data/explorer.py
```diff
@@ -780,3 +782,129 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+        edge_biases = self._temporal_biases(
+            [t for _, _, t in edges],
+            None,
+            bias_type=initial_edge_bias,
+            is_forward=False,
+        )
+
+        # loop runs until we have enough context windows in total
+        while num_cw_curr < num_cw:
```
kjun9: this may not terminate

huonw:
> this is now blocked by: #751 I could probably do something for this just for the networkx graph for now, what do you think @huonw ?

kjun9:
> If there is at least one walk longer than num_cw_curr, it will terminate... it just might take a long time

Makes sense 👍 thanks for spelling out the details! So if I understood correctly, I just need to do something like:
```python
def not_progressing_enough(successes, failures, threshold):
    # (this uses Beta(1, 1) as the prior, since it's uniform on p)
    posterior = stats.beta(1 + successes, 1 + failures)
    return posterior.ppf(0.95) < threshold

while num_cw_curr < num_cw:
    ...
    if not_progressing_enough(success, failures, threshold):
        raise RuntimeError
```

huonw:
Seems reasonable. 👍

stellargraph/data/explorer.py
```diff
@@ -780,3 +783,163 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+        failures = 0
+
+        def not_progressing_enough():
+            # (this uses Beta(1, 1) as the prior, since it's uniform on p)
+            posterior = stats.beta(1 + successes, 1 + failures)
+            return posterior.ppf(0.95) < p_walk_success_threshold
+
+        # loop runs until we have enough context windows in total
+        while num_cw_curr < num_cw:
+            (src, dst), t = self._sample(temporal_edges, edge_biases, np_rs)
```
huonw: Maybe we could have `_sample` return the index, and then write:

```python
first_edge_index = self._sample(edge_biases, np_rs)
src, dst = edges[first_edge_index]
t = times[first_edge_index]
```

This in particular lets us remove the `temporal_edges = ...` line, where the O(num edges) `zip` with a numpy array is unnecessarily expensive (https://github.com/stellargraph/stellargraph/pull/750#discussion_r374377997).

tests/data/test_temporal_random_walker.py
```diff
@@ -0,0 +1,57 @@
...
+        2 -[10]-> 4 -[12]-> 6
+        3 -[2]-> 2 -[10]-> 4
+        5 -[4]-> 4 -[12]-> 6
+    """
+    expected = {(1, 2, 4), (2, 4, 6), (3, 2, 4), (5, 4, 6)}
+
+    rw = TemporalRandomWalk(temporal_graph)
+    num_cw = 20  # how many walks to be sure we're getting valid temporal walks
+
+    for walk in rw.run(num_cw=num_cw, cw_size=3, max_walk_length=3, seed=None):
```
huonw: What about:

```suggestion
    for walk in rw.run(num_cw=num_cw, cw_size=3, max_walk_length=4, seed=None):
```

and including `(1, 2, 4, 6)` and `(3, 2, 4, 6)` in `expected`, just to be sure that there's not something special about `cw_size == max_walk_length`. (I only suggest this since it should be a small change from the current test, if it turns out to be more than just those two changes, it's fine to leave it.)

stellargraph/data/explorer.py
```diff
@@ -780,3 +783,163 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+                    raise RuntimeError(
+                        f"Discarded {failures} walks out of {failures + successes}. "
+                        "Too many temporal walks are being discarded for being too short. "
+                        "Consider using a smaller context window size."
+                    )
+
+        return walks
+
+    def _sample(self, items, biases, np_rs):
+        chosen_index = (
```
huonw: I personally find this easier without the parens, i.e.

```python
if biases is not None:
    chosen_index = ...
else:
    chosen_index = ...
```

but totally up to you.

kjun9:
> The code seems to match the paper more closely now, but I haven't reviewed the notebook at all. Does the notebook give better results?

I also haven't run the notebook at all, since it requires using a 20gb dataset in its current form. I'd prefer to just remove it for now and work on a fresh demo with another dataset?

huonw:
> I also haven't run the notebook at all, since it requires using a 20gb dataset in its current form. I'd prefer to just remove it for now and work on a fresh demo with another dataset?

Sounds reasonable, although just for convenience of retrieval (i.e. minimise how much git magic we have to do to dig it out), maybe you could open a separate draft PR that contains only that notebook?

stellargraph/data/explorer.py
kjun9: Thanks, although I think I'll still have to do something like this for the `temporal_edges = ...` line for now:
```python
edges, times = self.graph.edges(include_edge_weights=True)
edges = list(edges) # convert networkx EdgeView to list
```
which might still be a bit unnecessarily expensive

I can add a note somewhere to remove that as part of removing the networkx stellargraph somewhere, unless you have other thoughts?

tests/data/test_temporal_random_walker.py
kjun9: That's a good idea, I'll give it a go

stellargraph/data/explorer.py
kjun9: Actually I can just use the new stellargraph in the test

kjun9:
I've opened https://github.com/stellargraph/stellargraph/pull/773 to keep track of the old notebook

stellargraph/data/explorer.py
```diff
@@ -854,13 +853,17 @@ def run(
         failures = 0
 
         def not_progressing_enough():
-            # (this uses Beta(1, 1) as the prior, since it's uniform on p)
+            # Estimate the probability p of a walk being long enough; the 95% percentile is used to
+            # be more stable with respect to randomness. This uses Beta(1, 1) as the prior, since
+            # it's uniform on p
             posterior = stats.beta(1 + successes, 1 + failures)
```
huonw: Minor thing, but since we're only querying the `ppf`, it turns out to be much faster to do that directly rather than construct the distribution object:
```python
from scipy import stats
%timeit -n 100 stats.beta.ppf(0.95, 11, 10)
# 198 µs ± 18.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
%timeit -n 100 stats.beta(11, 10).ppf(0.95)
# 674 µs ± 31.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

As such:

```suggestion
            posterior = stats.beta.ppf(0.95, 1 + successes, 1 + failures)
```

and `return posterior < p_walk_success_threshold`.

huonw:

kjun9:
New PR: https://github.com/stellargraph/stellargraph/pull/787

Issue #411, opened 2019-05-30
Chebyshev filter in FullbatchNodeGenerator returns transformed features as a list instead of a numpy array.


habiba-h:
### Description

Chebyshev filter in `FullbatchNodeGenerator` returns the transformed features as a list instead of a numpy array. This results in error when `FullBatchNodeGenerartor.feature` is accessed. Such as when `FullBatchNodeGenerartor.feature.shape` is passed to the `node_model` in `GCN`. 

### Done Checklist (Bug)
- [ ] Bug fixed
- [ ] Branch and Pull Request build on CI
- [ ] Branch and Pull Request pass unit tests on CI
- [ ] Branch and Pull Request pass integration tests on CI
- [ ] Version number reflects new status
- [ ] Peer Code Review Performed
- [ ] Code well commented
- [ ] CHANGELOG.md updated

timpitman:
See also #825 so that we're actually using this.

PR #558, opened 2019-12-12
Return transformed features as numpy array

kjun9:
Part of #411


stellargraph/core/utils.py
```diff
@@ -195,7 +195,7 @@ def preprocess_adj(adj, symmetric=True):
         if isinstance(k, int) and k >= 2:
             # default minimum degree for Chebyshev polynomials.
             T_k = chebyshev_polynomial(rescale_laplacian(normalized_laplacian(A)), k)
-            features = [features] + T_k
+            features = np.array([features] + T_k)
```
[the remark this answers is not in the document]
kjun9: Hmm I see.. so the tests were wrong as well then? It appears to be expecting the outer dimension to be 4 (as in `1 + (k + 1)` where the first `1` is from `[features]` and `(k +1)` are the chebyshev polynomials from orders 0 to k), but what you're saying makes more sense to me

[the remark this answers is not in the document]
habiba-h: @ huonw @kjun9  yes, converting to a numpy array does not work. Exactly this is what happens.  The `FullBatchNodeGenerator` should return a 2D matrix with one row of features for each node. Whereas the `FullBatchNodeGenerator` with `method='chebyshev'` results in a 2D numpy array with dimension` (1,k+2)`.
The unit tests aren't complete to check that. 
I caught this error at `node_model()` method with the `FullBatchNodeGenerator` using the chebyshev filter.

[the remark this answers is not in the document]
kjun9: Okay thanks for clarifying, I'll have another go 👍

stellargraph/core/utils.py
```diff
@@ -195,7 +195,7 @@ def preprocess_adj(adj, symmetric=True):
         if isinstance(k, int) and k >= 2:
             # default minimum degree for Chebyshev polynomials.
             T_k = chebyshev_polynomial(rescale_laplacian(normalized_laplacian(A)), k)
-            features = [features] + T_k
+            features = np.concatenate([features] + [t.todense() for t in T_k], axis=1)
```
huonw: This seems like it's going to absolutely explode the memory usage of this, since the `todense` calls will create arrays with `(k + 1) * n * n` elements in memory. Maybe we could instead have everything be sparse as it was: that is, convert `features` to the appropriate sparse format and concatenate them sparsely?

What do you and @habiba-h  think?

habiba-h: Yes, I agree,  we should convert `features` to sparse and then concatenate.

kjun9: I think there's going to be extra logic required downstream if `features` becomes a sparse matrix. For instance, there's currently a separate `FullBatchNodeSequence` vs `SparseFullBatchNodeSequence` that works with dense vs sparse adjacency matrices that get fed into keras. Does that sound right?

kjun9:
> @kjun9, The `FullBatchNodeGenerator` with `method=chebyshev` returns 2D features numpy ndarray whereas a 3D array is required. The first dimension is just a padding of 1. I don't remember the reason for that. Perhaps @adocherty can explain this. But in any case at the moment features array should be 3D (1, num_nodes, num_features)

@habiba-h I think the reshaping you're thinking of is done in [`FullBatchNodeSequence`](https://github.com/stellargraph/stellargraph/blob/b93459d7aa1b05956005cf4cbb034a62d20246c3/stellargraph/mapper/sequences.py#L371). For the function I'm updating, if `method=gcn` or `method=sgcn` the output `features` remains as a 2D matrix as well, so 2D seems correct here?

stellargraph/core/utils.py
habiba-h: I know a little bit about how and where sparse and dense adjacency matrices are used but not sure how the sparse features will be affected downstream.

habiba-h:
@kjun9 You are right. Its the `FullBatchNodeSequence`.

stellargraph/core/utils.py
adocherty: I agree, there should not be a transformation to dense matrices here. It was done because at the time there was no support for sparse matrices. You'll need to return sparse matrices here (not convert them to dense) and then convert a list of sparse matrices to dense in the `FullBatchNodeSequence` class or to index/value arrays in the `SparseFullBatchNodeSequence` class.

adocherty: One issue I've had for a long time with this is that the `GCN_Aadj_feats_op` accepts features & adjacency matrix A and returns "transformed" features & new matrix "A". However, what the Chebyshev function does is add the Chebyshev matrix polynomials of A to the list of features, which is wrong IMHO.

I think the function should return a list of matrices as A, and for the Chebyshev function this will be a the transformed list. As far as I know, none of the methods actually transform the features.

Note that the higher the polynomial order the denser these matrices will become.

adocherty:
I've looked at the code and added some comments which will hopefully clarify the issues here.  

A good place to start might be to compare `SparseFullBatchNodeSequence` in and `RelationalFullBatchNodeSequence` in `sequences.py`.

In `FullBatchNodeSequence` a single matrix is assumed:
```
        self.A_indices = np.expand_dims(
            np.hstack((A.row[:, None], A.col[:, None])), 0
        ).astype("int64")
        self.A_values = np.expand_dims(A.data, 0)
```

@kieranricardo extended this to multiple matrices in `RelationalFullBatchNodeSequence` i
```
            self.A_indices = [
                np.expand_dims(np.hstack((A.row[:, None], A.col[:, None])), 0)
                for A in As
            ]
            self.A_values = [np.expand_dims(A.data, 0) for A in As]
            self.As = self.A_indices + self.A_values
```

I think that this is what we need to do to support the Chebyshev method, where we change the functions to not take the features matrix and return a list of matrices (GCN, GAT etc will return a list of one).

stellargraph/core/utils.py
kjun9: > I think the function should return a list of matrices as A, and for the Chebyshev function this will be a the transformed list. As far as I know, none of the methods actually transform the features.

> I think that this is what we need to do to support the Chebyshev method, where we change the functions to not take the features matrix and return a list of matrices (GCN, GAT etc will return a list of one).

Ah, thanks for the clarification! So if I understood correctly... this method can be something like `def transform_adj(adj: Adj, method="gcn", k=1) -> List[Adj]`, and I can update the `GCN` layer, `FullBatchNodeSequence` and `SparseFullBatchNodeSequence` to always use a list of adjacency matrices?

kjun9: Actually, after looking through the `rgcn` implementation a bit, I'm suspecting that plugging the chebyshev matrix polynomials of A into RGCN might basically have the desired effect? In which case, it might make sense to just combine the two classes `GCN` and `RGCN` instead of implementing the multi-adjacency-matrix-handling logic twice

kjun9:
I think doing this properly actually requires refactoring `GCN` and `RGCN`, likely just unifying them to be 1, which seems like something to be tackle as a bigger separate task https://github.com/stellargraph/stellargraph/issues/621

Will close this for now

PR #562, opened 2019-12-16
Pin coverage version

kjun9:
Latest release of coverage 5.0 is incompatible with coveralls python client

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ pytest==3.9.3
 pytest-benchmark>=3.1
 pytest-cov>=2.6.0
 coveralls>=1.5.1
+coverage==4.5.4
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -13,3 +13,4 @@ pytest==3.9.3
 pytest-benchmark>=3.1
 pytest-cov>=2.6.0
 coveralls>=1.5.1
+coverage==4.5.4
```
huonw: Did you consider
```suggestion
coverage<5.0
```
to match `coveralls`'s dependency specification (https://github.com/coveralls-clients/coveralls-python/blob/c23508dd017677c2f69c1b2bc1561671d7767d84/setup.py#L36)?

or 

```suggestion
coverage>=4.4,<5.0
```
to match the intersection of that and https://github.com/pytest-dev/pytest-cov/blob/86b0e6f5e66fba82be7ad6b6733dfe966e443335/setup.py#L125 ?

kjun9: Ah yes that seems better, thanks

Issue #546, opened 2019-12-06
tensorflow should not be listed as dependency in setup.py


amaiya:
**Describe the bug**
The `stellargraph` library includes `tensorflow` (i.e., `"tensorflow>=1.14,<1.15"`) as a dependency in setup.py.  If a user has `tensorflow-gpu` installed, then `pip3 install stellargraph` installs `tensorflow`, which overwrites things and causes all models to be trained on CPU instead of GPU.  To fix, users must remove the `tensorflow*` installations and reinstall `tensorflow-gpu`.   This was true for TF 1.14 at least, and I think may hold true with TF 2.0.

**To Reproduce**
On a system where `tensorflow-gpu` is installed, install stellargraph with: `pip3 install stellargraph`

**Expected behavior**
Installing `stellargraph` shouldn't corrupt existing `tensorflow-gpu` installations.

**Solution**
The  `tensorflow` entry should be removed as a dependency in `setup.py` and requirements.  Instead, the **Installation** section in the `stellargraph` README should include a note about installing `tensorflow` (or `tensorflow-gpu`) separately when installing `stellargraph`.  This is how other libraries typically handle it.  For instance, see the installation instructions for [graph_nets](https://github.com/deepmind/graph_nets):

>To install the Graph Nets library for CPU, run:

>$ pip install graph_nets "tensorflow>=1.15,<2" tensorflow_probability

>To install the Graph Nets library for GPU, run:

> $ pip install graph_nets "tensorflow_gpu>=1.15,<2" tensorflow_probability

In `setup.py` of **graph_nets**:
```
    # Additional "tensorflow" and "tensorflow_probability" requirements should
    # be installed separately (See README).
    install_requires=[
        "absl-py",
        "dm-sonnet<2",
        "future",
        "networkx",
        "numpy",
        "setuptools",
        "six",
    ],
```

adocherty:
Hi @amaiya,

Thanks for trying the library! We understand the issue and will look at updating the `setup.py` file to remove a hard dependency on `tensorflow`.  I am thinking instead of removing tensorflow as a requirement we can add it as a 'extra' requirement, so that you can specify stellargraph either with gpu or cpu support, for example:

To install `stellargraph` + `tensorflow-gpu` you can use the following:
```
pip install stellargraph[gpu]
```

To install `stellargraph` + `tensorflow` you can use the following:
```
pip install stellargraph[cpu]
```

Finally, to install stellargraph without change to your existing tensorflow libraries:
[...]

PR #567, opened 2019-12-16
Move tensorflow to extras

kjun9:
Part of #546 

I'm not sure if we would land this as is, since this feels like a big breaking change, so I'm open to suggestions for how to make migration easier.

I'm hoping to figure out a way to print a warning if tensorflow is not installed with stellargraph as @adocherty has suggested in #546 but let me know if you have ideas.

I also didn't add a `stellargraph[gpu]` option for now since we're not verifying it in CI currently, although the user would have the option to try use their own gpu tensorflow installation regardless.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,10 @@ StellarGraph depends upon.
 The StellarGraph library requires [Keras](https://keras.io/), so you'll need to install Keras and a selected backend (we recommend tensorflow, which is used to test StellarGraph).  Other requirements are the NetworkX library (to create and modify graphs and networks), numpy (to manipulate numeric arrays), pandas (to manipulate tabular data), and gensim (to use the Word2Vec model), scikit-learn (to prepare datasets for machine learning), and matplotlib (for plotting).
 -->
 
+Stellargraph requires a compatible version of Tensorflow installed. A compatible
+CPU version of Tensorflow can be installed together with Stellargraph by referring
+to `stellargraph[cpu]` in any of the installation steps below.
+
 The StellarGraph library can be installed from PyPI, from Anaconda Cloud, or directly from GitHub, as described below.
 
 #### Install StellarGraph using PyPI:
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ URL = "https://github.com/stellargraph/stellargraph"
 
 # Required packages
 REQUIRES = [
+    "tensorflow>=1.14,<1.15",
     "numpy>=1.14",
     "scipy>=1.1.0",
     "networkx>=2.2",
@@ -42,7 +43,7 @@ REQUIRES = [
 EXTRAS_REQURES = {
     "demos": ["numba", "jupyter", "seaborn"],
     "test": ["pytest", "pytest-benchmark"],
-    "cpu": ["tensorflow>=1.14,<1.15"],
+    "gpu": ["tensorflow-gpu>=1.14,<1.15"],
 }
 
 # Long description
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,6 @@ URL = "https://github.com/stellargraph/stellargraph"
 
 # Required packages
 REQUIRES = [
-    "tensorflow>=1.14,<1.15",
     "numpy>=1.14",
     "scipy>=1.1.0",
     "networkx>=2.2",
@@ -43,6 +42,7 @@ REQUIRES = [
 EXTRAS_REQURES = {
     "demos": ["numba", "jupyter", "seaborn"],
     "test": ["pytest", "pytest-benchmark"],
+    "cpu": ["tensorflow>=1.14,<1.15"],
 }
 
 # Long description


kieranricardo:
This could potentially be confusing to less experienced end users because currently we only support `tensorflow==1.14 `. Ideally a simple `pip install stellargraph` will just work. 

Could we keep the current behaviour of `pip install stellargraph` and just add a `pip install stellargraph[gpu]' option? That way we avoid a breaking change as well

kjun9:
@kieranricardo 

Hmm yes I agree that ideally `pip install stellargraph` still works out of the box without any extra steps.. but I think there still needs to be a way for users to install `stellargraph` without `tensorflow` bundled together, and unfortunately it seems that using `extras_require` only lets us add _additional_ dependencies rather than remove/replace one. And this makes it a difficult to achieve something like:
```
pip install stellargraph  # installs with tensorflow-cpu
pip install stellargraph[-tf]  # installs without tensorflow
pip install stellargraph[gpu]  # installs with tensorflow-gpu (and not tensorflow-cpu)
```
Example use-case being someone wanting to write a `requirements.txt` file that contains `stellargraph` and `tensorflow-gpu`

Maybe another option is for us to publish a separate pypi package `stellargraph-no-tf` that doesn't include tensorflow?

kieranricardo:
Good point! Two separate packages might be the way to go. 

Not sure if I'm missing something but I think a `stellargraph` and a `stellargraph-gpu` package should cover all bases? This seems to be a pretty common way of getting around this issue

kjun9:
> Not sure if I'm missing something but I think a stellargraph and a stellargraph-gpu package should cover all bases?

> Why would we need this option: pip install stellargraph[-tf] # installs without tensorflow?
I agree with Kieran that stellargraph and stellargraph-gpu should be enough - or am I missing some edge case?

Yeah it does cover all bases, but I thought it might be worthwhile to try avoid packaging releases that are tightly coupled to either of `tensorflow` and `tensorflow-gpu` implementations and instead just depend on the tensorflow API in general. If a single installation of `stellargraph` works with either tensorflow CPU or GPU, then switching between the two remains as a concern addressed by `tensorflow` rather than us.

> For a separate stellargraph-no-tf package: what would happen if a user installs it with a tf version we don't support (!=1.14)?

I now realise `pip` wouldn't show the error message about incompatible versions in this case if `tensorflow` isn't listed as a dependency at all... so yes this will be problematic, and so now I'm also leaning towards just going with `stellargraph` and `stellargraph-gpu` 😅

kieranricardo:
I found a quick but slightly hacking work-around: 

https://github.com/ska-sa/montblanc/blob/rime-tf/setup.py 

This `setup.py` detects whether CUDA is installed and whether there's any available GPU devices. If a GPU and CUDA is detected it install `tensorflow-gpu` and installs `tensorflow` otherwise.

kjun9:
> I'm not sure that we'd want to use this, is seems like a worse solution than just having another package but it would be easier/quicker to do. Thoughts?

Nice find, but I'd prefer to just have another package to keep things simple

adocherty:
I think having a second package seems like a good solution, especially having `stellargraph-gpu` as it allows us to specify the library version.

kjun9:
If we're all happy with the second package option, I can close this 👍 

I'm not so familiar with the release procedure - is there somewhere we can update a script/documentation to make sure we release the second package?

youph:
> If we're all happy with the second package option, I can close this

Shouldn't `setup.py` be updated to have `gpu` in extras (instead of `cpu`)?

kjun9:
@youph doesn't seem like there's an easy way to "replace a dependency" (`tensorflow` -> `tensorflow-gpu`) using `extras_require`. Some relevant discussion [here](https://stackoverflow.com/questions/53376135/install-by-default-optional-dependencies-in-python-setuptools) about using a post install script, but I think it's too hacky.

We're now saying we switch out `tensorflow` for `tensorflow-gpu` "manually" when we publish the second package `stellargraph-gpu`.

youph:
@kjun9 I've updated my comment, would the suggested structure of `setup.py` work?

kjun9:
>  If I remember correctly, installing tensorflow-gpu after tensorflow is ok.

What if someone already has `tensorflow-gpu` installed, will it still be okay with `pip install stellargraph[gpu]` with that structure?

youph:
> > If I remember correctly, installing tensorflow-gpu after tensorflow is ok.
> 
> What if someone already has `tensorflow-gpu` installed, will it still be okay with `pip install stellargraph[gpu]` with that structure?

Not sure, but I think it would re-install `tensorflow==1.14`, then re-install `tensorflow-gpu==1.14` in that case. One way to check would be to test it on `fractal`.

kjun9:
> Not sure, but I think it would re-install tensorflow==1.14, then re-install tensorflow-gpu==1.14 in that case. One way to check would be to test it on fractal.

I think with `pip` it might just install `tensorflow` and not install `tensorflow-gpu` if it thinks that requirement is already satisfied

youph:
Actually, `tensorflow-gpu` doesn't have the same versions as `tensorflow`: 
```
ERROR: Could not find a version that satisfies the requirement tensorflow-gpu==1.14 (from versions: 0.12.1, 1.0.0, 1.0.1, 1.1.0rc0, 1.1.0rc1, 1.1.0rc2, 1.1.0)
```

kjun9:
Hmm not really sure what's going on, but I can see all the versions here, which seem to match `tensorflow` https://pypi.org/project/tensorflow-gpu/#history

And this page also seems to think so: https://www.tensorflow.org/install/gpu

youph:
Ok, I've done some tests on `fractal` with the `setup.py` structure suggested above, and here are the results:
- `pip install -e .` in a clean environment installs `tensorflow==1.14.0` (cpu version)
- `pip install -e .[gpu]` in a clean environment installs both `tensorflow==1.14.0` and `tensorflow-gpu==1.14.0`
- `pip install -e .[gpu]` in an environment with `tensorflow` only installs `tensorflow-gpu` 
- `pip install -e .` in an environment with both `tensorflow==1.14.0` and `tensorflow-gpu==1.14.0` does not touch tensorflow libraries, and only installs `stellargraph`

What cases have I missed?

kjun9:
What cases have I missed?

`pip install -e .[gpu]` in an environment with only `tensorflow-gpu` installed?

youph:
> What cases have I missed?
> 
> `pip install -e .[gpu]` in an environment with only `tensorflow-gpu` installed?

- If only `tensorflow-gpu==1.14` is installed, this installs `tensorflow==1.14.0`, then `stellargraph`. Doesn't touch `tensorflow-gpu`
- If only `tensorflow-gpu!=1.14` is installed, this installs `tensorflow==1.14.0` and `tensorflow-gpu==1.14.0`, then `stellargraph`.

kjun9:
> If only tensorflow-gpu==1.14 is installed (the supported version), this installs tensorflow==1.14.0, then stellargraph. Doesn't touch tensorflow-gpu as it is already the required version.

So isn't this quite a common situation and will be problematic?

youph:
Why would it be problematic?

kjun9:
> Why would it be problematic?

https://github.com/stellargraph/stellargraph/issues/546

youph:
Testing this scenario on `fractal`...

youph:
Should we then do
```
# Required packages
REQUIRES = [
    "numpy>=1.14",
    "scipy>=1.1.0",
    "networkx>=2.2",
...
]
...
EXTRAS_REQURES = {
    "demos": ["numba", "jupyter", "seaborn"],
    "test": ["pytest", "pytest-benchmark"],
    "cpu": ["tensorflow-cpu>=1.14,<1.15"],
    "gpu": ["tensorflow-gpu>=1.14,<1.15"]
}
```
?

kjun9:
That was the original idea, but this would mean users need to install with `pip install stellargraph[cpu]` for the most obvious standalone installation, and is a big breaking change. It might also mean a `requirements.txt` file like
```
tensorflow==2.0
stellargraph
```
doesn't produce any errors about incompatible versions.

I think just publishing a second package called `stellargraph-gpu` is probably the safest option.

youph:
Ok, so then the only difference between the 2 packages would be the TF requirement in `setup.py`, right? I.e., `stellargraph` requires `tensorflow`, and `stellargraph-gpu` requires `tensorflow-gpu`?

kjun9:
> Ok, so then the only difference between the 2 packages would be the TF requirement in setup.py, right? I.e., stellargraph requires tensorflow, and stellargraph-gpu requires tensorflow-gpu?

Yep

youph:
Another alternative would be to always require `tensorflow-gpu==1.14`: if it's installed on a system without CUDA and Nvidia GPU, it will fall back to using CPU. However, the downside of this would be that it won't work on MacOS, since there is no `tensorflow-gpu==1.14` for MacOS

kjun9:
> And would we have to maintain two github repos? Or how do we handle the 2 packages?

I'm thinking we could maintain two `setup.py`s (or 1 `setup.py` + a script that generates the other one by switching out `tensorflow`->`tensorflow-gpu`) in one repo. Then when we need to publish we use the second one for `stellargraph-gpu`. And we can automate this in BK

youph:
Ok, sounds good, although I'm out of depth on this one

adocherty:
Just for fun, note that tensorflow v1.15 only has one package with both CPU and GPU, while tensorflow 2.0.0 has `tensorflow` and `tensorflow-gpu` packages 😑

See: https://www.tensorflow.org/install/gpu

Who knows where Google is going with this!

README.md
```diff
@@ -138,6 +138,10 @@ StellarGraph depends upon.
 The StellarGraph library requires [Keras](https://keras.io/), so you'll need to install Keras and a selected backend (we recommend tensorflow, which is used to test StellarGraph).  Other requirements are the NetworkX library (to create and modify graphs and networks), numpy (to manipulate numeric arrays), pandas (to manipulate tabular data), and gensim (to use the Word2Vec model), scikit-learn (to prepare datasets for machine learning), and matplotlib (for plotting).
 -->
 
+Stellargraph requires a compatible version of Tensorflow installed. A compatible
+CPU version of Tensorflow can be installed together with Stellargraph by referring
```
adocherty: Should this be "GPU" as there is no CPU in the `EXTRAS_REQUIRES`?

kjun9:
Based on the discussion above, I think we don't merge this at all, but instead update the release procedure. I'll bring this up for discussion during the next sprint planning

PR #569, opened 2019-12-18
fixed unit test for symmetric normalized laplacian

kieranricardo:
This branch fixes the unit test `test_normalized_laplacian` in `tests/core/test_utils.py`.

Previously the unit test erroneously asserted that `normalized_laplacian(Aadj).sum() == 1`. For a symmetrically normalised Laplacian this is false. Instead, I replaced this assertion with a check that the smallest eigenvalue of `normalized_laplacian(Aadj)` is 0, and that its largest eigenvalue is less than 2 as these are properties of symmetrically normalised Laplacians. 

Reviewer checklist:

- [ ] all unit tests pass
- [ ] `test_normalized_laplacian` passes consistently
- [ ] the eigenvalue checks in `test_normalized_laplacian` are implemented correctly

### DIFF
--- a/tests/core/test_utils.py
+++ b/tests/core/test_utils.py
@@ -70,14 +70,18 @@ def test_normalize_adj():
 
 
 def test_normalized_laplacian():
-    node_list = list(pytest.G.nodes())
     Aadj = pytest.G.to_adjacency_matrix()
-    laplacian = normalized_laplacian(Aadj)
-    assert 1 == pytest.approx(laplacian.sum(), 0.2)
-    assert laplacian.get_shape() == Aadj.get_shape()
+    laplacian = normalized_laplacian(Aadj).todense()
+    eigenvalues, _ = np.linalg.eig(laplacian)
+
+    # min eigenvalue of normalized laplacian is 0
+    # max eigenvalue of normalized laplacian is <= 2
+    assert eigenvalues.min() == pytest.approx(0, abs=1e-7)
+    assert eigenvalues.max() <= (2 + 1e-7)
+    assert laplacian.shape == Aadj.get_shape()
 
     laplacian = normalized_laplacian(Aadj, symmetric=False)
-    assert 1 == pytest.approx(laplacian.sum(), 0.2)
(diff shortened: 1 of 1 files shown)


tests/core/test_utils.py
```diff
@@ -70,14 +70,18 @@ def test_normalize_adj():
...
     Aadj = pytest.G.to_adjacency_matrix()
-    laplacian = normalized_laplacian(Aadj)
-    assert 1 == pytest.approx(laplacian.sum(), 0.2)
-    assert laplacian.get_shape() == Aadj.get_shape()
+    laplacian = normalized_laplacian(Aadj).todense()
+    eigenvalues, _ = np.linalg.eig(laplacian)
+
+    # min eigenvalue of normalized laplacian is 0
+    # max eigenvalue of normalized laplacian is <= 2
+    assert 0 == pytest.approx(eigenvalues.min(), abs=1e-7)
```
huonw: I know you're reproducing the existing code, but `pytest.approx` is an asymmetric comparison, with the reference/expected/true/exact value passed to the function and the computed value on the other side of the comparison operator: https://docs.pytest.org/en/latest/reference.html#pytest-approx

> Because the relative tolerance is only calculated w.r.t. b, this test is asymmetric and you can think of b as the reference value.

```suggestion
    assert eigenvalues.min() == pytest.approx(0, abs=1e-7)
```
and to match this, I'd personally do:

```python
    assert eigenvalues.max() <= pytest.approx(2, abs=1e-7)
```
on the next line, too.

kieranricardo: thanks for the quick review! ah I didn't release it was asymmetric, I've implemented your first suggestion.

Your second suggestion yields this error: `TypeError: '<=' not supported between instances of 'numpy.float64' and 'ApproxScalar'`
I could be missing it, but I had a quick scan through the `pytest` docs and couldn't find `pytest.approx` support for inequalities.

huonw: Oh, ok, I think I misread the "Warning" at the end of the `approx` docs; sorry.

Issue #575, opened 2020-01-01
Check demo scripts and notebooks on CI


huonw:
### Description

Currently, our notebooks are not validated on CI, which can lead to problems like #570 / #571, where they aren't updated after a change.

### User Story

**As** a data scientist

**I want to** be able to pick up any working demo

**So that** I learn to use stellargraph

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

timpitman:
We could avoid the block on #610 by starting with just cora and the notebooks test script at  https://github.com/stellargraph/stellargraph/blob/develop/scripts/test_demos.py

PR #576, opened 2020-01-01
Test notebooks on CI

huonw:
See: #575


huonw:
This will be much easier after #610, because the notebooks can then automatically download and load the datasets.

huonw:
This is blocked on #610, and will likely require changes after doing so. Thus, going to close this for now.

PR #582, opened 2020-01-02
Build the docker image using the buildkite docker-compose plugin

huonw:
This makes our docker image building more declarative by using https://github.com/buildkite-plugins/docker-compose-buildkite-plugin

It has to add a `pre-command` hook to set the environment variable used as an image tag.


.buildkite/hooks/pre-command
```diff
@@ -0,0 +1,19 @@
...
+echo "~~~ setting docker image tag"
+case "${BUILDKITE_BRANCH}" in
+  develop)
+    IMAGE_TAG="latest"
+    ;;
+  *)
+    if [ "$BUILDKITE_PULL_REQUEST" != "false" ]; then
+      IMAGE_TAG="PR-$BUILDKITE_PULL_REQUEST"
+    else
+      IMAGE_TAG="testing"
```
kjun9: Shouldn't this be unique as well? Or we don't care if they step on each other unless it's a PR?

TheDen: This could be `stellargraph-public-stellargraph-build-${BUILDKITE_BUILD_NUMBER}` so it's the same auxiliary image that's pushed anyway?

huonw: This is currently just preserving existing behaviour; but I can change it here too.

PR #583, opened 2020-01-02
Replace print("Warning: ...") with warnings.warn("...")

huonw:
This ensures:

- warnings can be controlled, e.g. https://docs.python.org/3/library/warnings.html#overriding-the-default-filter , `warnings.simplefilter("ignore")` to silence them all, `warnings.simplefilter("error")` to turn them into hard errors (it's easier to fix them with full stack traces)
- warnings get printed to `stderr` and so do not contaminate `stdout`
- there's more context for what an warning is caused by, due to printing line numbers

See: #584

### DIFF
--- a/stellargraph/core/graph_networkx.py
+++ b/stellargraph/core/graph_networkx.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2019 Data61, CSIRO
+# Copyright 2019-2020 Data61, CSIRO
 #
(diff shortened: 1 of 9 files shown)


huonw:
@PantelisElinas I've requested a review from you because I don't have the full context for why warnings were emitted using `print`, and there may be a great reason for it. Since you've been working on the library for much longer, you're more likely to know it 😄

PantelisElinas:
@huonw 

we never discussed how warnings should be printed. 

P.

PR #588, opened 2020-01-02
Add labels automatically when using issue templates

huonw:
This extends the YAML front matter for the existing templates to automatically add `bug`, `enhancement` or `sg-library` labels, as appropriate. This will help with high-level triage and the zenhub view.


huonw:
@timpitman does this fit into your ideas of how we'll manage our issues? (@PantelisElinas I had requested a review from you while Tim was away, but now he's back he can take it on.)

PR #590, opened 2020-01-02
Tweak the Dockerfiles for better caching

huonw:
This uses the technique of installing dependencies in one layer, and then copying the source code in another, so that changes to the source don't require reinstalling all the dependencies. To do this effectively, it has to turn the multi-stage build into a single stage one. However, this only increases the compressed image size by 7MB (326MB -> 333MB) because the only thing that was being excluded from the final image was the actual `stellargraph` source.

This then results in the majority of the image being shared, because the layer containing the dependencies is rarely recomputed.

<table><tr><th>situation<br></th><th>branch</th><th>build (s)</th><th>push (s)</th><th>total build &amp; publish on BK (s)</th><th>compressed (MB)</th><th>uncompressed (MB)</th></tr><tr><td rowspan="2">same deps &amp; source</td><td>this PR</td><td>5</td><td>7</td><td>60 (31+29)</td><td>0</td><td>0</td></tr><tr><td>develop*</td><td>150</td><td>73</td><td>341 (260+71)</td><td>270</td><td>953</td></tr><tr><td rowspan="2">same deps<br>(source changes)</td><td>this PR</td><td>14</td><td>18</td><td>83 (52+31)</td><td>7.2</td><td>54.4</td></tr><tr><td>develop</td><td>150</td><td>73</td><td>341 (260+71)</td><td>270</td><td>953</td></tr><tr><td rowspan="2">different deps</td><td>this PR</td><td>121</td><td>69</td><td>285 (210+75)</td><td>277</td><td>1019</td></tr><tr><td>develop</td><td>150</td><td>73</td><td>341 (260+71)</td><td>270</td><td>953</td></tr></table>

(* the computed installed dependencies are slightly different, so caching seems to be ineffective, and thus this ends up being recomputed.)

The "total build &amp; publish on BK" time is the time to run the build step and then the publish step. This captures some of the speed-up of sharing more layers, because a significant amount of this time is the time required to download the images: when the biggest layers are shared, most agents will have those layers available and so not have to redownload them.

This includes a substitution for [the `no-cache` setting](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin#no-cache-optional-build-only), so that we can have a scheduled build that refreshes the cache for any dependency updates.


huonw:
> My only concern is that we now have a COPY . but keeping .dockerignore updated will mitigate accidentally publishing sensitive data

Good point, but because this is being built on CI, I think we're probably ok: unless we do something weird with hooks, anything that gets into the repo on CI will also be in the source on GitHub and thus already be public?

PR #591, opened 2020-01-02
Use docker image for testing

huonw:
Experiment.


huonw:
I don't think this is the right approach: we need to test both 3.6 and 3.7, and potentially other configurations, and so could have a combinatorial explosion in the number of docker images we need to coordinate. Also, it likely won't save _that_ much time.

PR #592, opened 2020-01-03
Check yamllint and prettier YAML formatting on CI

huonw:
Yamllint will catch minor issues like `foo: yes` becoming `foo: true` (not `foo: "yes"`) implicitly, and using `prettier` to check the formatting will ensure our diffs are more focused, because there won't be any extraneous indentation or syntax changes.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -65,7 +65,7 @@ steps:
       docker#v3.4.0:
         image: "stellargraph/hadolint:snapshot"
 
-  - label: ":prettier::yaml::json::markdown: format YAML"
+  - label: ":prettier::yaml: format YAML"
     plugins:
       docker#v3.4.0:
         image: "stellargraph/prettier:${BIGDATA_DOCKER_TAG-snapshot}"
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -64,8 +65,25 @@ steps:
       docker#v3.4.0:
         image: "stellargraph/hadolint:snapshot"
 
+  - label: ":prettier::yaml: format YAML"
+    plugins:
+      docker#v3.4.0:
+        image: "stellargraph/prettier:${BIGDATA_DOCKER_TAG-snapshot}"
```
kjun9: Would it make sense to just have an environment variable for each image tag like
```suggestion
        image: "stellargraph/prettier:${PRETTIER_IMAGE_TAG-snapshot}"
```
and similarly for `yamllint`, rather than referring to `BIGDATA_DOCKER` which doesn't seem relevant in this repo

huonw: It would, indeed... or:

```suggestion
        image: "stellargraph/prettier:${STELLAR_DEVOPS_TAG-snapshot}"
```

to have them be the same for all of the images, and then set up a triggered build of this repo from stellar-devops to test any new images.

However, we're currently just using `snapshot` for the other images (`hadolint` and, in #551, `black`), so I'll leave it at:

```suggestion
        image: "stellargraph/prettier:snapshot"
```

PR #593, opened 2020-01-05
Format markdown with prettier

huonw:
This gives a consistent style, and ensures we don't have things like trailing whitespace that may cause spurious changes in diffs.


huonw:
Not going to work on this for a bit, so closing.

PR #594, opened 2020-01-05
Try running pylint on codeclimate

huonw:
Experiment.


huonw:
We probably don't need more of these checks for now, given we still have 153 existing issues.

PR #595, opened 2020-01-05
Disable code complexity checks on Code Climate

huonw:
These are very noisy with our current state of the repository and code complexity metrics seem to be low value, in many cases: https://en.wikipedia.org/wiki/Cyclomatic_complexity#Correlation_to_number_of_defects

If we improve/fix other higher-value warnings, we can potentially re-enable this in future and see what it would take to address.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -8,6 +8,9 @@ checks:
   file-lines:
     enabled: false
 
+  method-complexity:
+    enabled: false
+
   method-count:
     enabled: false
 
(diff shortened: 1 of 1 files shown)


huonw:
Per the link above, this "fixes" 228 of the cyclomatic and cognitive complexity issues that code climate flags:

image

huonw:
@PantelisElinas I added you as a reviewer because this is changing an existing check, and it would be good to have some thoughts on someone who has used interacted with it a lot.

PantelisElinas:
@huonw 

I have always and 100% ignored anything produced by this tool. As far as I am concerned, we can just get rid of it all together. In the meantime, reducing the noise it produces is fine by me.

P.

PR #597, opened 2020-01-06
New stellargraph backed by pandas, without networkx

huonw:
Updated version of #520 (separate for now to avoid notifications to people involved in that discussion).


huonw:
See https://github.com/stellargraph/stellargraph/pull/520#issuecomment-579027517 (in particular #668).

PR #598, opened 2020-01-06
Promote various node feature and type related methods to be on StellarGraph

huonw:
These methods were previously only on the NetworkXStellarGraph subclass of `StellarGraph`, but were used in many places on values of type `StellarGraph`.

This also does the FIXME of renaming it to match the naming convention of other methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ In particular, calls like `list(G)` will no longer return a list of nodes; use `
 - Passing a `NodeSequence` or `LinkSequence` object to `GraphSAGE` and `HinSAGE` classes is now deprecated and no longer supported [\#498](https://github.com/stellargraph/stellargraph/pull/498).
 Users might need to update their calls of `GraphSAGE` and `HinSAGE` classes by passing `generator` objects instead of `generator.flow()` objects.
 
+- Various methods on `StellarGraph` have been renamed to be more succinct and uniform:
+   - `get_feature_for_nodes` is now `node_features`
+   - `type_for_node` is now `node_type`
+
 **Fixed bugs:**
 
 
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -242,7 +242,7 @@ class StellarGraph(metaclass=StellarGraphFactory):
         Get the nodes of the graph with the specified node types.
 
         Args:
-            node_type: a type of nodes that exist in the graph
+            node_type (hashable, optional): a type of nodes that exist in the graph
 
         Returns:
             A list of node IDs with type node_type
@@ -276,7 +276,7 @@ class StellarGraph(metaclass=StellarGraphFactory):
         Get the feature sizes for the specified node types.
 
         Args:
-            node_types: (list) A list of node types. If None all current node types
+            node_types (list, optional): A list of node types. If None all current node types
                 will be used.
 
         Returns:
@@ -292,8 +292,8 @@ class StellarGraph(metaclass=StellarGraphFactory):
         for this method to be fast.
 
         Args:
-            n: (list or hashable) Node ID or list of node IDs
-            node_type: (hashable) the type of the nodes.
+            nodes (list or hashable): Node ID or list of node IDs
+            node_type (hashable): the type of the nodes.
 
         Returns:
             Numpy array containing the node features for the requested nodes.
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,10 @@ In particular, calls like `list(G)` will no longer return a list of nodes; use `
 - Passing a `NodeSequence` or `LinkSequence` object to `GraphSAGE` and `HinSAGE` classes is now deprecated and no longer supported [\#498](https://github.com/stellargraph/stellargraph/pull/498).
 Users might need to update their calls of `GraphSAGE` and `HinSAGE` classes by passing `generator` objects instead of `generator.flow()` objects.
 
+- Various methods on `StellarGraph` have been renamed to be more succinct and uniform:
+   - `get_feature_for_nodes` is now `node_features`
+   - `type_for_node` is now `node_type`
```
kjun9: Have you checked if notebooks require updating from this change?

kjun9: potentially in `readthedocs` too?

stellargraph/core/graph.py
```diff
@@ -237,6 +237,69 @@ def out_nodes(self, node: Any) -> Iterable[Any]:
...
+
+    def node_features(self, nodes, node_type=None):
+        """
+        Get the numeric feature vectors for the specified node or nodes.
+        If the node type is not specified the node types will be found
+        for all nodes. It is therefore important to supply the ``node_type``
+        for this method to be fast.
+
+        Args:
+            n: (list or hashable) Node ID or list of node IDs
```
```suggestion
            nodes: (list or hashable) Node ID or list of node IDs
```

kjun9: Hmm actually I think the type is supposed to come before the colon right?
```suggestion
            n (list or hashable): Node ID or list of node IDs
```
and similarly for below

stellargraph/core/graph.py
```diff
@@ -237,6 +237,69 @@ def out_nodes(self, node: Any) -> Iterable[Any]:
...
+            set of types
+        """
+        raise NotImplementedError
+
+    def node_feature_sizes(self, node_types=None):
+        """
+        Get the feature sizes for the specified node types.
+
+        Args:
+            node_types: (list) A list of node types. If None all current node types
```
```suggestion
            node_types (list, optional): A list of node types. If None all current node types
```

stellargraph/core/graph.py
```diff
@@ -237,6 +237,69 @@ def out_nodes(self, node: Any) -> Iterable[Any]:
         """
         raise NotImplementedError
 
+    def nodes_of_type(self, node_type=None):
+        """
+        Get the nodes of the graph with the specified node types.
+
+        Args:
+            node_type: a type of nodes that exist in the graph
```
```suggestion
            node_type (str, optional): a type of nodes that exist in the graph
```
I think?

CHANGELOG.md
huonw: > Have you checked if notebooks require updating from this change?
 
Yep:

```
$ git grep get_feature_for_node 
CHANGELOG.md:   - `get_feature_for_nodes` is now `node_features`
$ git grep type_for_node
CHANGELOG.md:   - `type_for_node` is now `node_type`
stellargraph/core/graph_networkx.py:        type_for_node = {}
stellargraph/core/graph_networkx.py:            type_for_node[n] = self._get_node_type(ndata)
stellargraph/core/graph_networkx.py:            node_types.add(type_for_node[n])
stellargraph/core/graph_networkx.py:                node_features, type_for_node, node_types, dtype
```

The last four are just a local variable.

> potentially in readthedocs too?

I'm not sure I understand? I think the docs are just generated from the code and so there's no separate "read the docs" entity?

stellargraph/core/graph.py
huonw: I don't think it's that strict; at the moment, I think it can be any type (or maybe any hashable type), e.g. some of our tests use integers for the labels:

https://github.com/stellargraph/stellargraph/blob/bdb3a13cb171f1965ad5279b021b3587d0560e93/tests/core/test_stellargraph.py#L451-L452

huonw: I've gone with

```suggestion
            node_type (hashable, optional): a type of nodes that exist in the graph
```

to match some other places that use this.

CHANGELOG.md
kjun9: Thanks

> I'm not sure I understand? I think the docs are just generated from the code and so there's no separate "read the docs" entity?

I was just referring to anything in the `docs` directory that might need to be updated.. but doesn't seem like there's any stellargraph usage code snippets there, so all good

huonw: Ah, makes sense 👍

PR #599, opened 2020-01-06
Add a to_networkx conversion function

huonw:
This ensures that algorithms built into networkx (like connected components) are convenient to perform, even as we move away from the networkx-backed graph type.


stellargraph/core/graph.py
```diff
@@ -24,6 +24,7 @@
 from typing import Iterable, Any, Mapping, Optional
 
 from .schema import GraphSchema
+from ..globalvar import FEATURE_ATTR_NAME
```
kjun9: What's this used for?

stellargraph/core/graph_networkx.py
```diff
@@ -772,6 +772,29 @@ def to_adjacency_matrix(self, nodes: Optional[Iterable] = None):
...
+        types = self.node_types
+
+        for ty in types:
+            node_ids = self.nodes_of_type(ty)
+            features = self.node_features(node_ids, node_type=ty)
+
+            for node_id, node_features in zip(node_ids, features):
+                graph.add_node(
+                    node_id,
+                    **{self._feature_attr: node_features, self._node_type_attr: ty},
```
kjun9: With #598 it seems like if `self._feature_attr` and `self._node_type_attr` are promoted, we can also promote this implementation - thoughts?

stellargraph/core/graph_networkx.py
```diff
@@ -772,6 +772,29 @@ def to_adjacency_matrix(self, nodes: Optional[Iterable] = None):
             self._graph, dtype="float32", weight=self._edge_weight_label, format="coo"
         )
 
+    def to_networkx(self):
+        # This is implemented generically to not be tied to the (outgoing) networkx format.
```
kjun9: What do you mean by this comment? Is this explaining why we don't just return `self._graph`?

stellargraph/core/graph.py
huonw: Ah, left over.

```suggestion
```

stellargraph/core/graph_networkx.py
huonw: Yeah, maybe something like the following would be clearer:

```suggestion
        # This is not tied to the (to-be-removed) networkx-based StellarGraph, so it can easily be reused.
```

stellargraph/core/graph_networkx.py
huonw: Good point, but I'm not sure I want to promote those attributes at the moment, they're a bit peculiar and maybe don't fit into a Pandas/Numpy world very well.

stellargraph/core/graph_networkx.py
huonw: I went with something similar but slightly different.

stellargraph/core/graph_networkx.py
```diff
@@ -772,6 +772,30 @@ def to_adjacency_matrix(self, nodes: Optional[Iterable] = None):
...
+        if self.is_directed():
+            graph = nx.MultiDiGraph()
+        else:
+            graph = nx.MultiGraph()
+
+        types = self.node_types
+
+        for ty in types:
+            node_ids = self.nodes_of_type(ty)
+            features = self.node_features(node_ids, node_type=ty)
```
habiba-h: Results  in error when the nodes have no features.

stellargraph/core/graph_networkx.py
```diff
@@ -772,6 +772,30 @@ def to_adjacency_matrix(self, nodes: Optional[Iterable] = None):
...
         )
 
+    def to_networkx(self):
+        # Despite this class using NetworkX, this implementation does not directly use that
+        # representation, so that it can be reused as we move away from being NetworkX-based.
+        if self.is_directed():
+            graph = nx.MultiDiGraph()
+        else:
+            graph = nx.MultiGraph()
+
```
habiba-h: why is it always converted to multi(di)graph?

huonw: I chose this because `StellarGraph` itself allows for multiple edges (and, the old `StellarGraph` types inherited from `nx.MultiGraph`/`nx.MultiDiGraph`). I don't think we record or check that edges are unique anywhere, so doing the check here might be slightly expensive.

stellargraph/core/graph_networkx.py
huonw: Good catch 👍

huonw:
I've fixed the problem with graphs with no features, and added the `test_to_networkx_no_features` test to validate it. Good find @habiba-h!

stellargraph/core/graph_networkx.py
```diff
@@ -775,6 +775,36 @@ def to_adjacency_matrix(self, nodes: Optional[Iterable] = None):
...
+                for node_id, node_features in zip(node_ids, features):
+                    graph.add_node(
+                        node_id, **ty_dict, **{self._feature_attr: node_features},
+                    )
+            else:
+                # no features, so just add the type
+                graph.add_nodes_from(node_ids, **ty_dict)
+
+        graph.add_edges_from(self.edges(triple=True))
```
habiba-h: Since `self.edges(triple=True)` returns node pairs with `edge_type` and not the actual weight/value over the edge, `graph.add_edges_from(self.edges(triple=True))` results in a networkx multigraph with values that are `edge_type` and not the actual values over the edges. I have tested it only for one `edge_type`, not sure what happens in case of multi-edge type graph. 
Moreover, as @huonw you mentioned in #642 the `edge_type` is not the actual label of the type of edge.

huonw: Ah, yeah; this is broken for two reasons:

1. `.edges(triple=True)` is wrong, but that's being fixed separately in #642 #643 
2. even when it returns the right tuples, the edge type will be passed in as a key, not a data dictionary

PR #600, opened 2020-01-06
Promote info method to StellarGraph class

huonw:
This method is used in various places that manipulate a `StellarGraph` instance, and so should be part of the overall interface, not just the `NetworkXStellarGraph` subclass.

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -249,6 +249,7 @@ class StellarGraph(metaclass=StellarGraphFactory):
         time for a large graph.
 
         Args:
+            show_attributes (bool, default True): If True, include attributes information
             sample (int): To speed up the graph analysis, use only a random sample of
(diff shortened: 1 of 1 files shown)


stellargraph/core/graph.py
```diff
@@ -240,6 +240,23 @@ def out_nodes(self, node: Any) -> Iterable[Any]:
...
+    def info(self, show_attributes=True, sample=None):
+        """
+        Return an information string summarizing information on the current graph.
+        This includes node and edge type information and their attributes.
+
+        Note: This requires processing all nodes and edges and could take a long
+        time for a large graph.
+
+        Args:
+            sample (int): To speed up the graph analysis, use only a random sample of
```
```suggestion
            show_attributes (bool, default True): If True, include attributes information
            sample (int): To speed up the graph analysis, use only a random sample of
```

Issue #604, opened 2020-01-07
Support sample weights when using `fit_generator`


adocherty:
### Description

When fitting data with different weights for each data point the `sample_weight=` argument can be supplied to the Keras `Model.fit` method to provide weights to the samples. This can be used, for example, with the `sample_weight_mode='temporal'` argument to `Model.compile` to specify different weights for different classes for 3-D outputs (see #603 ).

However, all demos in the library are based around the use of `Model.fit_generator` which does not support this argument.  The generators are required to supply the appropriate sample weights but this functionality is not currently implemented.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
@adocherty had done some work along these lines, which is displayed in https://github.com/stellargraph/stellargraph/pull/605.

huonw:
This may be related to/obsoleted by #803.

PR #605, opened 2020-01-07
Bugfix/sample weights for fullbatch

huonw:
PR noting work that @adocherty had done for #604, but couldn't be completed before moving teams; for ease of reference.


huonw:
This was only to make it possible to delete the branch without permanently losing the changes for future reference, so closing (and deleting the branch).

PR #608, opened 2020-01-07
Feature/190 homogenous stellargraph

huonw:
This is investigatory code for the new StellarGraph class.


huonw:
This PR only exists to preserve the code even after the branch is deleted, so closing.

Issue #611, opened 2020-01-07
Flag incompatible dependency changes via automated CI runs


huonw:
### Description

There's been a few examples of dependencies doing a release that is incompatible with stellargraph and requiring a hot fix to restrict the dependencies, e.g. #505 fixed a problem with both networkx and tensorflow.

These were only noticed incidentally, when a downstream user had their code break. It would be good to flag these early, e.g. a daily scheduled build on CI that uses the latest dependencies, which would mean we notice problems soon because our own builds would fail.

### Done Checklist

- [x] Scheduled build created
- [ ] Version numbers printed on CI

huonw:
I've added a scheduled build: https://buildkite.com/stellar/stellargraph-public/settings/schedules/ee7d883a-e9aa-4ed2-8f93-e1e8f5a5ecab

I've also opened https://github.com/stellargraph/stellargraph/pull/613 to have CI print the versions it uses, to make it easy to identify problems.

PR #613, opened 2020-01-07
Add pip freeze on CI to ease debugging problems from version changes

huonw:
This invocation shows the versions of each library installed with `pip`.

Doing this on CI makes it easy to find which version(s) of dependencies changed between passing and failing builds, helping to narrow down the possible cause of a failure. This thus ties into #611 and the daily scheduled build.

### DIFF
--- a/.buildkite/steps/script.sh
+++ b/.buildkite/steps/script.sh
@@ -8,6 +8,9 @@ upload_tests() {
   buildkite-agent artifact upload "$junit_file" "s3://${AWS_LOGS_BUCKET}/pytest/${BUILDKITE_BRANCH}/${BUILDKITE_BUILD_NUMBER}"
 }
 
+echo "--- installing required packages"
+apt install -y git
+
 echo "--- installing dependencies"
 pip install -q --no-cache-dir -r requirements.txt -e .
 
--- a/.buildkite/steps/script.sh
+++ b/.buildkite/steps/script.sh
@@ -11,6 +11,9 @@ upload_tests() {
 echo "--- installing dependencies"
 pip install -q --no-cache-dir -r requirements.txt -e .
 
+echo "--- listing dependency versions"
+pip freeze
+
 echo "+++ running tests"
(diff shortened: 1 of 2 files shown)


.buildkite/steps/script.sh
```diff
@@ -9,7 +9,13 @@ upload_tests() {
...
 
 echo "--- installing dependencies"
-pip install -q --no-cache-dir -r requirements.txt -e .
+pip install -q --no-cache-dir -r requirements.txt
+
+echo "--- listing dependency versions"
+pip freeze
+
+echo "--- installing stellargraph"
+pip install -q --no-cache-dir -e .
```
huonw: I split this part of the `pip install` to run after the `freeze`, because the `freeze` tries to use `git` to get info about the editable `.` install, and the `python:3.x-slim` images do not contain git. Not having `git` when doing `freeze` on an `.` installation results in a non-problematic (but disconcerting) pair of ERROR/WARNING lines: https://buildkite.com/stellar/stellargraph-public/builds/314#a98235b9-2092-4852-98f4-19baf1a0c452/108-110

By running the `freeze` first, that command only lists the actual dependencies, which is fine.

kjun9: Does it need to be an editable install in the first place? We also have dependencies listed in `setup.py` so I think technically the `.` installation could install additional dependencies

huonw: > Does it need to be an editable install in the first place? 

Ah, no it does not need to be, and `pip freeze` doesn't use `git` for a non-editable install. 👍 

> We also have dependencies listed in setup.py so I think technically the . installation could install additional dependencies

That's true, although I think it's theoretically a mistake in `requirements.txt` to have a difference here.

timpitman: Since we've got "REQUIRES = [..." in setup.py, I'm not sure this does what's intended?  Would we be better off with an image including git?

PR #618, opened 2020-01-07
Experiment: use dask for parallelism

huonw:
This is experimental work (from a while ago) that uses dask to do graph walks in parallel.


huonw:
@PantelisElinas said (summarised, any errors are mine):

- random walk sampling worked well on a multicore machine even with a non-configuration dask client, and gave a reasonable speed-up
- subgraph sampling with graphsage worked, but resulted in a slow down (due to overhead of coordinating the parallelism), because the bottleneck is the data transfer of the sampled subgraphs, not the sampling

Closing this PR, because we're unlikely to revisit/rebase this for a while.

Issue #603, opened 2020-01-07
Class weights not supported for full-batch methods (GCN/GAT/?PPNP)


adocherty:
Class weights are not supported for full-batch models as they are for GraphSAGE/HinSAGE.  This is due to the shape of the outputs being 3 dimensional `(1 ⨉ num_output ⨉ num_classes)` which is not supported by Keras.

There is a workaround for binary classification where the outputs can be reshaped to `(1 ⨉ num_output)`.

Example:

In the [hateful-twitters.ipynb](https://github.com/stellargraph/stellargraph/blob/develop/demos/use-cases/hateful-twitters.ipynb) demo notebook the class_weights option is switched off when `model ≠ 'graphsage'` (execution cell 40).

Setting `model = 'gcn` and `class_weight=train_class_weights` in execution cell 40 we get the following error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-48-cb658319ad28> in <module>
      7     verbose=0,
      8     shuffle=False,
----> 9     class_weight=class_weight,
     10 )
...
site-packages/tensorflow_core/python/keras/engine/training_utils.py in standardize_weights(y, sample_weight, class_weight, sample_weight_mode)
    923   if isinstance(class_weight, dict):
    924     if len(y.shape) > 2:
--> 925       raise ValueError('`class_weight` not supported for '
    926                        '3+ dimensional targets.')
    927 

ValueError: `class_weight` not supported for 3+ dimensional targets.
```

huonw:
https://github.com/stellargraph/stellargraph/pull/619 contains an old version of one approach to fixing this.

PR #619, opened 2020-01-07
Class weights flow

huonw:
This is one way to handle unbalanced data.

See: #603


huonw:
@PantelisElinas said (summarised, any errors are mine):

- this is outdated due to the change to have batch dimension in the full batch methods
- unbalanced can currently be handled by oversampling

Since it seems like we may need to essentially start from scratch, and this isn't currently prioritised, I'm going to close.

Issue #508, opened 2019-10-24
Creating embedding is not reproducible


ericbroda:
I am using StellarGraph to create embeddings for a particular graph/feature set.  Unfortunately, the embeddings are different each time I create/train the graph despite providing identical information each time.

Is this bug, or am I using StellarGraph incorrectly?

Below is the code that demonstrates the issue:

    import networkx as nx
    import random
    import numpy as np
    import pandas as pd
    import keras
    import stellargraph as sg
    from stellargraph.mapper import GraphSAGELinkGenerator, GraphSAGENodeGenerator
    from stellargraph.layer import GraphSAGE, link_classification
    from stellargraph.data import UnsupervisedSampler
    
    # Establish random seed
    RANDOM_SEED = 42
    random.seed(RANDOM_SEED)
    
    # Create a graph from well-known karate club data
    print(f"Creating graph")
    graph = nx.karate_club_graph()
    
    # Create features for each node
    print(f"Creating features")
    features = []
    nodes = list(graph.nodes)
    columns = ["c-" + str(x) for x in range(10)]
    nodes.sort()
    for node in nodes:
        f = {c: random.random() for c in columns}
        features.append(f)

    features_df = pd.DataFrame(features)
    print(f"features_df: \n{features_df}")
    
    for i in range(2):
        print(f"----- Iteration: {i} -----")
    
        # Create the model and generators
        print(f"Creating the model and generators")
        Gs = sg.StellarGraph(graph, node_features=features_df)
        unsupervisedSamples = UnsupervisedSampler(Gs, nodes=graph.nodes(), length=5, number_of_walks=3, seed=RANDOM_SEED)
        train_gen = GraphSAGELinkGenerator(Gs, 50, [5, 5], seed=RANDOM_SEED).flow(unsupervisedSamples)
        graphsage = GraphSAGE(layer_sizes=[100, 100], generator=train_gen, bias=True, dropout=0.0, normalize="l2")
        x_inp_src, x_out_src = graphsage.node_model(flatten_output=False)
        x_inp_dst, x_out_dst = graphsage.node_model(flatten_output=False)
    
        x_inp = [x for ab in zip(x_inp_src, x_inp_dst) for x in ab]
        x_out = [x_out_src, x_out_dst]
        edge_embedding_method = "l2"
        prediction = link_classification(output_dim=1, output_act="sigmoid", edge_embedding_method=edge_embedding_method)(x_out)
    
        # Create and train the Keras model
[...]

PR #620, opened 2020-01-07
Make unsupervised graphsage reproducible

kjun9:
This may be one of a number of PRs to come around reproducibility - this adds a bit of setup for testing for reproducibility, and updates various classes involved in unsupervised GraphSAGE.

The interaction between `UnsupervisedSampler` and `OnDemandLinkSequence` has changed a fair bit in order to make sure batches can be obtained in parallel without any guaranteed order, so I'm interested in feedback on the approach I've taken.

Related to #508 

I'm not entirely sure I haven't completely broken unsupervised learning, so I'll go through the demos and maybe think about adding more tests, but I think this can still be reviewed in the mean time.


stellargraph/mapper/sampled_link_generators.py
```diff
@@ -254,7 +278,9 @@ def get_levels(loc, lsize, samples_per_hop, walks):
         # of 2 nodes, so we are extracting 2 head nodes per edge
         batch_feats = []
         for hns in zip(*head_links):
-            node_samples = self.sampler.run(nodes=hns, n=1, n_size=self.num_samples)
+            node_samples = self._sampler(batch_num).run(
```
huonw: Are batch numbers ever passed multiple times to `sample_features`? If not, it seems like this might be a memory leak: we end up caching `SampledBreadthFirstWalk` instances despite them not being used again.

As an alternative to having to have multiple `SampledBreadthFirstWalk` objects, what do you think about moving the `SampledBreadthFirstWalk.__init__` `seed` argument to `run`, or, adding an optional `seed` argument to `run` too (that gets mixed into the `__init__` seed)?

stellargraph/mapper/sequences.py
```diff
@@ -273,20 +272,12 @@ def __init__(self, sample_function, batch_size, walker, shuffle=True):
...
-            * self.walker.length
-            * self.walker.number_of_walks
-        )
-
-        # the generator method from the sampler with the batch-size from the link generator method
-        self._gen = self.walker.generator(self.batch_size)
+        self.shuffle = shuffle
+        self._batches = self._create_batches()
+        self.length = len(self._batches)
+        self.data_size = sum([len(batch[0]) for batch in self._batches])
```
huonw:
```suggestion
        self.data_size = sum(len(batch[0]) for batch in self._batches)
```

stellargraph/mapper/sequences.py
```diff
@@ -273,20 +272,12 @@ def __init__(self, sample_function, batch_size, walker, shuffle=True):
...
-            2
-            * len(self.walker.nodes)
-            * self.walker.length
-            * self.walker.number_of_walks
-        )
-
-        # the generator method from the sampler with the batch-size from the link generator method
-        self._gen = self.walker.generator(self.batch_size)
+        self.shuffle = shuffle
+        self._batches = self._create_batches()
```
huonw: It seems like putting this call here is converting this `__init__` constructor from being efficient and on-demand per the class's name (O(1) time and memory?) to doing all of the sampling and storing it in memory (i.e. O(size of sample)). This seems suboptimal. Can we avoid it?

stellargraph/mapper/sequences.py
```diff
@@ -273,20 +272,12 @@ def __init__(self, sample_function, batch_size, walker, shuffle=True):
...
-        self.data_size = (
-            2
-            * len(self.walker.nodes)
-            * self.walker.length
-            * self.walker.number_of_walks
-        )
-
-        # the generator method from the sampler with the batch-size from the link generator method
-        self._gen = self.walker.generator(self.batch_size)
+        self.shuffle = shuffle
```
huonw: Was the shuffle argument previously being ignored?

stellargraph/data/unsupervised_sampler.py
```diff
@@ -121,69 +121,45 @@ def generator(self, batch_size):
...
-        while not done:
-            self.random.shuffle(self.nodes)
-            for node in self.nodes:  # iterate over root nodes
-                # Get 1 walk at a time. For now its assumed that its a uniform random walker
-                walk = self.walker.run(
-                    nodes=[node],  # root nodes
-                    length=self.length,  # maximum length of a random walk
-                    n=1,  # number of random walks per root node
-                )
+        walks = self.walker.run(nodes=self.nodes, length=self.length, n=1, seed=0)
```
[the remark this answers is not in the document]
kjun9: Ah oops nice catch I think this is a left over, seed is already set to `seed` when `self.walker` is constructed, I'll remove it 👍

stellargraph/mapper/sequences.py
kjun9: yes, because there was nothing to shuffle, since each batch is computed on the fly

stellargraph/mapper/sampled_link_generators.py
kjun9: Every epoch results in each batch number being reused, so if there are 100 batches per epoch, `self._samplers` will have exactly 100 samplers. 

I did consider something like what you're proposing, but in order to "mix" an optional `seed` with the init `seed` it seemed to me like we'd have to create a new `random` state purely off that mixed seed, which means every sample obtained from the same `seed` + optional `seed` uses an identical random state which felt like we're limiting the "random space" we use for sampling. Instead I wanted each `batch_num` to reuse a random state and mutate it each time. 

Hopefully I'm making sense!

stellargraph/mapper/sequences.py
kjun9: I guess one way would be putting in a check in `__getitem__` to see if batches exist, otherwise create them.

I'll double check if `fit_generator` is supposed to call any sort of pre-train method on the sequence

stellargraph/data/unsupervised_sampler.py
```diff
@@ -121,69 +121,45 @@ def generator(self, batch_size):
...
-                        edge_ids_labels = list(zip(all_pairs, all_targets))
-                        self.random.shuffle(edge_ids_labels)
-                        edge_ids, edge_labels = [
-                            [z[i] for z in edge_ids_labels] for i in (0, 1)
-                        ]
-
-                        yield edge_ids, edge_labels
+        # convert to batches
+        return [
+            tuple(zip(*edge_ids_labels[i : i + batch_size]))
```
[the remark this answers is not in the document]
kjun9: Hmm yes I think that could be possible 👍I'll give it a go

stellargraph/mapper/sampled_link_generators.py
huonw: > Every epoch results in each batch number being reused, so if there are 100 batches per epoch, self._samplers will have exactly 100 samplers.

Ah; definitely worth a comment, if we go with this approach 😄 

> I did consider something like what you're proposing, but in order to "mix" an optional seed with the init seed it seemed to me like we'd have to create a new random state purely off that mixed seed, which means every sample obtained from the same seed + optional seed uses an identical random state which felt like we're limiting the "random space" we use for sampling. Instead I wanted each batch_num to reuse a random state and mutate it each time.

Ah, yeah, that makes sense. Maybe it could be mixed in to a sample from the random state (e.g. `random_object.randrange(2**64) + seed`), or even just overwrite it, and it becomes the caller's (i.e. here) responsibility to mix in and have global state.

Or maybe that hits all the same ordering/reproducibility problems...

stellargraph/mapper/sequences.py
huonw: Couldn't (shouldn't?) it still shuffle the order of the head IDs? (This is probably not a question that's relevant to this PR.)

stellargraph/mapper/sequences.py
huonw: Why not continue to create them on demand, doing the sampling for each batch in `__getitem__`?

stellargraph/mapper/sampled_link_generators.py
kjun9: > Ah; definitely worth a comment, if we go with this approach 😄

I can add to the docstring for `def _sampler`?
```
For its intended use in a Keras/TF workflow, if there are N batches in an epoch, there will be N samplers created, each corresponding to a particular ``batch_num``.
```

> Ah, yeah, that makes sense. Maybe it could be mixed in to a sample from the random state (e.g. random_object.randrange(2**64) + seed), or even just overwrite it, and it becomes the caller's (i.e. here) responsibility to mix in and have global state. Or maybe that hits all the same ordering/reproducibility problems...

Yeah I think there will be multiple threads on the "caller" side accessing its random object with no guaranteed order. I guess one alternative might be to only keep a dictionary of random objects instead of sampler objects and generate new seeds to call `sampler.run` with on a single `sampler` object.

stellargraph/mapper/sequences.py
kjun9: Hmmm.. I guess we could but then it would require similar sort of dictionary of samplers logic in `unsupervised_sampler` as well, since the order of `batch_num` is not guaranteed. Since this is just creating the IDs associated with each training example without features, I didn't think it was particularly worth doing that.

stellargraph/mapper/sequences.py
kjun9: I think it was always just shuffling without end, there wasn't really a concrete idea of an "epoch" the way `UnsupervisedSampler.generator` was implemented - just an infinite loop that shuffles things when it runs out of nodes to go through and it wasn't guaranteed that an actual training epoch went through an "epoch" in the sampler.

stellargraph/data/unsupervised_sampler.py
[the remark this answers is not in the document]
kjun9: Just as further explanation - if seed is None for `self.walker.run` it just uses the random state created with the seed passed to the constructor so it's fine to remove it here

stellargraph/data/unsupervised_sampler.py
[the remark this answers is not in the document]
kjun9: Now it returns a tuple of numpy arrays, it seems like all the downstream usage still works

kjun9:
# Cora unsupervised GraphSAGE demo
Things seem roughly correct - I guess I should also commit the new notebook outputs?
## TSNE visualisation
### develop
image

### this branch
image

## Accuracy of logistic regression classifier on top of embeddings

develop: 0.7555382821609017
this branch: 0.7652545666537116

kjun9:
Seems like the attri2vec embeddings notebook is broken due to this - working on a fix

stellargraph/data/unsupervised_sampler.py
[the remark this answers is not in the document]
kjun9: Actually I might just revert this, since I realised that IDs are not necessarily integers, so then we'd need to have a numpy array of arbitrary python objects?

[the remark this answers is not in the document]
huonw: I think that should work ~automatically after removing `dtype`s? It may make sense to manipulate an array of IDs an array of labels separately (I think that's why the final return has `[..., :2]`, `[..., 2]`, right?).

```python
>>> import numpy as np
>>> np.array(["foo", 1, [], {}])
array(['foo', 1, list([]), {}], dtype=object)
```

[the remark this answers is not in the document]
kjun9: Do you still suspect the "manipulations" will be faster though? 

The array of IDs and array of labels are zipped together for shuffling purposes - I guess we could just create another array of indices to shuffle instead, but it didn't occur to me that would necessarily be better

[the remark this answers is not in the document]
huonw: Ah, hm... I don't know. It could be done with pandas or even a numpy `recarray` https://docs.scipy.org/doc/numpy/user/basics.rec.html 

```
>>> np.rec.array([("foo", "barbaz", 1), ("qux", "1234567890", 0)])
rec.array([('foo', 'barbaz', 1), ('qux', '1234567890', 0)],
          dtype=[('f0', '<U3'), ('f1', '<U10'), ('f2', '<i8')])
```

However, I suspect just shuffling the arrays "together" would be fine too, e.g. https://stackoverflow.com/q/4601373/1256624 (not the first answer, though, since that seems to rely on dtypes matching).

stellargraph/mapper/sequences.py
huonw: > I guess we could but then it would require similar sort of dictionary of samplers logic in unsupervised_sampler as well

Right.

> Since this is just creating the IDs associated with each training example without features, I didn't think it was particularly worth doing that.

Good point, but I think the explosion in memory use is still potentially concerning: going from a number of pairs of nodes of size `O(batch_size)` (since only a single batch is buffered at a time) to `O(number_of_walks * nodes * length)`; this seems like it *could* approach or even exceed the size of the features if `number_of_walks * length` is a significant fraction of the feature dimension, although in practice, in our demos, it seems like that's at most `5 * 4`.

If we are going to do this, I think we should potentially consider renaming the class, since it's no longer on demand? (Possibly with some sort of nice deprecation warnings.) Either that, or land this as is but have future work to refactor/improve it.

What do you think?

tests/reprod/test_unsupervised_graphsage.py
```diff
@@ -0,0 +1,103 @@
...
+        use_multiprocessing=False,
+        workers=4,
+        shuffle=True,
+    )
+    weights = model.get_weights()  # list of np matrices
+    return weights
+
+
+def embedding_equals(emb1, emb2):
+    return all([np.array_equal(w, w_new) for w, w_new in zip(emb1, emb2)])
```
huonw:
```suggestion
    return all(np.array_equal(w, w_new) for w, w_new in zip(emb1, emb2))
```

tests/reprod/fixtures.py
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+
+
+def assert_reproducible(func, equals, num_iter=20):
+    """
+    Assert results from calling ``func`` are reproducible. The ``equals`` function is used to test equality of results.
+
+    Args:
+        func (callable): Function to check for reproducible "result"
+        equals (callable): Function to take two arguments (of "result" type) and check that they are equal
```
huonw: Is it worth making this a bit more structured/restricted? E.g. require that `func` returns a trained keras model, and then do the `get_weights`/`all`-equal thing in here, and not need `equals` at all?

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -214,18 +214,44 @@ def __init__(self, G, batch_size, num_samples, seed=None, name=None):
...
+        particular ``batch_num``.
 
-    def sample_features(self, head_links):
+        Args:
+            batch_num (int): Batch number
+
+        Returns:
+            SampledBreadthFirstWalk object
+        """
+        if batch_num in self._samplers:
```
huonw: It would be very slightly faster (and more pythonic) to EAFP rather than LYBL here (https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/):

```python
try:
    return self._samplers[batch_num]
except KeyError:
    seed = ...
    self._samplers[batch_num] = ...
    return self._samplers[batch_num]

File: stellargraph/data/unsupervised_sampler.py [hunk 10]
```diff
@@ -121,69 +123,52 @@ def generator(self, batch_size):
...
+                    positive_pairs, negative_samples
                 )
+            ]
+        )
+
+        # shuffle
+        pairs = np.concatenate((positive_pairs, negative_pairs), axis=0)
+        labels = np.repeat([1, 0], len(positive_pairs))
+
+        indices = self.np_random.permutation(len(pairs))
```
huonw: I think this very slightly changes behaviour: previously, I think there was always an equal number of positive and negative examples in each batch, but now they're shuffled together, arbitrarily, and they may be uneven (although, equal on average).

Is that a problem?

tests/reprod/test_unsupervised_graphsage.py
```diff
@@ -0,0 +1,103 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
kjun9: Hmm I'm not sure where else it would fit in, since I think there's currently no "integration"-like tests for an algorithm (i.e. a combination of a pipeline made from components of `layer`, `mapper`, and `data`)?

I did also consider calling this module `eval` instead of `reprod` to keep it more generic

tests/reprod/fixtures.py
kjun9: Hmm yeah that's probably better, will update

stellargraph/mapper/sequences.py
kjun9: Hmm.. I'm not sure 🤷‍♂ I'm kind of leaning towards just keeping this "on-demand" by adding the batch num logic here too.

tests/reprod/__init__.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
kjun9: Sure I don't mind the longer name 👍

stellargraph/data/unsupervised_sampler.py
```diff
@@ -121,69 +123,52 @@ def generator(self, batch_size):
...
+
+        positive_pairs = np.array(
+            [
+                (target, positive_context)
+                for target, walk in zip(targets, walks)
+                for positive_context in walk[1:]
+            ]
+        )
+
+        negative_samples = self.random.choices(
```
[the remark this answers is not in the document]
kjun9: Can I just update this here now? I think it's clear enough that `column_stack` will *at least* be as fast as a manual list comprehension?

stellargraph/mapper/sequences.py
kjun9: Hmm actually I think doing each batch on demand is more difficult here than in the link generator... basically if `batch_size % (number_of_walks * (length - 1) * 2) != 0` then it gets kind of awkward (the `* 2` is for each positive context having a negative context)  For instance:

With 1 walk of length 3, i.e. 2 positive context pairs per node + 2 negative, we'd be generating 4 examples per node. If `batch_size` doesn't divide evenly into 4's, then we can't just assign nodes to batches, we actually have to assign individual context pairs to batches, in which case we'd basically end up with the same sort of memory footprint? 

To avoid the memory footprint I think we'd need all threads to share one "state" of context pairs being generated on the fly (similarly to how it was implemented previously where `next(generator)` was wrapped around a lock)... but I'm not sure that can be made reproducible unless we just force batches to be processed in order (i.e. `UnsupervisedSampler` keeps track of which `batch_num` to process next, and has to block other threads)

stellargraph/data/unsupervised_sampler.py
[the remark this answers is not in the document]
huonw: Fine by me; just make sure you didn't feel like I was expecting you to do it, if you wanted to stop this from having an endless series of one-more-thing adjustments.

stellargraph/data/unsupervised_sampler.py
huonw: That sounds reasonable to me, I wonder if it's worth having a brief comment?

```suggestion
        # don't ensure an equal number of positive/negative examples in each batch,
        # just an equal number overall
        indices = self.np_random.permutation(len(pairs))
```

or if that's pointless?

stellargraph/mapper/sampled_link_generators.py
huonw: Yes, but the granularity of interleaving can mean things can be invalidated/confused if things are happening in parallel, e.g. it might be that an operation like `D[x] = 1` could be interrupted half way through by another thread (running `D[y] = 2`) which might cause the write to be dropped or otherwise misbehave.

However, it looks like that shouldn't be the case, due to http://effbot.org/pyfaq/what-kinds-of-global-value-mutation-are-thread-safe.htm . In particular:

1. only one worker for each `batch_num` operates concurrently
2 mutations of global values are restricted to operations on the `self._samplers` (not individual samplers) because of 1
3. those mutations are writes like `self._samplers[batch_num] = ...`, which are atomic
4. that mutation is guaranteed to never be an overwrite: if the existence check in the `try` fails, it will never "stop failing" before the write in 3 (that is, no other thread will fill in that element concurrently), because of 1

Thus, assuming 1, I think this is fine.

An easier way to be sure of this would be to have a lock around the body of `_sampler`, which is probably fine, since that happens once at the start of the expensive sampling operations and typically runs fast.

stellargraph/mapper/sequences.py
huonw: Ah, great point.

Do the batches all have to be exactly the same size? If not, I think we can get down to `O(number_of_walks * length)` (or something) by having a dummy batch with any left overs every time we overfill a batch. That is, if it doesn't divide evenly, we end up generating something like 2&times; more batches: one full data and one with leftovers, and they have to be generated at the same time and cached until the other one is requested, e.g. if `batch_size = 3` in your example:

| batch | size |
|---|---:|
|0|3|
|1|1|
|2|3|
|3|1|
|...|...|

I guess in the worst case, this can result in caching half of the batches, if it requests 0, 2, 4, ... without requesting the others, but usually won't?

I'd be happy to leave this work for future, given it seems moderately complicated.

> UnsupervisedSampler keeps track of which batch_num to process next, and has to block other threads

I'd be concerned that this might deadlock, if for whatever reason, there was no thread trying to process the next batch in order (e.g. if it starts with `batch_num` 10, 20, 30, they'll all be waiting forever for 0, 1, ..., 9 to be finished).

kjun9: I've created #681 for now

Things *might* still work without ensuring all batches have the same size... but even if it does work, it feels like we'd breaching the contract of using Keras's API and I'd be concerned about running into more subtle problems, unless Keras explicitly says this is ok.

stellargraph/mapper/sampled_link_generators.py
kjun9: I've added the lock

stellargraph/mapper/sequences.py
huonw:
```suggestion
        # FIXME(#681): all batches are created at once, so this is no longer "on demand"
        self._batches = self._create_batches()
```

tests/reproducibility/fixtures.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Args:
+        func (callable): Function to check for reproducible model
+        num_iter (int, default 20): Number of iterations to run through to validate reproducibility.
+
+    """
+    model = func()
+    for i in range(num_iter):
+        model_new = func()
+        if not models_equals(model, model_new):
```
huonw: Did you consider:

```suggestion
        assert models_equals(model, model_new)
```

I think pytest's introspect of assert statements should do a reasonable job of showing the two models: https://pytest.readthedocs.io/en/latest/assert.html#assert

kjun9: I'm guessing that only works when the assert is in the form `assert value1 == value2`? I tried a few alternatives but couldn't get it to print out anything about the model... which also led me to find the current form also just looks like:
```
>               assert False, (model, model_new)
E               AssertionError: (<tensorflow.python.keras.engine.training.Model object at 0x1432c6210>, <tensorflow.python.keras.engine.training.Model object at 0x13fb97990>)
```
so I might change it to be `assert False, (model.get_weights(), model_new.get_weights())`

huonw: https://pytest.readthedocs.io/en/latest/example/reportingdemo.html#tbreportdemo shows various things that aren't just `assert a == b`, e.g. 

```
______________________ TestMoreErrors.test_startswith ______________________

self = <failure_demo.TestMoreErrors object at 0xdeadbeef>

    def test_startswith(self):
        s = "123"
        g = "456"
>       assert s.startswith(g)
E       AssertionError: assert False
E        +  where False = <built-in method startswith of str object at 0xdeadbeef>('456')
E        +    where <built-in method startswith of str object at 0xdeadbeef> = '123'.startswith

failure_demo.py:227: AssertionError
```

but maybe methods are special and functions aren't.

Instead of `assert False` (and the `assert True` later), what about:

```python
assert models_equals(model, model_new), (model.get_weights(), model_new.get_weights())
```

Doing `if f(): assert False` instead of `assert f()` seems, to me, similar to doing `True if f() else False` instead of just `f()`. 😄

kjun9:
@PantelisElinas I've added you instead of Habiba for reviewing now that she's away for a while

Issue #626, opened 2020-01-08
'test_hinsage_from_generator' test fails


huonw:
**Describe the bug**

The `test_hinsage_from_generator` test in tests/layer/test_hinsage.py is failing due to

```
>       assert actual == pytest.approx(expected)
E       assert array([[24., 31.],\n       [32., 32.]], dtype=float32) == approx([[26.0 ± 2.6e-05, 29.0 ± 2.9e-05], [32.0 ± 3.2e-05, 31.0 ± 3.1e-05]])
```

In particular, 24 != 26, 31 != 29, and 32 != 31.

**To Reproduce**
1. Go to CI https://buildkite.com/stellar/stellargraph-public/builds/339

```
============================= test session starts ==============================
platform linux -- Python 3.6.10, pytest-5.3.1, py-1.8.1, pluggy-0.13.1
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /workdir, inifile: pytest.ini
plugins: benchmark-3.2.2, cov-2.8.1
collected 334 items                                                            
 
tests/core/test_schema.py .....                                          [  1%]
tests/core/test_stellargraph.py .........................                [  8%]
tests/core/test_types.py ..                                              [  9%]
tests/core/test_utils.py .....                                           [ 11%]
tests/data/test_biased_random_walker.py ............                     [ 14%]
tests/data/test_breadth_first_walker.py .........                        [ 17%]
tests/data/test_converters.py ............                               [ 20%]
tests/data/test_directed_breadth_first_sampler.py ......                 [ 22%]
tests/data/test_edge_splitter.py .........                               [ 25%]
tests/data/test_epgm.py .........                                        [ 28%]
tests/data/test_heterogeneous_breadth_first_walker.py ......             [ 29%]
tests/data/test_metapath_walker.py ......                                [ 31%]
tests/data/test_node_splitter.py ..........                              [ 34%]
tests/data/test_uniform_random_walker.py ......                          [ 36%]
tests/data/test_unsupervised_sampler.py ....                             [ 37%]
tests/layer/test_appnp.py .....                                          [ 39%]
tests/layer/test_attri2vec.py ...                                        [ 40%]
tests/layer/test_cluster_gcn.py .......                                  [ 42%]
tests/layer/test_gcn.py .........                                        [ 44%]
[...]

PR #629, opened 2020-01-09
Xfail (temporarily) hinsage test that fails with tensorflow 2.1

huonw:
This is a temporary work-around to unbreak CI until we can properly determine the cause and thus the fix.

It's marked as a `strict` xfail (that is, causes tests to fail, if it unexpectedly passes), so that we don't accidentally fix it without noticing, and are thus reminded to remove the `xfail` marker.

See: #626

### DIFF
--- a/tests/layer/test_hinsage.py
+++ b/tests/layer/test_hinsage.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2018-2019 Data61, CSIRO
+# Copyright 2018-2020 Data61, CSIRO
 #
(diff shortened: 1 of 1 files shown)


kjun9:
Would it be better to just pin tensorflow to 2.0 for now? How do we know it's not "actually" broken?

huonw:
I've got that as an option in https://github.com/stellargraph/stellargraph/pull/630, but I think this is better:

- only one test is failing, which hints (to me) that it's likely not major breakage
- restricting the version makes it (very slightly) harder to test any fixes
- sticking to an older version of dependencies is unfortunate and may restrict progress along unrelated dimensions

PR #630, opened 2020-01-09
Restrict tensorflow < 2.1 to unfail hinsage test

huonw:
See: #626


huonw:
We went with #629 instead.

Issue #602, opened 2020-01-06
Build documentation for each pull request


huonw:
### Description

When making changes that involve documentation (e.g. new features), it's helpful to be able to accurately preview that documentation as it will be shown on read the docs. This is theoretically possible by talking to the RTD people, as described in https://blog.readthedocs.com/building-docs-for-pull-requests/

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
I've emailed them to ask for access to the beta feature.

huonw:
They've enabled it for us. I've switched our readthedocs web hook (https://github.com/stellargraph/stellargraph/settings/hooks/50326049) to respond to PR pushes too:

![image](https://user-images.githubusercontent.com/1203825/72027586-bc7cb880-32d3-11ea-8350-0c8ba6652c35.png)

It's correctly picking them up and starting builds, but those builds are failing when installing dependencies, e.g. https://readthedocs.org/projects/stellargraph/builds/10225724/ is a build for https://github.com/stellargraph/stellargraph/pull/628:

```
/home/docs/checkouts/readthedocs.org/user_builds/stellargraph/envs/628/bin/python -m pip install --exists-action=w --cache-dir /home/docs/checkouts/readthedocs.org/user_builds/stellargraph/.cache/pip -r requirements.txt
Collecting networkx>=2.2
  Using cached https://files.pythonhosted.org/packages/41/8f/dd6a8e85946def36e4f2c69c84219af0fa5e832b018c970e92f2ad337e45/networkx-2.4-py3-none-any.whl
Requirement already satisfied: setuptools>=39.1.0 in /home/docs/checkouts/readthedocs.org/user_builds/stellargraph/envs/628/lib/python3.6/site-packages (from -r requirements.txt (line 2)) (41.0.1)
Collecting tensorflow>=2.0.0
  Downloading https://files.pythonhosted.org/packages/85/d4/c0cd1057b331bc38b65478302114194bd8e1b9c2bbc06e300935c0e93d90/tensorflow-2.1.0-cp36-cp36m-manylinux2010_x86_64.whl (421.8MB)
Killed


Command killed due to excessive memory consumption
```

huonw:
[...]

PR #631, opened 2020-01-09
Install tensorflow-cpu package when building docs on Read the Docs

huonw:
With tensorflow 2.1, the `tensorflow` package includes GPU support by default. This seems to be enough to exceed readthedocs's memory limits and cause the build to fail. `tensorflow-cpu` 2.1 does not hit this problem, and so we can fall back to that when building documentation.

We probably want to make this more generally configurable (#546), but that's a more intricate problem, and this gets us back to having working documentation.

This also switches to only using the dependencies in `setup.py`, as building documentation doesn't need any of the dev dependencies listed in `requirements.txt`.

See: #602

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,6 +5,5 @@ version: 2
 python:
   version: 3.6
   install:
-    - requirements: requirements.txt
     - method: pip
       path: .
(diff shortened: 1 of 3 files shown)


huonw:
Also, it's not immediately obvious but this means we now build our documentation for every pull request, so it's easy to check that it looks good. E.g., for this PR:

image

The "Details" link goes to https://external-builds.readthedocs.io/html/stellargraph/631/index.html

PR #638, opened 2020-01-09
Test on Python 3.5 in addition to 3.6 and 3.7

huonw:
Do we still work with Python 3.5?


huonw:
The tests failed, so making the requirement formal in #641

PR #640, opened 2020-01-09
remove unnecessary documentation in the CI section

TheDen:
PR removes the `docker images` section in the docs under `CI`. Reasons being

* It's currently wrong since we've migrated to the `slim` variants
* It's an implementation detail that isn't useful


TheDen:
I think quality gates that contribution should be in https://github.com/stellargraph/stellargraph/blob/develop/CONTRIBUTING.md which is already be in place, i.e., `black` is listed there as a requirement, so there isn't really a need to mention the black docker image either

PR #641, opened 2020-01-09
Require Python >= 3.6, since 3.5 no longer works

huonw:
In https://github.com/stellargraph/stellargraph/pull/638, it was observed that stellargraph no longer supports Python 3.5, due to use of 3.6 features like [f-strings](https://docs.python.org/3/whatsnew/3.6.html#pep-498-formatted-string-literals).

This makes this requirement formal, in `setup.py`.

This may have been accidental, so maybe we want to not land this and instead fix our support for 3.5?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 **Breaking changes:**
 - The stellargraph library now only supports `tensorflow` versions 2.0 and above [\#518](https://github.com/stellargraph/stellargraph/pull/518). Backward compatibility with earlier versions of `tensorflow` is not guaranteed.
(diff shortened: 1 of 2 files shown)


huonw:
I can't see any other relevant instances of "3.5" in the repo:

```bash
% git grep '3\.5' -- '*.py'
tests/utils/test_calibration.py:    x_train = [[1, 1], [2, 3.5]]
tests/utils/test_calibration.py:    x_train = np.array([[1, 1], [2, 3.5]])
tests/utils/test_calibration.py:    x_train = np.array([1, 3.5])
tests/utils/test_calibration.py:    x_train = [[1, 1], [2, 3.5]]
tests/utils/test_calibration.py:    x_train = np.array([[1, 1], [2, 3.5]])
```

So I guess the Meta section in the sidebar of pypi is populated from here automatically, however I do not know for sure.

Issue #642, opened 2020-01-09
StellarGraph.edges(triple=True) does not return edge types, as documented


huonw:
**Describe the bug**

The `triple` argument to the method is documented to make the method return edge types https://github.com/stellargraph/stellargraph/blob/5e087faaa0fa207cd9dd2ad2d91d79232c0858c0/stellargraph/core/graph.py#L173-L183 

It seems to instead return the "edge multigraph index" (i.e. the index of the edge among all the edges between its src and dst)

**To Reproduce**
1. Run:
   ```python
   import stellargraph as sg
   import networkx as nx
   nxg = nx.Graph()
   nxg.add_edge(1, 2, label="x")
   nxg.add_edge(2, 3, label="y")
   g = sg.StellarGraph(g, edge_type_name="label")
   list(g.edges(triple=True))
   # [(1, 2, 0), (2, 3, 0)]
   ```
2. Note that the third element of the tuple is not the edge types `x` and `y`

**Expected behavior**
The code above should return `[(1, 2, "x"), (2, 3, "y")]`.

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Data used**
N/A

**Additional context**
N/A

PR #643, opened 2020-01-09
Return edge type in edges(triple=True) as documented

huonw:
This fixes `StellarGraph.edges` to return the edge type, as was intended based on the documentation. Previously it was returning an integer that had no connection to the edge type: the multigraph key, that is, the index of the given edge within all of the edges between the same source and destination.

This key could be set to an arbitrary hashable value using an appropriate version of `add_edge`/`add_edges_from`'s constructor, which worked ok in reduced tests, but doesn't match how the rest of the library works, and doesn't work if there's multiple edges of a particular type:

```python
import networkx as nx
g = nx.MultiGraph()
g.add_edges_from([(1, 2, "A"), (1, 2, "A")])
# ['A', 'A']
g.number_of_edges()
# 1
```

Using the data dictionary to store the label (like the rest of the library) means the edge keys are autogenerated and multiple edges are thus fine:
```python
g2 = nx.MultiGraph()
g2.add_edges_from([(1, 2, {"label": "A"}), (1, 2, {"label": "A"})])
# [0, 1]
g2.number_of_edges()
# 2
```

See: #642


huonw:
Hm, this currently seems to cause the accuracy of the RGCN notebook to plummet from 0.9 to 0.3-0.4, so something isn't quite right.

huonw:
Hm, it seems like it's probably not a problem caused by this PR: I checked out the commit that added RGCN (a2a2ef28), and ran the notebook there a few times, and found (validation) accuracies 0.63, 0.66, 0.5, 0.58, which seems quite different to the 0.94 that's currently checked in. (It also hit #649 consistently.) The training plot usually look like:

image

image

Which is qualitatively quite different to the output that's checked in on `develop`:

image

Those numbers are *somewhat* similar to the numbers from this commit, and from the latest develop. In summary, the sorted results for 4 runs of each:

| code | acc 0 | acc 1 | acc 2 | acc 3 |
|---|--:|--:|---:|---:|
|a2a2ef28|0.5|0.58|0.63|0.66|
|develop|0.36|0.38|0.47|0.53|
|this PR|0.41|0.44|0.5|0.56|

As such, I opened an issue: https://github.com/stellargraph/stellargraph/issues/677 (it's likely to be a duplicate of #649, or at least related, but on the chance that it isn't I've opened it).

Thus, I think this is now ready for review to be merged, because it isn't causing the notebook to regress.

Issue #644, opened 2020-01-09
Simplify StellarGraph type


huonw:
### Description

The `StellarGraph` type currently uses a metaclass to slot in an appropriate subclass of `StellarGraph`. This is flexibility we probably don't need for now; even if we have other types of stellargraph we can instead have an abstract superclass (similar to the old `StellarGraphBase`, maybe even called that), with different subclasses. This will also make it much easier to specify and document arguments (e.g. for IDEs).

We also have two classes (`StellarGraph`, and `StellarDiGraph`) that we don't need: we can have everything keyed off `StellarGraph.is_directed()`, but this change can happen separately.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #645, opened 2020-01-09
Remove the metaclass from StellarGraph

huonw:
This replaces the use of the metaclass with storing a `NetworkXStellarGraph` and then calling its methods. This does not move the implementations directly into the class, since that will be a lot of noise just to support the to-be-removed `NetworkXStellarGraph` class better. Improvements to the functionality can land directly in this class.

This copies the more experimental methods from `NetworkXStellarGraph` onto `StellarGraph` too, so that no other changes are required to the code (other than a couple of places that poke into the NetworkX internals (#599)).

~TODO: check that the notebooks still work.~ `python scripts/test_demos.py` passes.

See: #644

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -363,7 +363,7 @@ class StellarGraph:
         """
         return self._graph.to_adjacency_matrix(nodes)
 
-    # Experimental/special-case methods that need to be considered more
+    # FIXME: Experimental/special-case methods that need to be considered more
     def get_index_for_nodes(self, nodes, node_type=None):
         """
         Get the indices for the specified node or nodes.
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -114,8 +114,11 @@ class StellarGraph:
             a numeric feature vector for each node in the graph.
 
     """
+
     def __init__(self, *args, **kwargs):
+        # Avoid a circular import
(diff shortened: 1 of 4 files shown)


stellargraph/core/graph.py
```diff
@@ -134,14 +115,21 @@ class StellarGraph(metaclass=StellarGraphFactory):
 
     """
 
+    def __init__(self, *args, **kwargs):
+        # Avoid a circular import
+        from .graph_networkx import NetworkXStellarGraph
+
+        is_directed = kwargs.pop("is_directed", False)
```
kjun9: Do we need to handle this particular parameter explicitly like this? Seems like `NetworkXStellarGraph` should default to `is_directed=False` anyway?

stellargraph/core/graph.py
```diff
@@ -343,9 +362,67 @@ def to_adjacency_matrix(self, nodes: Optional[Iterable] = None):
         Returns:
              The weighted adjacency matrix.
         """
-        raise NotImplementedError
+        return self._graph.to_adjacency_matrix(nodes)
+
+    # Experimental/special-case methods that need to be considered more
```
kjun9: Is this a `FIXME`?

huonw: I guess so

```suggestion
    # FIXME: Experimental/special-case methods that need to be considered more
```

Issue #635, opened 2020-01-09
Neighbour based methods in Stellargraph class


habiba-h:
### Description
Adding methods in the `StellarGraph` class to access neighbours for a node or a list (subset of nodes of a graph) with or without a specified `edge_type`.

 The `StellarGraph` class has the following three methods defined. 

`def neighbors(self, node: Any) -> Iterable[Any]`
`def in_nodes(self, node: Any) -> Iterable[Any]`
`def out_nodes(self, node: Any) -> Iterable[Any]`

We need to extends these to be in the form `def neighbors(self, node: Any, data: Optional[List[str]] = None, edge_types: Optional[List[str]] = None)` or something along those lines. This is needed to do things like edge-type or edge-attribute aware random walks efficiently.

### User Story
As a: research engineer 
I want: to access neighbors/in_nodes/out_nodes for a specific node
so that: 
I can implement random walks.

### Done Checklist (Development)
- [x] Assumptions of the user story met
- [x] Produced code for required functionality
- [x] Branch and Pull Request build on CI
- [x] Branch and Pull Request pass unit tests on CI
- [x] Branch and Pull Request pass integration tests on CI
- [ ] Version number reflects new status
- [x] Peer Code Review Performed
- [x] Code well commented
- [ ] Documentation on Google Docs
- [ ] Documentation in repo
- [x] CHANGELOG.md updated
- [ ] Team demo
- [ ] Mini-meetup talk
- [ ] Stakeholder sign-off

kjun9:
Okay since we don't actually support a full set of edge features, only a single feature as a "weight", I'm now thinking we would have something like:

```python
    def neighbors(self, node: Any, weight=False, edge_types=None) -> Iterable[Any]:
        """
        Obtains the collection of neighbouring nodes connected
        to the given node.

        Args:
            node (any): The node in question.
            weight (bool, default False): If True, each neighbour in the output is a
                tuple of (node ID, edge weight)
            edge_types (list of str, optional): If provided, only traverse the graph
                via the provided edge types when collecting neighbours.

        Returns:
            iterable: The neighbouring nodes.
        """
```

habiba-h:
This would still be able to support multiple edges though? In temporal graphs, even if we ignore different types of edge features, except time, there are going to  be multiple edges with distinct times for a pair of nodes.

habiba-h:
[...]

PR #646, opened 2020-01-09
Allow specifying weight and edge_types for neighbor methods

kjun9:
Part of #635


tests/core/test_stellargraph.py
```diff
@@ -535,3 +535,71 @@ def f():
         return StellarGraph(g, node_features=node_features)
 
     allocation_benchmark(f)
+
+
+def example_weighted_hin():
+    graph = nx.MultiDiGraph()
+    graph.add_nodes_from([0, 1], label="A")
+    graph.add_nodes_from([2, 3], label="B")
+    graph.add_edges_from([(0, 1, 0.0), (0, 1, 1.0)], label="AA")
```
huonw: The `0.0` here is being used as a multigraph edge key, not as a weight. I think this should be:

```suggestion
    graph.add_edge((0, 1), label="AA", weight=0.0)
    graph.add_edge((0, 1), label="AA", weight=1.0)
```

See https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.MultiGraph.add_edges_from.html#networkx.MultiGraph.add_edges_from . In particular

> The edges can be:
> - ...
> - 3-tuples (u, v, k) for not iterable key k

and also https://networkx.github.io/documentation/stable/reference/classes/multigraph.html

> If an edge already exists, an additional edge is created and stored using a key to identify the edge. By default the key is the lowest unused integer.


- [ ] and similarly `weight=10.0` instead of the tuple in the next line.

kjun9: Ah nice catch, I meant to use https://networkx.github.io/documentation/networkx-2.2/reference/classes/generated/networkx.MultiDiGraph.add_weighted_edges_from.html#networkx.MultiDiGraph.add_weighted_edges_from here actually

stellargraph/core/graph_networkx.py
```diff
@@ -745,21 +745,49 @@ def edges(self, triple=False) -> Iterable[Any]:
...
+                else get_node(e)
+            )
+
+        def is_correct_type(e):
+            return (
+                edge_types_set is None
+                or e[2].get(self._edge_type_attr) in edge_types_set
+            )
+
+        return {get(e) for e in edges if is_correct_type(e)}
```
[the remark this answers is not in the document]
kjun9: Hmm ah I copied the existing behaviour without thinking about it, but yeah that makes sense. I guess we should return a list instead?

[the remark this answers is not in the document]
huonw: I guess so, but maybe it should just be an issue that we address later, given we're likely to want to change the existing behaviour too?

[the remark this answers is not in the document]
kjun9: Okay sure, I've created #653

stellargraph/core/graph.py
```diff
@@ -196,41 +196,53 @@ def has_node(self, node: Any) -> bool:
...
 
-    def neighbors(self, node: Any) -> Iterable[Any]:
+    def neighbors(self, node: Any, weight=False, edge_types=None) -> Iterable[Any]:
         """
         Obtains the collection of neighbouring nodes connected
         to the given node.
 
         Args:
             node (any): The node in question.
+            weight (bool, default False): If True, each neighbour in the output is a
```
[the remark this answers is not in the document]
habiba-h: I think 

>  it could be a filter (i.e. neighbors(weight=10) will find neighbours with weight 10)
is a very different function than what I would expect the `neighbors`, `in_nodes`, `out_nodes` should do.  Given a node it should return all neighbors of the node. Given a node and `weight=True`, it should return all neighbors nodes and the weights on their edges. If `edge_type` is none then, it should return all neighbors of the node with a dictionary of `edge_types` and (list of) weight(s) corresponding the the neighbor edge of the specific `edge_type`. 
And lastly, if the `edge_types` is provided as a list, then the weights for edges for only the types specified.

The filtering of neighborhood based on thresholding the weights could be something seperate. This would be quite handy for the temporal random walks.
However, I think the basic   `neighbors`, `in_nodes`, `out_nodes` should be about returning all neighbors, all in_nodes, all out_nodes with or without weights and for all edge types by default and subset of edge types if indicated.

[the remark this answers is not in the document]
kjun9: I did consider returning an output format that keeps things organised in terms of `edge_type`, but I didn't think it was necessary for now so I stuck to something simpler

stellargraph/core/graph.py
```diff
@@ -196,41 +196,59 @@ def has_node(self, node: Any) -> bool:
...
+    ) -> Iterable[Any]:
         """
         Obtains the collection of neighbouring nodes with edges
         directed to the given node. For an undirected graph,
         neighbours are treated as both in-nodes and out-nodes.
 
         Args:
             node (any): The node in question.
+            include_edge_weight (bool, default False): If True, each neighbour in the
+                output is a tuple of (node ID, edge weight)
```
huonw: With the named tuple, this could be

```suggestion
                output is a named tuple with fields `node` (the node ID) and `weight` (the edge weight)
```

or similar.

stellargraph/core/graph.py
```diff
@@ -196,41 +196,59 @@ def has_node(self, node: Any) -> bool:
...
+        self, node: Any, include_edge_weight=False, edge_types=None
+    ) -> Iterable[Any]:
         """
         Obtains the collection of neighbouring nodes connected
         to the given node.
 
         Args:
             node (any): The node in question.
+            include_edge_weight (bool, default False): If True, each neighbour in the
+                output is a tuple of (node ID, edge weight)
```
huonw: This and `out_nodes` can have the named tuple documentation fix applied too.

stellargraph/core/graph_networkx.py
```diff
@@ -745,21 +748,61 @@ def edges(self, triple=False) -> Iterable[Any]:
...
+    def neighbors(
+        self, node: Any, include_edge_weight=False, edge_types=None
+    ) -> Iterable[Any]:
         if self.is_directed():
-            in_nodes = {e[0] for e in self._graph.in_edges(node)}
-            out_nodes = {e[1] for e in self._graph.out_edges(node)}
+            in_nodes = self._in(node, include_edge_weight, edge_types)
+            out_nodes = self._out(node, include_edge_weight, edge_types)
             return in_nodes | out_nodes
         return nx.neighbors(self._graph, node)
```
huonw: I think this branch in each of these functions needs to handle the new arguments too: if the graph is undirected, it'll go here and not do any of the filtering/transforming:

```python
from stellargraph import StellarGraph
import networkx as nx

def example_weighted_undirected_hin():
     graph = nx.MultiGraph()
     graph.add_nodes_from([0, 1], label="A")
     graph.add_nodes_from([2, 3], label="B")
     graph.add_weighted_edges_from([(0, 1, 0.0), (0, 1, 1.0)], label="AA")
     graph.add_weighted_edges_from([(1, 2, 10.0), (1, 3, 10.0)], label="AB")
     return StellarGraph(graph)

list(example_weighted_undirected_hin().in_nodes(1, include_edge_weight=True))
# [0, 2, 3]
```

kjun9: Thanks, somehow completely looked past that

stellargraph/core/graph.py
```diff
@@ -196,41 +196,59 @@ def has_node(self, node: Any) -> bool:
...
             node (any): The node in question.
+            include_edge_weight (bool, default False): If True, each neighbour in the
+                output is a named tuple with fields `node` (the node ID) and `weight` (the edge weight)
+            edge_types (list of hashable, optional): If provided, only traverse the graph
+                via the provided edge types when collecting neighbours.
 
         Returns:
             iterable: The neighbouring nodes.
         """
         raise NotImplementedError
```
huonw: Github isn't noticing it as a formal merge conflict, but I think there's a conceptual one here: on `develop`, these all now call `self._graph.neighbors(...)` etc., and so they'll need to be updated to pass the arguments through:

https://github.com/stellargraph/stellargraph/blob/69dac010e106c93a6036638dc2bf2abfb82e84a9/stellargraph/core/graph.py#L197

- [ ] Similarly for all of the other ones

kjun9:
Sorry, realised I messed up the tests (`assert_items_equal` was `return`ing instead of `assert`ing), fixed now.

kjun9:
@habiba-h This is ready for your review

stellargraph/core/graph_networkx.py
```diff
@@ -39,6 +39,9 @@
 from .utils import is_real_iterable
 
 
+NeighbourWithWeight = namedtuple("NeighbourWithWeight", ["node", "weight"])
```
habiba-h: @kjun9 I think in this `namedtuple("NeighbourWithWeight", ["node", "weight"])` the label "weight" is hard coded? When I create a graph with a different attribute name and try to access edge weights, I get None. 
Here is a small example: 

```
G = nx.MultiDiGraph()
G.add_edge('a', 'b',  t = 1, label='temporal')
G.add_edge('a', 'c', t=2, label='temporal')
G.add_edge('c', 'd', t=3, label='temporal')
SG = sg.StellarGraph(G)
print(SG.info())
list(SG.neighbors('a',include_edge_weight=True))
list(SG.neighbors('c', include_edge_weight = True))
list(SG.in_nodes('c', include_edge_weight = True))
list(SG.out_nodes('c', include_edge_weight = True))

```
Also, 
`edge_weight` doesn't give back the weights. 
`[ (n, u, SG.edge_weights(n, u)) for u in SG.neighbors('a')]`

If the attribute is named `weight` all the methods work.

stellargraph/core/graph_networkx.py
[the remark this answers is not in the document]
habiba-h: @kjun9 & @huonw , not sure if the following behavior is related to the discussion above but seemed like it is, so adding the comment here: 
If an `edge_type` has multiple attributes, only one of the attributes accessible. Although the `StellarGraph` graph does list all the attributes in the `info()` method. In fact, its only the attributed with the name `weight` that one can access using `neighbours()`, `in_nodes()`, 'out_nodes()`, and `edge_weights()` methods.
Here is a short example: 
```
G = nx.MultiDiGraph()

G.add_edge('a', 'b',  wt = 3000, weight=6, label='temporal')
G.add_edge('a', 'c', weight=2, w = .001, label='temporal')
G.add_edge('c', 'd', weight=1, label='temporal')
G.add_edge('c', 'e', weight=7, label='temporal')
G.add_edge('c', 'f', weight=9, label='temporal')
G.add_edge('a', 'd', weight=3, label='temporal')

G.add_edge('b', 'f', closeness = 10, label='friend')
G.add_edge('c', 'd',  label='friend')

SG = sg.StellarGraph(G)
print(SG.info())

list(SG.neighbors('a',include_edge_weight=True))
list(SG.neighbors('b',include_edge_weight=True, edge_types = ['temporal']))

list(SG.neighbors('b',include_edge_weight=True, edge_types = ['friend']))
```

Currently, can't think of where multiple attributes of an `edge_type` would be useful but thought to point out the behavior just so that we know. 

I think the more critical point is whether the attribute has to have a name `weight` or it can be user specified and when `include_edge_weight` is set to `True` or `get_weights` is used, it returns the value of the attribute.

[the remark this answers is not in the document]
huonw: The name of the weight label can currently be specified by the user using the `edge_weight_label` argument to `StellarGraph`, however this doesn't seem to be documented anywhere and so, until #659, wasn't visible in the docs at all.

stellargraph/core/graph_networkx.py
kjun9: I think there's actually two different things here:

* If you want to assign a different networkx attribute as the "edge weight", the following should work:
```
>>> import networkx as nx
>>> nxg = nx.MultiDiGraph()
>>> nxg.add_edge('a', 'b', t=1, label="temporal")
0
>>> nxg.add_edge('a', 'c', t=2, label='temporal')
0
>>> nxg.add_edge('c', 'd', t=3, label='temporal')
0
>>> import stellargraph as sg
>>> g = sg.StellarGraph(nxg, edge_weight_label="t")
>>> print(g.info())
NetworkXStellarGraph: Directed multigraph
 Nodes: 4, Edges: 3

 Node types:
  default: [4]
    Edge types: default-temporal->default

 Edge types:
    default-temporal->default: [3]
        Attributes: {'t'}

>>> list(g.neighbors('a', include_edge_weight=True))
[NeighbourWithWeight(node='b', weight=1), NeighbourWithWeight(node='c', weight=2)]
```
but yes, as @huonw points out, the `edge_weight_label` parameter was undocumented so it would've been difficult to figure out the above
* I'd say that the `"weight"` within the named tuple is hardcoded, the same way the term "weight" is hardcoded in the parameter `include_edge_weight` - currently `StellarGraph` only supports weights on edges **by design** and **not** an arbitrary set of attributes on edges, but the user has the ability to specify `edge_weight_label` parameter to decide which networkx edge attribute gets used as the "weight". 

I think for now, for the temporal walk use-case, we'd just have to remember that we'd be using the term `weight` to refer to timestamps. If you think this is going to be too jarring, we'd have to consider supporting arbitrary edge attributes. What do you think?

habiba-h: We can stick to the term "weight", that would work as long as we know it. Also, for walks we don't need multiple attributes over the edges. So  that isn't necessary either.

PR #650, opened 2020-01-09
Add CI step to check whitespace

huonw:
There's a lot of trailing whitespace and missing (or multiple) newlines at the end of files that can cause diffs to have irrelevant changes or even cause problems with some tools (if newlines are missing).

This adds a `scripts/whitespace.sh` which can both correct whitespace issues (for local use) and check that they're correct (for CI). It adds a buildkite annotation to make it easy to understand the problem and fix it.

Example of a build with a failure (typically there will be fewer issues, since each pull request will only edit a relatively small number of files): https://buildkite.com/stellar/stellargraph-public/builds/409


huonw:
If this check is likely to be too annoying (although the fact that `scripts/whitespace.sh` can be run to fix the problems should mean that it isn't), we can always disable it in future.

PR #654, opened 2020-01-10
Update copyright in headers to 2020

timpitman:
basic update of copyright to cover the new year so that we don't have to manually trickle out the changes in each new PR


timpitman:
> If we're going to bulk updates, we can simplify `.buildkite/steps/check-copyright-headers.sh`

Thanks @huonw that's a great idea to simplify the CI and I'll just add to this PR 😎

PR #659, opened 2020-01-12
Use explicit argument list, not splats, for StellarGraph constructor

huonw:
An explicit argument list is nicer in a few ways than the fully dynamic `*args`/`**kwargs` version, even when there's many arguments (or, maybe, especially when there are):

- means it's clear what the interface is for our upcoming refactoring, so we're less likely to break something accidentally
- it makes it clearer to users too, for at-a-glance documentation reading (e.g. a user who already knows the library may just need a reminder of how to spell an argument), e.g. `help(StellarGraph)` now says:
  ```
  ...
   |  Methods defined here:
   |  
   |  __init__(self, graph=None, is_directed=False, edge_weight_label='weight', node_type_name='label', edge_type_name='label', node_type_default='default', edge_type_default='default', feature_name='feature', target_name='target', node_features=None, dtype='float32')
   |      Initialize self.  See help(type(self)) for accurate signature.
  ...
  ```
- it integrates better into static analysis and code tools, e.g. PyCharm's argument listing and autocomplete:
  
  image
  image

This is work along the lines of #644.

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -448,7 +448,8 @@ class StellarGraph:
 
 # A convenience class that merely specifies that edges have direction.
 class StellarDiGraph(StellarGraph):
-    def __init__(self,
+    def __init__(
+        self,
         graph=None,
         edge_weight_label="weight",
         node_type_name=globalvar.TYPE_ATTR_NAME,
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -448,6 +448,28 @@ class StellarGraph:
 
 # A convenience class that merely specifies that edges have direction.
 class StellarDiGraph(StellarGraph):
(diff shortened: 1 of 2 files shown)


stellargraph/core/graph.py
```diff
@@ -115,11 +116,36 @@ class StellarGraph:
 
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(
+        self,
+        graph=None,
```
kjun9: Does `graph` here need to be optional?
```suggestion
        graph,
```

huonw: Yeah, it's currently truly optional:

```python
>>> import stellargraph as sg
>>> sg.StellarGraph()
NetworkXStellarGraph: Undirected multigraph
    Nodes: 0, Edges: 0
```

Issue #622, opened 2020-01-08
setup.py and requirements.txt duplicate information (inconsistently)


huonw:
**Describe the bug**

`setup.py` and the various `requirements.txt` list dependencies for both normal development and testing (and demos). This duplication is unfortunate and can (and does) lead to inconsistencies between the versions:

https://github.com/stellargraph/stellargraph/blob/ca643349298a12ae7fd8d2239f0b2e20a44f534e/requirements.txt#L1-L16

https://github.com/stellargraph/stellargraph/blob/ca643349298a12ae7fd8d2239f0b2e20a44f534e/setup.py#L22-L46

E.g. `pytest` and `pytest-benchmark` are constrained differently in `requirements.txt` vs. `setup.py`.

**To Reproduce**
1. Observe `setup.py` and `requirements.txt`

**Expected behavior**
Changes and additions to dependencies should only have to happen in a single location.

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Data used**
N/A
**Additional context**
N/A

PR #661, opened 2020-01-13
Put all requirements into setup.py, not scattered requirements.txts

huonw:
This pulls all of the dependencies of the various bits of stellargraph (demos, testing, ...) into a single place: `setup.py`. There's no longer any `requirements.txt` other than the top level one that exists for developer convenience. Installing the dependencies required for all demos is `pip install stellargraph[demos,igraph]` or `pip install .[demos,igraph]` (or without the `igraph`, if not running the community detection examples).

See: #622


setup.py
```diff
@@ -35,22 +35,37 @@
...
+#
+# * demos/community_detection: mplleaflet, python-igraph (separate
+#   'extra', because it's only available on some platforms)
 #
 # * demos/ensembles/ensemble-node-classification-example.ipynb: seaborn
 #
 # * demos/link-prediction/hinsage/utils.py: numba
 #
 # Other demos do not have specific requirements
 EXTRAS_REQURES = {
```
timpitman: There is still a long-standing typo - could fix?

timpitman:
```suggestion
EXTRAS_REQUIRES = {
```

Issue #535, opened 2019-11-27
Consolidate example graphs in unit tests


adocherty:
### Description

Currently our unit tests are disorganized and each test creates example StellarGraph graphs in different or similar ways with no sharing of this code.

This issue is to improve the unit tests by making functions to create example graphs available to all unit tests by, for example, making them pytest fixtures at the top level of the tests (see https://docs.pytest.org/en/latest/fixture.html)

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

thanh-nm:
@huonw why is this being moved to IceBox?

huonw:
I don't know. Possibly a misclick in zenhub...

PR #663, opened 2020-01-13
Introduce graph fixtures

kjun9:
To start working on #535, for now I've basically just pulled out any example graphs that occur in more than 1 set of tests, WITHOUT:
* trying to combine any of them apart from very obvious cases where the only difference between two graphs was whether they're directed, etc.
* actually having to update any of the tests apart from simple naming changes.


kjun9:
> these aren't "fixtures" in the sense of https://docs.pytest.org/en/latest/fixture.html and so the name might be confusing; I'd be happy with test_utils.graphs or something (but also it doesn't matter that much)

Sure `test_utils.graphs` works 👍 

> there's a few more graphs that are very similar to existing ones (e.g. isomorphic other than the names of labels); are you planning to do follow up work to merge those too?

Yes I'm planning to do more follow up work, probably focusing on fewer graphs at a time

kjun9:
> Now that the test graphs are mostly consolidated in tests/test_utils/graphs.py it could be good to document them more clearly (future work, not needed for this PR) to make their test characteristics clearer for re-use in future tests.

Yeah I think a lot of graphs already in `test_utils.graphs` can be consolidated further and be renamed to be something more informative than just `example_graph_n`, etc so I'll start chipping away in some follow up PRs

Issue #662, opened 2020-01-13
New StellarGraph class backed by Pandas instead of NetworkX


timpitman:
### Description

For graph operations, stellargraph relies on networkx which has a very nice API and strong community support.  The heavy memory requirements of networkx however, limit the potential of StellarGraph to be used with larger graphs.

This issue covers work to change the StellarGraph class to use a more optimal Pandas representation, rather than NetworkX.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

timpitman:
See https://github.com/stellargraph/stellargraph/pull/597

PR #671, opened 2020-01-14
Compute GraphSchema more generically (and construct it more conventionally)

huonw:
This is now almost implemented generically (only in terms of functions that will exist on any StellarGraph), and so requires less change for the migration to a new StellarGraph implementation (#668). There's still a couple of places that poke into the internals of `NetworkXStellarGraph`, but fewer than there were.

This also removes the idiosyncratic approach to constructing the `GraphSchema` object, using mutation rather than `__init__`.

See: #662


stellargraph/core/graph.py
```diff
@@ -387,7 +387,82 @@ def create_graph_schema(self, create_type_maps=True, nodes=None):
...
+            }
+            edge_type_map = {
+                (src, tgt, key): edge_types.index(
+                    EdgeType(
+                        node_types[node_type_map[src]],
+                        self._graph._get_edge_type(data),
+                        node_types[node_type_map[tgt]],
+                    )
+                )
+                for src, tgt, key, data in self._graph._graph.edges(
```
kjun9: Shouldn't we *not* promote this to the generic `StellarGraph` yet unless we can get rid of this poke into `self._graph._graph`? What's the motivation for promoting this now?

huonw: I promoted it because being mostly generically implemented, even if not entirely, still means it requires less change for the migration to a new StellarGraph implementation, and so the work that lands the new stellargraph will have fewer changes. However, maybe I can just do this refactoring in `NetworkXStellarGraph` so that it can be copied over more easily, similar to `to_networkx`.

huonw: I've done this, and, as part of it, made the construction of `GraphSchema` more conventional by using `__init__` instead of mutations.

stellargraph/core/graph_networkx.py
```diff
@@ -646,31 +646,26 @@ def create_graph_schema(self, create_type_maps=True, nodes=None):
...
 
         if nodes is None:
-            nodes = self._graph.nodes()
-            edges = self._graph.edges(keys=True)
+            nodes = self.nodes()
+            edges = self.edges(triple=True)
 
         elif create_type_maps is False:
-            edges = self._graph.edges(nodes, keys=True)
+            edges = self.edges(triple=True)
```
kjun9: Hmm previously, `self._graph.edges(nodes, keys=True)` would only collect edges that are incident on `nodes`, whereas now I think it's collecting all edge triples? This `elif` path is when `nodes` is not `None`, and so I think we still want preserve the original behaviour?

It would effect the resulting `edge_types` set that gets passed into constructing the `GraphSchema`

huonw: I've switched back to a traversal over `self._graph.edges`.

PR #673, opened 2020-01-14
Experiment: Test on Python 3.8, on CI

huonw:
Does Python 3.8 work?


huonw:
Answer: no, Tensorflow doesn't support it and doesn't even try to run on it.

PR #680, opened 2020-01-15
Give nicer message for old NetworkX attribute on new StellarGraph

huonw:
The StellarGraph type no longer inherits from the `networkx.MultiGraph` or `networkx.MultiDiGraph` types. If a user updates to a new `stellargraph` version, they may have code like `some_stellargraph.successors` that accesses methods/properties on `StellarGraph` that were inherited from one of the networkx super-classes, which will stop working with the new code, giving a somewhat opaque error like:

```
AttributeError: 'StellarGraph' object has no attribute 'successors'
```

Fortunately, we can intercept the attribute look-up code path to put a more helpful addition into the error message:

```
AttributeError: 'StellarGraph' object has no attribute 'successors'. The 'StellarGraph' type no longer inherits from the NetworkX types: consider using one of its method, or, if that is not possible, the `.to_networkx()` conversion function.
```


stellargraph/core/graph.py
```diff
@@ -147,6 +147,34 @@ def __init__(
...
+                ) from None
+
+            # doesn't look like a NetworkX method so use the default error
+            raise
+        else:
+            # __getattr__ is only called for attribute accesses that don't exist,
+            # so the `__getattribute__` call should also fail, but, just in case it doesn't,
+            # let's not return None implicitly
+            raise AssertionError(
+                f"attribute '{item}' that was missing has magically appeared again"
```
kjun9: I'm thinking this should only happen when someone directly calls `graph.__getattr__('missing_field')` so the error message could talk about that instead if you want

huonw: I'm not sure I understand? Do you mean that this code path can only happen for a direct call like that? Or do you think that we should detect whether it's a direct call and only do this then? Or something else?

If the first one, that's not true, e.g. the test shows that `some_stellargraph.successors` takes this code path, and so would `getattr(some_stellargraph, "successors")`?

kjun9: Oops sorry I mean to say `graph.__getattr__('existing_field')`, and was referring to the code path that results in this message about "attribute that was missing has magically appeared again". 
```
>>> class A:
...   def __init__(self, x):
...     self.x = x
...   def __getattr__(self, item):
...     try:
...       super().__getattribute__(item)
...     except AttributeError as e:
...       print(e)
...       raise
...     else:
...       print("hmm")
...       raise
...
>>> a = A(1)
>>> a.x
1
>>> a.y
'A' object has no attribute 'y'
  File "<stdin>", line 6, in __getattr__
AttributeError: 'A' object has no attribute 'y'
>>> a.__getattr__('x')
hmm
  File "<stdin>", line 12, in __getattr__
RuntimeError: No active exception to reraise
>>>
```

huonw: Oh, I see. Yes, it seems this case isn't actually impossible...

Issue #677, opened 2020-01-14
RGCN demo results are typically much lower than checked in version


huonw:
**Describe the bug**
The RGCN demo notebook currently mentions an validation accuracy of 0.94:

https://github.com/stellargraph/stellargraph/blob/7689c1d16955db811085cc6ab9b1e6dc8cf84b98/demos/node-classification/rgcn/rgcn-aifb-node-classification-example.ipynb#L292-L294

However that does not seem easily reproducible, with validation accuracies closer to 0.4-0.6 on `develop` and even in the commit that landed the RGCN code:

| code | acc 0 | acc 1 | acc 2 | acc 3 |
|---|--:|--:|---:|---:|
|a2a2ef28|0.5|0.58|0.63|0.66|
|develop (13611b46651c453ceda48eb2e1b090e9f12b3da5)|0.36|0.38|0.47|0.53|

(Sorted validation accuracies as reported by 4 runs of the notebook on the relevant commit)

This could be related to #649.

**To Reproduce**
1. Open `demos/node-classification/rgcn/rgcn-aifb-node-classification-example.ipynb`
2. Run it
4. Observe the lower accuracy than the notebook in `develop`

**Expected behavior**
The checked-in notebook should be reproducible (or at least, a close result).

**Screenshots**
Training history with manual run in a2a2ef28:

![image](https://user-images.githubusercontent.com/1203825/72393134-1ecb3280-3786-11ea-9e5d-502a50962c50.png)

Training history of checked in notebook:

![image](https://user-images.githubusercontent.com/1203825/72393156-2e4a7b80-3786-11ea-9286-63a6fb0cf26b.png)

**Desktop (please complete the following information):**
 - OS: macOS

**Data used**
N/A

**Additional context**
N/A

kjun9:
Also encountered while doing https://github.com/stellargraph/stellargraph/pull/669 - getting tensorflow warnings during each epoch of training:
```
[...]

Issue #649, opened 2020-01-09
RGCN demo gives "Gradients do not exist for variables ['relational_graph_convolution_1/bases:0' ..." for every epoch


habiba-h:
### Description

The RGCN demo has tensorflow warnings about gradients not existing when training the model, e.g.

```
Epoch 1/20
WARNING:tensorflow:From /Users/wil9dw/.pyenv/versions/sg4/lib/python3.6/site-packages/tensorflow_core/python/ops/array_grad.py:502: _EagerTensorBase.cpu (from tensorflow.python.framework.ops) is deprecated and will be removed in a future version.
Instructions for updating:
Use tf.identity instead.
[...]

Issue #617, opened 2020-01-07
Establish an experimental/beta algorithm process


huonw:
### Description

In order to keep velocity high and friction low, we have the idea of landing algorithms and features as experimental or "beta", and then iterating for a little (e.g. improving the API, optimising speed and adding documentation/demos) before they graduate to be generally available.

If we're going to do this, we need an easy process to manage it and sure things don't get too stuck.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
(This is a proposal comment, everything here is just my opinion and fully open for discussion.)

I think there's two parts of this:

- how to indicate an API is experimental /beta (technical)
- how to decide to promote an API to non-experimental/beta (social)

### Indicating experimental/beta status

I have three ideas for this:

1. a submodule like `stellargraph.experimental` or `stellargraph.beta` that contains various experimental algorithms similar to `tensorflow.experimental` ([1](https://www.tensorflow.org/api_docs/python/tf/experimental), [2](https://stackoverflow.com/questions/58333491/what-does-experimental-in-tensorflow-mean)), and could potentially be more fine-grained, e.g. `stellargraph.layer.experimental`, `stellargraph.mapper.experimental` (TF also has `tensorflow.keras.experimental`). An experimental improvement would initially land in that submodule/subfolder, and then is moved to the appropriate non-experimental module
2. having an attribute like `@experimental` or `@beta` that modifies the class or function (e.g. adding `This is experimental, and may change at any moment` to the docs)
3. have some sort of actual check, so that users get an exception if they try to use (or import?) an experimental feature without setting a flag of some sort, for instance, we might have a `stellargraph.allow_experimental()`/`stellargraph.allow_beta()` and/or check an environment variable like `STELLARGRAPH_ALLOW_BETA`, and require them to be called (or set) before importing an `experimental`/beta` module (for 1) or using an `@experimental`/`@beta` class/function (for 2), otherwise they through an exception 

These can be combined.

### Promotion

The process for promotion could probably be lightweight for now, e.g., a checklist like:

- [ ] Has some tests
- [ ] Has documentation for all classes, functions and their parameters
- [ ] Has a demo

We could potentially add other criteria, for instance:

[...]

PR #685, opened 2020-01-16
Add @experimental to mark functions and classes as experimental

huonw:
This adds a lightweight annotation that can be placed on a function, method or class to automatically adjusts the documentation to include a big warning at the top. The annotation requires a `reason=...` argument to be more specific about why something is experimental, and (optionally) takes a list of issue numbers to link in the `issues=...` argument.

https://external-builds.readthedocs.io/html/stellargraph/685/api.html#stellargraph.core.StellarGraph.adjacency_types

image

https://external-builds.readthedocs.io/html/stellargraph/685/api.html#module-stellargraph.layer.rgcn

image

This change also spells out a basic checklist for promoting something out of "experimental".

Finally, to serve as examples, the annotation is used on some `StellarGraph` methods, and also on the RGCN layers and generators (due to #649 and #677).

Why use an annotation? There's two basic approaches that we discussed on #617:

1. annotation (+ documentation)
2. `experimental` submodules (e.g. `stellargraph.layers.experimental`)

Despite some consensus on approach 2, this PR takes approach 1, because it can be applied to individual methods, require fewer changes and file movements than 2 (as code moves in and out of experimental-ness), and, even if we take approach 2, having this annotation to automatically fill out the docs would still be nice. =

However, approach 2 is more visible in code, as a reminder that things may break: for example, `from stellargraph.layers.experimental import RGCN`. The approach here relies on users noticing the annotation in the documentation.

See: #617

### DIFF
--- a/stellargraph/core/experimental.py
+++ b/stellargraph/core/experimental.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018-2020 Data61, CSIRO
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from textwrap import dedent
+
+
+def experimental(*, reason):
+    def decorator(decl):
+        # add warning at the start of the documentation
+        # <https://docutils.sourceforge.io/docs/ref/rst/directives.html#caution>
+        decl.__doc__ = f"""\
+.. warning::
+
+   ``{decl.__qualname__}`` is experimental: {reason}. It may be difficult to use and may have major
+   changes at any time.
+
+{dedent(decl.__doc__)}
+"""
+        return decl
+
+    return decorator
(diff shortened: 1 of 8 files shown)


stellargraph/core/experimental.py
```diff
@@ -0,0 +1,59 @@
...
+        decl.__doc__ = f"""\
+.. warning::
+
+   ``{decl.__qualname__}`` is experimental: {reason}{issue_text}. It may be difficult to use and 
+   may have major
+   changes at any time.
+
+{dedent(decl.__doc__)}
+"""
+        return decl
```
kjun9: > However, approach 2 is more visible in code, as a reminder that things may break: for example, from stellargraph.layers.experimental import RGCN. The approach here relies on users noticing the annotation in the documentation.

Could we also print the warning in code? Something along the lines of:
```python
def decorator(decl):
  msg = f"""``{decl.__qualname__}`` is experimental: ... """
  decl.__doc__ = ".. warning:: " + msg + dedent(decl.__doc__)
  def new_decl(*args, **kwargs):
    warnings.warn(msg)
    return decl(*args, **kwargs)
  return new_decl
```

huonw: That's not a bad idea, although I think there's a few considerations:

- it will mean that the user will always see the warning, even when they're sure it's ok, and it won't be easy to silence (I guess we could have it be a custom `class ExperimentalWarning(Warning)` so that `ignore::ExperimentalWarning` works)
- getting it right for classes may be slightly subtle, since I think it'll have to override `__init__` somehow (I think we can probably assume that there's no metaclasses and so nothing tricky with `__new__` or however that works...)

I can give it a go here, or do it in a future PR?

kjun9: I'm happy to land this as is, but whatever you prefer

huonw:
I realised that we should call out RGCN being experimental in the changelog, if we're going to talk about its addition.

stellargraph/core/experimental.py
huonw: I'll do it as a follow-up.

PR #689, opened 2020-01-16
Add a dynamic warning to declarations marked `@experimental`

huonw:
A function, method or class are marked with `@experimental` will now have a highly visible dynamic warning (with the class `ExperimentalWarning`) when that function, method or class's `__init__` is called. This tries to address some of the downsides of the documentation only approach that we started with in #685.

This works reasonably well for public calls like constructing the RGCN types in `rgcn-aifb-node-classification-example.ipynb`.

~At the moment, this isn't so great: it emits warnings even for internal calls of functions like `StellarGraph.get_index_for_nodes` (see spew of warnings in `attri2vec-citeseer-node-classification-example.ipynb`) which are only experimental because we're not happy with how we're exposing it publicly. That is to say, internal calls inside `stellargraph` of these functions and types probably shouldn't get a warning, but I don't know how to achieve this.~ After discussing on slack, we decided to make these "package private" by adding `_` prefixes to the internal-facing methods.

See: #617


huonw:
Good idea; done 👍

Issue #610, opened 2020-01-07
Make it easier to get example datasets


huonw:
### Description

For our examples, it would be convenient if there was a one line way to download and load the datasets, e.g. `stellargraph.datasets.load_cora()` or similar.

### User Story

**As** a data scientist
**I want to** be able to run the examples smoothly
**So that** I can focus on the actual machine learning

### Work to be covered

1. add functions to download common datasets that can be directly downloaded
2. update corresponding demos (jupyter notebooks and python scripts)

Future work (see #758, #811, #812):

3. some datasets can't be directly downloaded due to licensing or location:
* DBLP
    * demos/link-prediction/attri2vec/stellargraph-attri2vec-DBLP.ipynb
* Hateful Twitter
    * demos/interpretability/gcn/hateful-twitters-interpretability.ipynb
    * demos/use-cases/hateful-twitters.ipynb
* GTD global terrorism database
    * demos/community_detection/attacks_clustering_analysis.ipynb
* YELP
    * demos/node-classification/hinsage/yelp-example.py

4. some scripts need larger changes:
demos/link-prediction/random-walks/main.py

### Done Checklist

- [x] Produced code for required functionality
- [x] Tests written and coverage checked
- [x] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [x] Documentation in repo
- [ ] Version number reflects new status
- [x] CHANGELOG.md updated
- [x] Team demo

Issue #686, opened 2020-01-16
Consider using static type annotations


huonw:
### Description

Python 3 static type annotations can tie into tooling like mypy https://mypy.readthedocs.io/en/latest/introduction.html to give us more assurance about our code, without having to test everything in great detail.

By default, these are static checks only, but they can be used to generate runtime checks also.

This allows for "gradual typing", where maybe only some core modules (e.g. the StellarGraph type) are annotated and checked.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #690, opened 2020-01-17
Datasets subpackage to provide easy downloading of demo datasets #610

timpitman:
This is the first PR for #610, addressing these parts:
1. add functions to download common datasets that can be directly downloaded
2. update corresponding demos (jupyter notebooks and python scripts)

Outline:

* a new 'datasets' subpackage which can download most of the packages required for demos
    * note: some are not supprted yet due to licensing / lack of simple URL: DBLP, Hateful Twitter, GTD global terrorism database, YELP
* notebooks now get the dataset descriptions from the new dataset classes, and download them (if already downloaded then they are re-used)
* lots of duplicate dataset messages from README's and notebooks have been removed.

What's not covered:
* code for loading/parsing the datasets (this is still mostly inconsistent cut-and-paste in each notebook)
* demos/link-prediction/random-walks/main.py (larger changes neeeded)
* scripts and notebooks mentioning the datasets mentioned above that can't easily be downloaded


stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+        """Returns true if the expected files for the dataset are present"""
+        for file in self.expected_files:
+            if not os.path.isfile(os.path.join(self.base_directory, file)):
+                return False
+        return True
+
+    def download(self, ignore_cache: Optional[bool] = False) -> None:
+        """Download the dataset (if not already downloaded, unless ignore_cache=True)"""
+        if not self._is_downloaded() or ignore_cache:
+            print(
```
huonw: My opinion is that it's not the best to do `print`s in a library. In particular:

- if the library is being used as part of a bigger application, it results in uncontrolled printing to the terminal
- it's awkward to capture/redirect

Doing this sort of thing as logging (with the `logging` module) seems like it would be better, although it's not entirely clear to me how to best set this up for a library (and, this is a problem that we need to fix elsewhere in `stellargraph` too).

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+    def _is_downloaded(self) -> bool:
+        """Returns true if the expected files for the dataset are present"""
+        for file in self.expected_files:
+            if not os.path.isfile(os.path.join(self.base_directory, file)):
+                return False
+        return True
+
+    def download(self, ignore_cache: Optional[bool] = False) -> None:
+        """Download the dataset (if not already downloaded, unless ignore_cache=True)"""
+        if not self._is_downloaded() or ignore_cache:
```
huonw:
```suggestion
        if ignore_cache or not self._is_downloaded():
```

since that'll avoid the disk IO if we're ignoring the cache.

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+    @staticmethod
+    def _all_datasets_directory() -> str:
+        """Return the path of the base directory which contains subdirectories for each dataset."""
+        return os.path.expanduser(
+            os.getenv("STELLARGRAPH_DATASETS_PATH", os.path.join("~", "data"))
+        )
+
+    def _is_downloaded(self) -> bool:
+        """Returns true if the expected files for the dataset are present"""
+        for file in self.expected_files:
```
huonw: This could be:

```python
return all(os.path.isfile(os.path.join(...)) for file in self.expected_files)
```

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+            if not os.path.isfile(os.path.join(self.base_directory, file)):
```
huonw: Maybe we could have a:

```python
@property
def files(self) -> List[str]:
    return [os.path.join(self.base_directory, file) for file in self.expected_files]
```

allowing

```python
for file in self.files:
    if not os.path.isfile(file):
        ...
```

or

```python
def resolve(self, file_name: str) -> str:
    return os.path.join(self.base_directory, file_name)
```

```python
for file in self.expected_files:
    if not os.path.isfile(self.resolve(file)):
        ...
```

and then these can also be used elsewhere.

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+                # archive of files
+                filename, _ = urlretrieve(
+                    self.url
+                )  # this will download to a temporary location
+                self._create_base_directory()
+                unpack_archive(
+                    filename, self._all_datasets_directory(), self.url_archive_format
+                )
+            if not self._is_downloaded():
+                print(f"{self.name} dataset failed to download")
```
huonw: I think this should be an error:

```suggestion
                raise ...(f"{self.name} dataset failed to download")
```

It might be good to be more specific in the error message too, e.g. try to determine why we think it failed. I don't know what exception would be most appropriate, but maybe `FileNotFoundError` if this only ever fails because of missing files.

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def _create_base_directory(self) -> None:
+        data_dir = self.base_directory
+        if not os.path.exists(data_dir):
+            os.makedirs(data_dir)
+
+    @staticmethod
+    def _all_datasets_directory() -> str:
+        """Return the path of the base directory which contains subdirectories for each dataset."""
+        return os.path.expanduser(
```
huonw: I think we should strongly consider using something other than `~/data` by default, because that's a directory that a user may well have ideas about the structure of, and already be using a lot. There's somewhat of a standard for a cache directory for various operating systems: e.g. XDG https://wiki.archlinux.org/index.php/XDG_Base_Directory on Linux

https://pypi.org/project/appdirs/ seems to expose this correctly for various platforms, so, if we had it as a dependency, it might be used like 

```python
import appdirs
...

default = appdirs.user_cache_dir("stellargraph", "stellargraph")
return os.getenv("STELLARGRAPH_DATASETS_PATH", default)
```

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+        expected_files: List[str],
+        description: str,
+        source: str,
+        data_subdirectory_name: Optional[str] = None,
+    ) -> None:
+        self.name = name
+        self.directory_name = directory_name
+        self.url = url
+        self.url_archive_format = url_archive_format
+        self.expected_files = expected_files
```
huonw: To save us from missing a file, should we validate:

```python
if url_archive_format is None and len(expected_files) != 1:
    raise ValueError("url_archive_format is None, which requires a single expected_file, found: {expected_files!r}")
```

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def download(self, ignore_cache: Optional[bool] = False) -> None:
+        """Download the dataset (if not already downloaded, unless ignore_cache=True)"""
+        if not self._is_downloaded() or ignore_cache:
+            print(
+                f"{self.name} dataset downloading to {self.base_directory} from {self.url}"
+            )
+            if self.url_archive_format is None:
+                # single file to download
+                self._create_base_directory()
```
huonw: I think this can be pulled out of the two `if`s, and I suspect we can even pull the url retrieve:

```python
filename, _ = url_retrieve(self.url)
self._create_base_directory()

if self.url_archive_format is None:
    # not an archive, so the downloaded file is our data and just needs to be put into place
    shutil.move(filename, self.files[0])
else:
    # archive, so it needs to be unpacked into the right directory
    unpack_archive(filename, self._all_datasets_directory(), self.url_archive_format)
```

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+                )
+                urlretrieve(self.url, filename=destination_filename)
+            else:
+                # archive of files
+                filename, _ = urlretrieve(
+                    self.url
+                )  # this will download to a temporary location
+                self._create_base_directory()
+                unpack_archive(
+                    filename, self._all_datasets_directory(), self.url_archive_format
```
huonw: I think this relies on the archive having a top-level directory named `{self.directory_name}` or `{self.directory_name}/{self.data_subdirectory_name}` inside that contains all the data. Is that true for all of them?

tests/datasets/test_datasets.py
```diff
@@ -0,0 +1,33 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+from stellargraph.datasets import download_all_datasets, Cora
+from urllib.error import URLError
+
+
+@pytest.mark.slow
```
huonw: Do we have anything that activates or deactivates these sort of tests?

demos/link-prediction/graphsage/README.md
```diff
@@ -20,31 +20,21 @@ of the library's [root directory](https://github.com/stellargraph/stellargraph).
...
-
 ## Running the notebook
 The narrated version of this example is available in the `cora-links-example.ipynb` notebook.
 To run the notebook:
  - Activate the python 3.6 environment in which the
 `stellargraph` library is installed
  - Start `jupyter-notebook`
    - note: you may need to first install `jupyter` by running `pip install jupyter` in your python environment
- - Navigate to the notebook (`/demos/link-prediction-hinsage/movielens-recommender.ipynb`), and click on
+ - Navigate to the notebook (`/demos/link-prediction/graphsae/cora-links-example.ipynb`), and click on
```
huonw:
```suggestion
 - Navigate to the notebook (`/demos/link-prediction/graphsage/cora-links-example.ipynb`), and click on
```

demos/embeddings/README.md
```diff
@@ -16,7 +16,6 @@ All examples demonstrate how to calculate embedding vectors for a graph's nodes
 The learned node representations can be used in numerous downstream tasks such as node attribute inference, link
 prediction, and community detection.
 
-
```
huonw: (I don't think there's a reason for this file to be changed?)

stellargraph/datasets/datasets.py
```diff
@@ -0,0 +1,138 @@
...
+        dataset.download(ignore_cache)
+
+
+class Cora(DatasetLoader):
+    def __init__(self) -> None:
+        super().__init__(
+            "Cora",
+            "cora",
+            "https://linqs-data.soe.ucsc.edu/public/lbc/cora.tgz",
+            url_archive_format="gztar",
```
huonw: I suspect all of these can be inferred from the URL itself; is it worth repeating this information? Maybe we could have `url_archive_format` allow `None` to mean "infer", so we don't have to specify it for "obvious" cases like this. That said, optimising this is probably low-value.

stellargraph/datasets/datasets.py
```diff
@@ -0,0 +1,138 @@
...
+    for cls in DatasetLoader.__subclasses__():
+        dataset = cls()
+        dataset.download(ignore_cache)
+
+
+class Cora(DatasetLoader):
+    def __init__(self) -> None:
+        super().__init__(
+            "Cora",
+            "cora",
```
huonw: It might be worth spelling out the name of at least one of these first two arguments, since they're so similar:

```suggestion
            directory_name="cora",
```

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,108 @@
...
+            print(
+                f"{self.name} dataset downloading to {self.base_directory} from {self.url}"
+            )
+            if self.url_archive_format is None:
+                # single file to download
+                self._create_base_directory()
+                destination_filename = os.path.join(
+                    self.base_directory, self.expected_files[0]
+                )
+                urlretrieve(self.url, filename=destination_filename)
```
huonw: Minor thing, but as https://stackoverflow.com/a/7244263/1256624 discusses, `urlretrieve` is "legacy". Is that something that we should think about?

stellargraph/datasets/datasets.py
```diff
@@ -0,0 +1,138 @@
...
+            description="The CiteSeer dataset consists of 3312 scientific publications classified into one of six classes. "
+            "The citation network consists of 4732 links. Each publication in the dataset is described by a 0/1-valued word vector "
+            "indicating the absence/presence of the corresponding word from the dictionary. The dictionary consists of 3703 unique words.",
+            source="https://linqs.soe.ucsc.edu/data",
+        )
+
+
+class PubMedDiabetes(DatasetLoader):
+    """
+    The Pubmed Diabetes dataset consists of 19717 scientific publications from PubMed database pertaining to diabetes classified into one of three classes.
```
huonw: It seems we're sometimes duplicating this description and URL field, and sometimes not. Maybe we could have it computed once? E.g. have `description` automatically filled out from `self.__doc__`

```
def __init__(self) -> None:
    super().__init__(
        ...
        description=self.__doc__
        ...
    )
```

Or have `__doc__` automatically filled out from the descriptions?

```python
class DatasetLoader:
    @staticmethod
    def compute_doc(cls):
        if cls.__doc__ is not None:
            raise ValueError("DatasetLoader docs should be empty before @compute_doc call")

        cls.__doc__ = f"""\
{cls.DESCRIPTION}

Further details at: {cls.SOURCE}"""
        return cls

@DatasetLoader.compute_doc
class PubMedDiabetes(DatasetLoader):
     description="...."
     source="..."

...
```

(with the latter the `description` and `source` arguments to the constructor become unnecessary, although `DatasetLoader.__init__` should probably check the properties exist.)

I'm sure there's tricks one could do with metaclasses to make this automatic too, but I would guess that's not worth it.

stellargraph/datasets/datasets.py
```diff
@@ -0,0 +1,138 @@
...
+"""
+
+from .dataset_loader import DatasetLoader
+from typing import Optional
+
+
+def download_all_datasets(ignore_cache: Optional[bool] = False) -> None:
+    """ download all datasets """
+    # obtain all the classes derived from DatasetLoader (note: must be direct descendants)
+    for cls in DatasetLoader.__subclasses__():
```
huonw: Cool, I didn't realise one could do this 👍

stellargraph/datasets/dataset_loader.py
timpitman: I agree that logging is more appropriate than print, I believe best practice is to log with the module name, and leave log configuration up to the calling application, eg: http://eric.themoritzfamily.com/learning-python-logging.html

stellargraph/datasets/datasets.py
timpitman: I thought about this, and it seemed more complexity for little gain (either attempting to parse and map the end of the URL, or the http request headers)

stellargraph/datasets/dataset_loader.py
timpitman: See also here: https://stackoverflow.com/questions/15035123/what-command-to-use-instead-of-urllib-request-urlretrieve/48483216#48483216

The most minimal substitutes appear to need to import and use shutil and tempfile and involve extra complexity for something that "might become deprecated".

stellargraph/datasets/dataset_loader.py
timpitman: Now using FileNotFoundError with a message listing the missing files

stellargraph/datasets/datasets.py
timpitman: This is a great idea, I've implemented slightly differently using '__init_subclass__'

timpitman:
> * notebooks should be formatted with `scripts/format_notebooks.py --all --overwrite` manually at the moment, although this will likely result in a spew of spurious changes, so maybe not unless #697 is fixed first
Given that there's many formatting issues with the current notebooks on develop, I think this should take place as part of #697.
> * these new classes should probably be added to our documentation in `docs/api.txt`
good idea, added.
> * the type annotations are great, but without any checking, they're sure to become out of date or have minor mistakes; do we have a plan for adding mypy or similar?
I think we should look at a gradual introduction of mypy as per #686
> * have you thought about how the automatic parsing code might work?
There's already code in the library that can move into the specific dataset classes (eg `load_dataset_BlogCatalog3` from `stellargraph/data/loader.py`), as well as a lot of cut-and-pasted dataset specific parsing code within the notebooks.

stellargraph/datasets/dataset_loader.py
timpitman: ~/data matches old behaviour in StellarGraph but is too generic.  I'm not sure that a standard cache directory is necessarily appropriate here, as it would be difficult for a user to get to.

stellargraph/datasets/dataset_loader.py
timpitman: Great suggestion, and this now works well in __init_subclass__

stellargraph/datasets/dataset_loader.py
timpitman: Yes, `self.directory_name` must match the directory name defined inside the archive - which is simplistic, but works with all of the current datasets.

tests/datasets/test_datasets.py
timpitman: no, this was a mistake (undefined so gave warnings) - have removed.

stellargraph/datasets/dataset_loader.py
huonw: > I'm not sure that a standard cache directory is necessarily appropriate here, as it would be difficult for a user to get to (eg on Mac it might be /Users/x/Library/Application Support/stellargraph and on Linux /home/x/.local/share/stellargraph)

I personally think this is fine because I cannot think of a compelling reason for the user getting to it. Can you give a little more details about why it's important?

> I've updated the path to be ~/stellargraph-datasets (which was already cross-platform due to the use of os.path.expanduser within _all_datasets_directory())
 
This is better since it's less likely to conflict, but it's still junking up a user's home directory for the sake of some demo datasets, that are likely to be only of transient interest to the user (as in, many users will have their own data to work with, and these datasets only serve to demonstrate the library, not solve any real task).

I think it's likely that people will notice and be mildly annoyed: we could reduce some of this annoyance by making it a `.` hidden directory, but I think even that is likely to get people frustrated (e.g. https://github.com/rust-lang/cargo/issues/1734 is a complaint about exactly this issue for Rust's cargo package manager, where it was using `~/.cargo`).

> this avoids dependency on another library too.

That's fair, although we might be able to get away with only installing it for demos and testing (although we'll likely have to put the import inside this function rather than at the top of the file, so it's not evaluated eagerly).

stellargraph/datasets/dataset_loader.py
huonw: Could you add a comment about this subtlety?

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,218 @@
...
+            self._delete_existing_files()  # remove any existing dataset files to ensure we re-download
+
+        if ignore_cache or not self._is_downloaded():
+            log.info(
+                "%s dataset downloading to %s from %s",
+                self.name,
+                self.base_directory,
+                self.url,
+            )
+            filename, _ = urlretrieve(
```
huonw: It looks like the comment at the end of this line is making black do this horrible line breaking; maybe:

```suggestion
            # this will download to a temporary location
            filename, _ = urlretrieve(self.url)
```

or potentially even remove the comment, e.g.:

```suggestion
            temporary_filename, _ = urlretrieve(self.url)
```

tests/datasets/test_datasets.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
```
huonw: There's a couple more tests we could do, I think:

- the `STELLARGRAPH_DATASETS_PATH` override too
- checking that the caching works (i.e. that it doesn't redownload if all the files are there)

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,218 @@
...
+                # archive, so it needs to be unpacked into the right directory
+                unpack_archive(
+                    filename, self._all_datasets_directory(), self.url_archive_format
+                )
+            self._verify_files_downloaded()
+        else:
+            log.info("%s dataset is already downloaded", self.name)
+
+
+def download_all_datasets(ignore_cache: Optional[bool] = False) -> bool:
```
huonw: If this function is only for testing, maybe we could not even have it here? That is, just have the `for cls in DatasetLoader.__subclasses__()` loop in the tests?

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,218 @@
...
+                os.remove(self._resolve_path(file))
+            except OSError:
+                pass
+
+    def download(self, ignore_cache: Optional[bool] = False) -> None:
+        """
+        Download the dataset (if not already downloaded)
+
+        Args:
+            ignore_cache bool, optional (default=False): Ignore a cached dataset and force a re-download.
```
huonw: I suspect this is meant to be something like:

```suggestion
            ignore_cache (bool, optional, default=False): Ignore a cached dataset and force a re-download.
```

but the RTD build isn't working to check.

In addition, I believe the default value should be listed in the function signature:

```suggestion
            ignore_cache (bool, optional): Ignore a cached dataset and force a re-download.
```

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,218 @@
...
+from shutil import unpack_archive, move
+from urllib.request import urlretrieve
+from typing import List, Optional, Any
+from urllib.error import URLError
+
+
+log = logging.getLogger(__name__)
+
+
+class DatasetLoader(object):
```
huonw: I don't think this is necessary in Python 3?

```suggestion
class DatasetLoader:
```

tests/datasets/test_datasets.py
timpitman: > * checking that the caching works (i.e. that it doesn't redownload if all the files are there)

I wasn't able to get this to work with pytest monkeypatch on urlretrieve; can you think of a better way?

stellargraph/datasets/dataset_loader.py
timpitman: I've now replaced this with @pytest.mark.parametrize in the tests to effectively create a test for downloading each dataset.

tests/datasets/test_datasets.py
huonw: Maybe capturing the logging and then checking for the second code path? Could also have `download` return some status flag (e.g. `True` if it downloaded, `False` if it used the cache). Neither are that great, though.

stellargraph/datasets/dataset_loader.py
timpitman: I feel a . (hidden) directory is even worse - secretly cluttering the home directory!  The library would be a bit fragile if appdirs wasn't a proper dependancy - passable with documentation but still extra friction (possibly better off as a separate package in that case, though I don't think the current value justifies the extra maintenance overhead of that option)

huonw: > The current behaviour roughly matches https://scikit-learn.org/stable/datasets/ which downloads to ~/scikit_learn_data/.

Ah, ok, that sounds like a fine precedent, then 👍

tests/datasets/test_datasets.py
timpitman: I managed to get unittest.mock to work wrapping urlretrieve

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,195 @@
...
+        """Convert dataset relative file names to their full path on filesystem"""
+        return os.path.join(self.base_directory, filename)
+
+    def _is_downloaded(self) -> bool:
+        """Returns true if the expected files for the dataset are present"""
+        return all(
+            os.path.isfile(self._resolve_path(file)) for file in self.expected_files
+        )
+
+    def _verify_files_downloaded(self) -> None:
```
huonw: This and `is_downloaded` have practically the same core. Maybe:

```python
def _missing_files(self) -> List[str]:
    return [file for file in self.expected_files if not ....]

def _is_downloaded(self) -> bool:
    return len(self._missing_files()) == 0

def _verify_files_downloaded(self) -> None:
    missing = self._missing_files()
    if missing:
        raise FileNotFoundError(...)
```

With the `_missing_files()` method, we potentially don't need `_verify_files_downloaded` because it's small enough to just inline into `download`.

tests/datasets/test_datasets.py
```diff
@@ -0,0 +1,79 @@
...
+
+    remove_dataset_directory()
+    # download - the url should work, but the files extracted won't be correct
+    with pytest.raises(FileNotFoundError):
+        dataset.download()
+    remove_dataset_directory()  # cleanup
+
+
+def test_environment_path_override(monkeypatch) -> None:
+    new_datasets_path = os.path.join("~", "test-sg-datasets")
```
huonw: If possible, I'd prefer we not junk up our own home-folders unnecessarily:

```suggestion
    with tempfile.TemporaryDirectory() as new_datasets_path:
```

Does that sound reasonable?

tests/datasets/test_datasets.py
```diff
@@ -0,0 +1,79 @@
...
+    dataset = Cora()
+    dataset.url = "http://stellargraph-invalid-url/x"
+    with pytest.raises(URLError):
+        dataset.download(ignore_cache=True)
+
+
+def test_missing_files() -> None:
+    # we start with Cora to a special path, but with the URL from CiteSeer
+    dataset = Cora()
+    dataset.directory_name = "test-missing-files"
```
huonw: Is this going to attempt to download to `~/stellargraph-datasets/test-missing-files`? Doesn't this not work as exactly intended because the download code relies on the directory matching the one in the archive, and this special directory doesn't?

Also, this could cause conflicts with concurrent test runs? I guess other tests (like the caching one) could fail in this case too so maybe we don't care?

tests/datasets/test_datasets.py
```diff
@@ -0,0 +1,79 @@
...
+import os
+from stellargraph.datasets import Cora, CiteSeer
+from urllib.error import URLError
+from stellargraph.datasets.dataset_loader import DatasetLoader
+from urllib.request import urlretrieve
+from unittest.mock import patch
+
+
+# use parametrize to automatically test each of the datasets that (directly) derive from DatasetLoader
+@pytest.mark.parametrize("dataset_class", list(DatasetLoader.__subclasses__()))
```
huonw: Nice!

stellargraph/datasets/dataset_loader.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def _create_base_directory(self) -> None:
+        data_dir = self.base_directory
+        if not os.path.exists(data_dir):
+            os.makedirs(data_dir)
+
+    @staticmethod
+    def _all_datasets_directory() -> str:
+        """Return the path of the base directory which contains subdirectories for each dataset."""
+        return os.path.expanduser(
```
huonw: Very minor thing, but I don't think the envvar should be expanded, strictly speaking, because the shell will do that:

```bash
$ X=~/foo
$ echo $X
/Users/.../foo
```

By expanding twice here we risk overruling the users intention, if they wanted to write to a directory that literally started with `~/`:

```bash
$ export X=\~
$ echo $X
~
$ python -c 'import os; print(os.path.expanduser(os.environ["X"]))'
/Users/wil9dw
```

That's very much an edge case, but:

```python
os.getenv("STELLARGRAPH_DATASETS_PATH", os.path.expanduser("~/stellargraph-datasets"))
```

(FWIW, I think `/` is fine as a literal directory separator, even on windows: https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#canonicalizing-separators.)

tests/datasets/test_datasets.py
```diff
@@ -0,0 +1,79 @@
...
+    new_datasets_path = os.path.join("~", "test-sg-datasets")
+    monkeypatch.setenv("STELLARGRAPH_DATASETS_PATH", new_datasets_path)
+    dataset = CiteSeer()
+    assert dataset.base_directory == os.path.join(
+        os.path.expanduser(new_datasets_path), dataset.directory_name
+    )
+
+
+def test_download_cache() -> None:
+    with patch(
```
huonw: I take it this works better than pytest's mocking?

demos/calibration/calibration-pubmed-link-prediction.ipynb
```diff
@@ -66,33 +68,35 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "PubMed Diabetes dataset is already downloaded\n"
```
huonw: It's probably annoying and unnecessary, but this looks like it was the old output before we switched to using `logging`. Is it worth rerunning the notebooks to switch back to the new output?

tests/datasets/test_datasets.py
timpitman: yes - it's a lot more full-featured (eg the call counting) and better documented!

demos/calibration/calibration-pubmed-link-prediction.ipynb
timpitman: notebooks have been reformatted and re-run (via format_notebooks with some changes in #733).

tests/datasets/test_datasets.py
timpitman: well spotted, I've replaced this with a much simpler test that doesn't need a special folder

timpitman:
> It's unfortunate the RTD build is failing because it'd be nice to verify that the `__init_subclass__` magic is working as we intend. Have you had a look locally? (Or even with `help(...)`?)

I've tested using make with sphinx to build local html files

Issue #695, opened 2020-01-19
DirectedGraphSAGENodeGenerator fails with 'TypeError: sample_features() missing 1 required positional argument: 'sampling_schema'' in version 0.8.3


huonw:
### Describe the bug

Per https://github.com/stellargraph/stellargraph/issues/657#issuecomment-575625215, `DirectedGraphSAGENodeGenerator` cannot be used in stellargraph 0.8.3, because it hits `TypeError: sample_features() missing 1 required positional argument: 'sampling_schema'`.

### To Reproduce

1. Run `demos/node-classification/graphsage/directed-graphsage-on-cora-example.ipynb`
2. Notice the failure in the `history = model.fit_generator(train_gen, ...` cell

### Observed behavior

```
Epoch 1/20
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-21-397acc8a3819> in <module>
      4     validation_data=test_gen,
      5     verbose=2,
----> 6     shuffle=False
      7 )

~/.pyenv/versions/sg4-0.8.3/lib/python3.6/site-packages/tensorflow/python/keras/engine/training.py in fit_generator(self, generator, steps_per_epoch, epochs, verbose, callbacks, validation_data, validation_steps, validation_freq, class_weight, max_queue_size, workers, use_multiprocessing, shuffle, initial_epoch)
   1431         shuffle=shuffle,
   1432         initial_epoch=initial_epoch,
-> 1433         steps_name='steps_per_epoch')
   1434 
   1435   def evaluate_generator(self,

~/.pyenv/versions/sg4-0.8.3/lib/python3.6/site-packages/tensorflow/python/keras/engine/training_generator.py in model_iteration(model, data, steps_per_epoch, epochs, verbose, callbacks, validation_data, validation_steps, validation_freq, class_weight, max_queue_size, workers, use_multiprocessing, shuffle, initial_epoch, mode, batch_size, steps_name, **kwargs)
    218     step = 0
    219     while step < target_steps:
--> 220       batch_data = _get_next_batch(generator, mode)
    221       if batch_data is None:
    222         if is_dataset:

~/.pyenv/versions/sg4-0.8.3/lib/python3.6/site-packages/tensorflow/python/keras/engine/training_generator.py in _get_next_batch(generator, mode)
    360   """Retrieves the next batch of input data."""
    361   try:
--> 362     generator_output = next(generator)
    363   except (StopIteration, errors.OutOfRangeError):
[...]

PR #696, opened 2020-01-19
Special-case DirectedGraphSAGENodeGenerator in NodeSequence.__getitem__

huonw:
This is a hotfix for `DirectedGraphSAGENodeGenerator` being broken in 0.8.3, because the dynamic type introspection used in the old `NodeSequence` type missed handling `DirectedGraphSAGENodeGenerator` in one location.

This PR is targeting the 0.8.4 release branch, and so will be released as 0.8.4.

See: #695


huonw:
Thanks. I removed them using the script: I used an adjacent checkout that had it (`python ../stellargraph-2/scripts/format_notebooks.py`), but there's numerous other approaches to avoid having to do it in an error-prone manual way, like download the file from GitHub or use `git show` to print the file from a newer commit (`git show develop:scripts/format_notebooks.py | python - --help`).

huonw:
Ah, never mind: a rereview is not required for this target branch.

Issue #697, opened 2020-01-20
Notebooks have expected formatting but this is not checked on CI


huonw:
### Describe the bug

There's a `scripts/format_notebooks.py` script that clears out warnings and formats code using black, but this has to be run manually at the moment, and there's nothing that catches a developer who forgets to run it.

### To Reproduce

1. Edit a notebook in a way that will be changed by `scripts/format_notebooks.py` (e.g. poor formatting)
2. Push it to CI

### Observed behavior

No failures

### Expected behavior

CI should fail if the notebook is formatted incorrectly.

### Environment

N/A

### Additional context

N/A

PR #674, opened 2020-01-14
Have an explicitly cached docker image for running on CI

huonw:

PR #698, opened 2020-01-20
Extend format_notebooks.py to validate formatting on CI

huonw:
This adds an option to `format_notebooks.py` to validate the formatting of the notebooks matches the expected format (by reformatting the notebooks to a temporary file and then comparing the original and the new). It can annotate a Buildkite build with a summary if there's failures.

This PR runs the check on CI, and thus has to reformat all of the notebooks. Review of the non-mechanical changes is likely to be easier after filtering out the `ipynb` files:

image

Example of a failure: https://buildkite.com/stellar/stellargraph-public/builds/669

See: #697


scripts/format_notebooks.py
```diff
@@ -239,15 +258,59 @@ def preprocess_cell(self, cell, resources, cell_index):
             # Write notebook file
             if overwrite_notebook:
                 nb_file_loc = str(file_loc.with_suffix(""))
+            elif check_notebook:
+                _fd, nb_file_loc = tempfile.mkstemp()
```
TheDen: Might be a good idea to close and remove the temp files after they're used, or maybe using [tempfile.TemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile) 🤔 

Another possibility is to use the `io` library to read to memory and compare the values—[nbconvert](https://nbconvert.readthedocs.io/en/latest/api/writers.html#nbconvert.writers.StdoutWriter) has a `nbconvert.writers.StdoutWriter` class.

In either case, the disk IO probably wouldn't affect CI speed with the number of files we have

.buildkite/steps/script.sh
```diff
@@ -18,6 +18,12 @@ echo "+++ running tests"
 exitCode=$?
 py.test -ra --cov=stellargraph tests/ --doctest-modules --doctest-modules --cov-report=term-missing -p no:cacheprovider --junitxml="./${junit_file}" || exitCode=$?
 
+if [ "${CHECK_NOTEBOOK_FORMATTING-0}" = 1 ]; then
+  echo "+++ checking notebook formatting"
+  # This script requires non-trivial dependencies, so run it here after installing them
+  python scripts/format_notebooks.py --all --ci demos/ || exitCode=$?
```
TheDen: Could this be separated in a different parallel step?

scripts/format_notebooks.py
```diff
@@ -239,15 +258,59 @@ def preprocess_cell(self, cell, resources, cell_index):
...
+        message = f"""\
+Found notebook(s) with incorrect formatting:
+
+{notebooks}
+
+Fix by running:
+
+    {command}"""
+
+        print(f"\n\033[1;91;40mError:\033[0m {message}")
```
TheDen: May be a good idea defining these ANSI colours in a variable?

.buildkite/steps/script.sh
huonw: The comment above is meant to describe why we aren't doing this, but I can see that it's unclear.

Basically, we'd have to spend 2 minutes of CPU time installing dependencies for this step that takes 20s, which seems like a waste.

scripts/format_notebooks.py
huonw: Yeah, probably, can I do it as a separate PR, because the codes in the rest of the script should be factored out too?

scripts/format_notebooks.py
huonw: > Might be a good idea to close and remove the temp files after they're used, or maybe using tempfile.TemporaryFile 🤔

I'd decided not to use `tempfile.TemporaryFile` because I thought that it had to be used as a context manager with `with`... but that's of course not true. I'll switch. 👍 

> Another possibility is to use the io library to read to memory and compare the values—nbconvert has a nbconvert.writers.StdoutWriter class.

Unfortunately, as the name suggests, that's only writes to `StdoutWriter`, and I think mutating `sys.stdout` to capture it is worse than using a temporary file.

scripts/format_notebooks.py
```diff
@@ -170,12 +174,23 @@ def preprocess_cell(self, cell, resources, cell_index):
...
-    parser.add_argument(
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
         "-o",
         "--overwrite",
         action="store_true",
         help="Overwrite original notebooks, otherwise a copy will be made with a .mod suffix",
     )
+    group.add_argument(
+        "--check",
```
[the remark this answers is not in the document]
huonw: I'm not so sure; I wouldn't expect us to be running `--check` that often, since it seems more likely that we'd be running with `--overwrite` locally to do the formatting; it's pretty safe and idempotent, so I don't think we need to encourage "check -> format if failed" vs. only "format", since the latter is 2&times; faster.

scripts/format_notebooks.py
huonw: I used `TemporaryDirectory`, because the output writes to a different file (it adds a `.ipynb` extension).

.buildkite/steps/script.sh
huonw: I've updated the comment.

huonw:
I personally find it quite annoying if I don't get feedback from real tests just because of a "silly" complaint like automated formatting, but I guess it's also annoying to have to wait 5 minutes to discover that the formatting is wrong. Maybe I can run it first but have the tests keep running, so that we get feedback about formatting immediately, and also about the tests.

If/when we land https://github.com/stellargraph/stellargraph/pull/674, we can split this out into a separate parallel step which also resolves this.

huonw:
I've done the above, it now checks the formatting (and adds the annotation, if required) and then runs tests even if the formatting was wrong, e.g. https://buildkite.com/stellar/stellargraph-public/builds/1015

image

PR #703, opened 2020-01-20
Remove GraphSchema edge-type look-ups

huonw:
Follow on to #702 that does the same to edges as that did for nodes: removes the per-element caching of edge types in the graph schema, since the graph schema seems like it should represent a whole-graph summary rather than anything about individual attributes.

This makes constructing a graph schema 60% faster because it has to do fewer traversals of the graph. This thus recovers the performance that was lost in #702 and more: this results in an overall speedup of 30% compared to `develop` before #702 landed.

This applies a "deprecation"/removal notice to future look-ups of these attributes, as well as to the `create_type_maps` parameter to `create_graph_schema`.

This simplification helps with #662, by reducing the API surface that has to be reimplemented.


stellargraph/core/graph_networkx.py
```diff
@@ -543,6 +543,24 @@ def node_types(self):
                 self._get_node_type(ndata) for n, ndata in self._graph.nodes(data=True)
             }
 
+    def _is_of_edge_type(self, edge, expected_triple):
```
kjun9: I find this method a bit unusual, compared to expressing it with functions that sound more generic:
```python
edge_type_triples_equals(self.edge_type(edge), expected_triple)
```
or something like that, which would basically be splitting this function into two? We already have `def node_type` so would it be good to have `def edge_type` anyway?

stellargraph/core/graph_networkx.py
```diff
@@ -543,6 +543,24 @@ def node_types(self):
...
+        source, target, _ = edge
+        triple = EdgeType(
+            self.node_type(source),
+            self._get_edge_type(self._graph.edges[edge]),
+            self.node_type(target),
+        )
+        if triple == expected_triple:
+            return True
+
+        if self.is_directed:
```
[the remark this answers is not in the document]
huonw: I think it matches `GraphSchema.is_of_edge_type`, so yes, it's a fix with respect to the one in this file.

stellargraph/core/graph_networkx.py
huonw: Yeah, I agree, but this code is disappearing with the new StellarGraph, so I don't think it's worth refactoring in a bigger way like that. I could be convinced otherwise.

kjun9: okay sure 👍

Issue #688, opened 2020-01-16
Refactor GraphWalk classes to not inherit `run` method


kjun9:
### Description

The individual subclasses implementing the `run` method from `GraphWalk` all take very different sets of parameters from each other, so I think there isn't really a need for them to pretend to have the same interface. From looking through all the usages of `run`, there doesn't seem to be anything obvious relying on the `GraphWalk` superclass having the `run` method defined.

### User Story

**As** a ...

**I want to** [accomplish some things]

**So that** [achieve a larger goal]

### Done Checklist

- [x] Produced code for required functionality
- [x] Tests written and coverage checked
- [x] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [x] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

kjun9:
Ended up updating the `run` method take `*args, **kwargs` so that we can still remove a lot of the default None values for parameters

PR #704, opened 2020-01-20
Remove run method from GraphWalk

kjun9:
#688


kjun9:
> As a potential alternative, could we make run def run(self, *args, **kwargs) to make it clear that that's the entry point, but still remove all the None default values on the parameters that aren't actually optional in the subclasses?

Ah yep that seems better

Issue #684, opened 2020-01-16
Make reproducible pipelines easier to create


kjun9:
### Description

It's often desirable to write a demo/workflow that is reproducible, but currently this means the user has to make sure whatever `stellargraph` methods and classes they use, they actually remember to provide a seed. Most `stellargraph` pipelines would also require numpy and tensorflow seeds to be set as well. It would be nice if this can all be done in one step.

### User Story

**As** a data scientist

**I want to** easily create pipelines with reproducible results

**So that** I'm not confused whether `stellargraph` supports reproducible results, and it lets me to incremental work more easily.

### Done Checklist

- [x] Produced code for required functionality
- [x] Tests written and coverage checked
- [x] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [x] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #705, opened 2020-01-21
Introduce a global seed for stellargraph

kjun9:
This ensures that someone trying to create a reproducible pipeline can do so with ease and in the same way they are being tested for reproducibility.

I've included this functionality in the `utils` module, which meant I had to remove the default importing in `utils.__init__.py` to avoid circular dependency. An alternative would be to create a separate `random` module outside of `utils` if that seems better.

#684


kjun9:
actually I need to update the reproducibility test to set the global seed back to `None` after it's finished - will fix

huonw:
> I've included this functionality in the utils module, which meant I had to remove the default importing in utils.__init__.py to avoid circular dependency. An alternative would be to create a separate random module outside of utils if that seems better.

Just as a quick review, I think we should work out how to avoid this (maybe using some `__all__`s in different places), because the imports are much better with the `from ... import *` re-exports. In any case, could you be a bit more specific about where the circularity is coming from? It doesn't seem like this patch changes the relative imports much?

---

Also, should we mutate the global seeds of all these libraries? An alternative approach might be to have `set_seed(...)` just control the default behaviour of the `seed=None` arguments. I'm not sure this suggestion is a _good_ idea, but it seems like one we should at least discard on purpose rather than accidentally. Have you thought about it?

kjun9:
> could you be a bit more specific about where the circularity is coming from? 

I'm actually not sure if I understand it properly, but what I'm guessing is a chain of imports like:

`import stellargraph` 
-> in `stellargraph.__init__.py`: `from stellargraph import utils` 
-> in `stellargraph.utils.__init__.py`: `from .ensemble import *` (along with others)
-> in `stellargraph.utils.ensemble.py`: `import stellargraph as sg`

Up until here is fine apparently so it doesn't matter that `ensemble` is importing itself? but now if we have in another module, say:
-> in `data.explorer`: `from ..utils.random import set_seed`
then this again imports everything in `utils` due to all the `import *`s in `utils.__init__.py` and this is when it starts giving the import error.

So what I've done is remove `from .ensemble import *` etc from `utils.__init__.py` then the error is gone.

Note that we still have in `stellargraph.__init__.py`:
```python
from stellargraph.utils.calibration import TemperatureCalibration, IsotonicCalibration
from stellargraph.utils.calibration import (
    plot_reliability_diagram,
    expected_calibration_error,
)
from stellargraph.utils.ensemble import Ensemble, BaggingEnsemble
```
which means doing a `from stellargraph import Ensemble, BaggingEnsemble` etc still works.

> Also, should we mutate the global seeds of all these libraries? An alternative approach might be to have set_seed(...) just control the default behaviour of the seed=None arguments. I'm not sure this suggestion is a good idea, but it seems like one we should at least discard on purpose rather than accidentally. Have you thought about it?

I didn't really consider doing that, but what I did consider was printing out more warnings about mutating the other seeds. I thought for most cases if you're wanting to set the global seed for stellargraph, your intention would be to set the seed for everything. Maybe another alternative would be to expose more parameters for `set_seed` to give the option to not mutate the other external ones?

stellargraph/utils/random.py
```diff
@@ -0,0 +1,65 @@
...
+
+import numpy as np
+import random
+import tensorflow as tf
+
+
+_sg_seed = None
+
+
+def get_seed(seed):
```
huonw: The docs say this is optional, but it doesn't seem to be.

```suggestion
def get_seed(seed=None):
```

stellargraph/utils/random.py
```diff
@@ -0,0 +1,65 @@
...
+        seed if not None, otherwise the global seed
+
+    """
+    if seed is None:
+        return _sg_seed
+    else:
+        return seed
+
+
+def set_seed(seed, set_np_seed=True, set_tf_seed=True, set_random_seed=True):
```
huonw: I think the `_seed` in these is somewhat redundant, and potentially even the `set_` (given the function is already called `set_seed`. Also, this function is rarely called, so I think it's probably ok to spell the names out to avoid any ambiguity:

```suggestion
def set_seed(seed, set_numpy=True, set_tensorflow=True, set_random=True):
```

stellargraph/utils/random.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Set the seed for all possible randomness in StellarGraph. Note that this
+    also sets the global random seed for the following external modules:
+        * numpy
+        * tensorflow
+        * random
+
+    Args:
+        seed (int, optional): seed value
+        set_np_seed (bool, default True): If true, mutate the global numpy seed
```
huonw: FWIW, the function signature already lists the defaults:

image

https://external-builds.readthedocs.io/html/stellargraph/705/api.html#stellargraph.utils.random.set_seed

So maybe

```suggestion
        set_np_seed (bool, optional): If true, mutate the global numpy seed
```

and for the others; or maybe not necessary?

stellargraph/utils/random.py
```diff
@@ -0,0 +1,65 @@
...
+        set_tf_seed (bool, default True): If true, mutate the global tensorflow seed
+        set_random_seed (bool, default True): If true, mutate the global random module seed
+
+    """
+    global _sg_seed
+    _sg_seed = seed
```
huonw: Hm, having thought about it, I think this, combined with `get_seed` (which notably doesn't exist for any of the other modules), is behaving in a way that is quite subtle and potentially surprising: it means that everything that calls `get_seed` will be using exactly the same seed.

This might ruin ensembles; in the following pseudo-ensembling (and pseudo-code) I think all of the `predictions_*` variables will be the same:

```
set_seed(123)

sampler_1 = SomeSampler() # uses seed = 123
sampler_2 = SomeSampler() # uses seed = 123

predictions_1 = predict(sampler_1)
predictions_2 = predict(sampler_2)

predictions_ensemble = mean(predictions_1, predictions_2)
```

For the other modules, it means that the sequence of random numbers for the global RNG will be the same, not that every value will be identical.

Maybe we should reconsider my initial suggestion and just go back to only (and always) mutating the global seeds?

stellargraph/utils/random.py
kjun9: I think we've discussed this before somewhere else, but I thought `something, optional` just meant a parameter is of type `Optional[something]`, rather than a parameter with a default value which is written as `x, default something`? I guess this might just be purely whatever convention we prefer, but my point at the time was how would you indicate that something is of optional type in the docstring if you reserved `optional` for things having default values?

kjun9: e.g. I intended `get_seed` to be an optional type, but not have a default value of `None`

kjun9:
@huonw

> Maybe we should reconsider my initial suggestion and just go back to only (and always) mutating the global seeds?

I'm not sure what you meant here - wasn't your initial suggestion to **not** mutate the global seeds? I've switched to mutating a global random state that keeps generating seeds on the fly as @timpitman 's suggested - hopefully this resolves the weirdnesses you've mentioned?

stellargraph/random.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Args:
+        seed (int, optional): seed value
+
+    Returns:
+        seed if not None, otherwise the global seed
+
+    """
+    if seed is None:
+        return _sg_random.randint(0, 2 ** 32 - 1)
```
huonw: Why 32-bit seeds, not 64-bit ones, out of curiosity?

tests/reproducibility/fixtures.py
```diff
@@ -39,3 +41,18 @@ def assert_reproducible(func, num_iter=20):
             model.get_weights(),
             model_new.get_weights(),
         )
+
+
+@contextmanager
+def use_seed(seed):
```
huonw: I think typically a context manager in this style would try to be invisible: that is, `a` and `b` would be equal in

```python
a = random()
b = random()
```

```python
a = random()
with use_seed(0):
    something_complicated()
b = random()
```

which isn't the case as this is written. However, I think it _is_ possible with `get_state` and `set_state` APIs:

- https://docs.python.org/3/library/random.html#random.getstate
- https://docs.scipy.org/doc/numpy-1.17.0/reference/random/legacy.html#seeding-and-state

but there doesn't seem to be something of this sort for TF (https://www.tensorflow.org/api_docs/python/tf/random), and this is probably unavoidable, if the randomness fits into its graph.

Do you think this is an important property?

If not, I'm not sure this function is that valuable: to me, there doesn't seem to be much difference between an explicit `set_seed(None)`, and just continuing after a single `set_seed(seed)`. That is, instead of `with use_seed(0): ...` we could do `set_seed(0); ...` and have very similar behaviour. What do you think?

stellargraph/random.py
```diff
@@ -0,0 +1,65 @@
...
+    else:
+        return seed
+
+
+def set_seed(seed, set_numpy=True, set_tensorflow=True, set_random=True):
```
huonw: > I'm not sure what you meant here - wasn't your initial suggestion to not mutate the global seeds? I've switched to mutating a global random state that keeps generating seeds on the fly as @timpitman 's suggested - hopefully this resolves the weirdnesses you've mentioned?

Sorry, I mean consider the suggestion as not a good one anymore, not consider reapplying. English is ambiguous... 😄 

In particular, the value of a user to passing `set_numpy=False` here now seems low to me, compared to the risk of making a mistake. I think both of the following have to be true of the use-case for `set_numpy=False` (or one of the others) to be correct:

- explicitly don't want to set one of the seeds (for some reason?!)
- either don't need full reproducibility, or are sure the particular algorithms they're using don't use that library's randomness

As such, maybe we could just switch to always setting the seeds for all of them, and then even simplify slightly by removing the `_sg_random` value in favour of `random.randint(0, ...)` (so there's less risk of any threadsafety problems, although I suspect there are not any such problems at the moment).

In summary, maybe we could do:

```suggestion
def set_seed(seed):
```

What do you think? (Sorry for leading you up a garden path...)

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Major features and improvements
 
 - New algorithms:
+- A global stellargraph random seed is now available via `stellargraph.utils.random.set_seed`. This can be used to create reproducible pipelines in the same way they are being tested. [\#684](https://github.com/stellargraph/stellargraph/issues/684)
```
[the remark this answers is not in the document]
kjun9: Yeah that's what I meant - do you think it's not an important detail?

stellargraph/random.py
kjun9: It seems like numpy only supports 32 bit seeds https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.random.seed.html#numpy.random.seed

and the seeds we generate from here does get used for numpy seeds

tests/reproducibility/fixtures.py
kjun9: > to me, there doesn't seem to be much difference between an explicit set_seed(None), and just continuing after a single set_seed(seed). That is, instead of with use_seed(0): ... we could do set_seed(0); ... and have very similar behaviour.

Hmm that's true 🤔  Okay I think I'll just remove this thing

stellargraph/random.py
kjun9: I was thinking of another use-case:

* the user wants to use some other utility of their own for managing seeds for numpy/tensorflow etc

I'm not sure how likely that is, but it felt like an easy option to include just in case. 

> simplify slightly by removing the _sg_random value in favour of random.randint(0, ...)

What do you mean by this? Don't we still need `_sg_random` to be a seedable `Random` object that acts as a seed generator?

huonw: > the user wants to use some other utility of their own for managing seeds for numpy/tensorflow etc

Fair enough. Do utilities like these exist, or is it all adhoc code around `set_seed` similar to what we're doing here?

> Don't we still need _sg_random to be a seedable Random object that acts as a seed generator?
 
I don't think so: if we're seeding the global Python one, then `random.randint(0, 2**32 - 1)` will also give a reproducible sequence of seeds even when interleaved with other operations. I could well be misunderstanding what you're getting at?

kjun9: > Do utilities like these exist, or is it all adhoc code around set_seed similar to what we're doing here?

Not that I know of, I was thinking of just something adhoc like here

> if we're seeding the global Python one, then random.randint(0, 2**32 - 1) will also give a reproducible sequence of seeds even when interleaved with other operations

Ah I see what you mean I forgot we were setting the global python one already 😅 okay that does simplify things, will update 👍

huonw:

PR #708, opened 2020-01-22
Consolidate example graphs 1 and 2

kjun9:
Example graphs 1 and 2 were very similar so I consolidated them into one. 

I chose to keep "example graph 2" since it was slightly more interesting - it has node feature values equal to the integer value of the node ID (`int(node_id) * np.ones`), whereas graph 1 just had `np.ones`.

Part of #535


tests/layer/test_graph_attention.py
```diff
@@ -528,12 +528,15 @@ def test_gat_node_model_no_norm(self):
         actual = model.predict_generator(ng)
 
         expected = np.ones((G.number_of_nodes(), self.layer_sizes[-1])) * (
-            self.F_in * self.layer_sizes[0] * self.attn_heads
+            self.F_in
+            * self.layer_sizes[0]
+            * self.attn_heads
+            * np.max(G.node_features(G.nodes()))
```
kjun9: Updated to this based on how I observed `actual` change with features in the graph. Currently going through the GAT implementation to figure out why...

kjun9: Okay I think I vaguely understand what's going on.. I was able to reproduce a very basic single forward pass of GAT with the weight initialisations done the same way as in the test cases:
```python
def gat(x, a, output_size, n_heads):
    input_size = len(x[0])

    # each row becomes sum(x) * ones(output_size)
    x1 = np.dot(x, np.ones((input_size, output_size)))

    # sum x1 across rows  (sum(x) * output_size)
    attn_self = np.dot(x1, np.ones((output_size, 1)))  
    attn_neigh = np.dot(x1, np.ones((output_size, 1)))

    # an N x N matrix where the magnitude of each value at (i,j) depends on the ith and jth row of the original feature matrix.
    attn_raw = attn_self + attn_neigh.T  

    # masking with adjacency matrix - our A is binary so this basically removes any non-neighbours from the "picking" process in the next line
    attn_masked = attn_raw + (-10e9 * (1.0 - a))  

    # "picks" the largest value in each row
    attn_softmaxed = softmax(attn_masked)  

    # `n_heads` number of these are concatenated
    out = np.tile(np.dot(attn_softmaxed, x1), (1, n_heads))  

    return out
```
It seems to basically have the effect of taking the largest transformed feature vector of your neighbours for each node, so then due to the connectivity in the graph, by the second layer, every node has ended up picking the largest possible so that's where `np.max(G.node_features(G.nodes()))` comes from. I realise this isn't what actually happens when weights aren't initialised with all ones, etc (especially the "picking" part), but for the test case the softmax only ends up assigning probabilities very close to 0 or 1

kjun9:
GAT test currently failing due to #763 I think

huonw:
> GAT test currently failing due to #763 I think

That's now fixed, so this looks like it's good to go.

Issue #718, opened 2020-01-22
New StellarGraph: return columnar/numpy-based representations from functions where possible


huonw:
### Description

In #662 #668, we add a new implementation of the `StellarGraph` class. This is backed by `numpy` and `pandas`, and we can potentially get significant efficiency gains if we propagate this through the rest of the library.

For example, `neighbours(include_edge_weight=True)` currently returns a list `neighbours` of tuples, of `(node_id, weight)`, these are then used in temporal random walks (currently https://github.com/stellargraph/stellargraph/pull/552) like `times = [t for _, t in neighbours]`. If we returned this in a columnar way, e.g. `node_ids, weights` where they're "paired" numpy array of the relevant data (as in, `weights[0]` is the weight for the edge to `node_ids[0]`), the downstream code could use it directly.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

Issue #717, opened 2020-01-22
New StellarGraph: convert tests and demos to use pandas not networkx


huonw:
### Description

In #662 #668, we add a new implementation of the `StellarGraph` class. Only a few tests and demos are converted to use the new functionality, we should convert the rest. NetworkX support (#715) can be tested explicitly, separately.

The new form looks like `StellarGraph(nodes=..., edges=...)`, where the arguments are some sort of dictionary involving pandas dataframes.

Files using the old form (in decreasing order of number of occurrences):

Main library/tests:

- [x] `tests/core/test_graph.py` (19) @huonw #998
- [x] `tests/mapper/test_full_batch_generators.py` (17) @huonw  #1000
- [x] `tests/test_utils/graphs.py` (9) @huonw #999
- [x] `tests/mapper/test_node_mappers.py` (9) @huonw #1005 #1014 
- [x] `tests/layer/test_gcn.py` (7) @huonw #999
- [x] `tests/layer/test_appnp.py` (7)  @huonw #999
- [x] `tests/data/test_biased_random_walker.py` (7) @huonw  #1004
- [x] `stellargraph/core/graph.py` (7) (false positives)
- [x] `tests/mapper/test_link_mappers.py` (4) @huonw #1005
- [x] `tests/utils/test_ensemble.py` (2) @huonw #1014
- [x] `tests/layer/test_ppnp.py` (2) @huonw #999
- [x] `tests/data/test_heterogeneous_breadth_first_walker.py` (2) @huonw #1004
- [x] `tests/mapper/test_directed_node_generator.py` (1) @huonw #1005 #1014
- [x] `tests/mapper/test_cluster_gcn_node_mapper.py` (1) @huonw #1005
- [x] `tests/mapper/test_benchmark_generators.py` (1) @huonw #1005
- [x] `tests/layer/test_cluster_gcn.py` (1) @huonw #999
- [x] `tests/data/test_temporal_random_walker.py` (1) @huonw
- [x] `tests/data/test_node_splitter.py` (1) @huonw #1004
- [x] `tests/data/test_metapath_walker.py` (1) @huonw #1004
- [x] `tests/data/test_directed_breadth_first_sampler.py` (1) @huonw #1004
- [x] `tests/data/test_breadth_first_walker.py` (1) @huonw #1004

Demos (many fixed as part of #812):

- [x] `demos/link-prediction/graphsage/cora-links-example.py` (3)
[...]

PR #719, opened 2020-01-23
Convert two more test graphs to use new StellarGraph

huonw:
There's a few adjustments required here to make this work by matching the existing behaviour:

- `node_types` needs to return a `set`
- `node_features` needs to support `None` as an ID, which gets filled out as a zero vector
- the node types in `GraphSchema` should be a sorted list
- `neighbours`, `in_nodes` and `out_nodes` should return a `list`, not an `numpy.ndarray` (changing this is #718)

See: #717

### DIFF
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -203,7 +203,7 @@ class UniformRandomWalk(GraphWalk):
         current_node = start_node
         for _ in range(length - 1):
             neighbours = self.neighbors(current_node)
-            if not neighbours:
+            if not len(neighbours):
                 # dead end, so stop
                 break
             else:
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -203,7 +203,7 @@ class UniformRandomWalk(GraphWalk):
         current_node = start_node
         for _ in range(length - 1):
             neighbours = self.neighbors(current_node)
-            if not len(neighbours):
+            if not neighbours:
                 # dead end, so stop
                 break
             else:
--- a/stellargraph/core/element_data.py
+++ b/stellargraph/core/element_data.py
@@ -59,11 +59,11 @@ class ExternalIdIndex:
         """
         return id in self._index
 
-    def is_missing(self, ilocs: np.ndarray) -> np.ndarray:
+    def is_valid(self, ilocs: np.ndarray) -> np.ndarray:
         """
-        Flags the locations of any ilocs that are missing (that is, to_iloc failed).
+        Flags the locations of all the ilocs that are valid (that is, where to_iloc didn't fail).
         """
-        return (ilocs < 0) | (ilocs >= len(self))
+        return (0 <= ilocs) & (ilocs < len(self))
 
     def to_iloc(self, ids, smaller_type=True) -> np.ndarray:
         """
--- a/stellargraph/core/element_data.py
+++ b/stellargraph/core/element_data.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import itertools
+from collections import defaultdict
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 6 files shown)


huonw:
I had that thought too, and unfortunately I don't think there's a simple way we can just drop it in. We could change these functions to return both types of graphs in a list, and have the tests iterate over them, or, change them to be `@pytest.fixture`s that are parameterised, but this gets a bit awkward with the arguments.

I decided against doing any of these, for a few reasons:

- the `NetworkXStellarGraph` code is fairly static at this point: we're not likely to be adding or changing much in it over the next week or so (before we remove it entirely)
- it's likely to be awkward and a non-trivial amount of dev time (I haven't actually tried any of the above, though)
- the opportunity cost of testing both seems high, with a lot of churn and short-term edits required

What do you think?

kjun9:
> We could change these functions to return both types of graphs in a list, and have the tests iterate over them, or, change them to be @pytest.fixtures that are parameterised, but this gets a bit awkward with the arguments.

Hmm yeah I can't really think of anything better :/

> the NetworkXStellarGraph code is fairly static at this point: we're not likely to be adding or changing much in it over the next week or so (before we remove it entirely)

I don't really expect there to be any further updates to NetworkXStellarGraph from now at all, but even updates to most of the other parts of the library generally want to interact with a graph (e.g. updates to generators/mappers, new random walks).

> the opportunity cost of testing both seems high, with a lot of churn and short-term edits required

How about we only switch over the graphs used in testing `StellarGraph` and related methods specifically, but keep the example graphs used for explorers and things as the old networkx one and only switch these as we remove the @experimental tag on the new `StellarGraph`?

huonw:
> I don't really expect there to be any further updates to NetworkXStellarGraph from now at all, but even updates to most of the other parts of the library generally want to interact with a graph (e.g. updates to generators/mappers, new random walks).

And so you're concerned about behavioural differences between the new and old `StellarGraph` classes? I think the scope for this is relatively small, assuming we can remove the old `StellarGraph` soon (before the next release):

- existing algorithms are relatively static
- the current state of the algorithms work with the old `StellarGraph`, since they passed the tests before this patch
- I think the only algorithm change that is likely before the 0.9 release is a minor tweak to RGCN that doesn't touch a `StellarGraph` object (#731), and maybe the temporal random walk work (but these are both `@experimental` for 0.9)
- we may need to do 0.9.x hotfixes to algorithms, in which case we do have a higher risk of problems, but, AFAICT, the main new algorithms are RGCN (experimental) and Cluster GCN (still using the old `StellarGraph` in tests)

As such, I think it's very unlikely that switching the tests over is going to open a hole that we'll hit over the next day or two for 0.9.0, and I think it's also unlikely for a 0.9.x hotfix.

> How about we only switch over the graphs used in testing StellarGraph and related methods specifically, but keep the example graphs used for explorers and things as the old networkx one and only switch these as we remove the @experimental tag on the new StellarGraph?

That's good in theory, but in practice the tests for `StellarGraph` itself are insufficient. I think most of the fixes I'm doing in these follow up PRs are flagged by tests outside of `test_stellargraph.py`, in the files for other algorithms. Something I want to do after the initial push to get things landed is flesh out the `StellarGraph` tests to be more complete in terms of what algorithms are using and expecting.

Another option would be to just land the fixes in this PR, since they're blocking other PRs, but not touch any of the tests. I'm not a huge fan of this, though, because it leaves more room for errors in the code that is being actively developed (the new `StellarGraph`).

stellargraph/data/explorer.py
```diff
@@ -203,7 +203,7 @@ def _walk(self, rs, start_node, length):
         current_node = start_node
         for _ in range(length - 1):
             neighbours = self.neighbors(current_node)
-            if not neighbours:
+            if not len(neighbours):
```
kjun9: I don't think this needed to change? Or if it does, does that mean the behaviour of `graph.neighbors` has changed?

huonw: Ah, yeah, it has: it now returns a `np.array`, instead of a `list` and so the implicit `bool(...)` call here hits:

```
>>> bool(np.array([1, 2]))
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

I'll change it back to returning a `list` so we're doing minimal breakage for now (at least until #718).

PR #720, opened 2020-01-23
Use the new StellarGraph for Cora examples

huonw:
Blocked on:

- [ ] a functional `info` method #710

This doesn't convert the link prediction examples that use Cora because they don't load the data in the same way.


huonw:

Issue #716, opened 2020-01-22
New StellarGraph: implement unimplemented methods


huonw:
### Description

In #662 #668, we add a new implementation of the StellarGraph class. This has some methods left unimplemented, which aren't needed for the core work, but are needed to run all algorithms:

- `info`: #710 
- `to_networkx`: #721
- `_adjacency_types`: #722 

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

Issue #679, opened 2020-01-15
Consider replacing the usage of `StellarGraph.adjacency_types` by extending other existing neighbourhood methods in `StellarGraph`


kjun9:
### Description

Currently, `StellarGraph.adjacency_types` is being used to do heterogeneous graph exploration, but we currently have this comment `# XXX This has not yet been standardised in the interface.` for this method and its functionality could probably just be achieved by extending the existing methods `neighbors`, `in_nodes` and `out_nodes`. 

### User Story

**As** a engineer / data scientist

**I want to** have one consistent set of methods to use for neighbourhood queries

**So that** it's easier to use, extend, maintain

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

kjun9:
This might make more sense to do after most of #662 has landed, since the neighbourhood methods need to be reimplemented with a pandas backend

huonw:
I think this relates to/is a duplicate of #1532.

PR #722, opened 2020-01-23
Implement _adjacency_types for the new StellarGraph

huonw:
This function is used by `SampledHeterogeneousBreadthFirstWalk`. I've manually verified that this matches the existing behaviour (the test doesn't pass with the old type because it has explicit entries for the ones with a `[]` list whereas the new form leaves those implicit, as part of the `defaultdict` lookup).

See: #716

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -924,9 +924,9 @@ class StellarGraph:
                 triples[other_triple][tgt].append(src)
 
         for subdict in triples.values():
-            for k, v in subdict.items():
+            for v in subdict.values():
                 # each list should be in order, to ensure sampling methods are deterministic
-                subdict[k] = sorted(v, key=str)
+                v.sort(key=str)
 
         return triples
 
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -21,7 +21,7 @@ a machine-learning ready graph used by models.
 __all__ = ["StellarGraph", "StellarDiGraph", "GraphSchema", "NeighbourWithWeight"]
 
 from typing import Iterable, Any, Mapping, List, Optional, Set
-from collections import namedtuple
+from collections import defaultdict, namedtuple
 import pandas as pd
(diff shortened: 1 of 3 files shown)


stellargraph/core/graph.py
```diff
@@ -904,7 +904,34 @@ def _adjacency_types(self, graph_schema: GraphSchema):
...
+            if not self.is_directed() and src != tgt:
+                # non-self loop in an undirected graph, so we should add the other direction too
+                triple_dict[tgt].append(src)
+
+                other_triple = EdgeType(tgt_type, rel_type, src_type)
+                if other_triple != triple:
+                    # the edge type differs too, so we need to add its other direction
+                    other_triple_dict = triples[other_triple]
+                    other_triple_dict[src].append(tgt)
+                    other_triple_dict[tgt].append(src)
```
kjun9: This means the result of grabbing `triples[(src_type, rel_type, tgt_type)][node]` can contain either `src_type` or `tgt_type` nodes rather than nodes of `tgt_type` only, which to me seems less useful for things like HinSAGE where you want to ensure doing a neighbourhood hop only returns nodes of one type. I'd consider `EdgeType(src_type, rel_type, tgt_type)`as inherently having some sort of directedness even in undirected graphs, did you consider just doing:
```python
if not self.is_directed() and src != tgt:
    other_triple = EdgeType(tgt_type, rel_type, src_type)
    if other_triple != triple:
        other_triple_dict = triples[other_triple]
        other_triple_dict[tgt].append(src)
```
I haven't looked closely at what the existing networkx based implementation is doing, but I would've thought it would be something along those lines

huonw: Ah, yeah, you're right, I must've not actually checked against the implementation correctly...

huonw:
> Just an FYI: #679 I originally filed an issue to remove this method altogether, but you might have a better idea whether that issue still makes sense

Yeah, I think it does still make sense 👍

PR #728, opened 2020-01-27
Reformat changelog for simplicity and experimental-ness

huonw:
This restructures the changelog to have fewer of the existing sections (from 5 to 3: "Major features and improvements", "Breaking changes" and "Bug fixes and other changes"), and adds an "Experimental features" section for any new `@experimental` code.

It also:

- switches to using markdown headers for the section headings, 
- puts each point on a single line (which diffs simpler when doing edits inside a sentence, since there's no word-wrapping that will change) and
- removes some extraneous whitespace inside the lists that caused them to be spaced out unusually

This is modelled after [tensorflow's release notes](https://github.com/tensorflow/tensorflow/blob/master/RELEASE.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,18 +13,11 @@
 
 - The stellargraph library now only supports `tensorflow` versions 2.0 and above [\#518](https://github.com/stellargraph/stellargraph/pull/518). Backward compatibility with earlier versions of `tensorflow` is not guaranteed.
 - The stellargraph library now only supports Python versions 3.6 and above [\#](). Backward compatibility with earlier versions of Python is not guaranteed.
-
-- The `StellarGraph` class no longer exposes `NetworkX` internals, only required functionality.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,37 +4,33 @@
 
 [Full Changelog](https://github.com/stellargraph/stellargraph/compare/HEAD...v0.8.2)
 
-**New algorithms:**
+### Major features and improvements
 
 - Cluster-GCN algorithm (an extension of GCN that can be trained using SGD) + demo [\#487](https://github.com/stellargraph/stellargraph/issues/487)
```
timpitman: I like the focus on major features, not just algorithms, but was wondering whether algorithms is important enough to have a sub-heading under this?

huonw: I've added a nested list. What do you think?

timpitman: looks good, thanks!

Issue #692, opened 2020-01-17
3 demo notebooks are broken with tensorflow 2.1


timpitman:
### Describe the bug
```demos/node-classification/attri2vec/attri2vec-citeseer-node-classification-example.ipynb``` appears broken by recent Tensorflow 2.1:

```ValueError: The two structures don't have the same nested structure.```
(more details below)

Other notebooks are similarly broken also:
* ```demos/embeddings/stellargraph-attri2vec-citeseer.ipynb```
* ```demos/embeddings/embeddings-unsupervised-graphsage-cora.ipynb```

### To Reproduce

1. install latest python packages (or even just upgrade tensorflow) ```pip install --upgrade tensorflow```
2. run the notebook test script: ```python scripts/test_demos.py```

### Observed behavior

```
ERROR in testing ./attri2vec-citeseer-node-classification-example.ipynb

An error occurred while executing the following cell:
------------------
history = model.fit_generator(
    generator.flow(unsupervised_samples),
    epochs=epochs,
    verbose=2,
    use_multiprocessing=False,
    workers=1,
    shuffle=True,
)
------------------

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/ia/stellargraph/venv/lib/python3.6/site-packages/tensorflow_core/python/util/nest.py in assert_same_structure(nest1, nest2, check_types, expand_composites)
    329     _pywrap_utils.AssertSameStructure(nest1, nest2, check_types,
--> 330                                       expand_composites)
    331   except (ValueError, TypeError) as e:

ValueError: The two structures don't have the same nested structure.

First structure: type=tuple str=(1, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 0, 0)

Second structure: type=tuple str=(None,)

[...]

PR #732, opened 2020-01-28
pin to tensorflow<2.1.0

timpitman:
While we investigate bugs related to tensorflow 2.1 (#692 and #626), we should release with tensorflow 2.0.x

Note: this PR has to remove the change for #631 since there's no tensorflow-cpu package for 2.0.x

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ URL = "https://github.com/stellargraph/stellargraph"
 # full tensorflow is too big for readthedocs's builder
 tensorflow = "tensorflow-cpu" if "READTHEDOCS" in os.environ else "tensorflow"
 REQUIRES = [
-    f"{tensorflow}>=2.0.0",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -21,10 +21,8 @@
 URL = "https://github.com/stellargraph/stellargraph"
 
 # Required packages
-# full tensorflow is too big for readthedocs's builder
-tensorflow = "tensorflow-cpu" if "READTHEDOCS" in os.environ else "tensorflow"
 REQUIRES = [
-    f"{tensorflow}>=2.0.0",
+    "tensorflow>=2.0.0, <2.1.0",
```
huonw:
```suggestion
    # FIXME(#626,#692): tensorflow 2.1 causes some test and demo failures
    "tensorflow>=2.0.0, <2.1.0",
```

Issue #723, opened 2020-01-23
New StellarGraph: streamline and document constructor


huonw:
### Description

In #662 #668, we add a new implementation of the StellarGraph class. The appropriate way to construct this is not yet settled, and is not documented. Both of these should be resolved.

(#715 relates, since that may change how we do construction from networkx.)

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #736, opened 2020-01-28
Add documentation for the new StellarGraph constructor

huonw:
This fleshes out the documentation enough to hopefully allow people to construct the new `StellarGraph` given pointers from us. We will need to do more work here once we fully switch to the new type, to emphasise the new form and de-emphasise the NetworkX form.

This is a start on: #723

Rendered at: https://external-builds.readthedocs.io/html/stellargraph/736/api.html#stellargraph.core.StellarGraph

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -98,6 +98,12 @@ class StellarGraph:
             [feature_vector_1, feature_vector_2, ..])
         Gs = StellarGraph(nx_graph, node_features=node_data)
 
+    .. warning::
+
+        The constructor variant using the ``nodes=..., edges=...`` arguments to create a "new"
+        StellarGraph is experimental: the type is insufficiently documented and does not support
+        some algorithms.
+
 
     Args:
         graph: The NetworkX graph instance.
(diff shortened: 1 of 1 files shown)


stellargraph/core/graph.py
```diff
@@ -126,6 +132,32 @@ class StellarGraph:
             required by some graph models. These are expected to be
             a numeric feature vector for each node in the graph.
 
+        nodes: DataFrame or dict of hashable to DataFrame
+            Features for every node in the graph. Any columns in the dataframe are taken as numeric
+            edge features of type ``dtype``. If there is only one type of node, a DataFrame can be
```
kieranricardo: I think you meant node features here

stellargraph/core/graph.py
```diff
@@ -126,6 +132,32 @@ class StellarGraph:
...
+        edges: DataFrame or dict of hashable to DataFrame
+            An edge list for each type of edges as a Pandas DataFrame containing a source, target
+            and (optionally) weight column (the names of each are taken from the ``source_column``,
+            ``target_column`` and ``edge_weight_label`` parameters). If there is only one type of
+            edges, a DataFrame can be passed directly, and the type defaults to the
+            ``edge_type_default`` parameter. Edges have an ID taken from the index of the dataframe,
+            and they have to be unique across all types. This must be used with the ``nodes``
+            argument. **Warning**: this is experimental (see warning above for more details).
+
+        source_column: str, optional
```
kieranricardo: It would be helpful to have the default values for `source_column` and `target_column` mentioned. Also documentation is missing for the `edge_weight_label` parameter.

huonw: The default value is shown in the class header:

image

I'm mildly against listing information redundantly, especially in string form, because it makes it very easy to become out of date.

> Also documentation is missing for the edge_weight_label parameter.

Ah, good catch; that parameter existed for the NetworkX type so I didn't think to double check it was documented.

kieranricardo: fair point! the class header is usually the first thing a user looks at anyway

PR #738, opened 2020-01-29
removed experimental tags from rgcn related classes

kieranricardo:
This PR removes the experimental tags from RGCN related classes. All known bugs were fixed with #731


kieranricardo:
@huonw nice catch, I removed all the references to experimental status that you mentioned

demos/node-classification/rgcn/rgcn-aifb-node-classification-example.ipynb
```diff
@@ -246,56 +246,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 9,
+   "execution_count": null,
```
huonw: This hasn't removed the warning in the text at the start of the file. In fact, I don't think any of this needs to change because the code isn't changing?

```
git checkout HEAD^ -- demos/node-classification/rgcn/rgcn-aifb-node-classification-example.ipynb
```

will get the file back to what it was before rerunning.

kieranricardo: Ah I misread your last review, I thought you were referring to code warnings appearing in the demo notebook. I re-ran notebook to be sure there were no experimental and so I could keep more output visible (less warnings so less cells I had to clear).

I'll delete that warning text at the start

huonw: Ah, makes sense 👍

PR #740, opened 2020-01-29
Bump version to 0.10.0b

kjun9:
Bump to next develop version


kjun9:
I think starting with 0.10.0b is fine for now regardless of whether we end up releasing 0.10 or 1.0 later since it's a smaller version than either of them. Tim's in a meeting til late so I might just merge this

PR #746, opened 2020-01-29
clear cell warnings after executing notebooks in format_notebooks

timpitman:
When `format_notebooks.py` is used to format and execute notebooks, cell warnings should be cleared *after* execution.

### DIFF
--- a/scripts/format_notebooks.py
+++ b/scripts/format_notebooks.py
@@ -207,9 +207,6 @@ if __name__ == "__main__":
     if renumber_code:
         preprocessor_list.append(RenumberCodeCellPreprocessor)
 
-    if clear_warnings:
-        preprocessor_list.append(ClearWarningsPreprocessor)
-
     if set_kernel:
         preprocessor_list.append(SetKernelSpecPreprocessor)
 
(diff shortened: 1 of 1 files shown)


scripts/format_notebooks.py
```diff
@@ -207,9 +207,6 @@ def preprocess_cell(self, cell, resources, cell_index):
     if renumber_code:
         preprocessor_list.append(RenumberCodeCellPreprocessor)
```
huonw: I think this one should go later too? Or are we sure that the renumber numbering is the same as the execution one?

timpitman: Running the entire notebook should run all of the cells in order - I thought that renumbering was to fix issues where cells had been run manually?

huonw: Yeah, I think that's why it exists, but I think there might be a detail about code cells vs. text cells, where our manual numbering handles text cells differently to execution. However, maybe it does match in which case it doesn't matter.

I guess I'm just asking us to confirm that the numbering from `ExecutePreprocessor` is the same as from `RenumberCodeCellPreprocessor`.

timpitman: Our numbering changes only code cells (specifically the field `execution_count`, see https://ipython.org/ipython-doc/3/notebook/nbformat.html).  In my testing, the renumber behaviour of ascending integer numbering from 1 matches Jupyter's execution.

huonw: Awesome; thanks for checking.

Issue #737, opened 2020-01-28
Community Detection Demo GraphSAGELinkGenerator Issue


dubbsbrandon:
### Describe the bug

Training generator produced from GraphSAGELinkGenerator is of type stellargraph.mapper.sequences.OnDemandLinkSequence when it should be GraphSAGENodeGenerator or GraphSAGELinkGenerator

### To Reproduce

To reproduce, follow the demo to cell 37, which will result in the error as the training generator is being created.

### Observed behavior

![Screen Shot 2020-01-28 at 4 04 43 PM](https://user-images.githubusercontent.com/54581642/73305394-ff9dbd80-41e7-11ea-9305-94c1673c59ea.png)

### Expected behavior

Link generator to be created.

huonw:
Hi @dubbsbrandon, thanks for using the library!

Could you say a bit more about the environment you're working in? In particular:

- what version of stellargraph are you using (`print(stellargraph.__version__)` in Python or `pip freeze` in a terminal)?
- how did you get the demo? E.g. did you download it from GitHub?

dubbsbrandon:
I just pulled the code directly from the examples, line by line. Nothing implemented myself. The version I have is the 0.9.0b version. Is there another, more updated one?

huonw:
> The version I have is the 0.9.0b version.

Ah, from that I'm guessing that you cloned the repository with `git` and ran something like `pip install .`? And then ran the examples out of the cloned directory too. Thanks for the info, it helps eliminate a possibility of other problems, like a version mismatch between the `stellargraph` library and which examples you're running 👍 

I can reproduce the problem, and I've put a fix for it at https://github.com/stellargraph/stellargraph/pull/747. I'm still verifying that it completely fixes the demo, but it at least fixes the problem here.

huonw:
That PR has been merged as 4d02ca3, so updating (e.g. `git pull` along with: `pip install --upgrade .` if it wasn't an `-e`/`--editable` install) should get the example working again.

Alternatively you can just apply the two cell fix there. Change:

```python
train_gen = GraphSAGELinkGenerator(Gs, batch_size, num_samples).flow(unsupervisedSamples)
```

to

```python
generator = GraphSAGELinkGenerator(Gs, batch_size, num_samples)
train_gen = generator.flow(unsupervisedSamples)
```
and

```python
layer_sizes=layer_sizes, generator=train_gen, bias=True, dropout=0.0, normalize="l2"
```

to

```python
[...]

PR #498, opened 2019-10-08
Changed GraphSAGE and HinSAGE to take the base generator object

adocherty:
I have updated the classes GraphSAGE and HinSAGE to accept a generator object not a sequence object. This makes them the same as GCN and other full-batch methods. In order to do this the HinSAGE generators now need to have the head node types specified.

Changes:
 * GraphSAGE/HinSAGE models require multiplicity argument if generator not passed
 * HinSAGENodeGenerator requires a head_node_type argument
 * HinSAGELinkGenerator requires a head_node_types argument

### Review Checklist
- [ ] Produced code for required functionality
- [ ] Branch and Pull Request build on CI
- [ ] Branch and Pull Request pass unit tests on CI
- [ ] Docstrings well written
- [ ] Code well commented
- [ ] Changelog updated
- [ ] Checked API documentation
- [ ] Version number reflects new status

PR #747, opened 2020-01-29
Pass Generator, not Sequence, to GraphSAGE in community detection demo

huonw:
In #498, GraphSAGE was changed to take the `GraphSAGE*Generator` types directly, rather than one of the sequence types. The community detection demo was missed in the updates, and so it had stopped working.

See: #737

**Reviewer note**: Unfortunately, despite the huge diff, the actual changes are tiny: they can be seen in the first commit: https://github.com/stellargraph/stellargraph/pull/747/commits/26063673c3138640b024df64d0907f90768dacfb

The second commit just reruns the whole notebook which results in the large amount of output.


huonw:
(I haven't yet run the rest of the whole notebook; working on it...)

PR #750, opened 2020-01-30
Feature/watch your step

kieranricardo:
This pull request contains an implementation of the Watch Your Step Algorithm (https://arxiv.org/pdf/1710.09599.pdf).

The current implementation is functional but it lacks unit tests and only has a bare-bones demo.

Reviewer checklist:

- [ ] The demo works correctly and replicates the classification results from the paper
- [ ] The `AdjacencyPowerGenerator` correctly creates a tensorflow dataset of rows of the the adjacency powers
- [ ] The `graph_log_likelihood` implements the graph log likelihood loss function correctly
- [ ] The tensorflow function in `tf_utils.py` are implemented correctly
- [ ] The `WatchYourStep` and `AttentiveWalk` correctly implement the trainable embeddings and attention functionality described in the paper


stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,199 @@
...
+        # always use all left vectors - currently wastes time looking up embeddings
+        # to keep keras happy
+        # TODO: replace the embedding layer with a custom layer to avoid lookups
+        # input cols but be somehow connected to the input to keep keras happy
+        input_cols = Lambda(
+            lambda x: tf.constant(np.arange(int(self.n_nodes)), dtype="int64")
+        )(input_rows)
+        vectors_right = right_embedding(input_cols)
+
+        dot_product = Lambda(lambda x: K.transpose(K.dot(x[0], x[1])))(
```
huonw: Just to confirm: the paper text uses the "outer product" here (end of page 2), but this seems to be using the inner product/normal matrix multiplication. However... it seems like the paper text is wrong (or at least misleading here), since this is operating on L (n &times; d/2) and R (n &times; d/2) to get a matrix compatible with A (n &times; n), and the outer product would give a tensor of shape n &times; d/2 &times; n &times; d/2 unlike the normal matrix multiplication L &times; R<sup>T</sup> in the equations, which gives shape n &times; n as desired.

huonw: I think this might be able to be a `Dot` layer https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/layers/Dot ?

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+
+        input_rows = Input(batch_shape=(None,), name="row_node_ids", dtype="int64")
+        input_powers = Input(batch_shape=(None, self.num_powers, self.n_nodes))
+
+        left_embedding = Embedding(
+            self.n_nodes,
+            self.embedding_dimension,
+            input_length=None,
+            name="LEFT_EMBEDDINGS",
```
huonw: Is it worth writing, just to avoid any ambiguities with other parts of a large model:

```suggestion
            name="WATCH_YOUR_STEP_LEFT_EMBEDDINGS",
```

stellargraph/mapper/adjacency_generators.py
```diff
@@ -0,0 +1,84 @@
...
+
+    """
+
+    def __init__(self, G, num_powers=5):
+
+        if not isinstance(G, StellarGraph):
+            raise TypeError("G must be a StellarGraph object.")
+
+        Aadj = G.to_adjacency_matrix()
+        indices = list(zip(Aadj.col, Aadj.row))
```
huonw: I think this is going to fail if `to_adjacency_matrix` returns something that's not `coo` format, so:

```python
Aadj = G.to_adjacency_matrix().tocoo()
```

Also, doing this processing via python iterators is unnecessarily expensive given these are already numpy arrays:

```suggestion
        indices = np.column_stack((Aadj.col, Aadj.row))
```

stellargraph/mapper/adjacency_generators.py
```diff
@@ -0,0 +1,84 @@
...
+
+        self.Aadj_T = tf.sparse.SparseTensor(
+            indices=indices,
+            values=Aadj.data.astype(np.float32),
+            dense_shape=(Aadj.shape[0], Aadj.shape[0]),
```
huonw: I think:

```suggestion
            dense_shape=Aadj.shape,
```

- [ ] and similarly in `self.transition_matrix_T`

stellargraph/utils/loss.py
```diff
@@ -0,0 +1,27 @@
...
+        This function is experimental: it is insufficiently tested.
+    '''
+    batch_adj = tf.gather(y_true, [0], axis=1)
+
+    expected_walks = tf.gather(y_pred, [0], axis=1)
+    sigmoids = tf.gather(y_pred, [1], axis=1)
+
+    adj_mask = tf.cast((batch_adj == 0), "float32")
+
+    loss = tf.math.reduce_mean(
```
huonw: I think matching the paper exactly would be:

```suggestion
    loss = tf.math.reduce_sum(
```

Could you say why you went for `reduce_mean`?

stellargraph/core/tf_utils.py
```diff
@@ -0,0 +1,63 @@
+import tensorflow as tf
+from tensorflow.keras import backend as K
+from ..core.experimental import experimental
+
+
+@experimental(reason="lack of unit tests")
+def partial_powers(one_hot_encoded_row, Aadj_T, num_powers=5):
```
huonw: The paper talks about doing an approximate SVD to compute this; I assume this code is not doing that?

I suspect that this is going to be O(k * n^3) where n is the number of nodes.

kieranricardo: Yeah you're write that the code isn't doing that. I explored using SVD and eigen dedcomposition. First up if I remember right I believe they're paper made an error bc their formula doesn't work with SVD it only works with an eiegen decomposition.

I experimented on cora, and even using the largest 1000 (out of 2700ish) eigenvectors the resulting reconstructed matrix had large errors that caused useless embeddings. I think this was because most of the eigenvalues were the same magnitude so there were no negligible eigenvalues to discard.  AFAIK the paper didn't actually test this idea out but only mentioned it.

Also, with the current tensorflow dataset when training on GPU the pre-processing of a batch on CPU (calculating the powers for the batch rows) takes approximately the same amount of time as training on a batch on GPU, so there's not a significant amount of time added by calculating the powers in parallel to training.

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,199 @@
...
+from ..core.experimental import experimental
+
+
+@experimental(reason="lack of unit tests")
+class AttentiveWalk(Layer):
+    """
+    This implements the graph attention as in Watch Your Step: Learning Node Embeddings via Graph Attention
+    https://arxiv.org/pdf/1710.09599.pdf.
+
+    .. warning::
```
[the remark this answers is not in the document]
kieranricardo: I wasn't aware of this, thanks I'll get rid of that.

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,199 @@
...
+        walk_length (int): the length of the random walks. Equivalent to the number of adjacency powers used.
+        attention_initializer (str or func): The initialiser to use for the attention weights;
+            defaults to 'glorot_uniform'.
+        attention_regularizer (str or func): The regulariser to use for the attention weights;
+            defaults to None.
+        attention_constraint (str or func): The constraint to use for the attention weights;
+            defaults to None.
+    """
+
+    def __init__(self, walk_length=5, **kwargs):
```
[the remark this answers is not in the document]
kieranricardo: Magic number? I'll change it to the default value of 10 in the paper.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
 
 Some new algorithms and features are still under active development, and are available as an experimental preview. However, they may not be easy to use: their documentation or testing may be incomplete, and they may change dramatically from release to release. The experimental status is noted in the documentation and at runtime via prominent warnings.
 
+- The Watch Your Step algorithm has been implemented including: new classes, generators, and utils functions. These classes and functions lack testing.
```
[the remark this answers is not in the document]
kieranricardo: Nice, agreed I'll put that in.

stellargraph/core/tf_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+
+    # make sure the transpose of the adjacency is used
+    # tensorflow requires that the sparse matrix is the first operand
+
+    batch_adj = select_row_from_sparse_tensor(one_hot_encoded_row, Aadj_T)
+
+    partial_powers_list = [batch_adj]
+
+    for i in range(1, num_powers):
```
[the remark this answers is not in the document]
kieranricardo: Oh nice, that's neater! I'll change the loop body but it might still be useful to have a `select_row_from_sparse_tensor` function for future use.

And yeah, we don't want the zeroth power/identity matrix

stellargraph/core/tf_utils.py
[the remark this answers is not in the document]
kieranricardo: Good point, I think it will mostly be used for WYS.  I can put the function in `adjaceny_generators.py`

stellargraph/core/tf_utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Args:
+        one_hot_encoded_row: one-hot-encoded row
+        sp_tensor_T: the transpose of the sparse matrix
+
+    returns:
+        The specified row from sp_tensor_T.
+    """
+    one_hot_encoded_row = tf.reshape(
+        tf.sparse.to_dense(one_hot_encoded_row), shape=(1, sp_tensor_T.shape[1])
```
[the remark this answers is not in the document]
kieranricardo: Unfortunately AFAIK it is - there's no tensorflow support for sparse sparse matmul only sparse dense matmul.

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,199 @@
...
+        )
+
+        right_embedding = Embedding(
+            self.n_nodes,
+            self.embedding_dimension,
+            input_length=None,
+            name="RIGHT_EMBEDDINGS",
+        )
+
+        vectors_left = Lambda(lambda x: K.transpose(x))(left_embedding(input_rows))
```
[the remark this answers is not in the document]
kieranricardo: To take the outer product of the right and left vectors `R \dot L^T`. You can see this formula in the final paragraph of page 2 of the paper (weirdly that formula isnt numbered)

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # always use all left vectors - currently wastes time looking up embeddings
```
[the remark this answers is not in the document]
kieranricardo: in each batch:

`vectors_right = embbedings.lookup[batch_rows]`
`vectors_left = embeddings[all_columns]` 

Only a subset of right vectors are looked up, while every left vector is looked up. This is pretty inefficient and I'm planning on writing a custom embedding layer for this.

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,199 @@
...
+        # to keep keras happy
+        # TODO: replace the embedding layer with a custom layer to avoid lookups
+        # input cols but be somehow connected to the input to keep keras happy
+        input_cols = Lambda(
+            lambda x: tf.constant(np.arange(int(self.n_nodes)), dtype="int64")
+        )(input_rows)
+        vectors_right = right_embedding(input_cols)
```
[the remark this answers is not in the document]
kieranricardo: It because I'm batching up the processing of calculating the loss function - the right vectors correspond to rows in the adjacency matrix/power and the left vectors correspond to the columns. Because we're using entire rows all of the left vectors are used (thus using `range(n_nodes)` as indexes instead of `input_rows`.

The ugly unecessary lambda function is to keep keras happy - every operation has to be linked to an input of the model otherwise keras crashes. Again a custom embedding layer will fix this ugliness

stellargraph/layer/watch_your_step.py
kieranricardo: I thought an outer product was just: `outer(A, B) = A \dot B^T` but maybe that only applies to vectors? But the paper had another inaccuracy so who knows :) 

Either way I think the desired behaviour is taking the inner product of every right vector with every left vector as a similarity measure

kieranricardo: I think you're right, also I'm not sure what that extra transpose is doing might be a bug.

stellargraph/layer/watch_your_step.py
kieranricardo: good idea!

stellargraph/utils/loss.py
kieranricardo: oh yep, thats a bug!

stellargraph/mapper/adjacency_generators.py
kieranricardo: I didn't know that python iterators were expensive! are they expensive in general or just relative to processing numpy arrays in numpy?

stellargraph/layer/watch_your_step.py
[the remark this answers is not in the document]
huonw: The paper does `L \dot R^T`, with the transpose on the other one? I'm sure it's mostly equivalent but why change it?

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,199 @@
...
+        if "input_shape" not in kwargs and "input_dim" in kwargs:
+            kwargs["input_shape"] = (kwargs.get("input_dim"),)
+
+        self.walk_length = walk_length
+
+        self._get_regularisers_from_keywords(kwargs)
+        super().__init__(**kwargs)
+
+    def compute_output_shape(self, input_shapes):
+        """
```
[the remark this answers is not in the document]
kieranricardo: that's a nifty trick, I'll give it a go

stellargraph/mapper/adjacency_generators.py
kieranricardo: yep, neater

stellargraph/layer/watch_your_step.py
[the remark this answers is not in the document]
kieranricardo: good point, I think that's left over from when I using TF 1 and I had to do something weird with the batch dimension.   `L \dot R^T` will simplify the code

stellargraph/layer/watch_your_step.py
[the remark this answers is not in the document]
kieranricardo: sorry I made a mistake in comment, should've been:

```
vectors_left = embeddings.lookup[batch_rows]
vectors_right = embbedings.lookup[all_columns]
```

stellargraph/core/tf_utils.py
kieranricardo: I think the training scales as `N^2` but as a ratio to GPU training CPU pre-processing will linearly increase with the number of nodes.. with the reddit dataset (200K nodes) the GPU training is still about the same as the CPU pre-processing but you're right that with larger graphs it is likely that CPU pre-processing will dominate the computation time.

stellargraph/layer/watch_your_step.py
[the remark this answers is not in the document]
kieranricardo: I think its pretty standard to have decent default parameter for algorithms (e.g. all the scipy algos have defaults). its also convenient for data scientists and means they can just test things out without having to look into an algorithm.

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,199 @@
...
+                Shape tuples can include None for free dimensions, instead of an integer.
+
+        Returns:
+            An input shape tuple.
+        """
+        return (input_shapes[0][-1],)
+
+    def _get_regularisers_from_keywords(self, kwargs):
+        self.attention_initializer = initializers.get(
+            kwargs.pop("attention_initializer", "glorot_uniform")
```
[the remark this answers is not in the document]
kieranricardo: I was just keeping things consistent with some of the other `stellargraph` stuff - but I think real arguments make more sense here. I'll keep "input_shape"/"input_dim" in the `kwargs` for now just so someone isn't tripped up by having to specify "input_shape" now when previously "input_dim" worked fine for other `stellargraph` layers.

kieranricardo:
@huonw I think I've addressed all your comments :)

stellargraph/layer/watch_your_step.py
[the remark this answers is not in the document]
huonw: > I was just keeping things consistent with some of the other stellargraph stuff - but I think real arguments make more sense here

Yeah, I don't blame you: consistency is good and important. However, in this case, I think we should be moving everything to arguments where possible, especially because this doesn't change how the user interacts with the API, just how it is documented. 👍 

> I'll keep "input_shape"/"input_dim" in the kwargs for now just so someone isn't tripped up by having to specify "input_shape" now when previously "input_dim" worked fine for other stellargraph layers.

I wasn't suggesting changing the functionality/the API, just how it is expressed, similar to how the `attention_initializer="glorot_uniform"` defaulted-parameter works the same as the manual defaults done by manually examining `kwargs`; I could be wrong, but I think the current `input_dim` behaviour is equivalent to:

```python
def __init__(self, ..., input_dim=None, **kwargs):
    if "input_shape" not in kwargs and input_dim is not None:
        kwargs["input_dim"] = (input_dim,)
```

stellargraph/layer/watch_your_step.py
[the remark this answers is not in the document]
huonw: Sounds reasonable. Could you have either a reference in the documentation or just a comment pointing out where the default comes from, so that it's not entirely a magic number?

stellargraph/core/tf_utils.py
huonw: Interesting.

> I think the training scales as N^2 

Out of curiosity, is it O(N^2) or O(E) (since E is proportional to N^2 in "dense" graphs).

kieranricardo: I'm think the training scales as O(N^2) because it computes the dot product of every left vector with every right vector i.e. compares each node to every other node.

stellargraph/layer/watch_your_step.py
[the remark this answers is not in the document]
kieranricardo: I've included `input_dim` as a real argument now :)

stellargraph/mapper/adjacency_generators.py
huonw: Relative to processing numpy arrays in numpy. Python iterators are relatively inexpensive compared to everything else in the rest of Python, but numpy wins because it cuts out all of general Python overhead. In particular, a numpy array is a sequence of small numbers in memory, e.g. `np.arange(10000)` places 8 byte `np.int64` integers contiguously into a 80000 byte chunk of memory. This makes processing very efficient, because CPUs are great at doing operations on big chunks of memory like that. For example:

```python
import numpy as np
data = np.arange(10000)

%timeit -n 100 data.sum()
# 8.67 µs ± 880 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit -n 100 sum(data)
# 1.39 ms ± 94.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# just iterate over 10000 in a Python list
L = [None]*10000
def f():
    for _ in L: pass
%timeit -n 100 f()
# 59.4 µs ± 5.04 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# what's the cost of just iterating over a numpy array?
def g():
    for _ in data: pass
%timeit -n 100 g()
# 738 µs ± 42.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

That is, the pure numpy summation is ~160&times; faster than using the Python `sum`, and is even ~7&times; faster than just iterating over a list. Why?

- `array.sum()`: simple memory layout: get access to the underlying data, and then loop across the array all in C (with full static types, no function calls, zero allocations, and even doing multiple additions in parallel with SIMD)
- `sum(array)` (and `for _ in data`): create a python iterate object for `array`, and for each element, call a python function to get the next element (likely needs a hash table look-up to find the function), which requires allocating a new python object value to hold the C integer, filling in the value, and returning that; then (for `sum(array)`), add that to the accumulator inside `sum` by calling other Python functions (`int.__add__`), which likely involves various dynamic overheads like hash table lookups
- `for _ in L`: create a python iterator object for `L`, and for each element, call a python function to get the next element (with the hash table lookups). This form doesn't have to allocate new objects for the return value (since the `None` values already exist in the list), nor do anything with them, so it's focusing on just the cost of iterating in Python.

There's also sorts of other Python overheads too, like just the pervasive cost of having to interpret Python byte code, rather than executing CPU instructions directly.

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,202 @@
...
+            self.n_nodes,
+            self.embedding_dimension,
+            input_length=None,
+            name="WATCH_YOUR_STEP_RIGHT_EMBEDDINGS",
+        )
+
+        vectors_left = left_embedding(input_rows)
+
+        # TODO: replace the embedding layer with a custom layer to avoid lookups
+        # input cols but be somehow connected to the input to keep keras happy
```
huonw: Is this line a continuation of the `TODO`, or is it a separate comment? In either case, I don't really understand what it's getting at. Could you clarify?

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,202 @@
...
+        right_embedding = Embedding(
+            self.n_nodes,
+            self.embedding_dimension,
+            input_length=None,
+            name="WATCH_YOUR_STEP_RIGHT_EMBEDDINGS",
+        )
+
+        vectors_left = left_embedding(input_rows)
+
+        # TODO: replace the embedding layer with a custom layer to avoid lookups
```
huonw: Does it have to be custom? We could potentially model it as a trainable `n × d/2` matrix that is multiplied with the left embeddings (i.e. a dense layer). Does that make sense? What do you think?

stellargraph/mapper/adjacency_generators.py
```diff
@@ -0,0 +1,152 @@
...
+
+        training_dataset = tf.data.Dataset.zip(
+            (row_index_adj_powers_dataset, batch_adj_dataset)
+        ).batch(batch_size)
+
+        return training_dataset.repeat()
+
+
+@experimental(reason="lack of unit tests")
+def partial_powers(one_hot_encoded_row, Aadj_T, num_powers=5):
```
huonw: Do these need to be public functions? If not, they could be renamed to:

```suggestion
def _partial_powers(one_hot_encoded_row, Aadj_T, num_powers=5):
```

and then even the `@experimental` removed; in particular the user will currently see the `ExperimentalWarning`s even for the internal calls, which isn't correct (they cannot do anything about it, and, calls to `@experimental` functions inside `stellargraph` are fine: if we change the function we can change the callers easily). What do you think?

huonw: Since this is mostly an internal function, I'm not sure it makes sense to default the `num_powers` argument. If this is useful more generally than WyS, I don't think there's any reason that 5 is a good default (and, even for WyS, it seems 10 is the better default); thus:

```suggestion
def partial_powers(one_hot_encoded_row, Aadj_T, num_powers):
```

stellargraph/utils/loss.py
```diff
@@ -0,0 +1,43 @@
...
+
+import tensorflow as tf
+
+from ..core.experimental import experimental
+
+
+@experimental(reason="lack of unit tests")
+def graph_log_likelihood(y_true, y_pred):
+    """
+    .. warning::
```
huonw: Not needed because of `@experimental`

stellargraph/layer/watch_your_step.py
kieranricardo: That's exactly what I was thinking of doing but wrapped in a custom keras layer (all trainable weights have to be in keras layers). Although maybe a `Dense` layer will do the trick here...

stellargraph/mapper/adjacency_generators.py
kieranricardo: true, I don't think a stellargraph user would ever cal this directly

kieranricardo: yeah, especially true once its a private function

stellargraph/utils/loss.py
kieranricardo: nice catch!

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,202 @@
...
+        returns:
+            A tuple of (inputs, outputs) to use with a keras model.
+        """
+
+        input_rows = Input(batch_shape=(None,), name="row_node_ids", dtype="int64")
+        input_powers = Input(batch_shape=(None, self.num_powers, self.n_nodes))
+
+        left_embedding = Embedding(
+            self.n_nodes,
+            self.embedding_dimension,
```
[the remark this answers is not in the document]
kieranricardo: yep you're right, I'll add that in. I'll add an a check in there as well - I think I'll round down to the nearest even number and output a warning.

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+    This implements the graph attention as in Watch Your Step: Learning Node Embeddings via Graph Attention
+    https://arxiv.org/pdf/1710.09599.pdf.
+
+    Args:
+        walk_length (int): the length of the random walks. Equivalent to the number of adjacency powers used. Defaults
+            to `10` as this value was found to perform well by the authors of the paper.
+        attention_initializer (str or func): The initialiser to use for the attention weights;
+            defaults to 'glorot_uniform'.
+        attention_regularizer (str or func): The regulariser to use for the attention weights;
```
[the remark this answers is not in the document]
kieranricardo: good point!

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,202 @@
...
+class AttentiveWalk(Layer):
+    """
+    This implements the graph attention as in Watch Your Step: Learning Node Embeddings via Graph Attention
+    https://arxiv.org/pdf/1710.09599.pdf.
+
+    Args:
+        walk_length (int): the length of the random walks. Equivalent to the number of adjacency powers used. Defaults
+            to `10` as this value was found to perform well by the authors of the paper.
+        attention_initializer (str or func): The initialiser to use for the attention weights;
+            defaults to 'glorot_uniform'.
```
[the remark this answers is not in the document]
kieranricardo: are you referring to the default values?

stellargraph/mapper/adjacency_generators.py
kieranricardo: No they don't, stellargraph users won't have any need to call these funcitons - I'll make them private

stellargraph/layer/watch_your_step.py
kieranricardo: nice! the dense layer worked

stellargraph/layer/watch_your_step.py
kieranricardo: Its a separate line - i've removed the comment now that I've replaced the right embedding layer with a dense layer. 

The comment was explaining that the creation of the column indexes (tensor of all ints up num nodes) had to be done through a lambda layer connected to an input to the model to enable keras to compile the model. If that makes sense?

```all_cols = Lambda(
            lambda x: tf.constant(np.arange(int(self.n_nodes)), dtype="int64")
        )(input_rows)```

File: stellargraph/layer/watch_your_step.py [hunk 24]
```diff
@@ -0,0 +1,202 @@
...
+        self,
+        walk_length=10,
+        attention_initializer="glorot_uniform",
+        attention_regularizer=None,
+        attention_constraint=None,
+        input_shape=None,
+        **kwargs
+    ):
+
+        if input_shape is None and "input_dim" in kwargs:
```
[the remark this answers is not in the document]
kieranricardo: woops I mixed up `input_shape` and `input_dim` :sweat_smile:  I've fixed this now

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,190 @@
...
+            raise TypeError("num_walks should be an int.")
+
+        if num_walks <= 0:
+            raise ValueError("num_walks should be a positive int.")
+
+        self.num_walks = num_walks
+        self.num_powers = generator.num_powers
+        self.n_nodes = int(generator.Aadj_T.shape[0])
+
+        if embedding_dimension % 2:
```
huonw: I personally find treating ints as booleans a bit confusing (it takes a little bit to translate the implicit `bool` call to "checking is odd") and would prefer to be explicit:

```suggestion
        if embedding_dimension % 2 != 0:
```

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,190 @@
...
+            embedding_dimension = embedding_dimension - 1
+            warnings.warn(
+                "embedding_dimension must be even. embedding_dimension will be rounded down to the nearest even number."
```
huonw: Maybe this could show the actual number as well

```suggestion
                "embedding_dimension ({embedding_dimension}) must be even. embedding_dimension will be rounded down to the nearest even number."
```

(You'll have to move it before the mutation, though.)

huonw: Also, it's up to you, but I personally have been trying to use a consistent style of `<name>: expected <description of valid>, found <invalid value>` for warnings and errors, e.g.

```suggestion
                f"embedding_dimension: expected even number, found odd number ({embedding_dimension}). It will be rounded down to {embedding_dimension - 1}."
```

This also applies to the `TypeError`s and `ValueError`s above, e.g. `raise TypeError("num_walks: expected int, found {type(num_walks)}").

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,190 @@
...
+
+        if num_walks <= 0:
+            raise ValueError("num_walks should be a positive int.")
+
+        self.num_walks = num_walks
+        self.num_powers = generator.num_powers
+        self.n_nodes = int(generator.Aadj_T.shape[0])
+
+        if embedding_dimension % 2:
+            embedding_dimension = embedding_dimension - 1
```
huonw: Minor thing but this could be:

```suggestion
            embedding_dimension -= 1
```

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,190 @@
...
+
+        left_embedding = Embedding(
+            self.n_nodes,
+            int(self.embedding_dimension / 2),
+            input_length=None,
+            name="WATCH_YOUR_STEP_LEFT_EMBEDDINGS",
+        )
+
+        vectors_left = left_embedding(input_rows)
+        outer_product = Dense(self.n_nodes, name="WATCH_YOUR_STEP_RIGHT_EMBEDDINGS")(
```
huonw: Nice 👍 

Could you add a comment about how this corresponds to the paper, because it's a significant departure from how it is phrased there (but, hopefully, not in its behaviour)?

stellargraph/layer/watch_your_step.py
```diff
@@ -0,0 +1,190 @@
...
+        self,
+        walk_length=10,
+        attention_initializer="glorot_uniform",
+        attention_regularizer=None,
+        attention_constraint=None,
+        input_dim=None,
+        **kwargs
+    ):
+
+        if (not "input_shape" in kwargs) and (not input_dim is None):
```
huonw: I think it's slightly more Pythonic to be more English-y:

```suggestion
        if "input_shape" not in kwargs and input_dim is not None:
```

demos/embeddings/watch-your-step-cora-demo.ipynb
```diff
@@ -0,0 +1,449 @@
...
+    "## Loading in the data"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data_dir = os.path.expanduser(\"~/data/cora\")\n",
```
huonw: This can now use the new `datasets` module, e.g.:

https://github.com/stellargraph/stellargraph/blob/205b6d81ccedb7be20c0d1cd8843bb4cf819a409/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb#L63-L65

demos/embeddings/watch-your-step-cora-demo.ipynb
```diff
@@ -0,0 +1,449 @@
...
+      "  \"embedding_dimension must be even. embedding_dimension will be rounded down to the nearest even number.\"\n",
+      "/Users/kieranricardo/PycharmProjects/origin-stellargaph/stellargraph/stellargraph/core/experimental.py:88: ExperimentalWarning: AttentiveWalk is experimental: lack of unit tests. It may be difficult to use and may have major changes at any time.\n",
+      "  warnings.warn(direct_msg, ExperimentalWarning)\n"
+     ]
+    }
+   ],
+   "source": [
+    "wys = WatchYourStep(\n",
+    "    generator, num_walks=80, \n",
+    "    embedding_dimension=65, \n",
```
huonw: Are you meaning to trigger the non-even warning here?

demos/embeddings/watch-your-step-cora-demo.ipynb
```diff
@@ -0,0 +1,449 @@
...
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/kieranricardo/PycharmProjects/origin-stellargaph/stellargraph/stellargraph/core/experimental.py:88: ExperimentalWarning: WatchYourStep is experimental: lack of unit tests. It may be difficult to use and may have major changes at any time.\n",
+      "  warnings.warn(direct_msg, ExperimentalWarning)\n",
+      "/Users/kieranricardo/PycharmProjects/origin-stellargaph/stellargraph/stellargraph/layer/watch_your_step.py:147: UserWarning: embedding_dimension must be even. embedding_dimension will be rounded down to the nearest even number.\n",
+      "  \"embedding_dimension must be even. embedding_dimension will be rounded down to the nearest even number.\"\n",
+      "/Users/kieranricardo/PycharmProjects/origin-stellargaph/stellargraph/stellargraph/core/experimental.py:88: ExperimentalWarning: AttentiveWalk is experimental: lack of unit tests. It may be difficult to use and may have major changes at any time.\n",
```
huonw: Ah, hm, is `AttentiveWalk` purely an implementation detail of WyS? If so, and a user should never call it directly, it might be better to not mark it `@experimental`, so that there's not this warning. If a user might call it directly, I guess it's probably ok.

demos/embeddings/watch-your-step-cora-demo.ipynb
```diff
@@ -0,0 +1,449 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import sklearn\n",
+    "from sklearn.preprocessing import OneHotEncoder\n",
+    "from sklearn.manifold import TSNE\n",
+    "\n",
+    "nodelist = list(Gnx.nodes)\n",
+    "\n",
+    "node_data = pd.read_csv(os.path.join(data_dir, \"cora.content\"), sep='\\t', header=None)\n",
```
huonw: Haven't we already read the data into `node_data` above?

```suggestion
```

demos/embeddings/watch-your-step-cora-demo.ipynb
```diff
@@ -0,0 +1,449 @@
...
+    "node_data.set_index(0, inplace=True)\n",
+    "\n",
+    "labels = node_data[node_data.columns[-1]].loc[nodelist]\n",
```
huonw: Why does this need to be `loc`'d? Aren't Gnx's nodes the same as the `node_data`s?

demos/embeddings/watch-your-step-cora-demo.ipynb
```diff
@@ -0,0 +1,449 @@
...
+    "Now we use TSNE to visualize the embeddings."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "embeddings = np.hstack(\n",
```
huonw: I wonder if there's a way that we can make this more automatic, so that the user can access the embeddings more directly without having to know that the right embeddings are a dense layer? Do you have any thoughts about how we could customise the model or the layers to achieve this?

stellargraph/layer/watch_your_step.py
huonw: Also, just to confirm, will this behave exactly like an embedding layer?

From the docs https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dense vs. https://www.tensorflow.org/api_docs/python/tf/keras/layers/Embedding it seems like there may be some tweaks needed:

- `use_bias` defaults to `True`, but I suspect we don't want a bias for this `Dense` layer (at least, `g(Y) = ...` didn't have a bias in it in the paper)
- `kernel_initializer` defaults to `glorot_uniform` for `Dense`, but `embeddings_initializer` defaults to `uniform` for `Embeddings`

Assuming that's all, maybe:

```suggestion
        outer_product = Dense(self.n_nodes, use_bias=False, kernel_initializer="uniform", name="WATCH_YOUR_STEP_RIGHT_EMBEDDINGS")(
```

demos/embeddings/watch-your-step-cora-demo.ipynb
kieranricardo: yeah good point, I don't think a user would ever call this directly

demos/embeddings/watch-your-step-cora-demo.ipynb
kieranricardo: I'm not entirely sure why, but the ordering is different in `node_data` compared to the embeddings. So without the loc to reorder `node_data` the accuracy drops to 22%

demos/embeddings/watch-your-step-cora-demo.ipynb
kieranricardo: I was, but perhaps its better to not trigger this in the demo and just mention it in the text..

demos/embeddings/watch-your-step-cora-demo.ipynb
kieranricardo: didn't realise we had this functionality, nice!

demos/embeddings/watch-your-step-cora-demo.ipynb
huonw: I would personally not demonstrate it here and only have it in the docs for `WatchYourStep` (which currently don't mention it, and so should be updated), but it's up to you: if you do decide to trigger it here, it would be good to mention explicitly in the surrounding text in addition to the docs.

demos/embeddings/watch-your-step-cora-demo.ipynb
huonw: Ah, right. I guess because the graph is created by adding edges (not explicitly adding nodes), so the order of the nodes is probably related to the order of the edges. Using the new StellarGraph class would change this.

demos/embeddings/watch-your-step-cora-demo.ipynb
huonw: New in #690.

stellargraph/layer/watch_your_step.py
kieranricardo: I've added in a comment - let me know if it needs more clarity :) 

weirdly changing to a `Dense` layer didn't speed things up. Maybe `tensorflow` had already optimized around the weird lambda layers and lookups

kieranricardo: yep true, I shouldve checked the defaults! with those defaults in place, the dense layer will just be a trainable matrix that mat muls the input of that layer (and doesn't apply any activation function)

demos/embeddings/watch-your-step-cora-demo.ipynb
kieranricardo: that's a good idea, I'll a `WatchYourStep.get_embeddings(self, model)` function that accepts a WyS keras model and returns the embeddings

huonw:

Issue #751, opened 2020-01-30
`StellarGraph.edges` should support `include_edge_weight` parameter similar to neighbourhood methods


kjun9:
### Description

As well as generally sounding like a useful feature, this is actually required for temporal random walks, since we need to apply biases across all edges in the graph using their weights (weights are used for timestamps). 

Since we already have a `include_edge_type` parameter, this means the returned list will grow from holding triples to holding quadruples, and this will probably make it more difficult for the user to figure out which element in the tuple corresponds to which attribute, so we might need to rethink the return type a bit.

### User Story

**As** a data scientist

**I want to** get the weights on all the edges in my graph

**So that** I can use them for useful things like running random walks based on weights

### Done Checklist

- [x] Produced code for required functionality
- [x] Tests written and coverage checked
- [x] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [x] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #754, opened 2020-01-31
Add include_edge_weight to StellarGraph.edges

kjun9:
Some unrelated trailing commas got removed by black, it still seems to pass our black format CI step, so not really sure what's going on there but I'll look into it further.

Given this FIXME in `StellarGraph.edges` currently:
```
        # FIXME: these would be better returned as the 2 or 3 arrays directly, rather than tuple-ing
        # (the same applies to all other instances of zip in this file)
```
I decided it might be best to just return weights as a separate list - it should have no impact on existing code, and avoids the weirdness of returning longer tuples described in #751

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -221,7 +221,7 @@ class StellarGraph:
             self._is_directed = is_directed
 
             self._nodes = convert.convert_nodes(
-                nodes, name="nodes", default_type=node_type_default, dtype=dtype
+                nodes, name="nodes", default_type=node_type_default, dtype=dtype,
             )
             self._edges = convert.convert_edges(
                 edges,
@@ -869,7 +869,9 @@ class StellarGraph:
             features = self.node_features(node_ids, node_type=ty)
 
             for node_id, node_features in zip(node_ids, features):
-                graph.add_node(node_id, **ty_dict, **{feature_name: node_features})
+                graph.add_node(
+                    node_id, **ty_dict, **{feature_name: node_features},
+                )
 
         iterator = zip(
             self._edges.sources,
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -221,7 +221,7 @@ class StellarGraph:
             self._is_directed = is_directed
 
             self._nodes = convert.convert_nodes(
-                nodes, name="nodes", default_type=node_type_default, dtype=dtype,
+                nodes, name="nodes", default_type=node_type_default, dtype=dtype
             )
             self._edges = convert.convert_edges(
                 edges,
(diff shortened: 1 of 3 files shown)


stellargraph/core/graph.py
```diff
@@ -303,19 +303,31 @@ def nodes(self) -> Iterable[Any]:
...
+            The graph edges. If edge weights are included then a tuple of (edges, weights)
         """
         if self._graph is not None:
-            return self._graph.edges(include_edge_type)
+            return self._graph.edges(
+                include_edge_weight=include_edge_weight,
+                include_edge_type=include_edge_type,
+            )
+
+        if include_edge_weight:
```
huonw: Given how you've implemented the `graph_networkx` version, I think this is possible in almost the same way, and so should be done here:

```python
if include_edge_type:
    edges = list(zip(...))
else:
    edges = list(zip(...))

if include_edge_weight:
    return edges, self._edges.weights

return edges
```

stellargraph/core/graph_networkx.py
```diff
@@ -709,16 +709,29 @@ def number_of_edges(self) -> int:
...
             )
         else:
             # returns pairs of format (node 1, node 2)
-            return self._graph.edges()
+            edges = self._graph.edges()
+
+        if include_edge_weight:
+            return (
+                edges,
+                [
```
huonw: How about returning this as a numpy array, so that we're doing #718 for all new APIs (and matching the new implementation)? Also, splitting out the computation into a separate variable may make this code slightly less explosively formatted:

```python
weights = np.array([data.get(self._edge_weight_label) for _, _, data in self._graph.edges(data=True)])
return edges, weights
```

kjun9:
It seems like I just had an older version of black, but it was a bit surprising to me that the latest version of black doesn't have an idea of a single "well formatted" version of a piece of code since either versions are passing CI (or was this always the case and is the case for all code formatters..?)

stellargraph/core/graph.py
```diff
@@ -303,32 +303,49 @@ def nodes(self) -> Iterable[Any]:
...
+            raise NotImplementedError(
```
huonw: This is implemented now.

stellargraph/core/graph_networkx.py
```diff
@@ -709,16 +709,29 @@ def number_of_edges(self) -> int:
...
                 for src, dst, data in self._graph.edges(data=True)
             )
         else:
             # returns pairs of format (node 1, node 2)
-            return self._graph.edges()
+            edges = self._graph.edges()
+
+        if include_edge_weight:
+            weights = np.array(
+                [
```
huonw: Ah, well, this formatting is still large. 🤷‍♂

tests/core/test_stellargraph.py
```diff
@@ -738,3 +738,13 @@ def test_info_heterogeneous():
...
+
+
+def test_edges_include_weights():
+    g = example_weighted_hin()
+    edges, weights = g.edges(include_edge_weight=True)
+    nxg = g.to_networkx()
+    assert len(edges) == len(weights) == len(nxg.edges())
+    for (src, dst), w in zip(edges, weights):
+        # check weight matches an edge in multi-edges
+        assert w in {data["weight"] for data in nxg.get_edge_data(src, dst).values()}
```
huonw: This set is going to be recomputed for every edge, so it's probably mildly cleaner to avoid doing that construction:

```python
assert any(data["weight"] == w for data in ...)
```

Another option would be to group by the `src, dst` pairs and compare the complete lists, because, as is, this is only verifying that it's returning a weight between the edges, not every weight between them. This could be built on pandas like `pd.DataFrame(edges, columns=["source", "target"]).assign(weight=weights).groupby(["source", "target"]).agg(list)` to get a data frame mapping the source/target pairs to list of weights, and then iterate through it or something, but that's kinda complicated.

stellargraph/core/graph.py
kjun9: Oops sorry, will remove

stellargraph/core/graph_networkx.py
kjun9: 🤷‍♂

tests/core/test_stellargraph.py
kjun9: I think I'll just do your pandas suggestion

PR #759, opened 2020-01-31
testing whether black formatting is working correctly

kjun9:


kjun9:
Running black locally:
```
$ black --version
black, version 19.3b0
```
`black --check .` passes for 46128c0896d23f29ef4a73eb9922b232a15638e6 but not for 63d4240b96fc333e816d8cf31207835bc8c71235

which seems more correct - there shouldn't be ambiguity

but on CI, currently both commits pass the check - @TheDen any thoughts on how to investigate this further?

TheDen:
Could be a version mismatch with your local one. The docker image on the CI uses `19.10b0`

```
$ docker run -it --entrypoint sh stellargraph/black:snapshot -c 'black --version'
black, version 19.10b0
```

Issue #748, opened 2020-01-30
Migrate code coverage from coveralls to codecov


timpitman:
Our coveralls statistics ( https://coveralls.io/github/stellargraph/stellargraph ) are outdated - they haven't been uploaded since June 2019.  After a good experience on Neon with codecov, we should change our CI pipeline (and github integration/badges).

PR #760, opened 2020-01-31
Migrate code coverage analysis to Codecov #748

timpitman:
migrating our coverage analysis from coveralls (which wasn't being updated since June 2019) to coveralls: #748

### DIFF
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -16,6 +16,9 @@ services:
       - BUILDKITE_BUILD_NUMBER
       - BUILDKITE_BRANCH
       - AWS_LOGS_BUCKET
+      - BUILDKITE
+      - BUILDKITE_PULL_REQUEST
+      - BUILDKITE_JOB_ID
       - CODECOV_TOKEN
 
   runner-3.7:
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -12,15 +12,10 @@ services:
     environment:
       - PYTHONDONTWRITEBYTECODE=1
       - GIT_BRANCH=$BUILDKITE_BRANCH
-      #- COVERALLS_REPO_TOKEN
       - BUILDKITE_AGENT_ACCESS_TOKEN
       - BUILDKITE_BUILD_NUMBER
       - BUILDKITE_BRANCH
       - AWS_LOGS_BUCKET
-      # for coveralls-python, to send more metadata to coveralls.io
-      - BUILDKITE
-      - BUILDKITE_PULL_REQUEST
-      - BUILDKITE_JOB_ID
       - CODECOV_TOKEN
 
   runner-3.7:
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -2,7 +2,7 @@
 version: "3.4"
 services:
   runner-3.6: &runner-config
-    image: "python:3.6-slim"
+    image: "python:3.6"
     working_dir: /workdir
     volumes:
       - $PWD:/workdir
@@ -24,4 +24,4 @@ services:
 
   runner-3.7:
     <<: *runner-config
-    image: "python:3.7-slim"
+    image: "python:3.7"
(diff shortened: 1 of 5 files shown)


.buildkite/steps/script.sh
```diff
@@ -16,10 +16,10 @@ pip freeze
 
 echo "+++ running tests"
 exitCode=$?
-py.test -ra --cov=stellargraph tests/ --doctest-modules --doctest-modules --cov-report=term-missing -p no:cacheprovider --junitxml="./${junit_file}" || exitCode=$?
+py.test -ra --cov=stellargraph tests/ --doctest-modules --cov-report=xml -p no:cacheprovider --junitxml="./${junit_file}" || exitCode=$
 
-echo "--- uploading coveralls"
-coveralls
```
huonw: I think this means we can remove `coveralls` from our dependencies.

.buildkite/steps/script.sh
```diff
@@ -16,10 +16,10 @@ pip freeze
 
 echo "+++ running tests"
 exitCode=$?
-py.test -ra --cov=stellargraph tests/ --doctest-modules --doctest-modules --cov-report=term-missing -p no:cacheprovider --junitxml="./${junit_file}" || exitCode=$?
+py.test -ra --cov=stellargraph tests/ --doctest-modules --cov-report=xml -p no:cacheprovider --junitxml="./${junit_file}" || exitCode=$
```
huonw:
```suggestion
py.test -ra --cov=stellargraph tests/ --doctest-modules --cov-report=xml -p no:cacheprovider --junitxml="./${junit_file}" || exitCode=$?
```

.buildkite/docker-compose.yml
```diff
@@ -2,7 +2,7 @@
 version: "3.4"
 services:
   runner-3.6: &runner-config
-    image: "python:3.6-slim"
+    image: "python:3.6"
```
TheDen: If we want to keep using slim we could add?

```apt-get update && apt-get install --no-install-recommends curl && rm -rf /var/lib/apt/lists/*```

File: .buildkite/steps/script.sh [hunk 4]
```diff
@@ -16,10 +16,10 @@ pip freeze
...
 
-echo "--- uploading coveralls"
-coveralls
+echo "--- :coverage::codecov::arrow_up: uploading coverage to codecov.io"
+bash <(curl https://codecov.io/bash)
```
TheDen: Do you think it's potentially unsafe pulling in a script from the internet to execute on our build machines? We could commit the script to the repo and execute it when needed?

timpitman: Codecov claim that since this is https it's no different to trusting things like npm (we already do this for neon...) . if we committed then we'd be manually updating when codecov makes changes.

.buildkite/docker-compose.yml
timpitman: With caching, I don't think there's any major advantage to the slim image?

.buildkite/steps/script.sh
huonw: I think the biggest downside here is the download sometimes seems to fail/timeout (at least it did on neon); but having to remember to manually update seems somewhat worse.

.buildkite/docker-compose.yml
TheDen: It's mainly the docker image size difference (after pulled once on an agent it should be cached):

54.93 MB for `python:3.6-slim` https://hub.docker.com/layers/python/library/python/3.6-slim/images/sha256-f0f38f594bc1babb0a09c0323bce8fba121e278661fe124c2f2b3dc69f450022

336.9 MB for `python:3.6`https://hub.docker.com/layers/python/library/python/3.6/images/sha256-5da77e7f0723cceeeb09d82093eea3a463c63d9a873648a4e8d53643027fc8e5

Running 
```bash
$ time docker run -it python:3.6-slim sh -c 'apt-get update && apt-get install -y --no-install-recommends curl && rm -rf /var/lib/apt/lists/*'
```

Running that a few times it averages at around 7 seconds, so it's whether it's faster to pull down `281.97`MBs of extra data on the instances.

Regardless, Huon is working on adding a builder docker image image in https://github.com/stellargraph/stellargraph/pull/674 so in either case it wouldn't matter

.buildkite/steps/script.sh
TheDen: Yeah makes sense. Perhaps when we solve adding https://github.com/stellargraph/stellargraph/pull/674 we can have the script pulled down in the docker image so we don't have to worry about network connections

PR #764, opened 2020-02-03
Update CHANGELOG.md for codecov #760

timpitman:
update to changelog with codecov change (this was missed in #760)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Some new algorithms and features are still under active development, and are ava
 ### Bug fixes and other changes
 
 - DevOps changes:
-  - CI: Code coverage analysis is now performed using codecov: [\#760](https://github.com/stellargraph/stellargraph/pull/760)
+  - CI: [\#760](https://github.com/stellargraph/stellargraph/pull/760)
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ Some new algorithms and features are still under active development, and are ava
 
 ### Bug fixes and other changes
 
+- DevOps changes:
+  - CI: Code coverage analysis is now performed using codecov: [\#760](https://github.com/stellargraph/stellargraph/pull/760)
```
huonw: My assumption is that most users will have little-to-no interest in DevOps things like this, and listing them is good to give a sense of velocity (and for our own reference), but I'm not sure it's worth spending the space on the specifics. If we were to list all the little changes like this, I think our release notes may end up very long and users trying to find relevant new features and bug fixes may be intimidated or overloaded. As such, maybe we could do something similar to what we did for the 0.9 release, of just linking to each one:

```suggestion
  - CI: [\#760](https://github.com/stellargraph/stellargraph/pull/760)
```

What do you think?

timpitman: you're right, it's internal and not something we need to call out to users, just link.

PR #765, opened 2020-02-03
Compute the correct adjacency matrix for undirected graphs

huonw:
A function is always truthy (e.g. `bool(lambda: None)` => `True`), so missing the method call parentheses in `self.is_directed()` meant we were treating all graphs as directed.

This adds tests for both undirected and directed graphs, and verifies the behaviour of:

- weighted edges: appear in the adjacency matrix as the weights, not 1s
- self-loops: only counted once, in undirected graphs
- repeated edges: weights are summed

See: #763

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -810,7 +810,7 @@ class StellarGraph:
         n = len(index)
 
         adj = sps.csr_matrix((weights, (src_idx, tgt_idx)), shape=(n, n))
-        if not self.is_directed:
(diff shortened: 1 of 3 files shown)


huonw:
> If it wasn't for matching the NetworkX API, I wonder if it might be a better idea to make is_directed a property?

Yeah, potentially, but as you say, there's the NetworkX compatibility thing.

> Also this is something that a linter (or MyPy?) should catch....

I would guess so.

Issue #766, opened 2020-02-03
New StellarGraph: in_nodes/out_nodes gives only directed edges for undirected graphs


huonw:
### Describe the bug

The `in_nodes` methods currently don't match the `networkx`-based implementation, where the `*_nodes` methods behave the same as `neighbours`, giving all incident edges (as in, every edge is an incoming edge, and every edge is an outgoing one).

### To Reproduce

```
import stellargraph as sg
import pandas as pd
import networkx as nx

g = nx.Graph([(0, 1)])
old = sg.StellarGraph(g)
print(old.in_nodes(0), old.out_nodes(0))
new = sg.StellarGraph(nodes=pd.DataFrame([], index=range(2)), edges=pd.DataFrame([(0, 1)], columns=["source", "target"]))
print(new.in_nodes(0), new.out_nodes(0))
```

### Observed behavior

```
[1] [1]
[] [1]
```

### Expected behavior

```
[1] [1]
[1] [1]
```

### Environment

Operating system: `Darwin-18.6.0-x86_64-i386-64bit`
Python version:
```
3.6.9 (default, Jul 10 2019, 12:25:55) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)]
```
Package versions:
<details>

```
absl-py==0.8.0
appnope==0.1.0
astor==0.8.0
attrs==19.3.0
backcall==0.1.0
bleach==3.1.0
boto==2.49.0
boto3==1.9.230
botocore==1.12.230
cachetools==4.0.0
certifi==2019.9.11
chardet==3.0.4
coverage==4.5.4
coveralls==1.10.0
cycler==0.10.0
decorator==4.4.0
defusedxml==0.6.0
docopt==0.6.2
docutils==0.15.2
entrypoints==0.3
gast==0.2.2
gensim==3.8.0
google-auth==1.10.0
google-auth-oauthlib==0.4.1
google-pasta==0.1.7
gprof2dot==2019.11.30
grpcio==1.23.0
h5py==2.10.0
idna==2.8
importlib-metadata==1.3.0
ipykernel==5.1.3
ipython==7.11.1
ipython-genutils==0.2.0
ipywidgets==7.5.1
isodate==0.6.0
jedi==0.15.2
Jinja2==2.10.3
jmespath==0.9.4
joblib==0.13.2
json5==0.8.5
jsonschema==3.2.0
jupyter==1.0.0
jupyter-client==5.3.4
[...]

PR #767, opened 2020-02-03
All neighbours are in- and out-neighbours in undirected new StellarGraphs

huonw:
The docs for `in_nodes` and `out_nodes` explicitly state

> For an undirected graph, neighbours are treated as both in-nodes and out-nodes.

but the new `StellarGraph` class did not match this.

See: #766

### DIFF
--- a/tests/core/test_stellargraph.py
+++ b/tests/core/test_stellargraph.py
@@ -608,12 +608,19 @@ def test_allocation_benchmark_creation(
 
 
 def example_weighted_hin(is_directed=True):
-    graph = nx.MultiDiGraph() if is_directed else nx.MultiGraph()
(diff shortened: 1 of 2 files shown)


tests/core/test_stellargraph.py
```diff
@@ -608,12 +608,19 @@ def f():
 
 
 def example_weighted_hin(is_directed=True):
```
huonw: The functionality is tested in `test_undirected_hin_neighbor_methods`, now that this graph has been switched over to create the new `StellarGraph` implementation:

https://github.com/stellargraph/stellargraph/blob/91d0aa5a77aaf3f92b9274ba66fe47e0f0f0d171/tests/core/test_stellargraph.py#L664-L667

Issue #774, opened 2020-02-05
New StellarGraph: creating with no nodes or no edges throws "ValueError: No objects to concatenate"


huonw:
### Describe the bug

A new StellarGraph that contains no nodes or no edges (in the form of having no node/edge types) fails to construct.

### To Reproduce

```
import stellargraph as sg
import pandas as pd
sg.StellarGraph(nodes={}, edges={})
sg.StellarGraph(nodes=pd.DataFrame(), edges={})
```

### Observed behavior

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-c3e41b02aac5> in <module>
      1 import stellargraph as sg
      2 import pandas as pd
----> 3 sg.StellarGraph(nodes={}, edges={})
      4 sg.StellarGraph(nodes=pd.DataFrame(), edges={})

~/projects/stellargraph/stellargraph-2/stellargraph/core/graph.py in __init__(self, graph, is_directed, edge_weight_label, node_type_name, edge_type_name, node_type_default, edge_type_default, feature_name, target_name, node_features, dtype, nodes, edges, source_column, target_column)
    222 
    223             self._nodes = convert.convert_nodes(
--> 224                 nodes, name="nodes", default_type=node_type_default, dtype=dtype,
    225             )
    226             self._edges = convert.convert_edges(

~/projects/stellargraph/stellargraph-2/stellargraph/core/convert.py in convert_nodes(data, name, default_type, dtype)
    120     )
    121     nodes, node_features = converter.convert(data)
--> 122     return NodeData(nodes, node_features)
    123 
    124 

~/projects/stellargraph/stellargraph-2/stellargraph/core/element_data.py in __init__(self, shared, features)
    209 
    210     def __init__(self, shared, features):
--> 211         super().__init__(shared)
    212         if not isinstance(features, dict):
    213             raise TypeError(f"features: expected dict, found {type(features)}")

~/projects/stellargraph/stellargraph-2/stellargraph/core/element_data.py in __init__(self, shared)
    135             type_dfs.append(type_data)
    136 
--> 137         all_columns = pd.concat(type_dfs)
    138         self._id_index = ExternalIdIndex(all_columns.index)
    139         self._columns = {

~/.pyenv/versions/sg2/lib/python3.6/site-packages/pandas/core/reshape/concat.py in concat(objs, axis, join, join_axes, ignore_index, keys, levels, names, verify_integrity, sort, copy)
[...]

PR #776, opened 2020-02-05
Handle no nodes and/or no edges in new StellarGraph

huonw:
If there's no node types or no edge types, the code ends up calling `pd.concat([])` which is an error, so this has to be made conditional and appropriately-shaped dataframe created manually.

Now that it works, this also switches the default empty `StellarGraph()` constructor call to use the new `StellarGraph` implementation.

See: #774


stellargraph/core/element_data.py
```diff
@@ -107,6 +107,8 @@ class ElementData:
         shared (dict of type name to pandas DataFrame): information for the elements of each type
     """
 
+    _SHARED_REQUIRED_COLUMNS = []
```
timpitman: I know this is called "shared" but possibly an extra comment to say that it's sometimes overridden in derived classes?

kieranricardo:
I think you meant "Handle no nodes and/or no edges in new StellarGraph" in the title? :)

huonw:

huonw:
New PR: https://github.com/stellargraph/stellargraph/pull/790

PR #779, opened 2020-02-05
Use float32 for new StellarGraph default edge weights

huonw:
This requires less memory than numpy's default `np.int64` for an integer like `1` (or `np.float64` for a float like `1.0`), and is generally a better type to be using with tensorflow.

This is confirmed by the benchmarks, e.g. the `test_allocation_benchmark_creation[None-1000-5000]` benchmark goes from 747k on `develop` to 725k with this PR, a reduction of 22k. This is close to the theoretical reduction of 5000 edges &times; 4 bytes/edge = 20k.


kieranricardo:
@huonw just to clarify, for graphs without edge weights this will set the edge weights to `np.float32(1)` but won't type caste when the edge weights are specified?

huonw:
It's a bit subtle:

- it'll set edges weights to `np.float32(1)` for types that don't have a weight specified in the dataframe (since each type is specified as its own dataframe)
- all of the dataframes get `pd.concat`'d into one long dataframe, which will then promote the types to the best super type
  https://github.com/stellargraph/stellargraph/blob/6da771e8d256bc47edadb7905064238dbdd2b569/stellargraph/core/element_data.py#L137
- this may promote some types, but should never lose precision accidentally. The following code snippet demonstrates what happens when concatenating a `float32` column (e.g. the default weight column) with a `float64` column (e.g. an explicit weight column):

  ```python
  f32 = pd.DataFrame(index=[0]).assign(a=np.float32(1))
  f64 = pd.DataFrame(index=[0]).assign(a=2.0)
  f32.dtypes
  # a    float32
  f64.dtypes
  # a    float64
  pd.concat((f32, f64)).dtypes
  # a    float64
  ```

huonw:

huonw:
New PR: #789

Issue #780, opened 2020-02-05
Test start-up is slow (4s) and could be faster


huonw:
### Describe the bug

The time required to collect `stellargraph`s tests is quite slow (5s), which makes a development loop like "edit code" -> "run one fast test" -> "edit code" -> ... slower than it needs to be.

### To Reproduce

1. Run `time pytest -k test_edge_weights_directed --durations=0` which will collect all the tests and run only `test_edge_weights_directed`, and then print how long it took

### Observed behavior

```
$ time pytest -k test_edge_weights_directed --durations=0
============================ test session starts =============================
platform darwin -- Python 3.7.3, pytest-5.3.1, py-1.8.1, pluggy-0.13.1
benchmark: 3.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/wil9dw/projects/stellargraph/stellargraph-2, inifile: pytest.ini
plugins: benchmark-3.2.3, cov-2.8.1
collected 414 items / 413 deselected / 1 selected                            

tests/core/test_stellargraph.py .                                      [100%]

============================== warnings summary ==============================
/Users/wil9dw/.pyenv/versions/3.7.3/envs/sg2-3_7/lib/python3.7/site-packages/tensorflow_core/python/pywrap_tensorflow_internal.py:15
  /Users/wil9dw/.pyenv/versions/3.7.3/envs/sg2-3_7/lib/python3.7/site-packages/tensorflow_core/python/pywrap_tensorflow_internal.py:15: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/latest/warnings.html
=========================== slowest test durations ===========================
0.01s call     tests/core/test_stellargraph.py::test_edge_weights_directed

(0.00 durations hidden.  Use -vv to show these durations.)
================ 1 passed, 413 deselected, 1 warning in 3.19s ================

real	0m3.893s
user	0m4.414s
sys	0m1.047s
```

### Expected behavior

The time to collect the tests in order to run one 0.01s long test should be much less (i.e. there's currently 319-389&times; overhead for running that one test).

### Environment

Operating system: `Darwin-18.6.0-x86_64-i386-64bit`
Python version:
```
3.7.3 (default, Jul 10 2019, 09:35:06) 
[...]

PR #781, opened 2020-02-05
Use fixtures in test_edge_splitter to cut 0.5s off pytest start-up

huonw:
Reducing pytest's start-up time smoothes the developer experience of editing some code and then running a subset of tests (especially a subset that runs quickly).

Before this patch, the Cora dataset was being loaded and parsed into a graph during pytest start-up. This patch uses `@pytest.fixture` to make that lazy, and so it only happen when the edge-splitter tests are actually executed.

Times for the test from #780 (`time pytest -k test_edge_weights_directed`):

| code | real time (from `time`) | internal `pytest` time |
|---|--:|--:|
|this PR|3.2|2.5|
|develop|3.7|3.0|

(A large chunk of the changes are just reformatting after removing the `self.` prefix from all instances of `self.es_obj` and `self.g`.)

This seems to be the largest low-hanging fruit in terms of test start-up time, as more than half the remaining start-up time is the time to import `tensorflow`:

```
$ time python -c 'import tensorflow'

real	0m1.778s
user	0m2.645s
sys	0m0.731s
```

See: #780


kieranricardo:
LGTM. 

fyi if you're looking to speed up running a particular subset of tests specifying the file will cause `pytest` to collect tests from only that file all, reducing start up time.

e.g. `time pytest tests/core/test_stellargraph.py -k test_edge_weights_directed` produced the same results for develop and this branch.

huonw:

Issue #715, opened 2020-01-22
New StellarGraph: convert from networkx


huonw:
### Description

In #662 #668, we add a new implementation of the `StellarGraph` class. This is in parallel to the `NetworkXStellarGraph`. We should:

- implement a conversion from the networkx graphs to the more efficient ones
- remove `NetworkXStellarGraph` to not have to maintain it
- consider deprecating (with a warning) `StellarGraph(something_networkx)` and replacing it with a `StellarGraph.from_networkx` static method or `stellargraph.from_networkx(...)` top level function
- consider making `networkx` an optional dependency, e.g. `to_networkx` could do the `import networkx` call internally, so that the dependency isn't required to use the rest of `stellargraph` (this import will fail if the module isn't installed, but we can catch the failure and emit a better error message)

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #782, opened 2020-02-05
Use columnar edge weights for BiasedRandomWalk

huonw:
WIP; probably blocked on #715 to be able remove the networkx implementation.

Part of: #718


huonw:

Issue #500, opened 2019-10-10
Interpretability of saliency maps: scale node importance scores and edge importance scores  to be on the same scale 


annitrolla:
### Description
Current implementation of gcn saliency maps outputs node importance scores and edge importance scores on different scales, that makes it difficult to compare. It would be very useful from a practical point of view to find a way to have values on a same scale. 
### User Story
As a: data scientist
I want: to compare importance scores of edge and node importance
so that: I will be able to answer a question: "what contributes to the prediction more: features of a node or a structure (e.g information from the neighborhood)"

### Done Checklist (Development)
- [ ] Assumptions of the user story met
- [ ] Produced code for required functionality
- [ ] Branch and Pull Request build on CI
- [ ] Branch and Pull Request pass unit tests on CI
- [ ] Branch and Pull Request pass integration tests on CI
- [ ] Version number reflects new status
- [ ] Peer Code Review Performed
- [ ] Code well commented
- [ ] Documentation on Google Docs
- [ ] Documentation in repo
- [ ] CHANGELOG.md updated

dubbsbrandon:
I'd also really like to have this implemented if possible.

kjun9:
So I think this is currently a problem because of the way we've defined "node importance"

```
The importance of the node is defined as the sum of all the feature importance of the node.
```

We need to allow for different ways to aggregate the set of feature importances of a node to produce a "node importance" so that the resulting value becomes comparable to the link importance value. 

It's not clear to me exactly which aggregation method would be "correct" - I think `mean` or `max` could make sense? If you have time, I'm curious if you have any other thoughts on this @annitrolla @youph since you're both probably more familiar with this than I am!

kjun9:
Or maybe taking the magnitude of the feature importances vector makes sense too?

kjun9:
Okay I've thought a bit more about this:

We've defined node importance as the sum of the individual feature importances of a node. Feature importance is the partial derivative of the prediction output w.r.t. a particular feature value. So then node importance is basically how the output moves if you nudge all of the features at once. 

For instance, if you have `feature_importances=[2, 1, -1, 1]`, if you nudge the first feature by `0.1` you'd expect the output to move up by roughly `0.1 * 2 = 0.2`. If you nudge all features by `0.1` you'd expect the output to move by roughly `0.1 * sum([2, 1, -1, 1]) = 0.3`. This is our definition of node importance currently:
```
node_importance = sum(feature_importances)
```

[...]

PR #784, opened 2020-02-05
Normalise node importance score

kjun9:
Node importance was defined as the sum of node feature importances,
making it not directly comparable to link importance values. This
normalises the sum by the number of features to resolve this issue.

See comments on #500 

I've verified on one notebook `node-link-importancer-demo-gcn.ipynb` that this does in fact give you node importance values that looks a bit more comparable to link importance ones:
image
Previously, the node importance was at least an order of magnitude larger than that.

I'll run and commit all the other interpretability notebooks once the approach has been reviewed


kjun9:
New PR due to git problem: https://github.com/stellargraph/stellargraph/pull/785

huonw:

PR #785, opened 2020-02-05
Normalise node importance score

kjun9:
Node importance was defined as the sum of node feature importances,
making it not directly comparable to link importance values. This
normalises the sum by the number of features to resolve this issue.

See comments in #500 

I've verified on one notebook node-link-importancer-demo-gcn.ipynb that this does in fact give you node importance values that looks a bit more comparable to link importance ones:
image
Previously, the node importance was at least an order of magnitude larger than that.

I'll run and commit all the other interpretability notebooks once the approach has been reviewed


kjun9:
Closing this, since it looks like this is incorrect - see discussion in #500

Slightly different importance value scaling issue will be looked at instead #831

PR #787, opened 2020-02-05
Add temporal random walk

kjun9:
Reference Paper: http://ryanrossi.com/pubs/nguyen-et-al-WWW18-BigNet.pdf

Marked as experimental since this currently lacks more thorough testing, documentation, and demo


stellargraph/data/explorer.py
```diff
@@ -780,3 +783,166 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
             self._raise_error(
                 "The number of hops for the in and out neighbourhoods must be the same."
             )
+
+
+@experimental(reason="requires more thorough testing and documentation")
+class TemporalRandomWalk(GraphWalk):
+    """
+    Performs uniform temporal random walks on the given graph
```
PantelisElinas: Should this explain that a "given graph" should have edge weights that correspond to timestamps and the format and interpretation of said timestamps? Are these expected to be seconds, minutes, hours, or another format, e.g., HH:MM:SS. 

If such an explanation is available elsewhere, then ignore this comment.

kjun9: Good point, currently we just expect edge weights to be numbers that represent time rather than some kind of timestamp string. I think the exact units (minutes or hours, etc) don't actually matter here, as long as all the time values can be compared relative to each other. I'll add a comment

stellargraph/data/explorer.py
```diff
@@ -780,3 +783,166 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+            initial_edge_bias (str, optional): distribution to use when choosing a random
+                initial temporal edge to start from. Available options are:
+                * None (default) - the initial edge is picked from a uniform distribution
+                * "exponential" - heavily biased towards more recent edges.
+            walk_bias (str, optional): distribution to use when choosing a random
+                neighbour to walk through. Available options are:
+                * None (default) - neighbours are picked from a uniform distribution
+                * "exponential" - exponentially decaying probability, resulting in a bias
+                    towards shorter time gaps
+            p_walk_success_threshold (float): Lower bound for the proportion of successful
```
PantelisElinas: The purpose of this parameter is not clear. If you are trying to avoid infinite loops (I suspect) because you cannot generate enough data for a given graph and random walk settings, then we need to explain more clearly how this parameter controls this behaviour.

Alternatively, we can display a progress bar (since we know how much data we need to generate, calculating the percent completed should be straightforward) and let the user decide if they want to kill the process if the progress bar is not moving. We can indicate this 'infinite loop' behaviour can happen in the class docstring.

stellargraph/data/explorer.py
```diff
@@ -780,3 +783,166 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+                src, dst, t, min(max_walk_length, remaining_length), walk_bias, np_rs
+            )
+            if len(walk) >= cw_size:
+                walks.append(walk)
+                num_cw_curr += len(walk) - cw_size + 1
+                successes += 1
+            else:
+                failures += 1
+                if not_progressing_enough():
+                    raise RuntimeError(
```
PantelisElinas: I would suggest that this should be a warning message rather than raising an exception. We should let the user know that his choices are causing the algorithm to run for too long. Too many failures does not necessarily mean that if one waits long enough the walker will not terminate with enough data to continue training the model. We should allow users to keep going if they so wish as we cannot predict the kind of datasets and the special circumstances people are working with.

stellargraph/data/explorer.py
```diff
@@ -780,3 +783,166 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
...
+                        f"Discarded {failures} walks out of {failures + successes}. "
+                        "Too many temporal walks are being discarded for being too short. "
+                        f"Consider using a smaller context window size (currently cw_size={cw_size})."
+                    )
+
+        return walks
+
+    def _sample(self, n, biases, np_rs):
```
PantelisElinas: Shall we have docstrings describing the parameters as for all these "private" methods? It might be useful for maintaining and extending the code in the future.

stellargraph/data/explorer.py
kjun9: > We should allow users to keep going if they so wish as we cannot predict the kind of datasets and the special circumstances people are working with.

This is the reasoning behind the parameter that you've asked about in your [other comment](https://github.com/stellargraph/stellargraph/pull/787/files#r376828022) (`p_walk_success_threshold`). I can make the documentation more clear on that parameter, but I don't really see the value in turning turn this error into a warning instead. The parameter exists so that if they see this error, and want to ignore it, they can reduce that value to however low they want to go (even zero I guess if they really want an infinite loop). I think it's *much* more likely that when someone hits this error, it's indicative of something actually wrong (wrong setup, wrong parameters) that should be fixed.

stellargraph/data/explorer.py
kjun9: Sure I'll add some docstrings

stellargraph/data/explorer.py
huonw: The current threshold implies that this will only fail if less 1% of the attempted random walks are successful. This is effectively a 100&times; slowdown, and, as Kevin says, probably indicates some deeper problem. Prior art in this space: `sklearn` models only emit a `ConvergenceWarning` warning, not raise an error, if fitting model fails to converge after a (user-specified) number of iterations (the `max_iter` parameter), but it does not go into an infinite/very slow loop, and instead returns the partially-trained model.

I think it's quite bad to go into an infinite loop, for anything other than interactive experimentation. If a process isn't being actively monitored (e.g. batch processing on a server/cluster), a infinite loop could translate into hundreds or thousands of dollars wasted on useless compute, along with whatever the consequences are of not finishing whatever task is being attempted; at least with a hard error, it's flagged in an obvious manner.

An approach more similar to the `sklearn` approach that still avoids catastrophic failures for misconfiguration (or, at least, very poor configuration) would be to emit a warning but still return as many walks as were performed, e.g.:

```python
def run(...):
    ...
    while num_cw_curr < num_cw:
        ...
       if not_progressing_enough():
            warnings.warn(...)
            # there's been too many non-productive walks, so give up and go with what's been found up to this point
            break
    return walks
```
Or, something that might be more intuitive than estimating the probability of a walk working would be to put a hard cap on the number of walks that are performed:

```python
def run(
    num_cw,
    ...
    maximum_walks_multiplier=100 # do at most 100*num_cw random walks
):
    ...
    maximum_number_of_walks = num_cw * maximum_walks_multiplier

    while num_cw_curr < num_cw:
        ...

        if success + failures >= maximum_number_of_walks:
            warnings.warn(...)
            # there's been too many non-productive walks, so give up and go with what's been found up to this point
            break

    return walks
```

stellargraph/data/explorer.py
```diff
@@ -780,3 +783,166 @@ def _check_neighbourhood_sizes(self, in_size, out_size):
             self._raise_error(
                 "The number of hops for the in and out neighbourhoods must be the same."
             )
+
+
+@experimental(reason="requires more thorough testing and documentation")
```
huonw: Do we have any issues for these tasks? If not, could we open one or more and then

```suggestion
@experimental(reason="requires more thorough testing and documentation", issues=[...issue number here...])
```

kjun9: Thanks, I've created #827 #828 #832

stellargraph/data/explorer.py
kjun9: I've added a bit more description

stellargraph/data/explorer.py
kjun9: > Prior art in this space: sklearn models only emit a ConvergenceWarning warning, not raise an error, if fitting model fails to converge after a (user-specified) number of iterations (the max_iter parameter), but it does not go into an infinite/very slow loop, and instead returns the partially-trained model.

> An approach more similar to the sklearn approach that still avoids catastrophic failures for misconfiguration (or, at least, very poor configuration) would be to emit a warning but still return as many walks as were performed,

Hmmmm, I think I'm still a bit unconvinced. I'd think it's currently a good thing that this method always returns `num_cw` number of context windows, especially given that the current use-case it's implemented for is generating context pairs as training examples. Having to catch cases where you end up with fewer training example seems like a hassle, and not really something the user would intentionally want to do..?

> Or, something that might be more intuitive than estimating the probability of a walk working would be to put a hard cap on the number of walks that are performed

Hmm yeah could be reasonable, but then we kind of lose the early stopping (e.g. if the graph is large and so `num_cw` is large, it could still run for ages without collecting a single successful walk), so I'm not sure if it's definitely an improvement 🤷‍♂

kjun9:
@PantelisElinas 

> I'm assuming that you are going to write more unit tests or is that it?

I've basically just written a happy-case to ensure the "temporal"-ness behaves as expected, and a separate test specifically to check that the break does avoid infinite loops. I thought this might be enough to land as `@experimental` for now?

> I noticed that we never check anywhere that the graph is suitable for performing temporal random walks.
As above, I was thinking of adding these kind of checks when I add tests to cover them with follow-up PRs.

Issue #786, opened 2020-02-05
All branches in the repository were force pushed to write their entire history


huonw:
There was an accidental force push of all branches in the repository, which rewrote the ~entire history of each branch.

For example, https://github.com/stellargraph/stellargraph/pull/674 was closed and lists:

![image](https://user-images.githubusercontent.com/1203825/73814590-9fd29280-4837-11ea-93c7-392f6c37f474.png)

The end of the PR says:

![image](https://user-images.githubusercontent.com/1203825/73814929-8a119d00-4838-11ea-9bdb-537e026db68a.png)

## Closed PRs

To track the work that was in-progress and may be lost, here's the table of the PRs that were closed and the corresponding reopened one:

| closed | reopened |
|---|---|
| #552 | #787 |
| #674 | #788 |
| #720 | #795 |
| #724 | #793 |
| #757 | #792 |
| #776 | #790 |
| #779 | #789 |
| #781 | #791 |
| #782 | #794 |
| #784 | #785 |
| #214 | #796 |
| #515 | #798 |
| #705 | #806 |
| #750 | #797 |
| #773 | #800 |
| #783 | #799  |

huonw:
The branches have been rewritten back to what they were originally, e.g. https://github.com/stellargraph/stellargraph/tree/feature/separate-builder-ci again says e5b425b is it's HEAD:

![image](https://user-images.githubusercontent.com/1203825/73815062-f2607e80-4838-11ea-9414-714f017ca455.png)

However, the PRs have been disconnected from their branches, and so we may have to manually open new PRs for each one.

PR #788, opened 2020-02-05
Have an explicitly cached docker image for running on CI

huonw:
This uses the Buildkite docker-compose plugin to eliminate the 1-2 minutes of "install dependencies" required to run tests and demos. It does this by prebuilding a docker image that contains the dependencies required for this, in a way that benefits from caching, and so only has to be recomputed if the dependencies themselves change (approximated by whether `setup.py` changes).

The image is called [`stellargraph-ci-runner`](https://hub.docker.com/repository/docker/stellargraph/stellargraph-ci-runner) and is public. It's quite large (605 MB compressed), but changes rarely, and so is likely to end up already cached on our build agents, and thus the time to pull/push is minimal (just the time to check that the layers already exist).

Because we test against both 3.6 and 3.7, this pushes an image for each of them under `3_6-<normal tag>` and `3_7-<normal tag>` (examples: `3_6-PR-788`, `3_7-latest`). Because this is just a cache image, the image is pushed under that final tag as soon as it is built (that is, the prebuild step includes `push: ...`), rather than done as part of the final publish after all testing.

This makes the individual steps noticeably faster (about 90s), e.g.

| step | [develop](https://buildkite.com/stellar/stellargraph-public/builds/1439) (m:ss) | [this PR](https://buildkite.com/stellar/stellargraph-public/builds/1447) (m:ss) |
|---|--:|--:|
| test 3.6 | 8:19 | 6:42 |
| test 3.7 | 8:00 | 6:35  |
| `calibration-pubmed-link-prediction.ipynb` | 4:10 | 2:58 |
| `attri2vec-citeseer-node-classification-example.ipynb` | 5:02 | 3:32 |
| `sgc-node-classification-example.ipynb` | 2:52  | 1:26 |

This doesn't make the end-to-end build time that much faster with a build with no waiting: the prebuild step takes about 40s-65s (~20s of setup, 20s of pulling the original image if the agent doesn't have it already, 5s of building the image from cache, 20s of pushing the unchanged image twice), and only then can the downstream steps start, and that's similar to the ~90s required to install dependencies before testing starts.

On the other hand, it does dramatically reduce the agent time required, per build:

- t2medium: 2&times;30 seconds = 1 minute, and
- t2large: 24 (number of notebooks that are actually tested)&times;90 seconds = 36 minutes

This likely also means that there's less agent waiting and so builds are faster in practice (because agents are spending less time working, and able to continue onto the next job sooner).

There's still some future work building on this, that will be done as separate PRs since they're not part of this core change:

- install `papermill` into the CI image (cuts another ~15s off the installation time for notebooks, i.e. another 6 minutes of t2large agent time saved) (#878)
- have the notebook formatting check run as a separate parallel step, outside the 3.6 testing one (#877)

(Reopening of #674 due to #786.)


.buildkite/pipeline.yml
```diff
@@ -9,39 +9,78 @@ plugins: &plugins
...
+      <<: *plugins
+      docker-compose#v3.1.0:
+        <<: *compose-config
+        build: runner-3_6
+        image-repository: index.docker.io/stellargraph/stellargraph-ci-runner
+        no-cache: ${REFRESH_BUILD-false}
+        cache-from:
+          - runner-3_6:index.docker.io/stellargraph/stellargraph-ci-runner:3_6-${IMAGE_TAG}
+          - runner-3_6:index.docker.io/stellargraph/stellargraph-ci-runner:3_6-latest
+
```
TheDen: I assume this whitespace is intentional for readability?

huonw: Yeah, it's separating the two separate sub-steps; above is `build`, and below is `push`.

.buildkite/steps/test-demo-notebooks.sh
```diff
@@ -39,8 +39,9 @@ echo "--- :python: installing papermill"
 # Pulling in https://github.com/nteract/papermill/pull/459 for --execution-timeout, which hasn't been released yet
 pip install https://github.com/nteract/papermill/archive/master.tar.gz
 
-echo "--- installing dependencies"
-pip install -q --no-cache-dir '.[demos]'
+echo "--- :python: installing stellargraph"
+# install stellargraph itself, which (hopefully) won't install any dependencies
+pip install .
```
timpitman: I was curious why `pip install .` is in this script, and not in `.buildkite/steps/script.sh` unless I missed it?

huonw: From slack (as referenced in your approval comment):

> I think: the notebooks run from a subdirectory, so `import stellargraph` in them cannot pick up the `stellargraph/__init__.py`, but tests run at the top level of the repo and so can.

PR #790, opened 2020-02-05
Handle no nodes and/or no edges in new StellarGraph

huonw:
If there's no node types or no edge types, the code ends up calling `pd.concat([])` which is an error, so this has to be made conditional and appropriately-shaped dataframe created manually.

Now that it works, this also switches the default empty `StellarGraph()` constructor call to use the new `StellarGraph` implementation.

See: #774

(Reopening of #776 due to #786.)

### DIFF
--- a/stellargraph/core/element_data.py
+++ b/stellargraph/core/element_data.py
@@ -107,6 +107,8 @@ class ElementData:
         shared (dict of type name to pandas DataFrame): information for the elements of each type
     """
 
+    _SHARED_REQUIRED_COLUMNS = []
+
     def __init__(self, shared):
         if not isinstance(shared, dict):
             raise TypeError(f"shared: expected dict, found {type(shared)}")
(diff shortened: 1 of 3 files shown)


stellargraph/core/element_data.py
```diff
@@ -107,6 +107,8 @@ class ElementData:
         shared (dict of type name to pandas DataFrame): information for the elements of each type
     """
 
+    _SHARED_REQUIRED_COLUMNS = []
```
huonw: On #776, @timpitman said:

> I know this is called "shared" but possibly an extra comment to say that it's sometimes overridden in derived classes?

The "shared" here is referring to the `shared` argument to `__init__`. Or is this not what you're getting at?

timpitman: I was referring to `_SHARED_REQUIRED_COLUMNS`, suggesting a comment that it may be overridden in derived classes

huonw: Right, that's what the comment is attached to. I'm more asking about your quoting of "shared" in the original comment: that word isn't referring to being shared among subclasses, but the argument to `__init__` that's called `shared`.

I'll add a comment, but I personally generally don't explain things like this when no-one should need to think about it outside a single file (that is, no other part of the library needs to know about this variable and its functionality, and certainly no external users: they should barely even know that `ElementData` and its subclasses exist).

huonw:
It doesn't need it. The features are passed as a separate argument, and the node ID is the pandas index, not a column.

PR #794, opened 2020-02-05
Use columnar edge weights for BiasedRandomWalk

huonw:
WIP; probably blocked on #715 to be able remove the networkx implementation.

Part of: #718

(Reopening of #782 due to #786.)


huonw:
This was done in #1596.

PR #214, opened 2018-09-06
Issue#186

youph:
Implemented test() method in `demos/link-prediction_hinsage/movielens-example.py`

#185 and #186

Issue #186, opened 2018-08-22
Movielens example: test() method should be implemented


youph:
### Description
The updated `demos/link-prediction_hinsage/movielens-example.py` have no test() method implemented - it should load the saved model/transforms from disk, and apply them to the graph. These should be written, to show users how to load and apply saved models.

### Done Checklist (Development)
- [x] Produced code for required functionality
- [ ] Branch and Pull Request build on CI
- [ ] Branch and Pull Request pass unit tests on CI
- [ ] Branch and Pull Request pass integration tests on CI
- [ ] Peer Code Review Performed
- [ ] Code well commented

Issue #185, opened 2018-08-22
Write the test() functions with model/transform re-loading


youph:
### Description
The updated `demos/node-classification/epgm-example.py` and `demos/node-classification/epgm-yelp-example.py` have no test() implemented - it should load the saved model/transforms from disk, and apply them to the graph. These should be written, to show users how to load and apply saved models.

### Done Checklist (Development)
- [ ] Produced code for required functionality
- [ ] Branch and Pull Request build on CI
- [ ] Branch and Pull Request pass unit tests on CI
- [ ] Branch and Pull Request pass integration tests on CI
- [ ] Peer Code Review Performed
- [ ] Code well commented

huonw:
With https://github.com/stellargraph/stellargraph/pull/889, we no longer have any of the above scripts, so moving this to done.

PR #796, opened 2020-02-05
Issue#186

huonw:
Implemented test() method in `demos/link-prediction_hinsage/movielens-example.py`

#185 and #186

(Reopening of #214 due to #786.)


huonw:
This script was removed in https://github.com/stellargraph/stellargraph/pull/889.

PR #783, opened 2020-02-05
Feature/neo4j subsampling

chgswin:
Neo4J subsampling:

Done:
- Craft Directed & Undirected GraphSAGE Node Generators.
- Eliminate the need to store the edges in NetworkX object.

To-dos:
- Fix build-image problem.
- Cache the query (no need to re-build the query).
- Create unit-tests.

PR #799, opened 2020-02-05
Feature/neo4j subsampling

chgswin:
Neo4J subsampling for undirected and directed GraphSAGE:

Done:
- Craft Directed & Undirected GraphSAGE Node Generators + demo notebooks for each on cora datasets.
- Eliminate the need to store the edges in NetworkX object.
- Allow one single head nodes to be sampled n times.
- Improve the Undirected GraphSAGE queries.
- Add demo on loading Cora dataset into Neo4J.

To-dos:
- Cache the query (no need to re-build the query).
- Create unit-tests.
* Note: Reopen of #783 because of #786


huonw:
I'm on leave tomorrow and won't have time to look at this in great detail, but @kjun9 kindly said he could, so he'll start the review.

stellargraph/connector/neo4j/sampler.py
```diff
@@ -0,0 +1,262 @@
...
+                # add the list of node ids into subgraph
+                "[apoc.coll.flatten(collect(id(head_node)))] AS subgraph",
+            )
+        )
+
+        for in_num, out_num in zip(in_samples, out_samples):
+            next_hop_query = "\n".join(
+                (
+                    # unpack cur_depth_list which is a list of list of nodes, into single list of nodes
+                    "UNWIND cur_depth_list AS cur_node_list",
```
kjun9: Can we write these big queries as one big string instead? like:
```python
next_hop_query = """
UNWIND cur_depth_list AS cur_node_list
CALL apoc.cypher.run('
...
"""
```
Although I guess that probably means we can't put comments on specific lines. I feel like it's a bit difficult to read at the moment.

kjun9: Also, do you think it's possible to take some parts of the query out into functions so we can reuse them in the two walker classes?

chgswin: Right, I think we could definitely turn this into a long string `f"""UNWIND {some_variables}"""` with parameters. 
I will look seriously into whether we could redesign the query for reusing purposes!

demos/connector/neo4j/undirected-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -0,0 +1,1117 @@
+{
```
kjun9: <p>Maybe the title should say something about this being an example using Neo4j? For example:</p><pre class="ql-syntax" spellcheck="false">GraphSAGE on the CORA citation network using StellarGraph and Neo4j
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/799/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

kjun9: <p>We now have a <code>datasets</code> module that we can use in our notebooks for downloading datasets like cora:</p><pre class="ql-syntax" spellcheck="false">from stellargraph import datasets
from IPython.display import display, HTML

dataset = datasets.Cora()
display(HTML(dataset.description))
dataset.download()
</pre><p>You can have a look at this notebook for reference: <a href="https://github.com/stellargraph/stellargraph/blob/develop/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb" target="_blank">https://github.com/stellargraph/stellargraph/blob/develop/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb</a></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/799/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

kjun9: <p>I think we need more explicit descriptions/explanation before this cell to talk about how to set up a neo4j instance to run this notebook and how to adjust the parameters you're passing into the <code>get_node_data_from_neo4j</code> method depending on your setup. It's probably good to assume that the user has installed stellargraph and knows nothing else, but is interested to learn how to run this notebook. You could also include some helpful links to external documentation instead of having to explain everything since it might be a lot to explain</p><p> </p><p>Also minor thing, but the print happens <em>after</em> all the vertices have been loaded, so could be better phrased as "loaded" instead of "loading"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/799/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

kjun9: <p>I think this looks a bit simpler with something like this:</p><pre class="ql-syntax" spellcheck="false"># flatten out the features
features_list = node_data["attr"].apply(pd.Series)["features"].values.tolist()

# create dataframe
node_features = pd.DataFrame(features_list, index=node_data["id"])
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/799/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

kjun9: <p>Instead of creating a <code>py2neo.Graph</code>  for the second time here, can we create it once at the top and reuse it?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/799/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

chgswin: Yes, totally agree! I will fix this problem soon!

kjun9: Some of this could be done in a README file instead too

stellargraph/connector/neo4j/mapper.py
```diff
@@ -0,0 +1,251 @@
...
+    Args:
+        G (StellarGraph): The machine-learning ready graph.
+        batch_size (int): Size of batch to return.
+        neo4j_graphdb (py2neo.Graph): the Neo4J Graph Database object
+        schema (GraphSchema): [Optional] Schema for the graph, for heterogeneous graphs.
+    """
+
+    def __init__(self, G, batch_size, neo4j_graphdb, schema=None):
+
+        import py2neo
```
huonw: `py2neo` isn't used inside this `__init__`, so it doesn't _strictly_ need to be imported. Are you importing it so that the user gets an error about needing `py2neo` as soon as possible? If you are, it would be good to have a comment here, or even intercept the error to improve the message:

```python
# check the py2neo module is installed (if it isn't, this code is unlikely to work)
try:
    import py2neo
except ModuleNotFoundError as e:
    raise ModuleNotFoundError(f"{e.msg}. StellarGraph can only connect to Neo4j using the 'py2neo' module; please install it", name=e.name, path=e.path) from None
```

<details><summary>What's the `from None` doing? (click for details)</summary>

It's making sure the user only sees one `ModuleNotFoundError` by suppressing the "context" of the first exception being caught. See https://docs.python.org/3/library/exceptions.html

No `from None`:

```python
try:
    import py2neo
except ModuleNotFoundError as e:
    raise ModuleNotFoundError(f"{e.msg}. StellarGraph can only connect to Neo4j using the 'py2neo' module; please install it", name=e.name, path=e.path)
```
```
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-20-56fdd076cde9> in <module>
      1 try:
----> 2     import py2neo
      3 except ModuleNotFoundError as e:

ModuleNotFoundError: No module named 'py2neo'

During handling of the above exception, another exception occurred:

ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-20-56fdd076cde9> in <module>
      2     import py2neo
      3 except ModuleNotFoundError as e:
----> 4     raise ModuleNotFoundError(f"{e.msg}. StellarGraph can only connect to Neo4j using the 'py2neo' module; please install it", name=e.name, path=e.path)

ModuleNotFoundError: No module named 'py2neo'. StellarGraph can only connect to Neo4j using the 'py2neo' module; please install it
```

With `from None`:

```python
try:
    import py2neo
except ModuleNotFoundError as e:
    raise ModuleNotFoundError(f"{e.msg}. StellarGraph can only connect to Neo4j using the 'py2neo' module; please install it", name=e.name, path=e.path) from None
```

```
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-21-cbbd8be25d60> in <module>
      2     import py2neo
      3 except ModuleNotFoundError as e:
----> 4     raise ModuleNotFoundError(f"{e.msg}. StellarGraph can only connect to Neo4j using the 'py2neo' module; please install it", name=e.name, path=e.path) from None

ModuleNotFoundError: No module named 'py2neo'. StellarGraph can only connect to Neo4j using the 'py2neo' module; please install it
```

</details>


---

We should also consider checking the type of the `neo4j_graphdb` argument, e.g.:

```python
if not isinstance(neo4j_graphdb, py2neo.Graph):
    raise TypeError(f"neo4j_graphdb: expected py2neo.Graph, found {type(neo4j_graphdb)}")

File: stellargraph/connector/neo4j/sampler.py (hunk 1 above)
Comment (huonw) [reply stellargraph/connector/neo4j/sampler.py]: Using multiline strings sounds like a great idea: the `"\n".join(...)` trick with literals is clever, but it results in a lot of extra syntax with all the quotes and commas and, as Kevin said, it makes it hard to read.

> I will look seriously into whether we could redesign the query for reusing purposes!

I think this is good as a "long term" thing, but for now it'd be great to focus on getting the queries into a state that can be more easily read (that is, multiline strings). It's relatively easy to factor out common code, after the code is landed, but it's hard to land the code if we can't easily review/understand it.

File: stellargraph/connector/neo4j/sampler.py [hunk 4]
```diff
@@ -0,0 +1,262 @@
...
+        """
+        self._check_neighbourhood_sizes(in_size, out_size)
+        self._check_common_parameters(nodes, n, len(in_size), seed)
+
+        bfsQuery = self.directed_breadth_first_search_query_builder(in_size, out_size)
+
+        result_records = neo4j_graphdb.run(bfsQuery, {"head_nodes": nodes})
+        samples = pd.DataFrame(result_records)[0][0]
+
+        return [samples]
```
huonw: Why does this only return a single sample? What happens if `n > 1`?

stellargraph/connector/neo4j/sampler.py
```diff
@@ -0,0 +1,262 @@
...
+        Returns:
+            A list of lists such that each list element is a sequence of ids corresponding to a BFW.
+        """
+        self._check_sizes(n_size)
+        self._check_common_parameters(nodes, n, len(n_size), seed)
+
+        bfsQuery = self.breadth_first_search_query_builder(n_size)
+
+        result_records = neo4j_graphdb.run(bfsQuery, {"head_nodes": nodes})
+        walks = pd.DataFrame(result_records)[0][0]
```
huonw: Could you say why you decided convert the `Cursor` object to a DataFrame?

stellargraph/connector/neo4j/mapper.py
```diff
@@ -0,0 +1,251 @@
...
+        Returns:
+            A list of the same length as ``num_samples`` of collected features from
+            the sampled nodes of shape:
+            ``(len(head_nodes), num_sampled_at_layer, feature_size)``
+            where num_sampled_at_layer is the cumulative product of `num_samples`
+            for that layer.
+        """
+        nodes_per_hop = self.sampler.run(
+            self.neo4j_graphdb, nodes=head_nodes, n=1, n_size=self.num_samples
+        )
```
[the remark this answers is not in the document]
chgswin: Yeah, thanks for raising this! I am happy to share a few things in my minds:

- At some point in the near future, we might only have Neo4J collect node features for us. This requires a slight tweak in the Neo4J cypher query. When we reach this, we might not need this part anymore.
```
    # Get features for sampled nodes
    batch_feats = [
        graph.node_features(layer_nodes, node_type)
        for layer_nodes in nodes_per_hop
    ]
```
But absolutely we still need to reshape our ```batch_feats``` with correct dimension.

demos/connector/neo4j/undirected-graphsage-on-cora-neo4j-example.ipynb
chgswin: My initial assumption is that we:

- Firstly, ingest the whole dataset into Neo4J database. I will upload the demo notebook for this for the next review request, using the ```dataset``` module).

- Secondly, from the database, we import node features into memory.

So I think it slightly makes more sense to use the ```dataset``` module at the first stage!

stellargraph/connector/neo4j/sampler.py
chgswin: My initial intention is to let Neo4J return all the lists of ids in a single result. 
It does not currently have sampling multiple times from a single node, but I will add this (to allow for ```n > 1``` in the next commit).

stellargraph/connector/neo4j/mapper.py
chgswin: Yes. I'll add import exception handling and type-checking now!

chgswin:
Recent improvements:
- Add readme.md on the demo folder.
- Add demo on loading Cora dataset into Neo4J.
- Allow one single head node to be sampled n times.
- Allow reusing the queries for in-neighbors, out-neighbors.

stellargraph/connector/neo4j/sampler.py
```diff
@@ -0,0 +1,202 @@
...
+
+                RETURN apoc.coll.flatten(collect(value.in_samples_list)) as next_samples
+            """
+
+        samples = [[head_node for head_node in nodes for _ in range(n)]]
+        neighbor_query = bfs_neighbor_query(sampling_direction="BOTH")
+
+        for num_sample in n_size:
+            cur_nodes = samples[-1]
+            result = neo4j_graphdb.run(
```
huonw: It looks like this code has gone from doing one round-trip to neo4j by sending the query as one huge string, to doing `len(n_size)` trips, each one sending back the list of nodes. Don't change it back just yet, but could you explain why you changed it?

stellargraph/connector/neo4j/mapper.py
```diff
@@ -54,11 +54,25 @@ class Neo4JBatchedNodeGenerator(BatchedNodeGenerator):
...
+
         # Create neo4J driver
-        self.neo4j_graphdb = neo4j_graphdb
+        if not isinstance(neo4j_graphdb, py2neo.Graph):
+            raise TypeError(
+                f"neo4j_graphdb: expected py2neo.Graph, found {type(neo4j_graphdb)}"
+            )
+
+        else:
+            self.neo4j_graphdb = neo4j_graphdb
```
huonw: I think it's better for this assignment to be unconditional. A strong convention in Python and in this library is to only use an `else` when it is necessary. For this code, it isn't necessary because execution of this function stops after the `raise` inside the `if`. That is:

```python
if not isinstance(...):
    raise TypeError(...)

self.neo4j_graphdb = neo4j_graphdb
```

Making the assignment unconditional is potentially slightly nicer from a code analysis/tooling perspective too, because then it's "obvious" that this `neo4j_graphdb` property is always assigned and always exists.

stellargraph/connector/neo4j/sampler.py
```diff
@@ -0,0 +1,202 @@
...
+            n (int, default 1): Number of walks per node id.
+            neighbours with replacement is always used regardless of the node degree and number of neighbours
+            requested.
+            seed (int, optional): Random number generator seed; default is None
+
+        Returns:
+            A list of lists, each list is a sequence of sampled node ids at a certain hop.
+        """
+
+        def bfs_neighbor_query(sampling_direction):
```
huonw: I think this `bfs_neighbor_query` function is exactly the same in both `run` methods. Maybe it could be moved to the top level, like:

```
def _bfs_neighbor_query(...):
    ...

class Neo4JSampledBreadthFirstWalk(GraphWalk):
    def run(...):
        ...
        neighbour_query = _bfs_neighbor_query(sampling_direction="BOTH")
        ...
```

stellargraph/connector/neo4j/sampler.py
```diff
@@ -0,0 +1,202 @@
...
+            direction = {"BOTH": "--", "IN": "<--", "OUT": "-->"}
```
huonw: To make the code slightly nicer inside the string, this could store the arrow directly:

```suggestion
            direction_arrow = {"BOTH": "--", "IN": "<--", "OUT": "-->"}[sampling_direction]
            
            f"""
                ...
                MATCH (cur_node){direction_arrow}(neighbors)
                ...
            """
```

stellargraph/connector/neo4j/sampler.py
```diff
@@ -0,0 +1,202 @@
...
+                    RETURN in_samples_list',
+                    {{ node_id: node_id, num_samples: {{num_samples}}  }}) YIELD value
+
+                RETURN apoc.coll.flatten(collect(value.in_samples_list)) as next_samples
+            """
+
+        self._check_neighbourhood_sizes(in_size, out_size)
+        self._check_common_parameters(nodes, n, len(in_size), seed)
+
+        samples = [[head_node for head_node in nodes for _ in range(n)]]
```
huonw: Repeating a list multiple times in Python can be done with `*`:

```suggestion
        samples = [nodes * n]
```

stellargraph/connector/neo4j/sampler.py
```diff
@@ -0,0 +1,202 @@
...
+
+        tree_cur_index = 0
+        tree_end_index = 0
+
+        in_sample_query = bfs_neighbor_query(sampling_direction="IN")
+        out_sample_query = bfs_neighbor_query(sampling_direction="OUT")
+
+        for in_num, out_num in zip(in_size, out_size):
+            # BFS traversal in the flattened binary tree
+            while tree_cur_index <= tree_end_index:
```
huonw: This is a neat idea, but I think it's quite opaque and hard to understand. For one, it might be better to use `range`:

```python
previous_length = 0

for in_num, out_num in zip(in_size, out_size):
    initial_length = len(samples)
    # expand the neighbourhoods of the nodes that were added at the previous hop
    for i in range(previous_length, len(samples)):
        cur_nodes = samples[i]
        ...
    previous_length = initial_length
```

Alternatively, maybe we could model the hop structure more directly:

```python
# list of hops, which are a list of neighborhoods, which are a list of nodes
hops = [[nodes * n]]
for in_num, out_num in zip(in_size, out_size):
    last_hop = hops[-1]
    this_hop = []
    for neighbourhoods in last_hop:
        for cur_nodes in neighbourhoods:
             neighbor_records = ...
             this_hop.append(neighbor_records.data()[0]["next_samples"])
             ...

    hops.append(this_hop)

# flatten all of the hops to be left with just a list of all of the neighborhoods
return sum(hops)
```

stellargraph/connector/neo4j/sampler.py
chgswin: Yes that's absolutely right! 

The main motivation is to make the queries neater and reusable across the module. Joining queries together is possible but this leads to much more syntax and is bad when maintaining the code. 

It is obviously easier to let Python stack the resulting hops together, but yes, this does come with the tradeoff of having to execute more transactions between Python and the databases. This ended up sending O(n) and O(2^n) queries to the graph databases (n is the number of hops). In practice, we experienced around 10% slower than sending one query that does all, which I think is acceptable.

stellargraph/connector/neo4j/sampler.py
chgswin: `samples = [nodes * n]` will collate the lists n times. E.g `[1,2,3]*2` will be `[1,2,3,1,2,3]`. Would there be any nicer way to generate repeated items that are next to each other, i.e. `[1,1,2,2,3,3]`?

stellargraph/connector/neo4j/sampler.py
chgswin: Sure! I am rewriting this part now.

stellargraph/connector/neo4j/sampler.py
huonw: Why do we need them to be next to each other?

stellargraph/connector/neo4j/sampler.py
huonw: > In practice, we experienced around 10% slower than sending one query that does all

"We"? Did multiple people test it?

Also, were you testing with a local neo4j, or over a network connection? If it's local, the overhead of connecting to it will be very small, whereas if it's a neo4j running on a different computer, there's likely to be noticeably more overhead.

In any case, it's fine to leave this as is, but it would be good to have a comment, like:

```python
# this does O(number of hops) queries to the database, because the code is cleanest like that
```

(With `O(number of hops)` -> `O(2^number of hops)` for the other one)

> Because n does not exceed 3 for most of the times, I guess it is still acceptable!

Doing 8 queries (with significant amounts of data set back and forth) instead of 1 is a big expansion. But yes, it's probably acceptable for now.

stellargraph/connector/neo4j/sampler.py
```diff
@@ -122,66 +131,36 @@ def run(
...
-                samples.append(neighbor_records.data()[0]["next_samples"])
+                this_hop.append(neighbor_records.data()[0]["next_samples"])
 
-                tree_cur_index += 1
-            # go to next binary tree depth
-            tree_end_index = tree_end_index * 2 + 2
+            hops.append(this_hop)
 
-        return samples
+        return list(it.chain(*hops))
```
huonw: I personally find this clearer as (and it's slightly faster):

```suggestion
        return sum(hops, [])
```

stellargraph/connector/neo4j/sampler.py
chgswin: Thanks for your question. As I observed in the `SampledBreadthFirstWalk` and `DirectedBreadthFirstNeighbours`, walks from the same head node are next to each other when they are sampled repeatedly. 
At this moment, both ways do not hurt, as by default, we only sample once from each `head_node`.

huonw: That's how the existing code is, yes, but I'm asking a slightly deeper question: do you think there's a reason that they're next to each other, or is it just coincidence?

chgswin: I personally think the main reason behind this arrangement amongst the head_nodes is to keep it consistent among other sampling algorithms (`UniformRandomMetaPathWalk`
, `TemporalRandomWalk`, etc.). 
I feel that it should be better if I can reproduce the result as similarly as we could for the original codes. But if there is any critical reason behind this, please let me know what you think.

huonw: That sounds good to me; using `nodes * n` is only a tiny improvement so I agree with your assessment that it's not worth being different. If the code required to match the existing algorithms was more complicated, the decision might be different.

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -0,0 +1,1158 @@
...
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "labels = np.array(node_data[\"subject\"])\n",
+    "\n",
+    "target_encoding = preprocessing.OneHotEncoder(sparse=False, categories=\"auto\")\n",
+    "targets = target_encoding.fit_transform(labels.reshape(-1, 1))\n",
+    "targets = pd.DataFrame(targets, index=node_data.index)"
```
huonw: There's two `target_encoding` objects in this file, and this `targets` object isn't used. I suspect either this or the `train_targets`/`test_targets` cell can be removed?

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -0,0 +1,1158 @@
...
+     },
+     "execution_count": 10,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from collections import Counter\n",
+    "\n",
+    "Counter(train_data[\"subject\"])"
```
huonw: I think we could consider removing a lot of this exploration stuff from this notebook, and just refer the user to the other directed/undirected graphsage notebooks. If we removed:

- this cell [10] (along with the explanation "Note using stratified sampling gives the following counts:"  and "The training set has class imbalance ...")
- `train_data[["subject"]]` [11]
- `print(train_targets)` end of [12]

The notebook would be a bit simpler, and we could keep the focus on the Neo4J code. What do you think?

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -0,0 +1,1158 @@
...
+   "metadata": {},
+   "source": [
+    "### Converting to numeric arrays"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For our categorical target, we will use one-hot vectors that will be fed into a soft-max Keras layer during training. To do this conversion ..."
```
huonw: This line ends with "...". Is it unfinished?

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -0,0 +1,1158 @@
...
+    "\n",
+    "\n",
+    "print(train_targets)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We now do the same for the node attributes we want to use to predict the subject. These are the feature vectors that the Keras model will use as input. The CORA dataset contains attributes 'w_x' that correspond to words found in that publication. If a word occurs more than once in a publication the relevant attribute will be set to one, otherwise it will be zero."
```
huonw: This says "now we do the same", but there's nothing being done here. Maybe this paragraph doesn't need to be here in this notebook?

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -0,0 +1,1158 @@
...
+   ],
+   "source": [
+    "print(G.info())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To feed data from the graph to the Keras model we need a data generator that feeds data from the graph to the model. The generators are specialized to the model and the learning task so we choose the `Neo4DirectedGraphSAGENodeGenerator` as we are predicting node attributes with a `DirectedGraphSAGE` model.\n",
```
huonw:
```suggestion
    "To feed data from the graph to the Keras model we need a data generator that feeds data from the graph to the model. The generators are specialized to the model and the learning task so we choose the `Neo4jDirectedGraphSAGENodeGenerator` as we are predicting node attributes with a `DirectedGraphSAGE` model.\n",
```

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -0,0 +1,1158 @@
...
+    "print(\"\\nTest Set Metrics:\")\n",
+    "for name, val in zip(model.metrics_names, test_metrics):\n",
+    "    print(\"\\t{}: {:0.4f}\".format(name, val))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Making predictions with the model"
```
huonw: I think typically we delete the last empty cell from the notebook.

demos/connector/neo4j/load-cora-into-neo4j.ipynb
```diff
@@ -0,0 +1,413 @@
...
+    "        subject: node.subject,\n",
+    "        features: node.features\n",
+    "    })\n",
+    "    \"\"\"\n",
+    "\n",
+    "# For efficient loading, we will load batch of nodes into Neo4J.\n",
+    "batch_len = 500\n",
+    "\n",
+    "for batch_start in range(0, len(node_list), batch_len):\n",
+    "    batch_end = min(len(node_list), batch_start + batch_len)\n",
```
huonw: Python slicing is relaxed so this `min` work isn't necessary: the endpoint is automatically clipped at the maximum value, e.g.

```
>>> pd.DataFrame(["a", "b", "c", "d"], columns=["X"]).iloc[2:99999999]
   X
2  c
3  d
```

That is, we can remove this line and write:

```python
records = node_list.iloc[batch_start:batch_start + batch_len].to_dict("records")
```

or, just simplify this line

```python
batch_end = batch_start + batch_len
records = node_list.iloc[batch_start:batch_end].to_dict("records")
```

The same applies to the edge loading.

demos/connector/neo4j/load-cora-into-neo4j.ipynb
```diff
@@ -0,0 +1,413 @@
...
+    "    \"\"\"\n",
+    "\n",
+    "batch_len = 500\n",
+    "\n",
+    "for batch_start in range(0, len(edge_list), batch_len):\n",
+    "    batch_end = min(len(edge_list), batch_start + batch_len)\n",
+    "    # turn edge dataframe into a list of records\n",
+    "    records = edge_list.iloc[batch_start:batch_end].to_dict(\"records\")\n",
+    "    tx = graph.begin(autocommit=True)\n",
+    "    tx.evaluate(loading_edge_query, parameters={\"edge_list\": records})"
```
huonw: How long does this take to run? I vaguely recall people saying that Neo4J is really slow to load data directly, and using CSVs is fastest.

chgswin: Yeah you're right! 
Cora dataset is tiny, so in my machine, it takes:
- 2.9 s to load all the nodes, 
- 12.5 s to load all the edges.

For bigger dataset, it is recommended to use ```neo4j admin``` tool. This loads the dataset directly from the CSVs.

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
chgswin: Yeah! I will delete this out right now!

demos/connector/neo4j/readme.md
```diff
@@ -0,0 +1,12 @@
+# Connect to Remote Neo4J Graph Databases
```
kjun9: Maybe we can add a note here to say that the functionality being demonstrated in these notebooks is still experimental?

Otherwise looks good to me!

chgswin: Sure! I will add this now!

kjun9:
@huonw I'll merge this unless we're waiting for anything else?

huonw:
Sounds good to me! :shipit:

Issue #801, opened 2020-02-06
Use real arguments instead of **kwargs where possible


huonw:
### Describe the bug

Many of the layers have layer-specific required or optional arguments that are passed by manipulating `**kwargs`, e.g., in GCN:

https://github.com/stellargraph/stellargraph/blob/2c513c19c8b8ba931eda3dbf9bf25197821f1ddb/stellargraph/layer/gcn.py#L54-L99

For a specific example, the `kernel_initializer` argument defaults to `"glorot_uniform"`; the current set up is unfortunate because:

- this default has to be specifically and manually documented
- users' mistakes are flagged less well (e.g. they have to be manually checked for, and IDEs/editors cannot easily autocomplete or validate)
- our mistakes are flagged less well (e.g. misspelling the argument in the documentation, or renaming it)

If the argument was specified in the `__init__` itself similar to `bias` and `dropout`, the documentation would list the default automatically, and tooling would be able to reason about it:

![image](https://user-images.githubusercontent.com/1203825/73895190-d5788980-48d2-11ea-9566-679974390631.png)

https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.layer.gcn.GCN

This is similar to Keras, which has Layer-specific arguments explicitly listed, and arguments for the Layer super-class passed in `**kwargs`: e.g. https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dense

![image](https://user-images.githubusercontent.com/1203825/73895873-34d79900-48d5-11ea-9bb9-89ce2d1e391d.png)

<details><summary>A list of probable suspects</summary>

(A search for `kwargs[`, `kwargs.get`, `kwargs.pop`)

```
$ git grep -n 'kwargs\(\[\|\.get\|\.pop\)'
stellargraph/layer/appnp.py:60:            kwargs["input_shape"] = (kwargs.get("input_dim"),)
stellargraph/layer/attri2vec.py:124:            self.input_node_num = kwargs["node_num"]
stellargraph/layer/attri2vec.py:125:            self.input_feature_size = kwargs["input_dim"]
stellargraph/layer/attri2vec.py:126:            self.multiplicity = kwargs["multiplicity"]
stellargraph/layer/cluster_gcn.py:84:            kwargs["input_shape"] = (kwargs.get("input_dim"),)
stellargraph/layer/cluster_gcn.py:318:            param_value = kwargs.pop(param_name, None)
[...]

PR #802, opened 2020-02-06
Replace GCN use of **kwargs with real arguments

huonw:
The headers for the `GCN` and `GraphConvolution` classes now shows all the information about the class:

`class stellargraph.layer.gcn.GCN(layer_sizes, generator, bias=True, dropout=0.0, activations=None, kernel_initializer=None, kernel_regularizer=None, kernel_constraint=None, bias_initializer=None, bias_regularizer=None, bias_constraint=None)`

`class stellargraph.layer.gcn.GraphConvolution(units, activation=None, use_bias=True, final_layer=False, input_dim=None, kernel_initializer='glorot_uniform', kernel_regularizer=None, kernel_constraint=None, bias_initializer='zeros', bias_regularizer=None, bias_constraint=None, **kwargs)`

Especially for GCN, this will reduce the chance of misspelled arguments being silently ignored.

See: #801

### DIFF
--- a/stellargraph/layer/gcn.py
+++ b/stellargraph/layer/gcn.py
@@ -57,10 +57,10 @@ class GraphConvolution(Layer):
         use_bias (bool): toggles an optional bias
         final_layer (bool): If False the layer returns output for all nodes,
                             if True it returns the subset specified by the indices passed to it.
-        kernel_initializer (str or func): The initialiser to use for the weights.
+        kernel_initializer (str or func, optional): The initialiser to use for the weights.
         kernel_regularizer (str or func, optional): The regulariser to use for the weights.
         kernel_constraint (str or func, optional): The constraint to use for the weights.
-        bias_initializer (str or func): The initialiser to use for the bias.
+        bias_initializer (str or func, optional): The initialiser to use for the bias.
         bias_regularizer (str or func, optional): The regulariser to use for the bias.
         bias_constraint (str or func, optional): The constraint to use for the bias.
     """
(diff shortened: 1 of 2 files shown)


stellargraph/layer/gcn.py
```diff
@@ -57,46 +57,45 @@ class GraphConvolution(Layer):
...
+        kernel_regularizer=None,
+        kernel_constraint=None,
+        bias_initializer="zeros",
+        bias_regularizer=None,
+        bias_constraint=None,
+        **kwargs
     ):
-        if "input_shape" not in kwargs and "input_dim" in kwargs:
-            kwargs["input_shape"] = (kwargs.get("input_dim"),)
+        if "input_shape" not in kwargs and input_dim is not None:
```
kjun9: Why does `input_shape` get special treatment and is still part of kwargs? Is it purely because of this parameter that we can't remove `**kwargs` from the argument list altogether?

huonw: No, `input_shape` is a kwarg to Keras's `Layer` (https://github.com/tensorflow/tensorflow/blob/a641fa1c5c10f0a278c7671fd6f7df550a74935d/tensorflow/python/keras/engine/base_layer.py#L278-L291
) (along with other kwargs), and it's typical for Keras layer subclasses to have a `**kwargs` to pass through this and anything else, e.g.: https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dense 's `__init__` has a `**kwargs` too.

Most of the stellargraph layers seem to have an `input_dim` argument to make defining single-dimensional input slightly easier, I guess.

PR #809, opened 2020-02-06
Implicitly convert NetworkX input to new StellarGraph

huonw:
This completes the removal of NetworkX as a `StellarGraph` backing store: when the `StellarGraph(graph, ...)` constructor is called, it converts the NetworkX value `graph` into the new `StellarGraph` representation.

This doesn't yet mark the constructor as deprecated, since the new form isn't fully documented (see open issues on #723). This also doesn't completely remove `networkx` as a required dependency, e.g. `EdgeSplitter` still uses it.

Some tests no longer make sense and are thus deleted.

See: #715


stellargraph/core/graph.py
```diff
@@ -211,27 +202,24 @@ def __init__(
...
-            if nodes is None:
-                nodes = {}
-            if edges is None:
-                edges = {}
-
-            self._is_directed = is_directed
+        if nodes is None:
+            nodes = {}
+        if edges is None:
+            edges = {}
```
kjun9: Isn't this overwriting the result from the networkx conversion if there was a graph provided?

huonw: `edges` won't be `None` in that case.

kjun9: Ah of course 😄 still asleep apparently

PR #816, opened 2020-02-07
fix missing dataset integration into demo notebooks

timpitman:
Discovered while implementing #575: 

`demos/ensembles/ensemble-node-classification-example.ipynb` is missing a download() of the dataset, so fails to run.

`demos/link-prediction/gcn/cora-gcn-links-example.ipynb` is missing any use of the new datasets module to use the `cora` dataset.

PR #817, opened 2020-02-07
Test demo notebooks on CI (#575)

timpitman:
Test demo notebooks on CI, using a Buildkite agent for each notebook.
* a new script `.buildkite/steps/test-demo-notebooks.sh` is run with parallel Buildkite agents (based on concept in `Neon`)
* some notebooks don't work and are skipped in the script, returning a soft-fail in Buildkite.  Issues to fix these in future are #818, #819, #816.
* some notebooks are too slow, and are skipped with soft-fail: #820
* agents are currently t2large (t2medium wasn't enough ram for some notebooks), potentially can be improved in future: #840


.buildkite/steps/test-demo-notebooks.sh
```diff
@@ -0,0 +1,49 @@
...
+    'ensemble-node-classification-example.ipynb' | 'cora-gcn-links-example.ipynb')
+    # FIXME: these notebooks do not yet work on CI (#818 - datasets can't be downloaded)
+    # FIXME: these notebooks do not yet work on CI (#819 - out-of-memory)
+    # FIXME: these notebooks do not yet work on CI (#833 - too slow)
+    # FIXME: these notebooks do not yet work on CI (#816 - missing dataset download)
+    exit 2 # this will be a soft-fail for buildkite
+    ;;
+  *) # fine, run this one
+    cd "$(dirname "$f")"
+    jupyter nbconvert --to notebook --execute --ExecutePreprocessor.timeout=2400 "$f"
```
huonw: Could you say why you went with this approach rather than the `papermill` one that `neon` uses?

I think we should strongly consider that one, or at least cribbing more of its features:

- `papermill` seems to give better command line output about the cells as they fail (in particular, better stdout and stderr handling)
- `papermill` will more easily generalise to handling #819, #820 and #833 if we need to reduce epochs or similar with its parameterisation functionality
- `neon` uploads the notebooks as buildkite artefacts so it's easy to download them and inspect the full output with colour etc.

.buildkite/steps/test-demo-notebooks.sh
```diff
@@ -0,0 +1,49 @@
...
+
+f=${NOTEBOOKS[$INDEX]}
+
+echo "+++ :python: running $f"
+case $(basename "$f") in
+  'attacks_clustering_analysis.ipynb' | 'hateful-twitters-interpretability.ipynb' | 'hateful-twitters.ipynb' | 'stellargraph-attri2vec-DBLP.ipynb' | \
+    'node-link-importance-demo-gat.ipynb' | 'node-link-importance-demo-gcn.ipynb' | 'node-link-importance-demo-gcn-sparse.ipynb' | 'rgcn-aifb-node-classification-example.ipynb' | \
+    'stellargraph-metapath2vec.ipynb' | \
+    'ensemble-node-classification-example.ipynb' | 'cora-gcn-links-example.ipynb')
+    # FIXME: these notebooks do not yet work on CI (#818 - datasets can't be downloaded)
```
huonw: Totally minor, but the convention in the rest of the library at the moment is to put the issue number next to the FIXME (this makes grepping mildly easier, and ensures that the number is close to the `FIXME` text even if the rest of the line is long and wrapped), e.g.

```python
# The notebooks do not yet work on CI:
# FIXME (#816): missing dataset download
# FIXME (#818): datasets can't be downloaded
# FIXME (#819): out-of-memory
# FIXME (#833): too slow
```

.buildkite/steps/test-demo-notebooks.sh
timpitman: no good reason: I just started with nbconvert from my outdated neon checkout - have updated with papermill and artefact upload from neon (the concept of behaviour changes is good)

timpitman:
> could we avoid running more of the longer ones to not regress quite so much?

the 4 notebooks listed in #820 (generally 20min+) are now disabled, brining CI time down to 16min total.  in addition, skipped notebooks now bail before pip which reduces overhead.

timpitman:
> CI wouldn't regress at all.

I've gone a compromise and disabled two of these, with CI now down to 9 minutes.

PR #821, opened 2020-02-10
Bugfix test_hinsage_from_generator

kieranricardo:
`test_hinsage_from_generator` started failing. This seems to be from a change in the randomness somewhere causing predicted output to differ from the expected despite the random seed being fixed. Running the test with the original `example_hin_1` graph without a random seed 1000 times yielded 153 unique model outputs, with one of these equaling the expected output.

This PR removes the random seed and instead calculates the expected output using numpy, Hinsage first principals, and knowledge of the graph and compares this with the predicted output.


tests/layer/test_hinsage.py
```diff
@@ -568,6 +565,100 @@ def test_hinsage_from_generator():
...
+
+    for i, feats in enumerate(batch_feats):
+        # 1st layer
+        # aggregate for the prediction node
+        layer_1_out = []
+        head = feats[0][0]
+        B_neighs = feats[0][1]
+
+        neighs_trans = (
+            B_neighs.reshape(
```
kjun9: I think we can pull out a function here to use across all the layers. Also, since the kernels are always ones, I think we can also just sum across that axis? It looks to me like that was the intention behind initialising all the weights to 1:
```python
def transform_neighbours(neighs, dim):
    neighs.reshape(
        1, dim, int(neighs.shape[1] / dim), neighs.shape[2]
    ).sum(axis=-1).mean(2)
```
I'm not sure if I've got that 100% correct, but something along those lines

kjun9: And maybe even
```python
def hinsage_layer(head, neighs_by_type):
    head_trans = head.sum(axis=-1)
    neigh_trains = sum(transform_neighbours(neigh, head.shape[1]) for neigh in neighs_by_type) / len(neighs_by_type)
    return np.concatenate([head_trans, neighs_trans], axis=-1)
```

kieranricardo: nice, yeah that's much simpler!

PR #836, opened 2020-02-10
Remove helpful handling of GraphSchema features removed in 0.9

huonw:
Various per-edge and per-node functionality was removed from `GraphSchema` in 0.9. We'd left some deprecations and other helpful hints to ease any migrations. I think it's somewhat unlikely that anyone would be depending on these specifically, and those that do will hopefully upgrade to 0.9 and get the warnings. Thus, I think we can remove them for the next release and not have to maintain this code anymore.

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -759,12 +759,6 @@ class StellarGraph:
         Returns:
             GraphSchema object.
         """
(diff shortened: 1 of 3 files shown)


timpitman:
good idea to lose this now 👍

PR #839, opened 2020-02-10
Add hinsage documentation

kjun9:
Converted from docx using [pandoc](https://pandoc.org/index.html) and then I had to manually fix up a lot of the formatting and equations.

readthedocs link: https://external-builds.readthedocs.io/html/stellargraph/839/hinsage.html


docs/hinsage.txt
```diff
@@ -0,0 +1,450 @@
...
+
+If we treat the output from the GraphSAGE layers as node embeddings for
+the source and destination nodes, then we need to combine these
+embeddings with a binary operator to obtain embeddings that correspond
+to links. For directed edges, we could expect concatenation to produce
+reasonable embeddings. Each of these edges can then be fed through a
+simple classifier, and calculate cross entropy against a vector of ones
+for true edges and zeros for negative examples.
+
+   :math:`J = ( - \log(\sigma(CONCAT(z_{u}, z_{v})))) + ( - \log(\sigma( - CONCAT(z_{u},z_{v_{n}}))))`
```
timpitman: should this be lower case for concat?

timpitman (changes requested):
Well done, this looks great - the equations come up well in readthedocs!   One small issue with concat, also I suggest this is added to the CHANGELOG.

docs/hinsage.txt
```diff
@@ -0,0 +1,450 @@
...
+
+-  :math:`b^{k}` is an optional bias,
+
+-  :math:`d_{k}` is node feature dimensionality at layer :math:`k`,
+
+-  :math:`\sigma` is the nonlinear activation,
+
+-  :math:`N(v)` is the neighbourhood of node :math:`v`
+
+-  :math:`D_{\text{p\ }}\lbrack \cdot \rbrack`\ is a random dropout with probability :math:`p` applied to its argument vector.
```
kieranricardo: extra "\" following "p" in "`D_{\text{p\ }}"

```suggestion
-  :math:`D_{\text{p}}\lbrack \cdot \rbrack`\ is a random dropout with probability :math:`p` applied to its argument vector.
```

docs/hinsage.txt
```diff
@@ -0,0 +1,450 @@
...
+
+The number of trainable parameters in layer :math:`k` for the mean
+aggregator is
+
+   :math:`{{2d}_{k}}/2 \times d_{k - 1} + d_{k} = d_{k}d_{k - 1} + d_{k}` if concat=True, or
```
kieranricardo: the use of `/` here instead of `\frac` makes it hard to read. Similar issues with other equations below.

```suggestion
   :math:`\frac{{{2d}_{k}}}{2} \times d_{k - 1} + d_{k} = d_{k}d_{k - 1} + d_{k}` if concat=True, or
```

docs/hinsage.txt
```diff
@@ -0,0 +1,450 @@
...
+
+   :math:`{2d}_{k}d_{k - 1} + d_{k}` if concat=False.
+
+For the GCN aggregator, the feature update rule is:
+
+1. | Aggregation of features from the neighbours of node :math:`v`:
+
+   :math:`{h^{k}}_{N(v)} = \frac{1}{|N(v)| + 1}\left( {h_{u}}^{k - 1} + {h_{v}}^{k - 1} \right)`
```
kieranricardo: I think this equation should have a summation over neighbour node features

```suggestion
   :math:`{h^{k}}_{N(v)} = \frac{1}{|N(v)| + 1}\left({h_{v}}^{k - 1} + \sum_{u \in N_r(v)}{h_{u}}^{k - 1}\right)`
```

docs/hinsage.txt
```diff
@@ -0,0 +1,450 @@
...
+The resulting feature update rules on heterogeneous graphs, for mean and
+GCN aggregators, are shown below (compare with the feature update rules
+for homogeneous graphs above).
+
+HinSAGE with mean aggregator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+1. Aggregation (mean) of features from the neighbours of node :math:`v` via edges of type :math:`r`:
+
+   :math:`{h^{k}}_{N_{r}(v)} = \frac{1}{|N_{r}(v)|}D_{p}({h_{u}}^{k - 1})`
```
kieranricardo: This should probably have a summation as well:

```suggestion
   :math:`{h^{k}}_{N_{r}(v)} = \frac{1}{|N_{r}(v)|}\sum_{u \in N_r(v)}D_{p}({h_{u}}^{k - 1})`
```

kieranricardo (changes requested):
Thanks for doing this! It looks pretty good overall, just a few small equation + formatting issues

docs/hinsage.txt
kjun9: Good catch!

kjun9:
Thanks for the reviews @timpitman @kieranricardo I've addressed your comments!

I'll also see where there are existing links to the google doc in our issues and point to this instead

docs/hinsage.txt
```diff
@@ -37,20 +39,20 @@ Where:
...
 -  :math:`N(v)` is the neighbourhood of node :math:`v`
 
--  :math:`D_{\text{p\ }}\lbrack \cdot \rbrack`\ is a random dropout with probability :math:`p` applied to its argument vector.
+-  :math:`D_{p}\lbrack \cdot \rbrack`\ is a random dropout with probability :math:`p` applied to its argument vector.
 
 The number of trainable parameters in layer :math:`k` for the mean
 aggregator is
 
-   :math:`{{2d}_{k}}/2 \times d_{k - 1} + d_{k} = d_{k}d_{k - 1} + d_{k}` if concat=True, or
+   :math:`d_{k - 1} + d_{k} = d_{k}d_{k - 1} + d_{k}` if concat=True, or
```
kieranricardo: I think you accidently deleted ``{{2d}_{k}}/2 \times ` instead of wrapping it in `\frac`

```suggestion
   :math:`\frac{{2d}_{k}}{2} \times d_{k - 1} + d_{k} = d_{k}d_{k - 1} + d_{k}` if concat=True, or
```

kieranricardo (commented):
Looks good! just that one extra typo and then I think this can be merged

docs/hinsage.txt
kjun9: Oops, good catch! I think what I actually meant to do was remove the whole left hand side?
```suggestion
   :math:`d_{k}d_{k - 1} + d_{k}` if concat=True, or
```
since it seemed a bit unnecessary to me to write `2d_k/2 * d_k-1 + d_k = d_k d_k-1 + d_k` - I guess it was supposed to be showing how we arrived at the result, but without some additional explanation for how we arrive at the left-hand side in the first place, it kind of looked strange to me.

What do you think?

kieranricardo: Yeah true, it doesn't make a whole lotta sense just by itself. Sounds good to me :)

kieranricardo (approved):
LGTM 👍

timpitman (approved):
Looking good, LGTM!  May be worth mentioning in the CHANGELOG too?

PR #841, opened 2020-02-11
Support to_networkx(feature_name=None) to not include node features

huonw:
Node features can be large, may not be always useful and have complicated types (a ndarray or list). The latter can cause problems with some downstream algorithms that cannot handle completely arbitrary attributes, like [`nx.write_graphml`](https://networkx.github.io/documentation/stable/reference/readwrite/generated/networkx.readwrite.graphml.write_graphml.html#networkx.readwrite.graphml.write_graphml):

```
----> 2 nx.write_graphml(Gnx, os.path.join(dataset.data_directory, pred_fname))
[...]
NetworkXError: GraphML writer does not support <class 'list'> as data values.
```

```
----> 2 nx.write_graphml(Gnx, os.path.join(dataset.data_directory, pred_fname))
[...]
NetworkXError: GraphML writer does not support <class 'numpy.ndarray'> as data values.
```

This helps with some of our notebooks that use `nx.write_graphml` (see: #717).

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -871,6 +871,10 @@ class StellarGraph:
         """
         Create a NetworkX MultiGraph or MultiDiGraph instance representing this graph.
 
+        Args:
+            feature_name (str, optional): the name of the attribute to use to store a node's feature
+                vector; if ``None``, feature vectors are not stored within each node.
+
         Returns:
              An instance of `networkx.MultiDiGraph` (if directed) or `networkx.MultiGraph` (if
              undirected) containing all the nodes & edges and their types & features in this graph.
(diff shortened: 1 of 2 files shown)


stellargraph/core/graph.py
```diff
@@ -871,6 +871,10 @@ def to_networkx(
         """
         Create a NetworkX MultiGraph or MultiDiGraph instance representing this graph.
 
+        Args:
+            feature_name (str, optional): the name of the attribute to use to store a node's feature
```
kjun9: I wonder if this docstring is a bit misleading if the user decides to pass in a value without using keyword arguments, since it looks like `feature_name` is the first argument to this method, when it's actually the 4th.

huonw: I guess so 🤷‍♂ 

I added docs for the other args.

Issue #842, opened 2020-02-11
format_notebooks.py could normalise "stream" output to non-stream


huonw:
### Description

Jupyters models interactively updated output as `"output_type": "stream"`, which is great for progress bars in interactive output, but not so great when saving that to a file: the incremental updates are all saved, but only the final result is interesting.

For example,  https://github.com/stellargraph/stellargraph/blob/baff22bcee03382c83d7e9e9f104c5a5d46552d9/demos/embeddings/embeddings-unsupervised-graphsage-cora.ipynb?short_path=2c8374e#L401-L440 was run and saved with `format_notebooks.py --execute` and has 10k+ lines of progress bar updates.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

timpitman:
see #826 for another different issue with cell output

PR #843, opened 2020-02-11
Coalesce stream output to single chunk in format_notebooks.py

huonw:
Jupyter notebooks can model incremental/progressive output as a series of separate output chunks. Unfortunately, when using `--execute`,  a progress bar that updates itself ends up having every update saved (see description #842), rather than only the final state.

This patch avoids this, by coalescing all of the incremental progress bar updates into a single chunk, including overwriting earlier updates with later ones (this is driven by the `\r` carriage return used to power progress bars).

Our current set of notebooks seem to satisfy this check already already, but not if they're re-run with `--execute`.

See: #842

### DIFF
--- a/scripts/format_notebooks.py
+++ b/scripts/format_notebooks.py
@@ -181,6 +181,11 @@ if __name__ == "__main__":
         help="Set kernel spec to default 'Python 3'",
     )
     parser.add_argument(
+        "--coalesce_streams",
+        action="store_true",
+        help="Coalesce streamed output into a single chunk of output",
+    )
+    parser.add_argument(
         "-d",
         "--default",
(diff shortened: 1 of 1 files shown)


scripts/format_notebooks.py
```diff
@@ -220,6 +225,7 @@ def preprocess_cell(self, cell, resources, cell_index):
     on_ci = args.ci
     format_code = args.format_code or args.default
     clear_warnings = args.clear_warnings or args.default
+    coalesce_streams = args.coalesce_streams or args.default
```
timpitman: I think this requires a corresponding update to the help text for the "default" option?

PR #844, opened 2020-02-11
Make graphsage node attribute inference reproducible

kjun9:
This introduces `ConcurrentGraphWalk` class and tests graphsage node attribute inference for reproducibility.

@huonw I've left the new class as part of this PR for now, but if bigger issues come up around the other changes being made, I could pull it out to a separate one to try land it first


huonw:
What's the status of this? There's several PRs that would benefit from building some sort of abstraction around per-epoch random state generation (e.g. #848, #879).

kjun9:
There's still some failing tests I need to fix. Are you talking about `ConcurrentGraphWalk`? I can prioritise this if it's blocking

huonw:
Yeah, `ConcurrentGraphWalk`. It'd be convenient if there was some sort of abstraction to manage the randomness sources. Potentially that could be landed as a separate PR that just adds that abstraction without changes to `GraphSAGENodeGenerator`.

stellargraph/data/explorer.py
```diff
@@ -953,3 +954,31 @@ def _walk(self, src, dst, t, length, bias_type, np_rs):
                 break
 
         return walk
+
+
+class ConcurrentGraphWalk:
```
huonw: Instead of `GraphWalk`, maybe this could be:

```suggestion
class ConcurrentSeeded:
```

or

```suggestion
class SeededPerBatch:
```

because it'll be useful in `KGTripleSequence` in https://github.com/stellargraph/stellargraph/pull/848 too. (The docs will need updating too.)

It could potentially move to the `random` module too?

stellargraph/data/explorer.py
```diff
@@ -953,3 +954,31 @@ def _walk(self, src, dst, t, length, bias_type, np_rs):
...
+
+    def __getitem__(self, index):
+        self._lock.acquire()
+        try:
+            return self._walkers[index]
+        except IndexError:
+            # always create a new seeded sampler in ascending order of batch number
+            # this ensures seeds are deterministic even when batches are run in parallel
+            for n in range(len(self._walkers), index + 1):
+                seed = self._rs.randint(0, 2 ** 32 - 1)
```
huonw: I know this is just moved, but:

```suggestion
                seed = self._rs.randrange(2 ** 32)
```

stellargraph/data/explorer.py
```diff
@@ -953,3 +954,31 @@ def _walk(self, src, dst, t, length, bias_type, np_rs):
...
+
+
+class ConcurrentGraphWalk:
+    """
+    Internal utility class for using GraphWalk classes in a multi-threaded
+    environment.
+
+    """
+
+    def __init__(self, create_with_seed, rs):
```
huonw: Maybe instead of taking a `rs`, this could take a seed and manage its own randomness internally?

tests/reproducibility/test_graphsage.py
```diff
@@ -100,3 +158,14 @@ def test_reproducibility(petersen_graph, shuffle):
...
     )
+
+
+@pytest.mark.parametrize("shuffle", [True, False])
+def test_nai(petersen_graph, shuffle):
+    target_size = 10
+    targets = np.random.rand(len(petersen_graph.nodes()), target_size)
+    assert_reproducible(
+        lambda: gs_nai(
+            petersen_graph, targets, [2], tf.optimizers.Adam(1e-3), shuffle=shuffle
```
huonw: Is it worth checking that multiple layers are reproducible, to ensure the sampling is doing the right thing?

```suggestion
            petersen_graph, targets, [2, 2], tf.optimizers.Adam(1e-3), shuffle=shuffle
```

kjun9: good point!

stellargraph/data/explorer.py
kjun9: makes sense 👍

stellargraph/random.py
```diff
@@ -75,3 +76,31 @@ def set_seed(seed):
...
+        self._rs, _ = random_state(seed)
+
+    def __getitem__(self, batch_num):
+        self._lock.acquire()
+        try:
+            return self._walkers[batch_num]
+        except IndexError:
+            # always create a new seeded sampler in ascending order of batch number
+            # this ensures seeds are deterministic even when batches are run in parallel
+            for n in range(len(self._walkers), batch_num + 1):
```
huonw: This could be

```python
self._walks.extend(
    self.create_with_seed(self._rs.randrange(2 ** 32))
    for _ in range(len(self._walkers), batch_num + 1)
)
```
if you thought that was clearer

stellargraph/random.py
```diff
@@ -75,3 +76,31 @@ def set_seed(seed):
         _rs = _global_state()
     else:
         _rs = _seeded_state(seed)
+
+
+class SeededPerBatch:
```
huonw: Is it worth having some tests for this? E.g. checking that the seeds are always in the same order no matter the order of `batch_num` queries or something?

kjun9: Sure, I'll add one 👍

tests/test_random.py
```diff
@@ -0,0 +1,38 @@
...
+from stellargraph.random import SeededPerBatch
+import numpy as np
+
+
+def test_seeded_per_batch():
+    num_batches = 10
+    num_iter = 10
+    seed = 0
+
+    # different permutations of batch numbers
```
huonw: It could be a bit hard to reverse engineer the purpose of this test in weeks/months, so maybe this comment could be expanded slightly (or another one added)? e.g.:

```suggestion
    # different permutations of batch numbers should always give the same seeds
```

Issue #831, opened 2020-02-10
Interpretability: Integrated gradient-based importance values are not on the same scale as values from the brute force approach


kjun9:
### To Reproduce

1. *Go to '...'*
2. *Execute '...'*
3. *See error*

### Observed behavior

*Describe or paste the exact output that demonstrates the bug (for example, the stack trace if it is an exception).*

### Expected behavior

*A clear and concise description of what you expected to happen.*

### Additional context

Add any other context about the problem here. For example:

Data used: *if this occurs with a specific dataset, which dataset and how to acquire it*

Screenshots: *if applicable, add some screenshots to explain the behaviour*

PR #847, opened 2020-02-11
Fix integrated gradients node importance scores

kjun9:
This addresses #831 

The source of the bug was from forgetting to multiply all the gradients by `d alpha` while integrating w.r.t alpha. I've pulled out the numerical integration code so that we can use it for all importance score calculations.


kjun9:
I'm going to add some tests to compare against the brute-force/ground-truth approach to validate this. Related: https://github.com/stellargraph/stellargraph/issues/850

kjun9:
I think I'll leave this until #872 and #855 lands and then fix it up

kjun9:
Closing this and moving the task to backlog since there's unlikely to be progress anytime soon

Issue #856, opened 2020-02-12
Edge weights are not treated equally across different algorithms


kjun9:
### Description

For weighted graphs, I think the current behaviour might be a bit surprising for users:

* For algorithms like GraphSAGE that sample the local neighbourhood, edge weights are currently not taken into account at all.
* For algorithms like GCN that use the adjacency matrix, edge weights are always used to fill in the values in the adjacency matrix.

I think we need to keep the meaning of "edge weight" flexible since this is the only edge feature available (for example, temporal walks use weights as time), so it might make sense to update the `to_adjacency_matrix` to default to ignoring edge weights instead. 

### User Story

**As** a data scientist

**I want to** use different algorithms with my weighted graph in a consistent way

**So that** results are comparable

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #857, opened 2020-02-12
Ignore weights when creating adj matrix

kjun9:
Part of #856

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -842,7 +842,9 @@ class StellarGraph:
         """
         return self._edges.degrees()
 
-    def to_adjacency_matrix(self, nodes: Optional[Iterable] = None):
+    def to_adjacency_matrix(
+        self, nodes: Optional[Iterable] = None, ignore_weights=True
+    ):
         """
         Obtains a SciPy sparse adjacency matrix of edge weights.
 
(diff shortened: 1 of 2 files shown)


stellargraph/core/graph.py
```diff
@@ -842,7 +842,9 @@ def node_degrees(self) -> Mapping[Any, int]:
         """
         return self._edges.degrees()
 
-    def to_adjacency_matrix(self, nodes: Optional[Iterable] = None):
+    def to_adjacency_matrix(
+        self, nodes: Optional[Iterable] = None, ignore_weights=True
```
huonw: I personally find a positive phrasing clearer; that is, changing from the default is turning something on, rather than turning something off.

```suggestion
        self, nodes: Optional[Iterable] = None, weighted=False
```

and then `weighted=True` would use the weights from the graph. But I guess that makes the multi-edge explanation a bit more awkward?

(Spelling out an example here for my own exploration, you don't have to use.)


```python
        """
        Obtains a SciPy sparse adjacency matrix of edge weights.

        By default (``weighted=False``), each element of the matrix contains the number 
        of edges between the two vertices (only 0 or 1 in a graph without multi-edges).

        Args:
            nodes (iterable): The optional collection of nodes
                comprising the subgraph. If specified, then the
                adjacency matrix is computed for the subgraph;
                otherwise, it is computed for the full graph.
            weighted (bool): If true, use the edge weight column from the graph instead
                of edge counts (weights from multi-edges are summed).

        Returns:
        """
```

kjun9: Thanks, I used your example I think the multi-edge explanation sounds fine, and not far from "expected" behaviour.

Issue #820, opened 2020-02-09
Some notebooks take a long time to run on CI and therefore aren't tested


timpitman:
### Describe the bug

Some notebooks are very slow to run, which causes a problem for CI times:

```
 25min:  calibration/calibration-pubmed-link-prediction.ipynb
 32min:  ensembles/ensemble-link-prediction-example.ipynb
 20min:  link-prediction/hinsage/movielens-recommender.ipynb
>1hour: embeddings/stellargraph-metapath2vec.ipynb
  14min: node-classification/node2vec/stellargraph-node2vec-weighted-random-walks.ipynb
  12min: calibration/calibration-pubmed-node-classification.ipynb
  18min: ensembles/ensemble-node-classification-example.ipynb 

```

(times above are with t2large and include some fixed overhead installing packages etc - taken from https://buildkite.com/stellar/stellargraph-public/builds/1179#943514b6-3e46-4d0d-8b1e-119ef805e54f)

There are various possibilities for speeding up, eg:
* run fewer epocs
* substitute smaller datasets

Note: the notebooks are disabled from CI via specific exclusion in .buildkite/steps/test-demo-notebooks.sh.

https://github.com/stellargraph/stellargraph/pull/860 addressed the following notebooks, which are now tested:
```
calibration-pubmed-node-classification.ipynb
ensemble-link-prediction-example.ipyn
stellargraph-node2vec-weighted-random-walks.ipynb
ensemble-node-classification-example.ipynb
```

This means that now only 3 remain to be addressed (now covered in #874) :
```
calibration-pubmed-link-prediction.ipynb
stellargraph-metapath2vec.ipynb
movielens-recommender.ipynb
```

PR #860, opened 2020-02-13
Speed up notebooks with Papermill to add them to CI (#820)

timpitman:
Some notebooks were too slow for CI #820 - we can alter basic parameters like number of epochs, walk lengths etc to make them run faster and add to CI.

This PR adds the following notebooks to CI that were skipped for #820:
```
calibration-pubmed-node-classification.ipynb
ensemble-link-prediction-example.ipyn
stellargraph-node2vec-weighted-random-walks.ipynb
ensemble-node-classification-example.ipynb
```

Some notebooks are still too slow and will be addressed in future:
```
calibration-pubmed-link-prediction.ipynb
stellargraph-metapath2vec.ipynb
movielens-recommender.ipynb
```

Papermill is used to substitute new (smaller) values for:
```
epochs
walk_length
batch_size
n_estimators
n_predictions
```
This is done by adding a special tag to a single cell on each of the notebooks - Papermill inserts a new cell immediately after which replaces the values - this can be seen by viewing the Buildkite artefact generated when running each notebook, which will clearly show the inserted cell.


demos/node-classification/node2vec/stellargraph-node2vec-weighted-random-walks.ipynb
```diff
@@ -555,16 +575,16 @@
...
    "cell_type": "code",
-   "execution_count": 18,
+   "execution_count": 19,
    "metadata": {},
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "0.792604501607717"
+       "0.4340836012861736"
```
huonw: The accuracy here and on lines 805 and 1122 has dropped massively? Has something gone wrong here, or is it just very variable?

.buildkite/steps/test-demo-notebooks.sh
```diff
@@ -51,7 +51,8 @@ echo "+++ :python: running $f"
 cd "$(dirname "$f")"
 # run the notebook, saving it back to where it was, printing everything
 exitCode=0
-papermill --execution-timeout=600 --log-output "$f" "$f" || exitCode=$?
+# papermill will replace parameters on some notebooks to make them run faster in CI
+papermill --execution-timeout=600 -p epochs 2 -p walk_length 5 -p batch_size 5 -p n_estimators 2 -p n_predictions 2 --log-output "$f" "$f" || exitCode=$?
```
huonw: This might be clearer as the YAML format, e.g.:

```bash
parameters='
---
epochs: 2
walk_length: 5
batch_size: 5
n_estimators: 2
n_predictors: 2
'

papermill --execution-timeout=600 --parameters_yaml "$parameters" ...
```

Or, even have it as a separate file like `.buildkite/notebook-parameters.yml` and write:

```bash
papermill --execution-timeout=600 --parameters_file ".buildkite/notebook-parameters.yml" ...
```

timpitman: Using a YAML file keeps it clearest, I think 👍

demos/node-classification/node2vec/stellargraph-node2vec-weighted-random-walks.ipynb
timpitman: well spotted!  I'm not sure what happened here, however I re-ran and got much better results (now pushed); it's possible that I accidentally committed results from a run with altered parameters.

huonw: Ah, right, I guess `walk_length = 5` would reduce performance...

.buildkite/notebook-parameters.yml
```diff
@@ -4,4 +4,4 @@ epochs: 2
 walk_length: 5
 batch_size: 5
 n_estimators: 2
-n_predictors: 2
+n_predictors: 1
```
huonw: I don't think this is the best unless it's a huge speed-up. My thinking is that there's more likely to be a difference between 1 and 2, than between 2 and 3, or 2 and 4. E.g. axis reductions might not happen correctly, or there might be special cases for a single predictor.

That is,

```suggestion
n_predictors: 2
```

What do you think?

Issue #840, opened 2020-02-11
Investigate reducing notebook CI stages to t2medium buildkite agents once new SG class is rolled out


timpitman:
### Description

Currently the notebook CI (for #575) is using t2large instances, since a handful of notebooks take too much ram for t2medium.  Potential efficiencies with the new StellarGrpah class may allow us to switch to using t2medium, with associated cost savings.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
This is now irrelevant with the switch to Github Actions in #1687.

PR #864, opened 2020-02-13
Run notebooks on T2 medium instances on CI

huonw:
See: #840


huonw:
It seems everything works here except for `cora-links-example.ipynb`, which hits the agent killed error.

huonw:
This will likely require more work, so closing for now.

PR #867, opened 2020-02-13
Handle no edges in StellarGraph.create_graph_schema's np.unique calls

huonw:
`numpy.unique` behaves poorly on a 2D array with a zero dimension when uniquing whole rows (`axis=0`) instead of individual elements:

```python
>>> import numpy as np
>>> arr = np.zeros(shape=(0, 3))

>>> np.unique(arr) # individual elements is fine
array([], dtype=float64)

>>> np.unique(arr, axis=0) # whole rows is not
  File "<__array_function__ internals>", line 6, in unique
  File "/opt/anaconda3/lib/python3.7/site-packages/numpy/lib/arraysetops.py", line 274, in unique
    ar = ar.reshape(orig_shape[0], -1)
ValueError: cannot reshape array of size 0 into shape (0,newaxis)
```

This patch special-cases having no edges (which results in zero edge triples (rows) being passed to `np.unique`).

See: #869

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -715,9 +715,20 @@ class StellarGraph:
 
     def _unique_type_triples(self, *, return_counts, selector=slice(None)):
         all_type_ilocs = self._edge_type_iloc_triples(selector, stacked=True)
-        ret = np.unique(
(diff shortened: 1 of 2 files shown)


huonw:
I filed https://github.com/numpy/numpy/issues/15559 and will add it as a FIXME to the code.

PR #872, opened 2020-02-14
Add brute force importance

kjun9:
Currently, this only works for algorithms that use the adjacency matrix instead of querying for neighbours using other stellargraph methods.


kjun9:
Closing this and moving task back to backlog since there's unlikely to be progress soon

PR #874, opened 2020-02-16
Add remaining notebooks to CI that were too slow to test (#820)

timpitman:
Use Papermill to configure CI parameters for slow Jupyter demo notebooks, so that they run fast enough to be included in CI.  This PR addresses some notebooks that were too complicated to be included in the initial speedup of #860:

```
calibration-pubmed-link-prediction.ipynb
stellargraph-metapath2vec.ipynb
movielens-recommender.ipynb
``` 

A Papermill parameter cell is configured in each Notebook, and parameters substituted from: `.buildkite/notebook-parameters.yml`.


demos/calibration/calibration-pubmed-link-prediction.ipynb
```diff
@@ -65,12 +72,39 @@
...
+     "parameters"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "batch_size = 50\n",
+    "epochs = 20  # The number of training epochs for training the GraphSAGE model.\n",
+    "\n",
+    "# train, test, validation split\n",
+    "train_size: 0.2\n",
```
huonw: I think this syntax is YAML, but should be python?

```suggestion
    "train_size = 0.2\n",
```

.buildkite/notebook-parameters.yml
```diff
@@ -1,7 +1,18 @@
...
+# minimal ensembles:
 n_estimators: 2
 n_predictions: 2
+# reduce graph size for movielens-recommender.ipynb and calibration-pubmed-link-prediction.ipynb:
+train_size: 0.005
+test_size: 0.01
+val_size: 0.01
+# minimal metapath for demos/embeddings/stellargraph-metapath2vec.ipynb:
+metapaths:
+  - - "group"
```
huonw: Is `group`, `group` a metapath that exists in the graph? Maybe `group`, `user` would be more realistic? Also, by listing only one metapath, we're not validating that it correctly handles multiple. Is there a big slowdown by adding another short one?

Also, totally minor, but the double nesting of lists here was hard for me to understand, so maybe the inner one could use an inline list, e.g.

```yaml
metapaths:
  - ["group", "user"]
  - ["user", "user"]
```

.buildkite/notebook-parameters.yml
```diff
@@ -1,7 +1,18 @@
 ---
 # configure the parameters replaced in jupyter notebooks when running on CI with papermill
-epochs: 2
-walk_length: 5
-batch_size: 5
+# common parameters used in several notebooks:
+epochs: 1
```
huonw: Reducing this to `1` seems slightly unfortunate to me, in case there's some problem caused by an epoch-transition. In particular, a few of our data generators have an `on_epoch_end` function that is called to permute or otherwise modify data for the next epoch. If the notebooks have misconfigured this, there may be a problem that `epochs == 1` doesn't catch.

What do you think?

.buildkite/notebook-parameters.yml
```diff
@@ -1,7 +1,18 @@
...
+batch_size: 2
```
huonw: Decreasing the batch size doesn't necessarily make things faster... in fact, it can make things slower, because there's more per-batch overhead. Decreasing sometimes seems to make the accuracy increase faster, but I think that's not something we care about for this testing?

Did this decrease seem to help? If not, maybe we could leave it as it was

```suggestion
batch_size: 5
```

and maybe even investigate increasing it in a follow-up PR?

.buildkite/notebook-parameters.yml
timpitman: group-group gives just 47 paths, others give thousands and are much too slow (I'll confirm with a CI run though).

timpitman: confirmed it fails out after 12.5 minutes with a cell timeout adding user-user path (it times out doing the walks).

.buildkite/notebook-parameters.yml
timpitman: agree - with some of the other changes made to reduce graph size, epochs can be increased back to 2.

.buildkite/notebook-parameters.yml
timpitman: batch size is now back to 5, decrease didn't help.

.buildkite/notebook-parameters.yml
timpitman: have changed so that there are now two short metapaths for slightly better coverage: [["group", "group"], ["group", "user", "user", "group"]]

huonw: 👍 

> I feel that a minimal test here is better than no test.

Yeah, totally.

PR #875, opened 2020-02-17
Update to latest tensorflow (currently 2.1.0)

timpitman:
remove our pinning to tensorflow 2.0, so that we can use the latest current version (2.1.0).
* #692 was a transient issue not from tensorflow 2.1
* #626 also not due to tensorflow 2.1

This has the advantage that we're now using the version of tensorflow with inbuilt gpu optimisation.


timpitman:
> do we need to drop support for `tensorflow 2.0.0`?

No, we don't actually - good point.  I've updated to require tensorflow >= 2.0.1 (the .1 release addresses a security vulnerability).  2.0.x will be untested though - CI will be testing on latest (2.1.0+).

PR #878, opened 2020-02-17
Install papermill as a test dependency

huonw:
By having Papermill as a proper dependency, it will be cached along with the rest of our dependencies and doesn't need to be installed separate for testing the demos.

Follow-up work to #788


setup.py
```diff
@@ -54,6 +54,9 @@
         "black>=19.3b0",
         "nbconvert>=5.5.0",
         "treon>=0.1.2",
+        # Pulling in https://github.com/nteract/papermill/pull/459 for --execution-timeout, which
+        # hasn't been released yet
+        "papermill @ https://github.com/nteract/papermill/archive/master.tar.gz",
```
timpitman: is this a little suspect for our setup.py?

huonw: I agree it's less than good, but I thought it was fine for a developer-only dependency. In any case, nice catch. I checked https://github.com/nteract/papermill/releases but didn't think to check PyPI.

Issue #871, opened 2020-02-13
Link prediction with Directed GraphSage


kieranricardo:
### Description

Currently there's no support for link prediction with Directed GraphSage. This can be added by creating a `DirectedGraphSageLinkGenerator` object.

### User Story

As a data scientist I want to preform link prediction on directed graphs.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

Issue #174, opened 2018-08-20
Make the EdgeSplitter return StellarGraph instead of Networkx graph


youph:
### Description
Currently EdgeSplitter returns a networkx graph, which then needs to be converted to StellarGraph again. It should accept and return StellarGraph graphs.

### Done Checklist (Development)
- [ ] Produced code for required functionality
- [ ] Branch and Pull Request build on CI
- [ ] Branch and Pull Request pass unit tests on CI
- [ ] Branch and Pull Request pass integration tests on CI
- [ ] Peer Code Review Performed
- [ ] Code well commented

huonw:
This task is now more important that we're moving away from NetworkX, so I'm working on it under the umbrella of #662.

PR #879, opened 2020-02-17
Directed graphsage link generator

kieranricardo:
This PR adds support for link prediction with directed graph sage

See: #871

Reviewer checklist:

- [ ] the demo clearly shows how to use directed graph sage for link prediction
- [ ] `DirectedGRAPHSAGELinkGenerator` is implemented correctly
- [ ] `DirectedGRAPHSAGELinkGenerator` is thoroughly tests


stellargraph/data/edge_splitter.py
```diff
@@ -98,8 +98,10 @@ def _train_test_split_homogeneous(
         """
         # minedges are those edges that if removed we might end up with a disconnected graph after the positive edges
         # have been sampled.
-        if keep_connected:
+
+        if (not self.g.is_directed()) and keep_connected:
             self.minedges = self._get_minimum_spanning_edges()
+            self.minedges_set = set(self.minedges)
```
huonw:
```suggestion
```

It's slightly confusing, but `self.minedges_set` is set by `_get_minimum_spanning_edges`.

(I'm working on rewriting this, as part of #174.)

stellargraph/data/edge_splitter.py
```diff
@@ -98,8 +98,10 @@ def _train_test_split_homogeneous(
         """
         # minedges are those edges that if removed we might end up with a disconnected graph after the positive edges
         # have been sampled.
-        if keep_connected:
+
+        if (not self.g.is_directed()) and keep_connected:
```
huonw: Why do we ignore `keep_connected` for directed graphs?

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -511,3 +522,143 @@ def sample_features(self, head_links, batch_num):
...
+
+            node_type = self.head_node_types[0]
+
+            max_hops = len(self.in_samples)
+            max_slots = 2 ** (max_hops + 1) - 1
+            features = [None] * max_slots  # flattened binary tree
+
+            for slot in range(max_slots):
+                nodes_in_slot = list(
+                    it.chain(*[sample[slot] for sample in node_samples])
```
huonw: I think this be rewritten to use list comprehensions:

```python
nodes_in_slot = [
    element
    for sample in node_samples
    for element in sample[slot]
]
```

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -511,3 +522,143 @@ def sample_features(self, head_links, batch_num):
...
+                )
+                features_for_slot = self.graph.node_features(nodes_in_slot, node_type)
+                resize = -1 if np.size(features_for_slot) > 0 else 0
```
huonw: Why is this needed? Based on what I understand of `DirectedBreadthFirstNeighbours`, `np.size(features_for_slot) == 0` is impossible, because it always returns _something_ for every element of the sample (`None` if there's no other choice).

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -511,3 +522,143 @@ def sample_features(self, head_links, batch_num):
...
+            # NN input layer. Every hop potentially generates both in-nodes and out-nodes, held separately,
+            # and thus the slot (or directed hop sequence) structure forms a binary tree.
+
+            node_type = self.head_node_types[0]
+
+            max_hops = len(self.in_samples)
+            max_slots = 2 ** (max_hops + 1) - 1
+            features = [None] * max_slots  # flattened binary tree
+
+            for slot in range(max_slots):
```
huonw: Could you say why you took such a different approach to undirected `GraphSAGELinkGenerator` (and, also, the conceptually-similar `HinSAGELinkGenerator`) to do the rearrangement here?

demos/link-prediction/directed-graphsage/cora-links-example.ipynb
```diff
@@ -0,0 +1,734 @@
+{
```
huonw: Other demos for directed GraphSAGE are just categorised under `graphsage`, e.g. https://github.com/stellargraph/stellargraph/blob/develop/demos/node-classification/graphsage/directed-graphsage-on-cora-example.ipynb

so maybe this can be moved to `demos/link-prediction/graphsage`?

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -125,10 +131,15 @@ def flow(self, link_ids, targets=None, shuffle=False):
...
+                except KeyError:
+                    raise KeyError(
+                        f"Node ID {src} supplied to generator not found in graph"
+                    )
+                try:
                     node_type_dst = self.graph.node_type(dst)
                 except KeyError:
                     raise KeyError(
-                        f"Node ID {n} supplied to generator not found in graph"
+                        f"Node ID {dst} supplied to generator not found in graph"
```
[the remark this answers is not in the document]
kieranricardo: It's just a minor tweak in the error messaging - previously if there was an `dst` node not found in graph it would print:

`f"Node ID {n} supplied to generator not found in graph"`

stellargraph/data/edge_splitter.py
kieranricardo: unfortunately `nx.minimum_spanning_edges` doesn't support directed graphs.

stellargraph/data/edge_splitter.py
kieranricardo: ah I didn't realise

stellargraph/mapper/sampled_link_generators.py
kieranricardo: I kept the approach consistent (basically a copy paste) with `DirectedGraphSAGENodeGenerator`

stellargraph/mapper/sampled_link_generators.py
kieranricardo: Ah I copied this over from `DirectedGraphSAGENodeGenerator` - I've fallen into the trap of reusing old code without looking at it too closely

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -511,3 +522,143 @@ def sample_features(self, head_links, batch_num):
...
+                )
+
+            # Get features for the sampled nodes
+            batch_feats.append(features)
+
+        # Resize features to (batch_size, n_neighbours, feature_size)
+        # and re-pack features into a list where source, target feats alternate
+        # This matches the GraphSAGE link model with (node_src, node_dst) input sockets:
+        batch_feats = [
+            feats  # np.reshape(feats, (len(head_links), -1, feats.shape[1]))
```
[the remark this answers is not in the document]
kieranricardo: That should be deleted - its the reshaping that's now performed in:

`features[slot] = np.reshape(...)`

stellargraph/data/edge_splitter.py
kieranricardo: Found the equalivent algo for directed graphs! I've added the nx Edmond's algo for directed graphs in `EdgeSplitter`

stellargraph/mapper/sampled_link_generators.py
huonw: Copy-paste is ok, but blind copy-pasta isn't the best...

stellargraph/mapper/sampled_link_generators.py
huonw: Oh, that makes sense 👍 I didn't even think to compare against that, but that's much more logical.

tests/mapper/test_link_mappers.py
```diff
@@ -783,3 +783,390 @@ def test_Attri2VecLinkGenerator_unsupervisedSampler_sample_generation(self):
...
+
+            for ii in range(len(nf) // 2):
+                assert nf[2 * ii].shape == nf[2 * ii + 1].shape
+
+            # head nodes
+            for ii in range(2):
+                assert nf[ii].shape == (min(self.batch_size, data_size), 1, self.n_feat)
+
+            # first hop in samples
+            for ii in range(2, 4):
```
kjun9: Maybe we can combine all these individual for-loops that look mostly the same:
```python
ins, outs = self.in_samples, self.out_samples
dims = [
    1,
    ins[0], 
    outs[0], 
    ins[0] * outs[1], 
    ins[0] * outs[1], 
    outs[0] * ins[1], 
    outs[0] * outs[1]
]
for i, dim in zip(range(0, 14, 2), dims):
    assert nf[i].shape == nf[i+1].shape == min(self.batch_size, data_size), dim, self.n_feat
```
or something..?

and similarly for `test_DirectedGraphSAGELinkGenerator_unsupervisedSampler_sample_generation`

tests/mapper/test_link_mappers.py
```diff
@@ -783,3 +783,390 @@ def test_Attri2VecLinkGenerator_unsupervisedSampler_sample_generation(self):
...
+    """
+    Tests of GraphSAGELinkGenerator class
+    """
+
+    n_feat = 4
+    batch_size = 2
+    in_samples = [2, 4]
+    out_samples = [6, 8]
+
+    def test_LinkMapper_constructor(self):
```
kjun9: I think this could just be
```suggestion
    def test_constructor(self):
```

tests/mapper/test_link_mappers.py
```diff
@@ -783,3 +783,390 @@ def test_Attri2VecLinkGenerator_unsupervisedSampler_sample_generation(self):
...
+            batch_size=self.batch_size,
+            in_samples=self.in_samples,
+            out_samples=self.out_samples,
+        )
+        mapper = generator.flow(G.edges(), edge_labels)
+        assert generator.batch_size == self.batch_size
+        assert mapper.data_size == G.number_of_edges()
+        assert len(mapper.ids) == G.number_of_edges()
+
+    def test_DirectedGraphSAGELinkGenerator_1(self):
```
kjun9: How about:
```suggestion
    def test_batch_feature_shapes(self):
```

tests/mapper/test_link_mappers.py
```diff
@@ -783,3 +783,390 @@ def test_Attri2VecLinkGenerator_unsupervisedSampler_sample_generation(self):
...
+                assert nf[ii].shape == (
+                    min(self.batch_size, data_size),
+                    self.out_samples[0] * self.out_samples[1],
+                    self.n_feat,
+                )
+
+        with pytest.raises(IndexError):
+            nf, nl = mapper[2]
+
+    def test_DirectedGraphSAGELinkGenerator_shuffle(self):
```
kjun9: I think the word `DirectedGraphSAGELinkGenerator` can be removed from pretty much all of these test methods - it's already effectively namespaced by the class name
```suggestion
    def test_shuffle(self):
```

kjun9: Instead of the nested function in here, you could use `@pytest.mark.parametrize`
```suggestion
    @pytest.mark.parametrize("shuffle", [True, False])
    def test_shuffle(self, shuffle):
```

tests/mapper/test_link_mappers.py
```diff
@@ -783,3 +783,390 @@ def test_Attri2VecLinkGenerator_unsupervisedSampler_sample_generation(self):
...
+                e2 = edges[nl[1]]
+                assert nf[0][0, 0, 0] == e1[0]
+                assert nf[1][0, 0, 0] == e1[1]
+                assert nf[0][1, 0, 0] == e2[0]
+                assert nf[1][1, 0, 0] == e2[1]
+
+        test_edge_consistency(True)
+        test_edge_consistency(False)
+
+    def test_DirectedGraphSAGELinkGenerator_not_Stellargraph(self):
```
kjun9: IMO I don't think this test is necessary and could be removed to reduce the amount of code here... or to me it doesn't seem any more valuable than replacing this with:
```python
    def test_not_stellargraph(self):
        with pytest.raises(TypeError):
            DirectedGraphSAGELinkGenerator(None, ...)
```

tests/mapper/test_link_mappers.py
```diff
@@ -783,3 +783,390 @@ def test_Attri2VecLinkGenerator_unsupervisedSampler_sample_generation(self):
...
+        data_size = G.number_of_edges()
+        edge_labels = [0] * data_size
+
+        mapper = DirectedGraphSAGELinkGenerator(
+            G, batch_size=self.batch_size, in_samples=[0], out_samples=[0],
+        ).flow(G.edges(), edge_labels)
+
+        assert len(mapper) == 2
+
+        for ii in range(1):
```
kjun9: Maybe you meant to do:
```suggestion
        for ii in range(len(mapper)):
```
unless you really wanted to only check the first element, which means we don't need the for loop at all
```python
nf, nl = mapper[0]
...
```

tests/mapper/test_link_mappers.py
```diff
@@ -783,3 +783,390 @@ def test_Attri2VecLinkGenerator_unsupervisedSampler_sample_generation(self):
...
+            nf, nl = mapper[ii]
+            assert len(nf) == 2 ** (len([0]) + 2) - 2
+            for j in range(len(nf)):
+                if j < self.batch_size:
```
kjun9: I'm a bit confused here... how does `batch_size` relate to `j`? I'm thinking it's supposed to be:
```suggestion
                if j < 2:
```
since the first two elements of `nf` are the "head" nodes?

and maybe looks a bit clearer like this:
```python
batch_dim = min(self.batch_size, data_size)

# head nodes
for f in nf[:2]:
    assert f.shape == (batch_dim, 1, self.n_feat)

# neighbours
for f in nf[2:]:
    assert f.shape == (batch_dim, 0, self.n_feat)
```

tests/mapper/test_link_mappers.py
```diff
@@ -783,3 +783,390 @@ def test_Attri2VecLinkGenerator_unsupervisedSampler_sample_generation(self):
...
+
+        with pytest.raises(TypeError):
+            DirectedGraphSAGELinkGenerator(
+                G,
+                batch_size=self.batch_size,
+                in_samples=self.in_samples,
+                out_samples=self.out_samples,
+            )
+
+    def test_DirectedGraphSAGELinkGenerator_zero_samples(self):
```
kjun9: maybe this would be a bit more descriptive and distinguishable from the `no_samples` test?
```suggestion
    def test_zero_dim_samples(self):
```

tests/mapper/test_link_mappers.py
```diff
@@ -783,3 +783,390 @@ def test_Attri2VecLinkGenerator_unsupervisedSampler_sample_generation(self):
...
+                else:
+                    assert nf[j].shape == (
+                        min(self.batch_size, data_size),
+                        0,
+                        self.n_feat,
+                    )
+            assert len(nl) == min(self.batch_size, data_size)
+            assert all(nl == 0)
+
+    def test_DirectedGraphSAGELinkGeneratorr_no_samples(self):
```
kjun9: maybe we can parameterize this one too: 
```suggestion
    @pytest.mark.parametrize("samples", [([], []), ([], [0]), ([0], [])])
    def test_no_samples(self, samples):
        ins, outs = samples
        ...
        mapper = DirectedGraphSAGELinkGenerator(
            ..., in_samples=ins, out_samples=outs
        ).flow(...)
        assert len(mapper) == 2
        with pytest.raises(ValueError):
            nf, nl = mapper[0]
```

stellargraph/mapper/sampled_link_generators.py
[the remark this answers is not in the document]
kieranricardo: I see what you mean. Yeah I just noticed it but didn't think it was big enough for its own PR so I fixed it here.

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -511,3 +522,139 @@ def sample_features(self, head_links, batch_num):
...
+
+    Args:
+        G (StellarGraph): A machine-learning ready graph.
+        batch_size (int): Size of batch of links to return.
+        in_samples (list): The number of in-node samples per layer (hop) to take.
+        out_samples (list): The number of out-node samples per layer (hop) to take.
+        seed (int or str), optional: Random seed for the sampling methods.
+    """
+
+    def __init__(self, G, batch_size, in_samples, out_samples, seed=None, name=None):
```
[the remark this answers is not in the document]
kieranricardo: A reason other than forget-fullness?

stellargraph/mapper/sampled_link_generators.py
```diff
@@ -511,3 +522,139 @@ def sample_features(self, head_links, batch_num):
...
+        super().__init__(G, batch_size)
+
+        self.in_samples = in_samples
+        self.out_samples = out_samples
+        self.name = name
```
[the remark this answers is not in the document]
kieranricardo: yep, sure can.

tests/mapper/test_link_mappers.py
kieranricardo: nice!

tests/mapper/test_link_mappers.py
kieranricardo: uh.. thats meant to loop through them all!

tests/mapper/test_link_mappers.py
kieranricardo: nice!

kieranricardo:
thanks for your reviews @huon and @kjun9! i think I've addressed all your points

stellargraph/mapper/sampled_link_generators.py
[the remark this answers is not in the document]
huonw: I don't think any change is too small for a separate PR. For instance, by being in a separate PR this good improvement could've landed immediately (easy to review, easy to merge) rather than getting caught up in the review of the rest of this feature.

There's definitely some overhead of switching things around to be able to make a new PR. One good way to minimise this is to have multiple checkouts of the repo, so you can be working on a big feature in one (or more) directory, and do little changes in another without having to `git stash` or commit and then switch branches, and then switch back. Feel free to ask me if you want some tips about what has worked well for me.

kieranricardo:
@kjun9 nice catch! these both seem like easy fixes, so I'll add them here.

kieranricardo:
@kjun9 I added in the changes you recommended! but I still can't get the demo test accuracy above 67%. I think this might be a limitation of this algorithm on directed cora:

- whether two papers are connected in the undirected citation graph can be pretty accurately determined from node feature and neighbouring node features
- the direction of the link is determined by a temporal element i.e. which paper is first, which I don't think DirectedGraphSAGE can accurately capture

As you suggested we can land this now (with or without the demo) and add an issue to find a suitable demo. What do you think?

kjun9:
@kieranricardo yeah we don't currently have any benchmark for what a directed link prediction task on Cora should produce, so it's hard to say whether the low accuracy is the result of:
* some problem with our implementation or 
* the capability of the algorithm to capture that information or 
* even whether the data itself contains enough information to capture that difference?

Yeah I think we can just take the demo out from the PR and do it in a separate one so we can land this since I think the changes to the generator look good. I'd also prefer to cover any changes to edge splitter with some small tests (e.g. using a different algorithm to find the minimum spanning edges should be validated), so maybe we can take that out for now to a separate PR too?

kieranricardo:
@kjun9 good point! I've created #945 and #946 to cover these and removed the demo + changes to `edge_splitter.py`

Issue #868, opened 2020-02-13
`node-link-importance-demo-gat.ipynb` demo fails with `ValueError: 7 is not in list`


huonw:
### Describe the bug

The last cell of the `node-link-importance-demo-gat.ipynb` demo (the cell starting with `[X, _, A], y_true_all = all_gen[0]`) fails with `ValueError: 7 is not in list` and fails to produce the expected plot.

### To Reproduce

1. Open `demos/interpretability/gat/node-link-importance-demo-gat.ipynb`
2. Run all cells

### Observed behavior

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-37-288bd78e8a98> in <module>
     65     pos,
     66     nodelist=[target_nid],
---> 67     node_size=50 * abs(node_sizes[nodes.index(target_idx)]),
     68     node_color=[colors[nodes.index(target_idx)]],
     69     cmap=cmap,

ValueError: 7 is not in list
```

![image](https://user-images.githubusercontent.com/1203825/74405951-87d7c000-4e82-11ea-96e4-11f15d81645a.png)

### Expected behavior

The cell should work and display a plot similar to the one on `develop`:

![image](https://user-images.githubusercontent.com/1203825/74405940-80181b80-4e82-11ea-8dc0-b79517984abe.png)

### Environment

Operating system: `Darwin-18.6.0-x86_64-i386-64bit`
Python version:
```
3.6.9 (default, Jul 10 2019, 12:25:55) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)]
```
Package versions:
<details>

```
absl-py==0.9.0
alabaster==0.7.12
appdirs==1.4.3
appnope==0.1.0
astor==0.8.1
attrs==19.3.0
Babel==2.8.0
backcall==0.1.0
black==19.10b0
bleach==3.1.0
boto==2.49.0
boto3==1.10.45
botocore==1.13.45
cachetools==4.0.0
certifi==2019.11.28
chardet==3.0.4
Click==7.0
cloudpickle==1.2.2
coverage==4.5.4
coveralls==1.10.0
cycler==0.10.0
decorator==4.4.1
defusedxml==0.6.0
docopt==0.6.2
docutils==0.15.2
doit==0.31.1
entrypoints==0.3
gast==0.2.2
gensim==3.8.1
google-auth==1.10.0
google-auth-oauthlib==0.4.1
[...]

PR #880, opened 2020-02-17
Fix GAT interpretability demo for new StellarGraph order changes

huonw:
This does a small change to `node-link-importance-demo-gat.ipynb` to get it working again after the switch to the new `StellarGraph` class. In particular, the change to the new StellarGraph changed the order in which nodes appear. so:

1. the code that was (effectively) `G.nodes()[7]` was no longer selecting the node with ID `1109199`
2. the code that was translating between IDs and indices was sometimes doing it via the `Gnx` (the NetworkX graph) and sometimes via `G`/`graph_nodes` (the `StellarGraph`) which gave different results.

This switches 1 to starting with the actual node ID of interest, and 2 to always using a single list (`graph_nodes` or `nodes` as appropriate).

The code is quite poor in many ways here (all of the `.index` calls are O(n)), but in the interests of getting this fixed as soon as possible, this tries to keep the changes here minimal and not make any improvements. The work to get this notebook running on CI is likely to be a little intricate/fiddly, because it's quite slow at the moment.

See: #868


huonw:
> The work to get this notebook running on CI is likely to be a little intricate/fiddly, because it's quite slow at the moment.

(I've started this at https://github.com/stellargraph/stellargraph/pull/881, but more work than that seems to be needed...)

PR #884, opened 2020-02-18
Display class name more nicely in `StellarGraph.info`

huonw:
By writing just `type(self)`, the output includes `<class 'stellargraph.core.graph.StellarGraph'>`, so we should pick out the `__name__` specifically get just the `StellarGraph` part.

See: #883

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -758,7 +758,7 @@ class StellarGraph:
         """
         directed_str = "Directed" if self.is_directed() else "Undirected"
         lines = [
-            f"{type(self)}: {directed_str} multigraph",
+            f"{type(self).__name__}: {directed_str} multigraph",
             f" Nodes: {self.number_of_nodes()}, Edges: {self.number_of_edges()}",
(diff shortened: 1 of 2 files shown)


huonw:
> However, an Undirected multigraph is created. I was expecting that Stellargraph will automatically pick the graph type of the networkx graph and make is directed. This may be isn't too big of a deal but it took me a while to sort this out and I think someone using the library the first time may miss this. Meaning, you would assume naturally that converting from one graph class to another would preserve the properties of the original graph.

`StellarGraph(a_networkx_graph)` is going to be deprecated soon, replaced by `StellarGraph.from_networkx(...)`, which does pick up the directness.

> This warning doesn't look pretty and is dangling.

It's disappearing in #891.

> So the names looking better but all else is missing

There's no node or edge types, so there's no more info to display.

huonw:
Because you said:

>StellarGraph.info display string looks better!

I take it you approve the specific change in this PR (and other adjustments can happen in other PRs), so I'm going to merge.

Issue #812, opened 2020-02-06
Consolidate dataset parsing code into datasets module


timpitman:
### Description

Currently, the parsing code for common datasets is sprinkled around (cut-and-pasted) into many different demo notebooks.  BlogCatalog3 has shared parsing code in the data module.

Ideally, datasets could return StellarGraph objects for each dataset with a consistent API.  At the very least, the code for parsing each dataset could live in the specific dataset class, rather than in notebooks.

Datasets:

- [x] AIFB #982
- [x] BlogCatalog3 #888
- [x] CiteSeer #989
- [x] Cora #913
- [x] MovieLens #947
- [ ] PubMedDiabetes #986 

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #888, opened 2020-02-18
Move `load_dataset_BlogCatalog3` to a `BlogCatalog3.load` method

huonw:
This moves the existing `load_dataset_BlogCatalog3` function to be part of the new `datasets` module, and tweaks it to return a `StellarGraph` object instead of a NetworkX graph. The old function now emits deprecation warning and then calls through to the new code (and does a conversion to a NetworkX graph).

This is a start on #812.


timpitman (approved):
Looks good, glad this is going ahead! There may be some overlap with #875 which removes the script and most README content.

huonw:
> There may be some overlap with #875 which removes the script and most README content

Do you mean #889?

Issue #818, opened 2020-02-09
Some demo notebooks can't be tested on CI due to using datasets that have licensing restrictions preventing automatic download


timpitman:
### Describe the bug

Some demo notebooks use datasets that can't be automatically downloaded, generally for licensing reasons:

```
attacks_clustering_analysis.ipynb
ensemble-node-classification-example.ipynb
hateful-twitters-interpretability.ipynb
stellargraph-attri2vec-DBLP.ipynb
```

One possible fix could be to substitute artificial mini datasets that match the general format.

huonw:
We could add these to `stellargraph.datasets` with a `download` method that fails, but still have `def load(self, directory)` methods to parse the dataset given its location (along the lines of #812).

PR #889, opened 2020-02-18
Remove demo scripts that duplicate notebooks (#575)

timpitman:
Over time, StellarGraph has moved to Jupyter notebooks for demos, rather than scripts.

After some discussion, it appears that the demo scripts were generally used for development and debugging, and hence are no longer required.  Removing them will remove the need to add them to CI for #575 and will also reduce our code surface for future refactoring, allowing us to concentrate on demo notebooks.

The only script remaining is: `demos/node-classification/hinsage` - this has no Jupyter notebook equivalent, and also can't be tested on CI due to use of the `YELP` dataset, and is listed with other untestable notebooks in #818.


timpitman:
> Looks good to me 👍
> 
> As for the HinSAGE script, I think we should turn that into a notebook as well even if we can't run it in CI?

#892

PR #891, opened 2020-02-18
Streamline and redocument the new StellarGraph constructor

huonw:
This makes the final changes to the `StellarGraph` constructor that we need to consider it non-experimental. This effectively deprecates the NetworkX form (e.g. `StellarGraph(g_nx, node_features=...)`), replaced by `StellarGraph.from_networkx`, and moves the new form front-and-centre. Summary of changes:

- moves where the Pandas node and edge DataFrames are passed: they were late in the argument list and so required keywords `StellarGraph(nodes=n, edges=e)`, but are now first, and so passing as positional arguments is possible `StellarGraph(n, e)` (keywords still works, though)
- tweaks some argument names to make more sense with the new form (`edge_weight_label` -> `edge_weight_column`), and 
- forces everything other than the nodes and edges arguments to be passed using keyword arguments, e.g. `StellarGraph(n, e, True)` is an error, it has to be `StellarGraph(n, e, is_directed=True)`. This gives us more flexibility for changes in future.
- entirely rewrites the class documentation to reflect the new form; rendered docs (linked here for convenience): https://stellargraph--891.org.readthedocs.build/en/891/api.html#stellargraph.core.StellarGraph
- removes the experimental warning from the code (and from the docs)

The NetworkX form does not yet give a `DeprecationWarning`, because that would cause a significant chunk of our tests and examples to emit these warnings; once we've made progress on #717, we can add a proper warning here to encourage people to migrate.

**Is changing the argument order, and forcing keyword arguments safe (that is, would it break any existing code)?** I think it is safe. All of our tests and examples use keyword arguments for everything except the graph (first argument), so these do not break, and it seems like this has encouraged all of our users to do the same: based on a moderate sized sample, no (open source) code does `StellarGraph(g_nx, True, "weight_label", "node_type")` or similar. I did this sample by:

- searching in our dependent code: https://github.com/stellargraph/stellargraph/network/dependents?package_id=UGFja2FnZS0xNjI2NTY5NDU%3D
- reviewing the first 10 or so pages of a GitHub search for `StellarGraph` https://github.com/search?q=StellarGraph+-is%3Afork+-org%3Astellargraph&type=Code

#893 does some similar work to clean-up `from_networkx`.

See: #723


stellargraph/core/graph.py
```diff
@@ -41,166 +41,232 @@
...
+        nodes=None,
+        edges=None,
+        *,
         is_directed=False,
-        edge_weight_label="weight",
-        node_type_name=globalvar.TYPE_ATTR_NAME,
-        edge_type_name=globalvar.TYPE_ATTR_NAME,
+        source_column=globalvar.SOURCE,
+        target_column=globalvar.TARGET,
+        edge_weight_column=globalvar.WEIGHT,
```
huonw: I've named this argument `edge_weight_column` rather than `weight_column` so that it's not ambiguous about being a "weight" for nodes  (e.g. a class weight for machine learning, or a scaling parameter for node features or something), however this means it's somewhat inconsistent with `source_column` and `target_column`.

I see our options as:

1. leave it like this (`source_column`, `target_column`, `edge_weight_column`)
2. put `edge` on everything (`edge_source_column`, `edge_target_column`, `edge_weight_column`)
3. leave `edge` off everything (`source_column`, `target_column`, `weight_column`)
4. some other names that are magically not ambiguous?!

Any thoughts this?

Also, is it worth being consistent between this and `from_networkx`? After this change, `from_networkx`'s equivalent argument is different (`edge_weight_label` here, and after #893, `edge_weight_attr`). Does that matter?

kjun9: Option 1 doesn't occur to me as being particularly inconsistent - looks fine IMO!

As for the consistency between this and networkx, I see the parameters as referring to some property in the "input" rather than the result, so I think `column` makes sense here since we're referring to a column in the input dataframe, whereas in networkx it's an attribute in the input networkx graph.

stellargraph/core/graph.py
```diff
@@ -41,166 +41,232 @@
...
+    Edges of different types work in the same way. example instance, if edges have different types based
+    on their orientation::
+
+        horizontal_edges = pd.DataFrame(
+            {"source": ["a", "c"], "target": ["b", "d"]}, index=[0, 2]
+        )
+        vertical_edges = pd.DataFrame(
+            {"source": ["b", "d"], "target": ["c", "a"]}, index=[1, 3]
+        )
+        diagonal_edges = pd.DataFrame({"source": ["a"], "target": ["c"]}, index=[4])
```
kjun9: Is it important that there's no collision between edge indices across different edge types? I was under the impression that it wasn't, but this example kind of suggests that it is

stellargraph/core/graph.py
```diff
@@ -41,166 +41,232 @@
...
+        edge_weight_column=globalvar.WEIGHT,
         node_type_default=globalvar.NODE_TYPE_DEFAULT,
         edge_type_default=globalvar.EDGE_TYPE_DEFAULT,
+        dtype="float32",
+        # legacy arguments:
+        graph=None,
+        node_type_name=globalvar.TYPE_ATTR_NAME,
+        edge_type_name=globalvar.TYPE_ATTR_NAME,
         feature_name=globalvar.FEATURE_ATTR_NAME,
         target_name=globalvar.TARGET_ATTR_NAME,
```
kjun9: were these always unused?

stellargraph/core/graph.py
```diff
@@ -41,166 +41,232 @@
...
-        edge_weight_label="weight",
-        node_type_name=globalvar.TYPE_ATTR_NAME,
-        edge_type_name=globalvar.TYPE_ATTR_NAME,
+        source_column=globalvar.SOURCE,
+        target_column=globalvar.TARGET,
+        edge_weight_column=globalvar.WEIGHT,
         node_type_default=globalvar.NODE_TYPE_DEFAULT,
         edge_type_default=globalvar.EDGE_TYPE_DEFAULT,
+        dtype="float32",
+        # legacy arguments:
```
kjun9: As part of #717 if we're going to keep these legacy arguments with a deprecation warning, maybe we can consider using `**kwargs` instead - I'd imagine it to have a noticeable impact on usability if auto-complete sort of thing shows all these deprecated parameters with names that kind of sound like they might be relevant.

stellargraph/core/graph.py
```diff
@@ -41,166 +41,232 @@
...
-            an explicit type.
-
-        node_features: str, dict, list or DataFrame optional (default=None)
-            This tells StellarGraph where to find the node feature information
-            required by some graph models. These are expected to be
-            a numeric feature vector for each node in the graph.
-
-        nodes: DataFrame or dict of hashable to DataFrame
-            Features for every node in the graph. Any columns in the dataframe are taken as numeric
+        nodes (DataFrame or dict of hashable to Pandas DataFrame):
```
timpitman: technically this is optional?

stellargraph/core/graph.py
timpitman: I think option 1. `edge_weight_column` is good, since a node could have a weight, but not a source or target.

stellargraph/core/graph.py
```diff
@@ -41,166 +41,232 @@
...
         target_name=globalvar.TARGET_ATTR_NAME,
+        edge_weight_label=None,
         node_features=None,
-        dtype="float32",
-        nodes=None,
-        edges=None,
-        source_column=globalvar.SOURCE,
-        target_column=globalvar.TARGET,
     ):
+        import networkx
```
timpitman: If we're importing networkX here, then we have a dependancy on networkX - should it just be imported normally for now?  (presumably it ends up only imported in fromNetworkX / toNetworkX once #717 is completed.)

stellargraph/core/graph.py
timpitman: Agree - eg the init signature in the docs:

kjun9: > Just so I understand completely, you are not suggesting doing it here, but instead doing it later when we add a deprecation warning to this function as part of #717? (Either is fine by me, I just want to understand your suggestion.)

ah yeah I meant it as a suggestion for something to do in #717 once we stop using those arguments in our examples

stellargraph/core/graph.py
kjun9: Ah I see - is it worth an explicit mention here then? Or you intend for this to be not the case soon?

stellargraph/core/graph.py
kjun9: Haha thanks for tracking down the history, yeah it sounds like we could just remove them here

stellargraph/core/graph.py
huonw: I'll add a note about it. I don't personally have any intentions for relaxing it, but I guess it might be a good idea 🤔

stellargraph/core/graph.py
huonw: Ok, I'll leave it as is (option 1). 👍

stellargraph/core/graph.py
huonw: Ok, I added a note to #717 so we remember to do this.

stellargraph/core/graph.py
huonw: I'm mimicking the rest of the file: there's two other places that `import networkx` (`from_networkx` and `to_networkx` as you say) inside functions. The thinking with it doing it here is that it can turn into:

```python
try:
    import networkx
except ImportError:
    networkx = None
```

and then later code can do `if networkx is not None:` for all handling/detection of the deprecated case (`isinstance(nodes, networkx.Graph)` will always be false if the `networkx` package doesn't exist).

In any case, this is unfortunately not the only place that forces us to non-optionally depend on networkx: `stellargraph/data/edge_splitter.py` and `stellargraph/data/epgm.py` seem to use it in intricate ways.

stellargraph/core/graph.py
huonw: Done.

PR #893, opened 2020-02-19
Standarize from_networkx on 'attr' for attributes, and other tweaks

huonw:
This tweaks `from_networkx` to be ready for "prime time":

- similar to #891, switches everything except the graph to force keyword arguments
- switches to using `attr` in the parameter name for parameters that refer to one of the pieces of data on a NetworkX elements (that is, one of the values inside the data dictionary of a node or edge), rather than inconsistently `name` or `label`. This matches NetworkX's use of node/edge attribute: https://networkx.github.io/documentation/stable/reference/glossary.html

Link to rendered docs, for convenience: https://stellargraph--893.org.readthedocs.build/en/893/api.html#stellargraph.core.StellarGraph.from_networkx

Related to the new StellarGraph clean-up/documentation work: #723


stellargraph/core/graph.py
```diff
@@ -279,52 +283,75 @@ def from_networkx(
...
+            edge_type_name: Deprecated, use ``edge_type_attr``.
+            edge_weight_label: Deprecated, use ``edge_weight_attr``.
+
         Returns:
             A ``StellarGraph`` (if ``graph`` is undirected) or ``StellarDiGraph`` (if ``graph`` is
             directed) instance representing the data in ``graph`` and ``node_features``.
         """
+        if node_type_name is not None:
+            warnings.warn(
+                "the 'node_type_name' parameter is has been replaced by 'node_type_attr'",
```
```suggestion
                "the 'node_type_name' parameter has been replaced by 'node_type_attr'",
```
and similarly for the two below

huonw:
> the rename to be more consistent with networkX makes sense, though could the deprecated parameters be in kwargs as suggested in #891?

My inclination is that there's few enough here that they don't make the signature that much worse (and, these only need to be here for a single release IMO), that it's not worth the risk and (mild) complexity of managing `**kwargs`.

huonw:
> Thoughts?

I merged, but we can tweak as required separately, if you do have thoughts. 😄

PR #897, opened 2020-02-19
Test for warning about from_networkx 0-vector feature default

huonw:
When reading features from an attribute in a node, the `StellarGraph.from_networkx` functionality will implicitly fill in a 0 vector for nodes that are missing the attribute. It does this with a warning, and so to verify that this is behaving correctly and to avoid the warning appearing in our test logs (for example: https://buildkite.com/stellar/stellargraph-public/builds/1575#64220409-0450-476e-83c3-698926b32c10/254-413), we can and should check for it using [`pytest.warns`](https://docs.pytest.org/en/latest/warnings.html#warns).

### DIFF
--- a/tests/core/test_graph.py
+++ b/tests/core/test_graph.py
@@ -924,15 +924,19 @@ def test_from_networkx_smoke():
     g.add_edge(1, 2, edge_label="Y")
     g.add_edge(1, 1)
 
-    from_nx = StellarGraph.from_networkx(
-        g,
-        edge_weight_label="weight_attr",
-        node_type_name="node_label",
-        edge_type_name="edge_label",
-        node_type_default="b",
-        edge_type_default="X",
-        node_features="features",
-    )
+    with pytest.warns(
+        UserWarning,
+        match=r"found the following nodes \(of type 'b'\) without features, using 4-dimensional zero vector: 2",
+    ):
+        from_nx = StellarGraph.from_networkx(
+            g,
+            edge_weight_label="weight_attr",
+            node_type_name="node_label",
+            edge_type_name="edge_label",
+            node_type_default="b",
+            edge_type_default="X",
+            node_features="features",
+        )
 
     raw = StellarGraph(
         nodes={


huonw:
1. Node 1 is a different node type (`node_label="a"`, vs. the default of `"b"`), so can have features of a different dimension.
2. StellarGraph features can only be numeric at the moment (I think this was true even before the change to the new StellarGraph).

huonw:
Also, for 1, the list of features is a n-dimensional feature vector. That is, for each node type T in the graph, there's a dimension d<sub>T</sub> such that every node has an associated R<sup>d<sub>T</sub></sup> vector (In that example/test, we have d<sub>a</sub> = 1 and d<sub>b</sub> = 4.).

The graph's nodes can be thought of as collection of matrices R<sup>n<sub>T</sub> &times; d<sub>T</sub></sup> for each node type T, where n<sub>T</sub> is the number of nodes of that type. These matrices can be passed explicitly as a dictionary of data frames (each with n<sub>T</sub> rows, and d<sub>T</sub> columns), or implicitly by having a vector stored as an attribute in each node, which is what that test is doing.

habiba-h:
Thanks for the clarification.  I still get mixed up with `node_label` and `node_type` terminology. 

> StellarGraph features can only be numeric at the moment (I think this was true even before the change to the new StellarGraph)
Basically `networkx` has no problems with the type of feature values. We stuck to numeric features but don't remember if I tried creating a `stellargraph` object from `networkx` where features were non-numeric.

huonw:
You're correct that NetworkX can store anything, and so can Pandas and Numpy, however I was specifically thinking of places in the old StellarGraph that had some explicit casts to numeric types, and flagged the errors. This is visible in:

```python
import stellargraph as sg
import networkx as nx
import pandas as pd

gnx = nx.Graph()
gnx.add_node(0)
gnx.add_node(1)
# throws exception:
G = sg.StellarGraph(gnx, node_features=pd.DataFrame(["foo", "bar"], index=[0, 1]))
```

```
ValueError                                Traceback (most recent call last)
~/projects/stellargraph/stellargraph-4/stellargraph/core/graph_networkx.py in _convert_from_node_data(data, node_type_map, node_types, dtype)
    187                 try:
--> 188                     data_arr = arr.values.astype(dtype)
    189                 except ValueError:

ValueError: could not convert string to float: 'foo'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-27-9e0f09199cb0> in <module>
     12 gnx.add_node(0)
     13 gnx.add_node(1)
---> 14 G = sg.StellarGraph(gnx, node_features=pd.DataFrame(["foo", "bar"], index=[0, 1]))
     15 G.node_features([0, 1])

...

~/projects/stellargraph/stellargraph-4/stellargraph/core/graph_networkx.py in _convert_from_node_data(data, node_type_map, node_types, dtype)
    189                 except ValueError:
    190                     raise ValueError(
--> 191                         "Node data passed as Pandas arrays should contain only numeric values"
    192                     )
    193 

ValueError: Node data passed as Pandas arrays should contain only numeric values
```

However, it seems, potentially accidentally, storing features in nodes didn't have the same check, and so did in fact work:

```python
gnx = nx.Graph()
gnx.add_node(0, feature="foo")
gnx.add_node(1, feature="bar baz")
# all ok:
G = sg.StellarGraph(gnx, node_features="feature")
G.node_features([0, 1]) # array(['foo', 'bar baz'], dtype=object)
```

However, I don't think this would be particularly useful with the current library, since it wouldn't be able to feed into any (I think?) of our models.

(The above tests were done by `git checkout 172edad0` to go back to 172edad0, the last commit before 2dfff02d which removed the networkx-based StellarGraph.)

PR #899, opened 2020-02-19
Store counts of edge types in graph schema

huonw:
WIP.


huonw:
I don't think this is worth the effort, which could be better spent elsewhere.

Issue #862, opened 2020-02-13
ComplEx: reproduce the paper results


huonw:
### Description

In #848, the algorithm is tested to match the paper's description, but not tested to exactly match the paper's reported results on the WN18 or FB15k datasets. We should do this validation.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #865, opened 2020-02-13
Implement DistMult

huonw:
This implements the DistMult algorithm from Yang, B., Yih, W.T., He, X., Gao, J. and Deng, L., 2014. Embedding entities and relations for learning and inference in knowledge bases https://arxiv.org/abs/1412.6575

This is very similar to ComplEx, but for now I'm not factoring out any commonalities per the ["rule of three"](https://andrewbrookins.com/technology/the-rule-of-three/):

> when reusing code, copy it once, and only abstract the third time

That is, once we add a third algorithm, we'll have a better idea of what is actually useful common code, and what is just coincidental sharing.

See: #755

PR #901, opened 2020-02-20
Add a notebook that reproduces the results from the ComplEx paper

huonw:
This adds a notebook that reproduces the results from the ComplEx paper (http://jmlr.org/proceedings/papers/v48/trouillon16.pdf), in Table 2 in particular:

image

In particular, the notebook does an evaluation of the "raw" ranks of a edge `(s, r, o)` (source, relation type, object) by comparing the model's predicted score for that edge against the scores for all "mutated-source" versions of that link `(n, r, o)` (for every node `n`), and for all "mutated-object" versions `(s, r, n)`. It does this on the WN18 and FB15k datasets.

The notebook is phrased as a bit of a tutorial for WN18, but omits the description for FB15k, since the code is otherwise identical. I decided against defining a function to do all the work for pedagogical reasons: it's easier to interleave the discussion of the code (and the results) if it's not all in a single big function that is used for both WN18 and FB15k. (This somewhat relates to #910; this notebook is both a how-to and a paper reproduction.)

**Why only raw ranks?** The paper (like most knowledge graph papers) focuses more on the "filtered" ranks, which are ranking a link `(s, r, o)` against only the mutated-source and mutated-object instances that aren't known edges (that is, that don't appear in the train, test or validation set). However, as yet, I've been unable to reproduce the large jump in performance one sees when comparing only these edges, despite trying (and manually validating) several methods of ignoring the known edges. (Follow-up issue: #1060)

**Why so much custom ranking-computing code?** The naive approach to implement this validation would be to create DataFrames storing all of the mutated-source and mutated-object links and running `model.predict` to get a score for each of the mutated edges. This would mean 2 (mutating source and object, separately) &times; 5000 (number of edges in the test set) = 10000 `predict` calls, each of which is on 40943 triples (total number of nodes, to slot in to the mutated source or object column). This is inefficient, and it's much better to work with the underlying embedding matrices directly, to phrase computing scores against all nodes as bulk operations on simple matrices (rather than having to go through embedding layers, etc.). This is implemented in the `ComplEx.rank_edges_against_all_nodes` function, and the optimised form is validated to match the naive/manual form in the unit tests.

**Why work with WN18 and FB15k, since those datasets have test-set leakage of inverse relations?** It's what the paper uses, and so for validating our results against the author, we should match.

See: #862


stellargraph/layer/knowledge_graph.py
```diff
@@ -137,6 +137,96 @@ def embeddings(model):
...
+            # for each column, compare all the scores against the score of the true edge
+            greater = pred > true_scores
+
+            # the raw rank is the number of elements scored higher than the true edge
+            raw_rank = 1 + greater.sum(axis=0)
+            assert raw_rank.shape == (batch_size,)
+
+            return raw_rank
+
+        n, all_r = ComplEx.embeddings(model)
```
kieranricardo: just to make the connection between `n` and `all_r` clear below. Could you change this to:

 ```suggestion
        all_n, all_r = ComplEx.embeddings(model)
```

or maybe:

 ```suggestion
        all_node_embs, all_rel_embs = ComplEx.embeddings(model)
```

stellargraph/layer/knowledge_graph.py
```diff
@@ -137,6 +137,96 @@ def embeddings(model):
...
+            os = n[objects, :]
+
+            # reproduce the scoring function for ranking the given subject and relation against all
+            # other nodes (objects), and similarly given relation and object against all
+            # subjects. The bulk operations give speeeeeeeeed.
+            # (num_nodes x k, batch_size x k) -> num_nodes x batch_size
+            mod_o_pred = np.inner(n_conj, ss * rs).real
+            mod_s_pred = np.inner(n, rs * os.conj()).real
+
+            mod_o_raw = ranks(mod_o_pred, true_ilocs=objects, true_is_source=True)
```
kieranricardo: `true_is_source` seems a bit confusing to me. when `true_is_source=True` the `true_ilocs` are the `objects`, when it seems to me that it should be the other way around. Can you explain why you chose this?

stellargraph/layer/knowledge_graph.py
```diff
@@ -137,6 +137,96 @@ def embeddings(model):
...
+            return raw_rank
+
+        n, all_r = ComplEx.embeddings(model)
+        n_conj = n.conj()
+
+        raws = []
+
+        # run through the batches and compute the ranks for each one
+        num_tested = 0
+        for i in range(len(test_data)):
```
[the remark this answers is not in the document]
huonw: Huh, I thought not, but yes, it can be.

stellargraph/layer/knowledge_graph.py
huonw: Yeah, no need to be so brief. I'll go with the second 👍

stellargraph/layer/knowledge_graph.py
huonw: It's only relevant for the filtered ranks, so I'm going to remove it from this PR... however, maybe this is why the filtering wasn't behaving as expected.

huonw:
> Just to be clear, is this meant to remove the experimental status from ComplEx or are you waiting to reproduce the filtered metrics?

Yeah, I think it'd be better to reproduce the filtered metrics before calling this done, because they're what almost all papers use. I marked #1060 as `experimental blocker` for this reason, but we can remove that/revise if we think it's not worth it. However, given our discussion and your better results with https://github.com/stellargraph/stellargraph/compare/feature/862-complex-notebook...feature/complex-filtered-scores maybe this can be done sooner rather than later.

stellargraph/layer/knowledge_graph.py
```diff
@@ -137,6 +137,96 @@ def embeddings(model):
...
+        """
+
+        if not isinstance(test_data, KGTripleSequence):
+            raise TypeError(
+                "test_data: expected KGTripleSequence; found {type(test_data).__name__}"
+            )
+
+        num_nodes = known_edges_graph.number_of_nodes()
+
+        def ranks(pred, true_ilocs, true_is_source):
```
kieranricardo: I think this function could be pulled out of `ComplEx` and re-used for all our other knowledge graph algorithms. But this function is likely the main experimental blocker so I'm happy for this do be done as part of #1060

demos/link-prediction/knowledge-graphs/complex.ipynb
```diff
@@ -0,0 +1,708 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "wn18_raw_ranks = ComplEx.rank_edges_against_all_nodes(\n",
+    "    wn18_model, wn18_gen.flow(wn18_test), wn18_g\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "code",
```
kieranricardo: Minor optional thing, but the mis-alignment of the tables below is kinda jarring. Maybe you make and print out a dataframe with the papers results e.g.

```
pd.DataFrame({
    "mrr": [0.587], 
    "filtered mrr": [0.941], 
    "hits at 1": [0.936], 
    "hits at 3": [0.945],
    "hits at 10": [0.947],
})
```

demos/link-prediction/knowledge-graphs/complex.ipynb
```diff
@@ -0,0 +1,708 @@
...
+    "summarise({\"raw\": wn18_raw_ranks})"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For comparison, Table 2 in the paper gives the following results for WN18:\n",
+    "\n",
+    "| raw MRR | filtered MRR | hits at 1 | hits at 3 | hits at 10 |\n",
```
kieranricardo: I think we should mention that these are filtered hits

```suggestion
    "| raw MRR | filtered MRR | filtered hits at 1 | filtered hits at 3 | filtered hits at 10 |\n",
```

stellargraph/layer/knowledge_graph.py
huonw: Similar to #865, I'd prefer to not start abstracting before we're more sure about what's useful, but yes, this is definitely likely to be able to be shared in some form.

demos/link-prediction/knowledge-graphs/complex.ipynb
huonw: I've combined this with the suggestion of including these results as a dataframe, to have it have the raw and filtered numbers on separate rows.

PantelisElinas:
Notebook looks good and it is great to see the results from the paper confirmed. I only have a couple of minor suggestions to make.

I think using early stopping with low patience, e.g., 10, and increasing the number of training epochs to something like 200 would be better than assuming 20 epochs are sufficient. 

When you create the generators, you set the batch size as follows `batch_size=len(wn18_train) // 100`. If your goal is to have a batch size of 100 then you should set it explicitly `batch_size=100` as is common practice.

I would rename variable `wn18_g` and `fb15k_g` to `wn18_graph` and `fb15k_graph` respectively for clarity. Just making it a bit more obvious that this is the knowledge graph.

You should be able to calculate the filtered metrics by first calculating the raw metrics and then counting the number of known triplets (those in the data) that rank higher than the test triplet and subtracting from the test triplet rank.

huonw:
@PantelisElinas Is there a reason you didn't comment on the file itself either on the JSON or on ReviewNB? Is there a process/tooling thing that can be optimised?

> I think using early stopping with low patience, e.g., 10, and increasing the number of training epochs to something like 200 would be better than assuming 20 epochs are sufficient.

👍 

> When you create the generators, you set the batch size as follows batch_size=len(wn18_train) // 100. If your goal is to have a batch size of 100 then you should set it explicitly batch_size=100 as is common practice.

Per the comment on that line, that is not what that code is doing, and if we wanted 100 elements per batch, of course `batch_size=100` would be better. The goal is to have 100 batches per epoch, because that's what the paper uses. On the page with Table 2:

> We also tried varying the batch size but this had no impact and we settled with 100 batches per epoch

I could hardcode the size based on the known size of the data (e.g. `batch_size=1414` for WN18), but that seems to be much less clear.

> I would rename variable wn18_g and fb15k_g to wn18_graph and fb15k_graph respectively for clarity. Just making it a bit more obvious that this is the knowledge graph.

👍 

> You should be able to calculate the filtered metrics by first calculating the raw metrics and then counting the number of known triplets (those in the data) that rank higher than the test triplet and subtracting from the test triplet rank.

Unfortunately, per the PR description, I tried various ways to compute it and none of them were working correctly so restating a definition doesn't lend much light. @kieranricardo did some investigations and wrote some code that did seem to be working better (see https://github.com/stellargraph/stellargraph/pull/901#issuecomment-597999484 above), so maybe I just had bugs in all of the version I'd tried.

PantelisElinas:
I don't have ReviewNB setup, so I didn't use it.

I misunderstood what you were trying to do with the batch size so it is fine as it is then.

Issue #898, opened 2020-02-19
Many duplicate 'plot_history' functions in demos


huonw:
### Description

According to `git grep --files-with-matches 'def plot_history'`, there's 21 notebooks that define a `plot_history` function:

- `demos/calibration/calibration-pubmed-link-prediction.ipynb`
- `demos/calibration/calibration-pubmed-node-classification.ipynb`
- `demos/ensembles/ensemble-link-prediction-example.ipynb`
- `demos/ensembles/ensemble-node-classification-example.ipynb`
- `demos/interpretability/gat/node-link-importance-demo-gat.ipynb`
- `demos/interpretability/gcn/hateful-twitters-interpretability.ipynb`
- `demos/interpretability/gcn/node-link-importance-demo-gcn-sparse.ipynb`
- `demos/interpretability/gcn/node-link-importance-demo-gcn.ipynb`
- `demos/link-prediction/gcn/cora-gcn-links-example.ipynb`
- `demos/link-prediction/graphsage/cora-links-example.ipynb`
- `demos/link-prediction/hinsage/movielens-recommender.ipynb`
- `demos/node-classification/cluster-gcn/cluster-gcn-node-classification.ipynb`
- `demos/node-classification/gat/gat-cora-node-classification-example.ipynb`
- `demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb`
- `demos/node-classification/graphsage/directed-graphsage-on-cora-example.ipynb`
- `demos/node-classification/graphsage/graphsage-cora-node-classification-example.ipynb`
- `demos/node-classification/graphsage/graphsage-pubmed-inductive-node-classification-example.ipynb`
- `demos/node-classification/ppnp/ppnp-cora-node-classification-example.ipynb`
- `demos/node-classification/rgcn/rgcn-aifb-node-classification-example.ipynb`
- `demos/node-classification/sgc/sgc-node-classification-example.ipynb`
- `demos/use-cases/hateful-twitters.ipynb`

Many of these seem to be identical or at least similar. It'd be good to not have quite so much duplication, by moving this functionality into the library.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

Issue #849, opened 2020-02-11
Install neo4j to run on CI


hd-chuong:
### Description
We need to install Neo4J on CI so that neo4j-related code and demos can be thoroughly tested. 

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #902, opened 2020-02-20
Add a 'plot_history' function for viewing metrics across epochs

huonw:
This factors out the `plot_history` function that has been, essentially, copy-and-pasted into every notebook into a function in the library. It places it in `stellargraph.model.plot_history`.

The plot allows for multiple histories (for ensemble models) and validation history (optional). It puts all of the plots in a column, sharing the x-axis (epochs).

This is added to `utils` since I wasn't sure of a better place at the moment (related to #909).

This doesn't rerun all of the notebooks that were updated:

- `rgcn-aifb-node-classification-example.ipynb` fails: #911
- `hateful-twitters.ipynb` fails: #912
- `directed-graphsage-on-cora-neo4j-example.ipynb` and `undirected-graphsage-on-cora-neo4j-example.ipynb`: I haven't got Neo4j running locally (relates to #849 and #904)

See: #898

## Examples

Train and validation history (`gcn-cora-node-classification-example.ipynb`):
image

Train history only (`gcn-cora-node-classification-example.ipynb` with `validation_data=val_gen` removed from the `model.fit_generator` call):
image

Ensemble with multiple training and validation histories (`ensemble-node-classification-example.ipynb`):
image


stellargraph/utils/history.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    # explicit colours are needed if there's multiple train or multiple validation series, because
+    # each train series should have the same color. This uses the global matplotlib defaults that
+    # would be used for a single train and validation series.
+    colors = plt.rcParams["axes.prop_cycle"].by_key()["color"]
+    if color_train is None:
+        color_train = colors[0]
+    if color_validation is None:
+        color_validation = colors[1]
```
kjun9: Should we make sure that the two colours aren't the same? It is pretty unlikely.. but technically if someone wanted a particular colour for `train` but didn't care about the validation colour, it might be kind of convenient.

huonw: Oh, like if they set `color_train="blue"` and `color_validation`'s default is `"blue"` too. That's a good point.

I think I might just remove the colour customisation entirely for now: the default colours are good enough for our examples (e.g. they seem distinguishable for all colourblindness types supported by https://www.color-blindness.com/coblis-color-blindness-simulator/), and I suspect someone wanting different colours is likely to want other customisations too, like changing axis names and fonts and ....

Issue #905, opened 2020-02-20
Long running/hung tests can lock up CI machines indefinitely


huonw:
### Describe the bug

If a test hangs, the CI machines will sit there doing nothing, but fully occupied, e.g. https://buildkite.com/stellar/stellargraph-public/builds/1589 ran for 2 hours on a PDB prompt.

### To Reproduce

1. Run a build that causes tests to take a long time.

### Observed behavior

https://buildkite.com/stellar/stellargraph-public/builds/1589 ran for 2 hours before I cancelled it.

<img width="1169" alt="image" src="https://user-images.githubusercontent.com/1203825/74974894-2616e780-547a-11ea-8e25-099e141c135f.png">

### Expected behavior

Very slow builds should time out after a while.

### Environment

CI, commit on develop: https://github.com/stellargraph/stellargraph/commit/928f80b33f8567165d3900c6043c3667f4326a46

### Additional context

N/A

PR #906, opened 2020-02-20
Add a timeout to every CI step to halt hung builds

huonw:
If a step in build hangs or takes an unusually long time, previously CI would let it continue, occupying machines forever. In lieu of a global timeout (https://github.com/buildkite/feedback/issues/170, https://forum.buildkite.community/t/pipeline-timeouts/722), we can manually apply a timeout to every step, as a last resort to catch slow/hung builds. This uses the [`timeout_in_minutes`](https://buildkite.com/docs/pipelines/command-step#command-step-attributes) optional attribute:

> The number of minutes a job created from this step is allowed to run. If the job does not finish within this limit, it will be automatically canceled and the build will fail. 

Our steps currently range from 30 seconds to 8 minutes, so 30 minutes should be a safe "something serious is wrong" time-out.

See: #905

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -3,7 +3,7 @@ timeout: &timeout
   # steps can run for half an hour, maximum: if they haven't finished by then, there's probably some
   # deeper problem. A global timeout would be better
   # (https://github.com/buildkite/feedback/issues/170).
-  timeout_in_minutes: 30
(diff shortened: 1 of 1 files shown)


huonw:
Example build with timeouts reduced to 2 minutes: https://buildkite.com/stellar/stellargraph-public/builds/1596

image

image

PR #913, opened 2020-02-21
Add datasets.Cora.load to package up parsing and StellarGraph-ing

huonw:
This adds a `load()` function to the `Cora` dataset loader, that returns a `StellarGraph` with the nodes (and features) and edges, and a `pandas.Series` with the node labels. This is rolled out to several of the notebooks that use the Cora dataset, but not all, because some of them do more than just create the graph and so it's a harder migration.

Part of: #812


stellargraph/datasets/datasets.py
```diff
@@ -43,7 +43,29 @@ class Cora(
     "indicating the absence/presence of the corresponding word from the dictionary. The dictionary consists of 1433 unique words.",
     source="https://linqs.soe.ucsc.edu/data",
 ):
-    pass
+    def load(self):
```
PantelisElinas: I wonder if it would be useful to offer an option here to only return the largest connected component and also whether the user wants a directed or undirected graph. Even though Cora is a directed graph, it is often used as an undirected one. 

In the current notebooks we load Cora as undirected. I run the GCN node classification notebook a few times and it looks like it performs better, e.g. higher accuracy on the test set, when the graph is undirected. 

Unless it is easy for users to cast a StellarDiGraph object to StellarGraph, we should give them the option in the load method.

huonw: (Worked started on #948 at https://github.com/stellargraph/stellargraph/pull/958.)

stellargraph/datasets/datasets.py
```diff
@@ -44,7 +44,45 @@ class Cora(
...
+        required.
+
+        The node feature vectors are included, and the edges are treated as directed or undirected
+        depending on the ``directed`` parameter.
+
+        Args:
+            directed (bool): if True, return a directed graph, otherwise return an undirected one.
+
+        Returns:
+            A tuple where the first element is the :class:`StellarGraph` object with the nodes, node
```
timpitman: should this mention returning a StellarDiGraph?

Issue #914, opened 2020-02-21
Include link to nbviewer.jupyter.org for result notebooks computed on CI


huonw:
### Description

CI runs notebooks and then uploads the result as a buildkite artifact, e.g. in https://buildkite.com/stellar/stellargraph-public/builds/1614#72d33e68-6b9b-460c-bc52-8c1dcbe58f6c:

![image](https://user-images.githubusercontent.com/1203825/74998197-35168d80-54ac-11ea-93cd-c0b985c5ae61.png)

These artifacts have public URLs (that one is https://buildkite.com/organizations/stellar/pipelines/stellargraph-public/builds/1614/jobs/72d33e68-6b9b-460c-bc52-8c1dcbe58f6c/artifacts/8e1f5178-b1d7-4f2a-b05b-a6503f731538 ) which can be fed into https://nbviewer.jupyter.org to view the rendered results, without having to download and open the notebook, or copying the link into that site manually, e.g.:

https://nbviewer.jupyter.org/urls/buildkite.com/organizations/stellar/pipelines/stellargraph-public/builds/1614/jobs/72d33e68-6b9b-460c-bc52-8c1dcbe58f6c/artifacts/8e1f5178-b1d7-4f2a-b05b-a6503f731538

We could potentially even annotate the build with the links for notebooks that failed.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #915, opened 2020-02-21
Provide convenient link for seeing CI notebooks on nbviewer

huonw:
This parses [the `buildkite-agent artifact upload` output](https://buildkite.com/docs/agent/v3/cli-artifact#uploading-artifacts) to pull out the artifact UUID, and then constructs the appropriate https://nbviewer.jupyter.org/ URL for seeing the notebook rendered.

If the notebook failed, this information is also added as [an annotation](https://buildkite.com/docs/agent/v3/cli-annotate) at the top of the build, with clickable links, so one doesn't need to dig through the logging in the individual step.

(An alternative implementation strategy would be to read the artifact ID using [the artifacts API](https://buildkite.com/docs/apis/rest-api/artifacts), however having a token and JSON parsing code just for this seems somewhat like complexity we don't need.)

See: #914

## Example

https://buildkite.com/stellar/stellargraph-public/builds/1631. Failing notebooks get an annotation

image

and passing notebooks have the URL listed in the logs if someone does want to look at it: https://buildkite.com/stellar/stellargraph-public/builds/1631#cae0ed07-f2c5-411f-a87c-9e24959961a8/987-1001

> This notebook can be viewed at <https://nbviewer.jupyter.org/urls/buildkite.com/organizations/stellar/pipelines/stellargraph-public/builds/1631/jobs/cae0ed07-f2c5-411f-a87c-9e24959961a8/artifacts/3507eef6-0bb3-4037-9ca7-2a0de2501dc9>

### DIFF
--- a/.buildkite/steps/test-demo-notebooks.sh
+++ b/.buildkite/steps/test-demo-notebooks.sh
@@ -59,7 +59,7 @@ filename="$(basename "$f")"
 # can include a link to the latter automatically
 buildkite-agent artifact upload "$filename" 2>&1 | tee agent-output.txt
 
-artifact_id="$(sed -n "s/.*Uploading artifact \\(.*\\) $filename/\\1/p" agent-output.txt)"
+artifact_id="$(sed -n "s/.*Uploading artifact \\(.*\\) $filename.*/\\1/p" agent-output.txt)"
 if [ -z "$artifact_id" ]; then
   echo "failed to find artifact ID; this may be an error in the script ($0)"
   exit 1
--- a/.buildkite/steps/test-demo-notebooks.sh
+++ b/.buildkite/steps/test-demo-notebooks.sh
@@ -53,6 +53,26 @@ exitCode=0
 # papermill will replace parameters on some notebooks to make them run faster in CI
 papermill --execution-timeout=600 --parameters_file "${stellargraph_dir}/.buildkite/notebook-parameters.yml" --log-output "$f" "$f" || exitCode=$?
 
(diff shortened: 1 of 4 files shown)


.buildkite/steps/test-demo-notebooks.sh
```diff
@@ -53,6 +53,26 @@ exitCode=0
...
 
-# and also upload the notebook with outputs, for someone to download and look at
-buildkite-agent artifact upload "$(basename "$f")"
+echo "+++ :jupyter: making result viewable"
+filename="$(basename "$f")"
+# and also upload the notebook with outputs, for someone to view by downloading or via nbviewer; we
+# can include a link to the latter automatically
+buildkite-agent artifact upload "$filename" 2>&1 | tee agent-output.txt
+
+artifact_id="$(sed -n "s/.*Uploading artifact \\(.*\\) $filename.*/\\1/p" agent-output.txt)"
```
TheDen: Would be be better to grep for the UUID format? Something like `grep -oE '[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}'`

huonw: That's a good idea. I went with something slightly more relaxed (just `[0-9a-f-]*`), but combined it with the filename check, because I thought it was good to match the `<UUID> <filename>` bit specifically, in case other UUIDs appear in the output for some reason (e.g. the job ID in a URL if it ends up having to retry something and reports a problem).

PR #916, opened 2020-02-21
Add buildkite annotation when notebook parallelism setting is wrong

huonw:
If the number of notebooks doesn't match the `parallelism` setting used for testing notebooks, all of the steps will fail, and one needs to go into them to deduce what the problem was, and then interpret the message. This PR (a) rewrites the message to be a bit clearer, including mentioning the common case of needing to merge with develop, and (b) adds it as [an annotation](https://buildkite.com/docs/agent/v3/cli-annotate) at the top of the build for convenience.

## Example

https://buildkite.com/stellar/stellargraph-public/builds/1628

image

### DIFF
--- a/.buildkite/steps/test-demo-notebooks.sh
+++ b/.buildkite/steps/test-demo-notebooks.sh
@@ -18,7 +18,9 @@ done < <(find "${stellargraph_dir}/demos" -name "*.ipynb" -print0)
 NUM_NOTEBOOKS_TOTAL=${#NOTEBOOKS[@]}
 
 if [ "$NUM_NOTEBOOKS_TOTAL" -ne "$SPLIT" ]; then
-  msg="Buildkite step parallelism is set to $SPLIT but found $NUM_NOTEBOOKS_TOTAL notebooks. Please update the notebook testing \`parallelism\` setting in \`.buildkite/pipeline.yml\`  to match the number of notebooks!"
+  msg="Buildkite step parallelism is set to $SPLIT but found $NUM_NOTEBOOKS_TOTAL notebooks. Please update the notebook testing \`parallelism\` setting in \`.buildkite/pipeline.yml\`  to match the number of notebooks!
+
+If a pull request adding or removing notebooks was merged recently, this may be fixed by doing a merge with \`develop\`; for example: \`git fetch origin && git merge origin/develop\`."
   echo "$msg"
   # Every step will do this annotation, but they all use the same context and there's no --append
   # flag here, and so they replace each other. Thus, there should only ever be a single instance of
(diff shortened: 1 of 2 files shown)


.buildkite/steps/test-demo-notebooks.sh
```diff
@@ -18,7 +18,14 @@ done < <(find "${stellargraph_dir}/demos" -name "*.ipynb" -print0)
 NUM_NOTEBOOKS_TOTAL=${#NOTEBOOKS[@]}
 
 if [ "$NUM_NOTEBOOKS_TOTAL" -ne "$SPLIT" ]; then
-  echo "Buildkite step parallelism is set to $SPLIT but found $NUM_NOTEBOOKS_TOTAL notebooks. Please update parallelism to match the number of notebooks!"
+  msg="Buildkite step parallelism is set to $SPLIT but found $NUM_NOTEBOOKS_TOTAL notebooks. Please update the notebook testing \`parallelism\` setting in \`.buildkite/pipeline.yml\`  to match the number of notebooks!
+
+If a pull request adding or removing notebooks was merged recently, this may be fixed by doing a merge with \`develop\`; for example: \`git fetch origin && git merge origin/develop\`."
```
kieranricardo: just to clarify, when will merging with develop fix this?

huonw: When a PR adding or removing notebooks was merged into `develop`, and the current build's commit doesn't inherit from that merged commit, e.g.: if the git DAG for `develop` and the PR in question looks like

```
D (develop)
|
|  P (PR)
| /
B
|
older
|
commits
|
...
```

Suppose commit `D` on `develop` adds a notebook and so has `parallelism: 40`, vs. the PR commit `P` that has `parallelism: 39`. When `P` is built against `D`, the `pipeline.yml` will be uploaded from `P` (and so `parallelism: 39`). The notebook steps will be run after doing a transient merge with `D` on the CI machines, the `plugins` anchor pulls in the https://github.com/stellargraph/github-merged-pr-buildkite-plugin plugin:


https://github.com/stellargraph/stellargraph/blob/913cf976dbfadd0320a16535e0285fe3424f8572/.buildkite/pipeline.yml#L82-L89



After doing the merge with `D`, the code being tested for `P` will have 40 notebooks (the merge will bring in that extra notebook), and thus the parallelism setting of 39 will be wrong. If an explicit merge is done and pushed to the PR, the parallelism in the `pipeline.yml` will be correct (40):

```
   M (PR)
 / |
D (develop)
|  | 
|  P
| /
B
|
|
older
|
commits
|
...

```

kieranricardo: ah makes sense, I wasn't aware of the transient merge going on that would pull in extra notebooks from develop 👍

Issue #749, opened 2020-01-30
Test all algorithms with randomness for reproducibility


kjun9:
### Description

Now that we have a reproducibility testing framework setup, we can start adding tests for all algorithms so that we can be sure they are all reproducible

### User Story

**As** a data scientist

**I want to** make reproducible pipelines with a variety of stellargraph algorithms

**So that** I can do incremental work more easily

### Algorithm Checklist

- [ ] GraphSAGE
- [ ] HinSAGE
- [ ] Node2Vec
- [ ] attri2vec
- [ ] Cluster GCN
- [ ] Ensembles

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

thaisneubauer:
What about this reproducibility issue for the UniformRandomMetaPathWalk?

Issue #803, opened 2020-02-06
Tensorflow `Model.*_generator` methods are deprecated


kjun9:
### Description

For example, for `fit_generator` https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit_generator:

> Warning: THIS FUNCTION IS DEPRECATED. It will be removed in a future version. Instructions for updating: Please use Model.fit, which supports generators.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed

PR #925, opened 2020-02-24
Make graphsage link prediction reproducible

kjun9:
Part of #749


tests/reproducibility/fixtures.py
```diff
@@ -23,13 +23,13 @@ def models_equals(model1, model2):
     return all(np.array_equal(w, w_new) for w, w_new in zip(w1, w2))
 
 
-def assert_reproducible(func, num_iter=20):
+def assert_reproducible(func, num_iter=10):
```
kjun9: I've had to reduce this since the tests take too long to run. On my local machine, with `num_iter=20`, the slowest reproducibility tests:
```
====================================== slowest 4 test durations =======================================
25.36s call     tests/reproducibility/test_graphsage.py::test_unsupervised[True]
24.63s call     tests/reproducibility/test_graphsage.py::test_unsupervised[False]
20.32s call     tests/reproducibility/test_graphsage.py::test_link_prediction[False]
18.54s call     tests/reproducibility/test_graphsage.py::test_link_prediction[True]
```

and with `num_iter=10`:
```
====================================== slowest 4 test durations =======================================
14.47s call     tests/reproducibility/test_graphsage.py::test_unsupervised[True]
13.79s call     tests/reproducibility/test_graphsage.py::test_unsupervised[False]
9.69s call     tests/reproducibility/test_graphsage.py::test_link_prediction[False]
9.18s call     tests/reproducibility/test_graphsage.py::test_nai[False]
```
I think 10 should still be enough? There's a lot of setup time in building a tensorflow model that we can't avoid, so we might have to consider some other options later.

tests/reproducibility/fixtures.py
```diff
@@ -23,13 +23,13 @@ def models_equals(model1, model2):
...
     """
     Assert Keras models produced from calling ``func`` are reproducible.
 
     Args:
         func (callable): Function to check for reproducible model
-        num_iter (int, default 20): Number of iterations to run through to validate reproducibility.
+        num_iter (int, default 10): Number of iterations to run through to validate reproducibility.
```
huonw: FWIW, I think it's fine to not list the defaults in doc strings, since documentation tools (even the basic `help(...)`) will list them in the function definition, e.g. `help(np.array)` starts with `array(object, dtype=None, copy=True, order='K', subok=False, ndmin=0)`.

```suggestion
        num_iter (int): Number of iterations to run through to validate reproducibility.
```

tests/reproducibility/test_graphsage.py
```diff
@@ -145,6 +145,67 @@ def gs_nai(
...
+    epochs=4,
+    bias=True,
+    dropout=0.0,
+    normalize="l2",
+    seed=0,
+    shuffle=True,
+):
+    set_seed(seed)
+    tf.random.set_seed(seed)
+    if shuffle:
```
huonw: Why does `shuffle == True` require more than just `stellargraph.random.set_seed`? (I know this applies to other models in this file too, but I apparently missed asking about it until now.)

tests/reproducibility/test_graphsage.py
```diff
@@ -145,6 +145,67 @@ def gs_nai(
...
+        random.seed(seed)
+
+    generator = GraphSAGELinkGenerator(g, batch_size, num_samples)
+    train_gen = generator.flow(edge_ids, edge_labels, shuffle=True)
+
+    model = gs_link_pred_model(
+        num_samples, generator, optimizer, bias, dropout, normalize
+    )
+
+    model.fit_generator(
```
huonw: In light of #803, maybe we shouldn't add new code that uses `fit_generator`?

tests/reproducibility/test_graphsage.py
```diff
@@ -145,6 +145,67 @@ def gs_nai(
...
+        train_gen,
+        epochs=epochs,
+        verbose=1,
+        use_multiprocessing=False,
```
huonw: Based on the docs, I suspect `workers=4` is ignored if `use_multiprocessing=False`:

> workers: Integer. Used for generator or keras.utils.Sequence input only. Maximum number of processes to spin up **when using process-based threading**. ...
>
> use_multiprocessing: Boolean. Used for generator or keras.utils.Sequence input only. **If True, use process-based threading**. ...

(Emphasis mine)

tests/reproducibility/test_graphsage.py
kjun9: It was introduced in #806 I mentioned it briefly in the description

> One slightly unfortunate thing is that using Model.fit_generator(..., shuffle=True) seems to rely on the global python random - so now we test that seeding the global python random is only needed when shuffle=True.

I'll double check if I get the same behaviour using `fit` instead.

tests/reproducibility/test_graphsage.py
kjun9: I was under the impression that `workers` refers to number of threads by default, but switches to number of processes if using multiprocessing. The docs aren't worded very well IMO but that's the behaviour I thought I saw.

> workers: Integer. Used for generator or keras.utils.Sequence input only. Maximum number of processes to spin up when using process-based threading. If unspecified, workers will default to 1. If 0, will execute the generator on the main thread.

huonw: Ah, yeah, it seems you're right: https://github.com/tensorflow/tensorflow/blob/eb3dc469b0bf00f770960b1a265beb05af244083/tensorflow/python/keras/utils/data_utils.py#L732 . The `get_pool_class` function returns a `Pool` from either `multiprocessing` or [`multiprocessing.dummy`](https://docs.python.org/3.8/library/multiprocessing.html#module-multiprocessing.dummy) as appropriate.

tests/reproducibility/test_graphsage.py
kjun9: Hmm actually I think it might be better to do the switch all at once and keep using `fit_generator` here. The deprecation of `fit_generator` only occurred with tensorflow 2.1, and I'm getting different behaviour from using `fit` with a generator depending on the tensorflow version - tensorflow 2.1 works fine, but 2.0 gives me errors, so I think we want to have `tensorflow>=2.1` then make the switch?

kjun9: Error for reference:
```

    model = func()
tests/reproducibility/test_graphsage.py:247: in <lambda>
    shuffle=shuffle,
tests/reproducibility/test_graphsage.py:204: in gs_link_prediction
    shuffle=shuffle,
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/training.py:728: in fit
    use_multiprocessing=use_multiprocessing)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/training_v2.py:224: in fit
    distribution_strategy=strategy)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/training_v2.py:547: in _process_training_inputs
    use_multiprocessing=use_multiprocessing)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/training_v2.py:606: in _process_inputs
    use_multiprocessing=use_multiprocessing)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/data_adapter.py:613: in __init__
    output_shapes=nested_shape)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/data/ops/dataset_ops.py:540: in from_generator
    output_types, tensor_shape.as_shape, output_shapes)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/data/util/nest.py:471: in map_structure_up_to
    results = [func(*tensors) for tensors in zip(*all_flattened_up_to)]
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/data/util/nest.py:471: in <listcomp>
    results = [func(*tensors) for tensors in zip(*all_flattened_up_to)]
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:1216: in as_shape
    return TensorShape(shape)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:776: in __init__
    self._dims = [as_dimension(d) for d in dims_iter]
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:776: in <listcomp>
    self._dims = [as_dimension(d) for d in dims_iter]
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:718: in as_dimension
    return Dimension(value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <[AttributeError("'Dimension' object has no attribute '_value'") raised in repr()] Dimension object at 0x144d2df90>
value = (None, None, None)

    def __init__(self, value):
      """Creates a new Dimension with the given value."""
      if value is None:
        self._value = None
      elif isinstance(value, Dimension):
        self._value = value.value
      elif isinstance(value, dtypes.DType):
        raise TypeError("Cannot convert %s to Dimension" % value)
      else:
>       self._value = int(value)
E       TypeError: int() argument must be a string, a bytes-like object or a number, not 'tuple'

../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:193: TypeError
---------------------------------------- Captured stdout call -----------------------------------------
link_classification: using 'ip' method to combine node embeddings into edge embeddings
_____________________________________ test_link_prediction[False] _____________________________________

petersen_graph = <stellargraph.core.graph.StellarGraph object at 0x145114a90>, shuffle = False

    @pytest.mark.parametrize("shuffle", [True, False])
    def test_link_prediction(petersen_graph, shuffle):
        num_examples = 4
        edge_ids = np.random.choice(petersen_graph.nodes(), size=(num_examples, 2))
        edge_labels = np.random.choice([0, 1], size=num_examples)
        assert_reproducible(
>           lambda: gs_link_prediction(
                petersen_graph,
                edge_ids,
                edge_labels,
                [2, 2],
                tf.optimizers.Adam(1e-3),
                shuffle=shuffle,
            )
        )

tests/reproducibility/test_graphsage.py:241:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/reproducibility/fixtures.py:35: in assert_reproducible
    model = func()
tests/reproducibility/test_graphsage.py:247: in <lambda>
    shuffle=shuffle,
tests/reproducibility/test_graphsage.py:204: in gs_link_prediction
    shuffle=shuffle,
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/training.py:728: in fit
    use_multiprocessing=use_multiprocessing)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/training_v2.py:224: in fit
    distribution_strategy=strategy)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/training_v2.py:547: in _process_training_inputs
    use_multiprocessing=use_multiprocessing)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/training_v2.py:606: in _process_inputs
    use_multiprocessing=use_multiprocessing)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/data_adapter.py:613: in __init__
    output_shapes=nested_shape)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/data/ops/dataset_ops.py:540: in from_generator
    output_types, tensor_shape.as_shape, output_shapes)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/data/util/nest.py:471: in map_structure_up_to
    results = [func(*tensors) for tensors in zip(*all_flattened_up_to)]
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/data/util/nest.py:471: in <listcomp>
    results = [func(*tensors) for tensors in zip(*all_flattened_up_to)]
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:1216: in as_shape
    return TensorShape(shape)
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:776: in __init__
    self._dims = [as_dimension(d) for d in dims_iter]
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:776: in <listcomp>
    self._dims = [as_dimension(d) for d in dims_iter]
../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:718: in as_dimension
    return Dimension(value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <[AttributeError("'Dimension' object has no attribute '_value'") raised in repr()] Dimension object at 0x1452ab5d0>
value = (None, None, None)

    def __init__(self, value):
      """Creates a new Dimension with the given value."""
      if value is None:
        self._value = None
      elif isinstance(value, Dimension):
        self._value = value.value
      elif isinstance(value, dtypes.DType):
        raise TypeError("Cannot convert %s to Dimension" % value)
      else:
>       self._value = int(value)
E       TypeError: int() argument must be a string, a bytes-like object or a number, not 'tuple'

../../../miniconda3/envs/stellargraph-dev/lib/python3.7/site-packages/tensorflow_core/python/framework/tensor_shape.py:193: TypeError
```
I'm guessing `fit` didn't support all the different ways the generator inputs can be formed until 2.1

tests/reproducibility/test_graphsage.py
huonw: Ah; I'd forgotten about that.

tests/reproducibility/test_graphsage.py
huonw: What a pest ☹️ ah well.

tests/reproducibility/fixtures.py
huonw: I guess we could reduce the sizes of the models (e.g. `layer_sizes`) if initialisation is proportional to that?

huonw:
Oh, also, should this have a changelog entry?

tests/reproducibility/fixtures.py
kjun9: It's a nice idea... but I just tried it and unfortunately it doesn't seem to make any difference 🤔

kjun9:
I'm trying a rebuild since I couldn't reproduce the failure on CI locally to check whether it's intermittent.. but it did seem like a legitimate failure for the specific test I'm adding here. I'm trying to run it locally with `num_iter=100` now

kjun9:
@huonw @kieranricardo I've rebuilt CI a few more times and run it locally after copying the `pip freeze` in CI to a new local environment, and I'm not able to reproduce the error in [this build](https://buildkite.com/stellar/stellargraph-public/builds/1766) - I'll file an issue if it ever comes up again?

huonw:
Hm, that's suspicious, but yeah, let's merge if we can't reproduce easily now.

kjun9:
Hmm I ended up reproducing the error locally, which also happened only for `test_link_prediction[shuffle=True]`...

It's not obvious to me at the moment where the problem is coming from since the mechanism for link prediction is pretty much the same as (or even a subset of) unsupervised learning. I think it's also unlikely to be a problem with the code in `Model.fit_generator(shuffle=True)` in general since we haven't seen the other tests fail intermittently. 

I think I might just take out `shuffle=True` from the test for now and include a FIXME. I don't think it's super critical that we need to support reproducibility with `shuffle=True` turned on, since all it does is mix up the batches before passing them to the model, but our sequence objects already have an internal shuffling mechanism, so it would be like double shuffling.

tests/reproducibility/test_graphsage.py
```diff
@@ -169,3 +230,21 @@ def test_nai(petersen_graph, shuffle):
             petersen_graph, targets, [2, 2], tf.optimizers.Adam(1e-3), shuffle=shuffle
         )
     )
+
+
+# FIXME (#970): This test fails intermittently with shuffle=True
+@pytest.mark.parametrize("shuffle", [False])
+def test_link_prediction(petersen_graph, shuffle):
+    num_examples = 4
```
huonw: Hm, `num_examples = 4` here with `batch_size = 4` means that there'll only ever be one batch? And thus this test isn't actually testing reproducibility with respect to batch orders?

kjun9: Good catch, I'll increase `num_examples` to 10

PR #926, opened 2020-02-24
Make HinSAGE reproducible

kjun9:
This is for both NAI and link prediction, which are the two tasks currently available for HinSAGE

Part of #749

### DIFF
--- a/stellargraph/mapper/sampled_link_generators.py
+++ b/stellargraph/mapper/sampled_link_generators.py
@@ -75,7 +75,7 @@ class BatchedLinkGenerator(abc.ABC):
     def sample_features(self, head_links, batch_num):
         pass
 
-    def flow(self, link_ids, targets=None, shuffle=False):
+    def flow(self, link_ids, targets=None, shuffle=False, seed=None):
         """
         Creates a generator/sequence object for training or evaluation
         with the supplied node ids and numeric targets.
@@ -99,6 +99,7 @@ class BatchedLinkGenerator(abc.ABC):
                 `(len(link_ids), target_size)`
             shuffle (bool): If True the links will be shuffled at each
                 epoch, if False the links will be processed in order.
+            seed (int, optional): Random seed
 
         Returns:
             A NodeSequence object to use with with StellarGraph models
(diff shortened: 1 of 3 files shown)


kjun9:
Hmm.. something is still flaky, investigating the test failure now

tests/mapper/test_node_mappers.py
```diff
@@ -112,6 +112,44 @@ def example_hin_3(feature_size_by_type=None):
...
+    """
+
+    def flatten_features(seq):
+        # check (features == labels) and return flattened features
+        batches = [
+            (np.ravel(seq[i][0][0]), np.array(seq[i][1])) for i in range(len(seq))
+        ]
+        features, labels = zip(*batches)
+        features, labels = np.concatenate(features), np.concatenate(labels)
+        assert all(features == labels)
```
kieranricardo: I think this can be:

```suggestion
        assert np.array_equal(features, labels)
```

tests/mapper/test_node_mappers.py
```diff
@@ -112,6 +112,44 @@ def example_hin_3(feature_size_by_type=None):
...
+        features = flatten_features(seq)
+        seq.on_epoch_end()
+        features_next = flatten_features(seq)
+        return features, features_next
+
+    comparison_results = set()
+
+    for i in range(max_iter):
+        f1, f2 = consecutive_epochs(seq)
+        comparison_results.add(all(f1 == f2))
```
kieranricardo: Same here:

```suggestion
        comparison_results.add(np.array_equal(f1, f2))
```

kieranricardo: could the assertions happen in this line?

```suggestion
        if shuffle:
            assert not np.array_equal(f1, f2)
        else:
            assert np.array_equal(f1, f2)   
```

Avalee21:
It should be blamed on random state use multiple times, which will generate different results. The following steps guarantee consistent results even at every epoch.
1. set stellargraph seed:  (1) `HinSAGELinkGenerator` (fix the seed in the `run` function of `SampledHeterogeneousBreadthFirstWalk` ) and (2) `flow` function 
2. [set tensorflow seed](https://stackoverflow.com/questions/50659482/why-cant-i-get-reproducible-results-in-keras-even-though-i-set-the-random-seeds/52897289#52897289)

PR #936, opened 2020-02-25
Make attri2vec reproducible

kjun9:
Part of #749


kjun9:
So it seems like #970 is also a problem here too, which does narrow things down a bit since this doesn't use `SeededPerBatch` at all 🤔 

Unlike Graph/HinSAGE I think there's actually nothing inherently random in this algorithm which I didn't realise before I started this. 

Also there's no node attribute inference in the tests, since `attri2vec` doesn't *directly* do node attribute inference - in the demo, it gets the embeddings using unsupervised learning, then you do NAI on the embeddings.

tests/reproducibility/test_attri2vec.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+    return model
+
+
+def attri2vec_link_prediction(
+    g,
+    edge_ids,
+    edge_labels,
+    optimizer,
```
huonw: Out of curiosity, what's the purpose of this parameter? It seems like it's always `tf.optimizers.Adam(1e-3)`?

Similarly, why have the other configurations (other than `shuffle`)? It seems like they're never customised?

huonw (approved):

tests/reproducibility/test_attri2vec.py
kjun9: hmm yeah that is very true 😅 

Maybe it makes more sense to have:
```python
def attri2vec_link_prediction(g, edge_ids, edge_labels, batch_size, shuffle)
```
We don't actually need any optional parameters

tests/reproducibility/test_attri2vec.py
```diff
@@ -45,18 +45,16 @@ def unsup_attri2vec_model(generator, optimizer, bias, normalize):
...
-    shuffle=True,
-):
+def unsup_attri2vec(g, batch_size, shuffle):
+
+    epochs = 4
+    bias = True
+    normalize = "l2"
+    seed = 0
+    optimizer = tf.optimizers.Adam(1e-3)
+    number_of_walks = 1
```
huonw: Oh, should we do 

```suggestion
    number_of_walks = 2
```

or higher, to catch any problems that only manifest across a transition (i.e. not when there's only one example)?

tests/reproducibility/test_attri2vec.py
```diff
@@ -45,18 +45,16 @@ def unsup_attri2vec_model(generator, optimizer, bias, normalize):
...
-    seed=0,
-    shuffle=True,
-):
+def unsup_attri2vec(g, batch_size, shuffle):
+
+    epochs = 4
+    bias = True
+    normalize = "l2"
+    seed = 0
+    optimizer = tf.optimizers.Adam(1e-3)
```
huonw: Do you think it's useful to specify these things upfront like this? I suspect it's probably ok to push these down to `unsup_attri2vec_model` and `attri2vec_link_pred_model`?

huonw (approved):
Still looks good 😄 but maybe some small tweaks.

PR #937, opened 2020-02-25
Make cluster gcn reproducible

kjun9:
Part of #749

### DIFF
--- a/tests/reproducibility/test_cluster_gcn.py
+++ b/tests/reproducibility/test_cluster_gcn.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 Data61, CSIRO
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tests/reproducibility/test_cluster_gcn.py
```diff
@@ -0,0 +1,72 @@
...
+    return model
+
+
+@pytest.mark.parametrize("shuffle", [True, False])
+def test_nai(petersen_graph, shuffle):
+    target_size = 10
+    targets = np.random.rand(petersen_graph.number_of_nodes(), target_size)
+    assert_reproducible(
+        lambda: cluster_gcn_nai(petersen_graph, targets, 4, 2, shuffle=shuffle)
+    )
```
PantelisElinas: I feel that we need a test to also make sure that if the seed is different, then the models are indeed different. Otherwise, we have not established a causal link between the random seed value and initial model weights.

Issue #909, opened 2020-02-20
Proposal: move components of `utils` into top-level and/or more descriptively-named packages


huonw:
### Description

We currently have a `stellargraph.utils` package, that defines https://github.com/stellargraph/stellargraph/tree/8e3c9b64983d57afcdd7780f31de75436c7df2de/stellargraph/utils :

1. interpretability/saliency maps
2. calibration
3. ensembling
4. loss functions

I think we could consider moving this current set of things into other locations, e.g. top level packages.

One arrangement might be:

- `stellargraph.interpretability` or `stellargraph.importance` that contains saliency maps (and brute force interpretability: #850)
- `stellargraph.calibration`
- `stellargraph.ensembles`
- `stellargraph.losses` (mimicking `tf.keras.losses`)

Another organisation might be:

- `stellargraph.interpretability` or `stellargraph.importance` that contains saliency maps (and brute force interpretability: #850)
- `stellargraph.calibration`
- `stellargraph.model` that contains ensembles, as well as models like `GCN` (as in, separate the individual layers like `GraphConvolution` from the actual model like `GCN`)
- `stellargraph.losses`

The move can and should be done with appropriate deprecation warnings, so that any code using the old one continues working.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

kjun9:
I think the idea behind the current `utils` package might've been as a collection of extra tools that can be used with any of the algorithms in the library? But yes, moving the inner modules to top-level sounds reasonable to me.

huonw:
> I think the idea behind the current utils package might've been as a collection of extra tools that can be used with any of the algorithms in the library? 

Ah, so it was meant to serve as a "one stop shop" for that sort of thing? I guess importance, calibration and ensembles could all be moved into a single module (the only ideas I have for a name is something like `stellargraph.enrich` or `stellargraph.augment`, since these "enrich" or "augment" how the model works... but that's not very good).

kjun9:
I was just guessing at the original intention, but I'd still prefer to move them all to top-level, I think it makes each of those features more visible that way and our top-level is not very crowded or anything currently.

timpitman:
I agree - let's move these to the top level, with appropriate renames.

PR #938, opened 2020-02-25
Move modules from `stellargraph.utils` to top-level ones

huonw:
This moves most of the code within `stellargraph.utils` to dedicated top-level modules, because `utils` is an overly generic name.

The old paths still work, but have runtime warnings to aid the migration. This is done by overriding `__getattr__` for the `stellargraph.utils` module (unfortunately this cannot use [PEP 562](https://www.python.org/dev/peps/pep-0562/) because that is Python 3.7+, and we support 3.6), and replacing each of the submodule files with, essentially:

```python
warnings.warn(...)
from path.to.new_module import *
```

(Unfortunately, I think these have to be files with the same name, so these new small files are stopping git from detecting the moves: all of the large new files are exact moves, with no other changes.)

Full list of changes:

| old | new |
|---|---|
| `stellargraph.utils.calibration` | `stellargraph.calibration` |
| `stellargraph.utils.IsotonicCalibration` | `stellargraph.calibration.IsotonicCalibration` |
| `stellargraph.utils.TemperatureCalibration` | `stellargraph.calibration.TemperatureCalibration` |
| `stellargraph.utils.expected_calibration_error` | `stellargraph.calibration.expected_calibration_error` |
| `stellargraph.utils.plot_reliability_diagram` | `stellargraph.calibration.plot_reliability_diagram` |
| `stellargraph.utils.ensemble` | `stellargraph.ensemble` |
| `stellargraph.utils.BaggingEnsemble` | `stellargraph.ensemble.BaggingEnsemble` |
| `stellargraph.utils.Ensemble` | `stellargraph.ensemble.Ensemble` |
| `stellargraph.utils.saliency_maps` | `stellargraph.interpretability.saliency_maps` |
| `stellargraph.utils.integrated_gradients` | `stellargraph.interpretability.saliency_maps.integrated_gradients` |
| `stellargraph.utils.integrated_gradients_gat` | `stellargraph.interpretability.saliency_maps.integrated_gradients_gat` |
| `stellargraph.utils.saliency_gat` | `stellargraph.interpretability.saliency_maps.saliency_gat` |
| `stellargraph.utils.GradientSaliencyGAT` | `stellargraph.interpretability.GradientSaliencyGAT` |
| `stellargraph.utils.IntegratedGradients` | `stellargraph.interpretability.IntegratedGradients` |
| `stellargraph.utils.IntegratedGradientsGAT` | `stellargraph.interpretability.IntegratedGradientsGAT` |

<details>

Computed by adding the following to `stellargraph/utils/__init__.py`:

```
for name, (new_module_name, new_value) in sorted(_MAPPING.items(), key=lambda x: ("stellargraph." + x[1][0] + "zzz", x[0]) if x[1][0] is not None else (x[1][1].__name__, "")):
    if isinstance(new_value, types.ModuleType):
        new_location = new_value.__name__
    else:
        new_location = f"stellargraph.{new_module_name}.{new_value.__name__}"

    print(f"| `stellargraph.utils.{name}` | `{new_location}` |")

raise ValueError()
```

</details>

See: #909


stellargraph/utils/loss.py
```diff
@@ -14,27 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import tensorflow as tf
+import warnings as warnings
```
timpitman: is this just `import warnings`?

huonw: Oh, yeah. It was previously `import warnings as _warnings` to avoid it being imported for something like `from stellargraph.utils.loss import *`, but I decided that wasn't worth it, apparently without actually doing the full simplification:

```suggestion
import warnings
```

stellargraph/utils/loss.py
```diff
@@ -14,27 +14,11 @@
...
 
-from ..core.experimental import experimental
+warnings.warn(
+    "'stellargraph.utils.loss' has been moved to 'stellargraph.losses'",
+    DeprecationWarning,
+)
```
timpitman: Is this still needed with the Wrapper in `stellargraph/utils/__init__.py`?

timpitman (commented):
The move looks good, but there's a lot of extra complexity to maintain compatibility with the old API

stellargraph/utils/loss.py
huonw: Yes; there's two code paths here:

1. `import stellargraph.utils` followed by `utils.foo`: this takes the `__getattr__` route
2. a direct import of a submodule `import stellargraph.utils.foo`: I believe this just does a file system traversal and opens/executes the given file (i.e. `stellargraph/utils/foo.py` or `stellargraph/utils/foo/__init__.py`) to create the submodule directly (rather than getting attributes of parent modules), and so needs handling in each file individually

(For this specific file, it isn't reexported in `utils/__init__.py`, and so `utils.loss` in 1 doesn't actually work, and only 2 works.)

huonw (commented):
> The move looks good, but there's a lot of extra complexity to maintain compatibility with the old API

I think it's not _that_ much complexity:

- it's only a relatively small chunk of code (although it is a little tricky)
- it's fully tested, and so future changes that break it should be caught
- it doesn't have complicated interactions with anything else, so the complexity is quite scoped
- we can remove it after it has been released

In any case, I think this sort of thing is worth it, because it makes users life much easier if they get a message "X.Y moved to ...", rather than just "module X doesn't have Y" and requiring them to carefully match release notes against their own code.

PantelisElinas (approved):
This looks good to me. The utils module came to existence because we needed a place for a bit of functionality that did not fit anywhere else. It has grown considerably since then and moving things to the top level make sense.

P.

timpitman (approved):

timpitman:
Thanks for the extra explanation @huonw, LGTM

Issue #940, opened 2020-02-25
Fix release procedure


kjun9:
### Describe the bug

There were a few problems with the release procedure identified during last release:
* changelog section heading needs to be updated in the release branch
* `meta.yaml` needs to be updated in the release branch
* publishing should happen before pushing the merge to master so that there is an opportunity to resolve any issues with publishing

PR #943, opened 2020-02-25
Update release procedure

kjun9:
Part of #940

### DIFF
--- a/RELEASE_PROCEDURE.md
+++ b/RELEASE_PROCEDURE.md
@@ -11,54 +11,29 @@
        - Version bumping: In stellargraph/version.py, change version from “X.X.Xb” to “X.X.X”. E.g. version=”0.2.0b” to version=”0.2.0”
        - Update Changelog section header to point to specific version tag instead of `HEAD`. Note: this link will be broken until the tag is pushed later.
        - Remove "Full Changelog" link from Changelog.
+       - Update version in `meta.yaml`
      - CAN do:
        - Minor bug fixes if necessary
      - NEVER do:
        - Add new features
 
-2. **Merge release branch into `master`**
+2. **Merge release branch into `master` locally**
 
-   ```shell
-   git checkout master
-   git merge --no-ff release-X.X.X -m "Release X.X.X"
-   git tag -a vX.X.X -m "Release X.X.X"
-   git push --follow-tags origin master
-   ```
+    This step gets your local `master` branch into release-ready state.
 
-3. **Merge `master` into `develop`**
+    ```shell
+    git checkout master
+    git merge --no-ff release-X.X.X -m "Release X.X.X"
+    git tag -a vX.X.X -m "Release X.X.X"
+    ```
 
-   - Do the merge
-     ```shell
-     git checkout develop
-     git merge master
-     ```
-   - Increase the version: in `stellargraph/version.py`, change version from `X.X.X` to `X.X+1.Xb`. E.g. `__version__ = "0.2.0"` to `__version__ = "0.3.0b"`. (To stay consistent we use `b` to indicate “beta”, but python will accept any string after the number. In semantic versioning: first number for major release, second number for minor release, third number for hotfixes.)
-   - Commit the version change:
-     ```shell
-     git commit -am "Bumped version"
-     ```
-   - Push the merge commit (and the version change):
-     ```shell
-     git push origin develop
-     ```
+3. **Upload new version to PyPI**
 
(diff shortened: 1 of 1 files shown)


RELEASE_PROCEDURE.md
```diff
@@ -9,7 +9,8 @@
    - Increase the version and apply other release-related changes
      - MUST do:
        - Version bumping: In stellargraph/version.py, change version from “X.X.Xb” to “X.X.X”. E.g. version=”0.2.0b” to version=”0.2.0”
-       - Other release-related changes e.g. changelog update
+       - Update Changelog section header to point to specific version tag instead of `HEAD`. Note: this link will be broken until the tag is pushed later.
+       - Remove "Full Changelog" link from Changelog.
```
huonw: I think this should be the same as the previous line, where `HEAD` is changed to the tag. For instance, https://github.com/stellargraph/stellargraph/blob/develop/CHANGELOG.md#080 tries to link to https://github.com/stellargraph/stellargraph/compare/v0.7.3...v0.8.0 (it uses https://github.com/stellargraph/stellargraph/compare/v0.8.0...v0.7.3 which has the wrong order). I guess we haven't done this for hotfixes.

That is, potentially the previous line can be:

```markdown
- Update Changelog section header and "Full Changelog" link to point to specific version tag instead of `HEAD`. Note: these links will be broken until the tag is pushed later.
```

and

```suggestion
```

RELEASE_PROCEDURE.md
```diff
@@ -88,6 +66,42 @@
      anaconda upload -u stellargraph /path/to/conda-package.tar.bz2
      ```
 
+5. **Make release on GitHub**
+
+    After successfully publishing to PyPi and Conda, we now want to make the release on GitHub.
+
+   - Push `master` branch
+    ```shell
```
huonw: I think this block needs one more level of indentation:

```suggestion
     ```shell
```

RELEASE_PROCEDURE.md
```diff
@@ -88,6 +66,42 @@
...
+
+6. **Get `develop` into correct state for next development version**
+
+    We want the merge any of the changes made during the release back into `develop`, and make sure the new version in `develop` is correct.
+
+   - Switch to `develop` branch:
+     ```shell
+     git checkout develop
+     ```
+   - Increase the version: in `stellargraph/version.py`, change version from `X.X.X` to `X.X+1.Xb`. E.g. `__version__ = "0.2.0"` to `__version__ = "0.3.0b"`. (To stay consistent we use `b` to indicate “beta”, but python will accept any string after the number. In semantic versioning: first number for major release, second number for minor release, third number for hotfixes.)
```
huonw:
```suggestion
   - Increase the version: in `stellargraph/version.py` and `meta.yaml`, change version from `X.X.X` to `X.X+1.Xb`. E.g. `__version__ = "0.2.0"` to `__version__ = "0.3.0b"`. (To stay consistent we use `b` to indicate “beta”, but python will accept any string after the number. In semantic versioning: first number for major release, second number for minor release, third number for hotfixes.)
```

I guess this will also need to change to `git add stellargraph/version.py meta.yaml` in the next line too.

RELEASE_PROCEDURE.md
```diff
@@ -11,55 +11,30 @@
        - Version bumping: In stellargraph/version.py, change version from “X.X.Xb” to “X.X.X”. E.g. version=”0.2.0b” to version=”0.2.0”
        - Update Changelog section header to point to specific version tag instead of `HEAD`. Note: this link will be broken until the tag is pushed later.
        - Remove "Full Changelog" link from Changelog.
+       - Update version in `meta.yaml`
```
huonw: Maybe this could be included as "Version bumping: In `stellargraph/version.py` and `meta.yaml`, ..." or similar.

RELEASE_PROCEDURE.md
kjun9: Do we need to update this in develop? It only gets used during a release specifically for publishing to conda cloud, i.e. `meta.yaml` with a `x.x.xb` version will never be used, until it gets updated to `x.x.x` for the next release again.

huonw: I thought it would be good to in case someone wants to install the Conda package from GitHub... but it turns out that's not how Conda works (e.g. https://stackoverflow.com/a/19071214/1256624 and the other answers here), and one generally just does a pip install inside the Conda environment. As such, yeah, I agree with you. 👍

Issue #948, opened 2020-02-25
Support loading only the largest connected component of Cora


huonw:
### Description

in #913, we gain a `datasets.Cora.load` function, which returns the whole Cora graph. Many papers use only the largest connected components, so it would be convenient if we supported extracting this conveniently.

The sizes of the 78 connected components are:

```
array([2485,    5,    2,   26,    3,    2,    2,    4,    5,    4,    6,
          2,    2,    2,    4,    3,    2,    2,    2,    2,    2,    2,
          2,    9,    2,    2,    2,    2,    2,    2,    2,    2,    2,
          2,    2,    5,    4,    3,    2,    2,    3,    2,    2,    3,
          2,    2,    2,    4,    2,    2,    2,    2,    8,    2,    2,
          2,    2,    2,    3,    2,    2,    2,    2,    2,    2,    2,
          2,    4,    3,    2,    2,    2,    2,    2,    2,    2,    2,
          2])
```

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #958, opened 2020-02-26
Add subgraph and connected_components to StellarGraph

huonw:
This adds two functions to `StellarGraph` that can be used together for some preprocessing tasks:

- `connected_components`: returns the nodes in each connected component of the graph as a lazy iterator in decreasing order of size
- `subgraph`: takes a collection of nodes and returns the subgraph induced by them (i.e. all of the nodes, plus any edges between them)

It then uses this to extend `Cora.load` to support returning only the largest connected-component because some papers/notebooks do this.

See: #948


stellargraph/core/graph.py
```diff
@@ -944,6 +944,75 @@ def to_adjacency_matrix(self, nodes: Optional[Iterable] = None, weighted=False):
...
+        next(graph.connected_components())``. The node IDs returned by this method can be used to
+        compute the corresponding subgraph with ``graph.subgraph(nodes)``.
+
+        Returns:
+            An iterator over sets of node IDs in each connected component, from the largest (most nodes)
+            to smallest (fewest nodes).
+        """
+
+        adj = self.to_adjacency_matrix()
+        count, cc_labels = sps.csgraph.connected_components(adj, directed=False)
```
kieranricardo: How about some extra functionality to handle directed graphs?

```suggestion
    def connected_components(self, connection="weak"):
        """
        Compute the connected components in this graph, ordered by size.

        The nodes in the largest component can be computed with ``nodes =
        next(graph.connected_components())``. The node IDs returned by this method can be used to
        compute the corresponding subgraph with ``graph.subgraph(nodes)``.
        
        Args:
            connection (str, optional): ['weak' | 'strong'] whether to calculate weakly or strongly connected components in the case of a directed graph 
        
        Returns:
            An iterator over sets of node IDs in each connected component, from the largest (most nodes)
            to smallest (fewest nodes).
        """

        adj = self.to_adjacency_matrix()
        count, cc_labels = sps.csgraph.connected_components(adj, directed=self.is_directed(), connection=connection)
```

stellargraph/core/graph.py
```diff
@@ -944,6 +944,75 @@ def to_adjacency_matrix(self, nodes: Optional[Iterable] = None, weighted=False):
...
+
+        node_frames = {
+            type_name: pd.DataFrame(
+                self._nodes.features(type_name, ilocs),
+                index=self._nodes.ids.from_iloc(ilocs),
+            )
+            for type_name, ilocs in node_type_to_ilocs
+        }
+
+        # FIXME: this is O(edges in graph) but could potentially be optimised to O(edges incident in
```
kieranricardo: can you create an issue for this (if you haven't already!) and document it here?

stellargraph/core/graph.py
huonw: I thought about it, but decided it was a YAGNI extension. That is, we could add (and thus test and support it) it in future if/when it's useful. Do you have something concrete in mind that wants strongly connected components?

(It's pretty unfortunately, because I was really hanging out for multiline suggestions but... it's seems it's not at all easy to understand the changes for big ones. 😦 )

stellargraph/core/graph.py
huonw: #985

stellargraph/core/graph.py
kieranricardo: That's a fair point, and no I don't have anything in mind. Lets document that this function returns weakly connected components though, its hard to tell that without digging into the scipy docs.

(I know, so sad :( every time I've tried it out it's been more confusing than adding seperate one line suggestions)

huonw: I've documented it :+1:

PR #960, opened 2020-02-26
Add MUTAG to datasets

PantelisElinas:
Hi @timpitman 

I have added the MUTAG dataset to the `stellargraph.datasets` module including a `load` method that returns a list of the graphs and corresponding labels.

For this dataset, edges have labels (categorical values, 1 of 4, represented as integers). Currently, StellarGraph only supports a single `weight` attribute for edges so I am storing the edge labels as weights. Semantically, weights and labels are not the same but it will have to suffice until `StellarGraph` adds support for more general edge attributes. Happy to remove the edge weights if too confusing.


stellargraph/datasets/datasets.py
```diff
@@ -304,3 +305,108 @@ class AIFB(
...
+
+        graphs = []
+        for graph_id in np.unique(df_graph_ids):
+            # find the subgraph with nodes that correspond to graph_id
+            node_ids = list(
+                df_graph_ids.loc[df_graph_ids["graph_id"] == graph_id].index
+            )
+
+            df_subgraph = df_graph[
+                df_graph["source"].isin(node_ids) | df_graph["target"].isin(node_ids)
```
huonw: It probably doesn't matter, but strictly, this should be "and", not "or", because otherwise a subgraph could end up with dangling edges pointing to unknown nodes (i.e. not in `node_ids`)

```suggestion
                df_graph["source"].isin(node_ids) & df_graph["target"].isin(node_ids)
```

However, I assume that the dataset is designed so that there's no such edges, and the source and target of an edge always have the same graph ID; as in, the graph ID of the source uniquely determines which graph it is in. As such, this could be simplified to:

```suggestion
                df_graph["source"].isin(node_ids)
```

(This suggestion applies to `graph_for_nodes` in the group-by comment above.)

tests/datasets/test_datasets.py
```diff
@@ -109,6 +110,30 @@ def test_blogcatalog3_deprecated_load() -> None:
...
+    n_avg_nodes = np.mean(n_nodes)
+    max_nodes = np.max(n_nodes)
+
+    # average number of nodes should be 17.93085... or approximately 18.
+    assert round(n_avg_nodes) == 18
+    # maximum number of nodes should be 28
+    assert max_nodes == 28
+
+    # There are two labels -1 and 1
+    assert len(np.unique(labels)) == 2
```
huonw: I think we could tweak this to be more specific, and thus make the comment above unnecessary:
```suggestion
    assert set(labels) == {-1, 1}
```

tests/datasets/test_datasets.py
```diff
@@ -109,6 +110,30 @@ def test_blogcatalog3_deprecated_load() -> None:
...
+
+    # get a list with the number of nodes in each graph
+    n_nodes = [g.number_of_nodes() for g in graphs]
+
+    # calculate average and max number of nodes across all graphs
+    n_avg_nodes = np.mean(n_nodes)
+    max_nodes = np.max(n_nodes)
+
+    # average number of nodes should be 17.93085... or approximately 18.
+    assert round(n_avg_nodes) == 18
```
huonw: Maybe instead of looking at the the averages, we could look at something exact:

```suggestion
    assert sum(n_nodes) == 3371
```

And similarly, verify that the edge selection is working appropriately, because it's slightly subtle:


```python
n_edges = [g.number_of_edges() for g in graph]
assert sum(n_edges) == 7442
````

File: tests/datasets/test_datasets.py [hunk 4]
```diff
@@ -109,6 +110,30 @@ def test_blogcatalog3_deprecated_load() -> None:
...
 
+def test_mutag_load() -> None:
+    graphs, labels = MUTAG().load()
+
+    n_graphs = 188
+
+    assert len(graphs) == n_graphs
+    assert len(labels) == n_graphs  # one label per graph
+
+    # get a list with the number of nodes in each graph
```
huonw: FWIW, I'm unsure of the value of comments like this, because they just rephrase what the code is saying. It's definitely valuable if the code is subtle, but most of the code here is straightforward. Explaining what the (exact) average is definitely useful, but (effectively) explaining that `np.mean` calculates an average doesn't seem quite as useful.

stellargraph/datasets/datasets.py
```diff
@@ -304,3 +305,108 @@ class AIFB(
...
+            filename="MUTAG_edge_labels.txt",
+            names=["weight"],
+            dtype=int,
+        )
+        df_graph = pd.concat([df_graph, df_edge_labels], axis=1)  # add edge weights
+        df_graph_ids = self._load_from_txt_file(
+            location=location,
+            filename="MUTAG_graph_indicator.txt",
+            names=["graph_id"],
+            index_increment=1,
```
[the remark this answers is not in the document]
PantelisElinas: Node IDs start at 1 and the dataframe's implicit index starts at 0. By setting the index to correspond to the node IDS when the data is loaded I don't have to worry about the difference later on.

stellargraph/datasets/datasets.py
```diff
@@ -304,3 +305,108 @@ class AIFB(
...
+            dtype="category",
+            index_increment=1,
+        )
+
+        # Let us one-hot encode the node labels because these are used as node features
+        # in graph classification tasks.
+        df_node_labels = pd.get_dummies(df_node_labels)
+
+        graphs = []
+        for graph_id in np.unique(df_graph_ids):
```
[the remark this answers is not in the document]
PantelisElinas: Yes, there are more than one ways to do things in Python.

stellargraph/datasets/datasets.py
PantelisElinas: Yes, for this dataset, either of the 3 methods works the same.

tests/datasets/test_datasets.py
```diff
@@ -109,6 +110,28 @@ def test_blogcatalog3_deprecated_load() -> None:
...
+    assert len(labels) == n_graphs  # one label per graph
+
+    n_nodes = [g.number_of_nodes() for g in graphs]
+    n_edges = [g.number_of_edges() for g in graphs]
+
+    n_avg_nodes = np.mean(n_nodes)
+    max_nodes = np.max(n_nodes)
+
+    # average number of nodes should be 17.93085... or approximately 18.
+    assert round(n_avg_nodes) == 18
```
timpitman: you could consider using ```pytest.approx``` here; eg
```
assert n_avg_nodes == pytest.approx(17.9, 0.1)
```

stellargraph/datasets/datasets.py
```diff
@@ -304,3 +305,106 @@ class AIFB(
...
+        "MUTAG_A.txt",
+        "MUTAG_graph_indicator.txt",
+        "MUTAG_node_labels.txt",
+        "MUTAG_edge_labels.txt",
+        "MUTAG_graph_labels.txt",
+        "README.txt",
+    ],
+    description="Each graph represents a chemical compound and graph labels represent 'their mutagenic effect on a specific gram negative bacterium.'"
+    "The dataset includes 188 graphs with 18 nodes and 20 edges on average for each graph. Graph nodes have 7 labels and each graph is labelled as belonging to 1 of 2 classes.",
+    source="https://ls11-www.cs.tu-dortmund.de/people/morris/graphkerneldatasets/",
```
huonw: I think the more human readable page is a slightly better source page, rather than the listing of the zip files.

```suggestion
    source="https://ls11-www.cs.tu-dortmund.de/staff/morris/graphkerneldatasets",
```

PantelisElinas:
@timpitman @huonw I think I have addressed your suggestions. Just let me know if you want anything else changed.

Issue #959, opened 2020-02-26
Remove deprecation warnings/compatibility functionality post-release


huonw:
### Description

We just did a 0.10 release, which included deprecation warnings and compatibility functionality for various features from 0.9 and earlier. We can now remove these, since users who upgrade to 0.10 will have these flagged and thus should no longer be using them in future releases.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #969, opened 2020-02-26
Remove deprecations included in 0.10 release

huonw:
This includes simplifying `BlogCatalog3.load` because we no longer need to support `load_dataset_BlogCatalog3`.

See: #959


stellargraph/core/graph.py
```diff
@@ -203,8 +203,6 @@ class StellarGraph:
             Deprecated, use :meth:`from_networkx`.
         edge_type_name:
             Deprecated, use :meth:`from_networkx`.
```
kieranricardo: There's a couple of deprecated arguments still here. (+ `graph` and `node_type_name` above that github wont let me suggest changes for)

```suggestion
```

stellargraph/core/graph.py
```diff
@@ -203,8 +203,6 @@ class StellarGraph:
...
-        edge_weight_label:
-            Deprecated, use :meth:`from_networkx`.
         node_features:
             Deprecated, use :meth:`from_networkx`.
```
kieranricardo:
```suggestion
```

stellargraph/core/graph.py
huonw: I had carefully left these ones here despite removing the adjacent `edge_weight_label` because these arguments are not properly deprecated: they don't trigger a runtime warning yet. I think it's important that we use a runtime warning to prompt people to move because:

- user with existing code won't re-read the docs, and
- they may not realise that a breaking change listed in the CHANGELOG affects their code (as in, they may not remember all the details about their code)

These arguments don't trigger a runtime warning yet because "FIXME(#717): this should have a deprecation warning, once the tests and examples have stopped using it" on line 253/241 below.

kieranricardo: cool that makes sense; everything looks good then

PR #974, opened 2020-02-27
Clear session after reproducibility test

kjun9:
Took inspiration from this issue: https://github.com/tensorflow/tensorflow/issues/14181

I logged the memory usage after each test using `resource.getrusage(resource.RUSAGE_SELF).ru_maxrss` and collected them into a list

## memory usage before
[648810496, 968466432, 1147457536, 1334071296]

## memory usage after
[647651328, 678608896, 717193216, 719740928]

### DIFF
--- a/tests/reproducibility/fixtures.py
+++ b/tests/reproducibility/fixtures.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import numpy as np
+import tensorflow as tf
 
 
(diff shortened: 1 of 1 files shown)


kjun9:
@huonw that's a good point. I've filed #975 - it might not make much difference for now since whatever models have been created in the global session will get cleared after the first reproducibility test, but probably good to do in the long run.

PR #977, opened 2020-02-27
Add WN18, WN18RR, FB15k & FB15k-237 knowledge graph datasets

huonw:
These knowledge graph datasets consist of triples, `subject` `relationship type` `object`, often as 3-column tab-separated files, each line being a triple. They are commonly used for knowledge-graph computation (link prediction is the usual task). For example, they're the top-4 most commonly used datasets at https://paperswithcode.com/task/link-prediction, and far ahead of anything else there.

They come from 3 papers:

1. [A Semantic Matching Energy Function for Learning with Multi-relational Data](https://arxiv.org/abs/1301.3485) (WN18, FB15k)
2. [Observed versus latent features for knowledge base and text inference](https://www.cs.princeton.edu/~danqic/papers/cvsc2015.pdf) (FB15k-237, called "FB15kSelected" in that paper)
3. [Convolutional 2D Knowledge Graph Embeddings](https://arxiv.org/abs/1707.01476) (WN18RR)

The WN18 and FB15k datasets are the originals and used in many papers. However, (2) and (3) observed that there's many "inverse relations", that is, `r1` and `r2` such that `T1 = (s, r1, o)` implies `T2= (o, r2, s)`, and there's overlap between the training and test sets (for example, `T1` in the train set, and `T2` in the test set) which even simple models can capture.

The datasets were acquired, without editing, from the following locations and the uploaded to http://figshare.com (rate limiting means it is suboptimal to download files from Github directly):

- https://github.com/villmow/datasets_knowledge_embedding (WN18, FB15k)
- https://github.com/TimDettmers/ConvE (WN18RR, FB15k-237)

Part of: #862


huonw:
Yeah, I thought about that and decided that the addition layers is not needed now; especially because we'd likely have to add special logic to our testing code that uses `__subclasses__`:

https://github.com/stellargraph/stellargraph/blob/2165dab4c71ddae41c750a4932f7e13d93ef35e1/tests/datasets/test_datasets.py#L39

Issue #972, opened 2020-02-26
ValueError: must have at least one node for inference, if `node_type` is not specified


sbhttchryy:
On trying to reproduce the RGCN example on AIFB dataset, I am running into the above error.
I am running the notebook on Google Colab and had tried with both installing the stellargraph package as well as the stellargraph[demo] package. Both of them show this error on the line:

generator = RelationalFullBatchNodeGenerator(G, sparse=True)

The entire trace is as follows:
ValueError                                Traceback (most recent call last)
<ipython-input-29-704246492183> in <module>()
----> 1 generator = RelationalFullBatchNodeGenerator(G, sparse=True)
      2 
      3 train_gen = generator.flow(train_data.index, targets=train_targets)
      4 test_gen = generator.flow(test_data.index, targets=test_targets)

1 frames
/usr/local/lib/python3.6/dist-packages/stellargraph/core/graph.py in node_features(self, nodes, node_type)
    737             if len(types) == 0:
    738                 raise ValueError(
--> 739                     "must have at least one node for inference, if `node_type` is not specified"
    740                 )
    741             if len(types) > 1:

ValueError: must have at least one node for inference, if `node_type` is not specified

timpitman:
Hi @sbhttchryy, thank you for reporting the issue with this notebook.  I believe this is the same as #911 which we'll be fixing soon - we'll update you once it's resolved.

huonw:
Thanks again for filing an issue; #983 has landed, so it should now be fixed on `develop`.

PR #983, opened 2020-02-28
Use str IDs for RGCN notebook, not rdflib types

huonw:
Previously the notebook wouldn't run because numpy was getting confused by the `rdflib` term types that were being used as the node IDs before; in particular, `rdflib.term.URIRef` and `rdflib.term.Literal` are subclasses of `str`, so `np.array(some_ids)` gets treated as an array of strings. However, the types get preserved as-is elsewhere (in particular, in a pandas Series or Index) and so there's a mismatch between `rdflib.term.URIRef` and `str` when doing node look-ups later.

Arguably, just converting these types to proper strings isn't the correct fix, and instead we should extend `StellarGraph` to support "weird" ID types like this, but that seems somewhat subtle. Thoughts?

See: #911, #972


huonw:
I opened https://github.com/stellargraph/stellargraph/issues/988 with a minimal reproducer, and a bit of a discussion about what we might do.

PR #986, opened 2020-02-28
Add PubMedDiabetes.load

huonw:
This adds a `load` method to `PubMedDiabetes` that returns the graph with its feature vectors, and a Series of the node labels.

For validation, this switches the notebooks that can use either Cora or PubMedDiabetes to use PubMedDiabetes, but I'll switch back before landing.

See: #812


demos/ensembles/ensemble-node-classification-example.ipynb
```diff
@@ -110,118 +110,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "use_cora = True  # Select the dataset; if False, then Pubmed-Diabetes dataset is used.\n",
+    "use_cora = False  # Select the dataset; if False, then Pubmed-Diabetes dataset is used.\n",
```
timpitman: the notebook now defaults to (and shows reaults for) pubmed rather than cora - was that intentional?

huonw: Yep, in the PR description:

> For validation, this switches the notebooks that can use either Cora or PubMedDiabetes to use PubMedDiabetes, but I'll switch back before landing.

demos/node-classification/cluster-gcn/cluster-gcn-node-classification.ipynb
```diff
@@ -226,78 +121,61 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [],
    "source": [
-    "dataset = \"cora\"  # can also select 'pubmed'\n",
+    "dataset = \"pubmed\"  # can also select 'pubmed'\n",
```
timpitman: same issue as previous notebook with swapped dataset exposing incorrect plot titles/labels.

demos/calibration/calibration-pubmed-node-classification.ipynb
```diff
@@ -1674,307 +303,13 @@
     "We now do the same for the node attributes we want to use to predict the subject. These are the feature vectors that the Keras model will use as input. The dataset contains attributes 'w-*' that correspond to words found in that publication."
```
PantelisElinas: Since loading the node features now happens in the load() method, you have correctly removed the code that performs the operation described here. The dangling comment can be removed as well.

huonw: Good point, although I think this comment was already outdated/dangling, since I don't think it was "doing the same" as any of the code working with the target variable.

demos/node-classification/graphsage/graphsage-pubmed-inductive-node-classification-example.ipynb
```diff
@@ -191,31 +140,31 @@
...
   },
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": 6,
    "metadata": {},
    "outputs": [],
    "source": [
-    "node_data_sampled = node_data.sample(frac=0.8, replace=False, random_state=101)"
+    "labels_sampled = labels.sample(frac=0.8, replace=False, random_state=101)"
```
PantelisElinas: You have renamed node_data_sampled to labels_sampled but the comment following this code still references node_data_sampled. Honestly, I thought the old name was more accurate since what is sampled here are nodes not labels.

huonw: I think this new name is accurate: the return value is a series with contents like:

```
pid
16221917    3
18591404    2
16424231    2
9925350     3
17448371    3
           ..
18579418    1
12819907    3
15142919    3
18728230    1
8786022     1
Name: label, Length: 15774, dtype: int64
```

That is, an index called `pid`, with values 1, 2, 3.

This is matching the old form, where the "`node_data`" in the name was referring to the actual node features since it was sampling whole rows of the DataFrame, not just the node IDs in the index. The sampling of node IDs is implicit in both old and new forms, as the index of the resulting structure.

You're 100% right about the following comment, though.

huonw:
Yeah, you're right. However, unfortunately that notebook is hard to update, because it uses EdgeSplitter, which still uses NetworkX. Removing (or, at least, hiding) the use of NetworkX is work in #174, after which we can update the various link prediction demos that haven't been updated to use the new `load` functions.

PantelisElinas:
> Yeah, you're right. However, unfortunately that notebook is hard to update, because it uses EdgeSplitter, which still uses NetworkX. Removing (or, at least, hiding) the use of NetworkX is work in #174, after which we can update the various link prediction demos that haven't been updated to use the new `load` functions.

Ok then, it's all good as far as I am concerned.

P.

PR #994, opened 2020-03-03
Check info() against literal expected output

huonw:
The `test_info_heterogeneous` test for `StellarGraph.info` on a heterogeneous graph was incorrect, because did not have some assertions when it should've.

This fixes that by checking the exact output for `info()`, meaning the full string including indentation/new lines. This is good because it means we're validating what the users see, and so will catch any mistakes that break the formatting.


huonw:
(Sorry @kjun9 if you happened to be reviewing this right now, I had to do some merge fixing.)

kjun9 (approved):
Looks fine, but could we also just compare the strings directly, rather than using `in_sequence`? If we're adding more things to `info` we probably want to update the tests anyway right?

huonw:
> Looks fine, but could we also just compare the strings directly, rather than using in_sequence? If we're adding more things to info we probably want to update the tests anyway right?

Do you mean type out the whole `info` string literally? I guess that's more reliable, in a sense.

kjun9:
> Do you mean type out the whole info string literally?

Yes that's what I meant. It also looks simpler IMO

huonw:
I'll switch to it.

huonw:
I switched to it, and like it, because the output of this method is designed for humans, so it's much better to check exactly what it looks like 👍 

As a minor detail, the literal strings previously had trailing whitespace which isn't allowed by our CI (https://buildkite.com/stellar/stellargraph-public/builds/1979#annotation-whitespace), and so I had to make an edit to `info` itself to write `Edge types: none` instead of `Edge types: ` (with a trailing space) when a node type has no incident edges/edge types.

kjun9 (approved):

PR #1000, opened 2020-03-03
Remove legacy construction from tests/mapper/...

huonw:
This switches away from the legacy `StellarGraph` constructor, and NetworkX itself, to use the new StellarGraph in the `tests/mapper/...` files as much as possible. This includes some instances of switching to a shared graph from `tests/test_utils/graphs.py` rather than each file creating its own.

See: #717

### DIFF
--- a/tests/test_utils/graphs.py
+++ b/tests/test_utils/graphs.py
@@ -25,29 +25,24 @@ import pytest
 
 def create_graph_features():
     # APPNP, ClusterGCN, GCN, PPNP, node_mappers
-    graph = nx.Graph()
-    graph.add_nodes_from(["a", "b", "c"])
-    graph.add_edges_from([("a", "b"), ("b", "c"), ("a", "c")])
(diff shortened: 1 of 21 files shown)


tests/test_utils/graphs.py
```diff
@@ -111,20 +111,27 @@ def example_hin_1_nx(feature_name=None, for_nodes=None, feature_sizes=None):
 
 
 def example_hin_1(
-    feature_sizes=None, is_directed=False, self_loop=False
+    feature_sizes=None,
+    is_directed=False,
+    self_loop=False,
+    node_types=("A", "B"),
```
kjun9: Thoughts on not exposing `node_types` and `edge_types` like this and just update the tests to use "A", "B", etc? I don't really see a particular use-case for any new tests wanting to use different names for the types for the same underlying graph, and the existing test probably doesn't need the `user` and `movie` context?

Issue #978, opened 2020-02-28
Implement Deep Graph Infomax


kieranricardo:
### Description

This issue is to implement the Deep Graph Infomax algorithm for scaleable unsupervised node inference (https://arxiv.org/pdf/1809.10341.pdf). 

This algorithm is of particular interest because it also has extensions to heterogeneous graphs (https://arxiv.org/pdf/1911.08538.pdf), temporal graphs (https://arxiv.org/pdf/1904.06316.pdf ), and graph classification (https://arxiv.org/pdf/1908.01000.pdf).

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation in repo
- [ ] CHANGELOG.md updated
- [ ] Team demo

Issue #614, opened 2020-01-07
Minimum version requirements are not validated


huonw:
### Description

Most of our version specifications are version ranges, usually as an lower bound `package >= min_version`:

https://github.com/stellargraph/stellargraph/blob/bcaa1fe660f3a217de590c456eb039761f7fcf76/setup.py#L24-L31

`pip` will typically install the highest version number that satisfies that constraint, and so our CI will effectively automatically be picking up the latest version of a dependency. This means we may accidentally rely on features from these newer packages, and a user that has a smaller version (but one that is still in the version range we specify) may not be able to use `stellargraph`, without `pip` flagging any problems.

Specifically, we say `scipy >= 1.1.0`, but the latest version is `1.4.1`. The `scipy.fft` module was added in `1.4.0`. If stellargraph started using that module, our CI would pass ([it installs `scipy == 1.4.1`](https://buildkite.com/stellar/stellargraph-public/builds/314#a98235b9-2092-4852-98f4-19baf1a0c452/108-200)), but a user who installed `stellargraph` into an environment with `scipy == 1.1.0` would not be able to use that code without upgrading `scipy`, and they would not have any incompatible version warnings from `pip`.

(#611 is a converse problem: problems caused by the latest versions of dependencies being incompatible with our code, rather than our code being incompatible with earlier versions of dependencies.)

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1001, opened 2020-03-03
Feature/gcn deep graph infomax

kieranricardo:
This PR implements Deep Graph Infomax for use with GCN as experimental (lacking unit tests). I have an implementation for GraphSage as well but this will be a seperate PR. At the moment I have a seperate `GCNInfomax` class but this could be merged with `GCN`; opinions on whether this is cleaner or not would be appreciated!

Reviewer checklist:

- [ ] the model is implemented correctecly
- [ ] the generator is implemented correctly
- [ ] the code is documented (including api.txt, changelog, and the readme)
- [ ] the demo is correct and easy to follow

See: #978


stellargraph/mapper/sequences.py
```diff
@@ -366,11 +366,13 @@ class FullBatchSequence(Sequence):
...
         indices (np.ndarray, optional): Array of indices to the feature and adjacency matrix
             of the targets. Required if targets is not None.
+        corruption (bool): If True return shuffled features along with the regular features for Deep Graph
+            Infomax.
     """
 
     use_sparse = False
 
-    def __init__(self, features, A, targets=None, indices=None):
+    def __init__(self, features, A, targets=None, indices=None, corruption=False):
```
huonw: It feels clunky to me to have to push down this corruption into each model manually, similar to it being clunky to have to duplicate GCN's construction with `GCNInfoMax`. I wonder if there's a way we can have have a wrapper like `CorruptingGenerator` around another Generator, similar to the proposal above for `InfoMax` to be a wrapper for `GCN`/`CorruptibleModel`?

tests/mapper/test_node_mappers.py
```diff
@@ -926,3 +926,34 @@ def test_generator_methods(self):
...
+    generator = CorruptedFullBatchNodeGenerator(G, sparse=sparse)
+
+    gen = generator.flow(G.nodes())
+
+    [features, shuffled_feats, *_], targets = gen.__getitem__(0)
+
+    assert features.shape == shuffled_feats.shape
+
+    # check shuffled_feats are feats
+    assert not np.array_equal(features, shuffled_feats,)
```
huonw: There's quite a few instances of trailing commas on a single line in this PR (search for `,)` and `,]` to find others); is this on purpose?

```suggestion
    assert not np.array_equal(features, shuffled_feats)
```

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,168 @@
...
+        node_feats = Lambda(lambda x: K.squeeze(x, axis=0,), name="NODE_FEATURES",)(
+            node_feats
+        )
+
+        node_feats_corrupted = self([x_corr, out_indices_t] + A_placeholders)
+        node_feats_corrupted = Lambda(lambda x: K.squeeze(x, axis=0,),)(
+            node_feats_corrupted
+        )
+
+        summary = Lambda(lambda x: tf.math.sigmoid(tf.math.reduce_mean(x, axis=0)))(
```
huonw: Is it worth having a `ReadOut` later similar to `Discriminator`?

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,168 @@
...
+
+        else:
+            # Placeholders for the dense adjacency matrix
+            A_m = Input(batch_shape=(1, self.n_nodes, self.n_nodes))
+            A_placeholders = [A_m]
+
+        x_inp = [x_corr, x_t, out_indices_t] + A_placeholders
+
+        node_feats = self([x_t, out_indices_t] + A_placeholders)
+        node_feats = Lambda(lambda x: K.squeeze(x, axis=0,), name="NODE_FEATURES",)(
```
huonw: It's probably good practice to factor this out:

```python
class GCNInfoMax:
    _NODE_FEATS = "GCN_INFO_MAX_NODE_FEATURES"

    ...
        node_features = Lambda(..., name=self._NODE_FEATS)
    ...
        x_emb_out = model.get_layer(self._NODE_FEATS).output
```

stellargraph/layer/deep_graph_infomax.py
kieranricardo: I feel like its simple enough to keep as a lambda, but I might create a `ReadOut` when I make the GraphSAGE infomax PR

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        features, summary = inputs
+
+        score = tf.linalg.matvec(tf.linalg.matmul(features, self.kernel), summary,)
+
+        return score
+
+
+@experimental(reason="lack of unit tests", issues=[1003])
+class GCNInfoMax(GCN):
```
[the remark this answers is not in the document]
kieranricardo: Nice, I like the idea of wrapping the base model. It seems like in this case the class structure doesn't really add anything (only storing the base model) so I'll try and implement this as two functions.

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
```diff
@@ -0,0 +1,291 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Test classification accuracy: 0.7473338802296965\n"
+     ]
+    }
+   ],
+   "source": [
+    "train_subjects, test_subjects = model_selection.train_test_split(\n",
```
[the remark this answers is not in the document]
kieranricardo: actually I realised that I don't use `train_targets` or `test_targets` here so I'll get rid of those.

I'll use your second suggestion tho, nice!

stellargraph/mapper/sequences.py
kieranricardo: I've got a wrapper for the generator (and the model as well) that can handle all full batch models and I think can be easily generalized to other model types.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,137 @@
...
+            constraint=None,
+        )
+        self.built = True
+
+    def call(self, inputs):
+        """
+        Applies the layer to the inputs.
+
+        Args:
+            inputs: a list of tensors with shapes [(N, F), (F,)] containing the node features and summary feature
```
PantelisElinas: This can be a 2-tuple as well not just a list.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,168 @@
...
+                vector.
+        """
+        features, summary = inputs
+
+        score = tf.linalg.matvec(tf.linalg.matmul(features, self.kernel), summary,)
```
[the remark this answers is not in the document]
kieranricardo: nice!

tests/mapper/test_node_mappers.py
kieranricardo: I had a fair bit of multiline stuff that I refactored (or maybe black did) but forgot to cut out the trailing commas, will delete

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,137 @@
...
+
+    if not isinstance(base_model, (GCN, GAT, APPNP, PPNP)):
+        raise TypeError(
+            f"base_model: expected GCN, GAT, APPNP or PPNP found {type(base_model).__name__}"
+        )
+
+    # Inputs for features
+    x_t = Input(batch_shape=(1, base_model.n_nodes, base_model.n_features))
+    # Inputs for shuffled features
+    x_corr = Input(batch_shape=(1, base_model.n_nodes, base_model.n_features))
```
PantelisElinas: Here we assume that the source and corrupted graphs are the same size. This will work on the single input graph transductive settings, e.g., Cora, and for the corruption function that only shuffles the features and keeps the adjacency matrix the same. This is the simplest setting for using this method and this should work. 

We should plan ahead for the multi-graph case where there is no guarantee that the graphs will have the same size. Our full batch GCN and GAT methods (not sure about APPNP and PPNP) cannot handle input graphs of different size. 

I don't want to suggest that you update the code now, but maybe create a new ticket to keep track of this issue. We will come across this problem sooner than later.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Applies the layer to the inputs.
+
+        Args:
+            inputs: a list of tensors with shapes [(N, F), (F,)] containing the node features and summary feature
+                vector.
+        """
+        features, summary = inputs
+
+        score = tf.linalg.matvec(tf.linalg.matmul(features, self.kernel), summary)
```
PantelisElinas: In the paper, equation 10, applies the sigmoid function to the `score` matrix. I don't see this anywhere in the code, here or in the `fullbatch_infomax_node_model` class where the below lines call the discriminator,

```
 scores = discriminator([node_feats, summary])
 scores_corrupted = discriminator([node_feats_corrupted, summary])
```

kieranricardo: I moved the application of sigmoid into the loss function for numerical stability, using this model now requires using `tf.nn.sigmoid_cross_entropy_with_logits` - but I should document this somewhere!

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,137 @@
...
+    scores_corrupted = discriminator([node_feats_corrupted, summary])
+
+    x_out = tf.stack([scores, scores_corrupted], axis=1)
+
+    x_out = K.expand_dims(x_out, axis=0)
+    return x_inp, x_out
+
+
+@experimental(reason="lack of unit tests", issues=[1003])
+def fullbatch_infomax_embedding_model(info_max_model):
```
PantelisElinas: Why is this and `fullbatch_infomax_node_model` implemented as methods? If we are to be consistent with the rest of the library, we should probably have  a `DeepGraphInfomax` class that we can initialise as necessary to implement the different variations of the method in the paper such as for example passing the base model (GCN, GAT, GraphSAGE, etc) and the setting (single or multi-graph) just to name a couple of options.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    A function to create the the inputs and outputs for an embedding model.
+
+    Args:
+        info_max_model (keras.Model): the base Deep Graph Infomax model
```
PantelisElinas: You might want to explain here what the base model is. It is not obvious what is a valid model and the method performs no checking just assumes that the models has the correct inputs.

stellargraph/mapper/full_batch_generators.py
```diff
@@ -23,6 +23,7 @@
     "FullBatchNodeGenerator",
     "FullBatchLinkGenerator",
     "RelationalFullBatchNodeGenerator",
+    "corrupt_full_batch_node_generator",
```
PantelisElinas: To be consistent I would name this `CorruptFullBatchNodeGenerator`.

stellargraph/mapper/full_batch_generators.py
```diff
@@ -503,3 +504,30 @@ def flow(self, node_ids, targets=None):
         return RelationalFullBatchNodeSequence(
             self.features, self.As, self.use_sparse, targets, node_indices
         )
+
+
+def corrupt_full_batch_node_generator(generator):
```
PantelisElinas: Why is this a method and not a class? You can just subclass `FullBatchNodeGenerator` and override its methods as necessary. This seems inconsistent with the rest of the library.

stellargraph/mapper/full_batch_generators.py
```diff
@@ -503,3 +504,30 @@ def flow(self, node_ids, targets=None):
...
+
+    if not isinstance(generator, FullBatchNodeGenerator):
+        raise TypeError(
+            f"generator: expected FullBatchNodeGenerator, found {type(generator).__name__}"
+        )
+
+    num_nodes = generator.features.shape[0]
+    node_indices = np.arange(num_nodes)
+    targets = np.zeros((num_nodes, 2))
+    targets[:, 0] = 1.0
```
PantelisElinas: What are `targets` used for?

stellargraph/mapper/sequences.py
PantelisElinas: Can we have a subclass `CorruptedFullBatchSequence(FullBatchSeqeuence)` that overrides the `__getitem__` method?

README.md
```diff
@@ -198,6 +198,8 @@ The StellarGraph library currently includes the following algorithms for graph m
 | Relational Graph Convolutional Network [11] | The RGCN algorithm performs semi-supervised learning for node representation and node classification on knowledge graphs. RGCN extends GCN to directed graphs with multiple edge types and works with both sparse and dense adjacency matrices.|
 | ComplEx[12] | The ComplEx algorithm computes embeddings for nodes (entities) and edge types (relations) in knowledge graphs, and can use these for link prediction |
 | GraphWave [13] | GraphWave calculates unsupervised structural embeddings via wavelet diffusion through the graph. |
+| Deep Graph Infomax [14] | Deep Graph Infomax trains GCN unsupervised to maximize the information in the final node features. |
```
PantelisElinas: Is this description correct? The base model in your code doesn't have to be GCN as you check for one of GCN, GAT, APPNP, and PPNP.

stellargraph/mapper/full_batch_generators.py
kieranricardo: I replaced this with a class in a commit pushed just before your comments - but I think this mustve been after you started the review process!

stellargraph/layer/deep_graph_infomax.py
kieranricardo: I replaced this with a class in a commit pushed just before your comments - but I think this mustve been after you started the review process!

stellargraph/mapper/full_batch_generators.py
kieranricardo: The mode outputs scores (w /out sigmoid applied) with shape (1 x N x 2), where the first column correpsonds to true nodes and the second column corresponds to corresponds to corrupted nodes. Used `targets`, the model output and `targets` can be passed to `tf.nn.sigmoid_cross_entropy_with_logits` to compute the Deep Graph Infomax loss function

stellargraph/layer/deep_graph_infomax.py
PantelisElinas: I just noticed that this has changed. The version I was reviewing became obsolete while I was reviewing it!

stellargraph/mapper/sequences.py
kieranricardo: The benefit of the wrapper is that it can handle corrupting multiple Sequences, currently it can handle `FullBatchNodeSequence`, `SparseFullBatchNodeSequence`, `RelationalFullBatchNodeSequence`. And it should be extendable to handle the GraphSage samplers as well. So in this case one class can handle the work of 3+ subclasses.

Also, users only access the `Sequence` objects through our generators, so we either have to subclass our generators or add extra arguments just to select whether to use a corrupted Sequence or not both of which feel clunky and end up duplicating code.

stellargraph/layer/deep_graph_infomax.py
kieranricardo: I stopped pushing when I realised you were reviewing, but I think at the point the damage had been done!

kieranricardo:
@PantelisElinas and @huonw thanks for your reviews!

@huonw I fixed my initial problem that yielded an accuracy of 30% - I was shuffling along the first axis of node feats forgetting it had a dimension of 1 and so wasn't shuffling at all! 

Now i'm getting 74-79% accurary. I suspect it may have something to do with the logistic regression, the paper reports a ~40% accuracy with logistic regression on raw features while I'm getting ~60% so I'm definitely doing something. I'm going to look into their code and replicate whatever they're doing for logistic regression and see whether that helps.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 Some new algorithms and features are still under active development, and are available as an experimental preview. However, they may not be easy to use: their documentation or testing may be incomplete, and they may change dramatically from release to release. The experimental status is noted in the documentation and at runtime via prominent warnings.
 
 - DistMult: computes embeddings for nodes and edge types in knowledge graphs, and use these to perform link prediction [\#755](https://github.com/stellargraph/stellargraph/issues/755). The implementation hasn't been validated to match the paper.
+- GCN Deep Graph Infomax has been implemented to perform unsupervised GCN training.
```
PantelisElinas: I think we need to update this as the implementation is not specific to GCN. Maybe something like, "Deep Graph Infomax for unsupervised node representation learning" is more suitable.

README.md
```diff
@@ -198,6 +198,8 @@ The StellarGraph library currently includes the following algorithms for graph m
 | Relational Graph Convolutional Network [11] | The RGCN algorithm performs semi-supervised learning for node representation and node classification on knowledge graphs. RGCN extends GCN to directed graphs with multiple edge types and works with both sparse and dense adjacency matrices.|
 | ComplEx[12] | The ComplEx algorithm computes embeddings for nodes (entities) and edge types (relations) in knowledge graphs, and can use these for link prediction |
 | GraphWave [13] | GraphWave calculates unsupervised structural embeddings via wavelet diffusion through the graph. |
+| Deep Graph Infomax [14] | Deep Graph Infomax trains GNNs unsupervised to maximize the information in the final node features. |
```
PantelisElinas: The "...to maximize the information in the final node features" part doesn't make much sense by itself. You might have to re-write.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        features, summary = inputs
+
+        score = tf.linalg.matvec(features, tf.linalg.matvec(self.kernel, summary))
+
+        return score
+
+
+@experimental(reason="lack of unit tests", issues=[1003])
+class InfoMax:
```
PantelisElinas: Should we call this class `DeepGraphInfoMax` since it is based on `Deep InfoMax`? The former name also directly links this model to the paper.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,146 @@
...
+            raise TypeError(
+                f"base_model: expected GCN, GAT, APPNP or PPNP found {type(base_model).__name__}"
+            )
+
+        self.base_model = base_model
+
+        # specific to full batch models
+        self.corruptible_inputs_idxs = [0]
+
+    def unsupervised_node_model(self):
```
PantelisElinas: We have been trying to standardise this method and always call it `build()`. The latter is supposed to build the correct model given the information that it has, e.g., if base model is GCN then go for unsupervised full batch node model or if base model is GraphSAGE then go for unsupervised mini-batch node model, etc.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        A function to create the the keras inputs and outputs for a Deep Graph Infomax model for unsupervised training.
+
+        Note that the tf.nn.sigmoid_cross_entropy_with_logits loss must be used with this model.
+
+        Example:
+
+            ```
+            x_in, x_out = infomax.unsupervised_node_model()
```
PantelisElinas: Maybe add one more line before this one to show how `infomax` was created because otherwise it might be confused with the class name `InfoMax`.

stellargraph/layer/deep_graph_infomax.py
kieranricardo: good point!

stellargraph/layer/deep_graph_infomax.py
PantelisElinas: You might be able to use the `tf.keras.layers.GlobalAveragePooling1D` layer here.

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +528,45 @@ def __len__(self):
...
+    """
+
+    def __init__(self, base_generator):
+
+        if not isinstance(
+            base_generator,
+            (
+                FullBatchSequence,
+                SparseFullBatchSequence,
+                RelationalFullBatchNodeSequence,
```
PantelisElinas: Have you tested this implementation with RGCN? Because if not, then shall we wait to add this when we are ready to extend to HINs?

README.md
kieranricardo: how about:

"Deep Graph Infomax trains unsupervised GNNs to maximize the shared information between node level and graph level features."

stellargraph/mapper/sequences.py
kieranricardo: yeah I think that makes sense to land all the heterogenous infomax stuff together

README.md
PantelisElinas: I think that's better although it still requires the reader to have some understanding of what it all means. For now, I think your last suggestion is better than the one before.

stellargraph/layer/deep_graph_infomax.py
kieranricardo: I'm hesitant to add this in here, `tf.keras.layers.GlobalAveragePooling1D` requires tensors with 3 dimension so won't work with mini-batch methods. It would also require keeping the batch dimension in . Squeezing out the batch dimension is convenient for calculating the scores and the embeddings. Also, squeezing out the batch dimension gives the node feature tensors the same number of dimension as the mini-batch methods which means `Discriminator` and `embedding_model` will work as is.

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
```diff
@@ -0,0 +1,284 @@
...
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "all_embeddings = emb_model.predict(generator.flow(G.nodes()))\n",
+    "\n",
+    "y = node_subjects.astype(\"category\").cat.codes\n",
+    "trans = TSNE(n_components=2)\n",
+    "emb_transformed = pd.DataFrame(trans.fit_transform(all_embeddings), index=list(G.nodes()))\n",
```
huonw:
```suggestion
    "emb_transformed = pd.DataFrame(trans.fit_transform(all_embeddings), index=G.nodes())\n",
```

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
```diff
@@ -0,0 +1,284 @@
...
+     },
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "es = EarlyStopping(monitor=\"loss\", min_delta=0, patience=20)\n",
+    "history = model.fit(gen, epochs=1000, verbose=0, callbacks=[es])\n",
```
huonw: FWIW, this fails with tensorflow 2.0.0 with `TypeError: int() argument must be a string, a bytes-like object or a number, not 'tuple'`. (I didn't try 2.0.1.)

I think it might be somehow related to the shape of the tuples/tensors returned by `__getitem__` below, because I saw something similar with `KGTripleGenerator`.

This is a broader problem (#614), and may affect other models too, so is potentially not something we need to fix here... but we should at least have an issue about "some models fail with tensorflow 2.0.0" to deal with in some form before the next release.

<details><summary>Full stack trace</summary>

```
TypeError                                 Traceback (most recent call last)
<ipython-input-5-2539c1ce905f> in <module>
      1 es = EarlyStopping(monitor="loss", min_delta=0, patience=20)
----> 2 history = model.fit(gen, epochs=1000, verbose=0, callbacks=[es])
      3 plot_history(history)

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/keras/engine/training.py in fit(self, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_freq, max_queue_size, workers, use_multiprocessing, **kwargs)
    726         max_queue_size=max_queue_size,
    727         workers=workers,
--> 728         use_multiprocessing=use_multiprocessing)
    729 
    730   def evaluate(self,

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/keras/engine/training_v2.py in fit(self, model, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_freq, **kwargs)
    222           validation_data=validation_data,
    223           validation_steps=validation_steps,
--> 224           distribution_strategy=strategy)
    225 
    226       total_samples = _get_total_number_of_samples(training_data_adapter)

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/keras/engine/training_v2.py in _process_training_inputs(model, x, y, batch_size, epochs, sample_weights, class_weights, steps_per_epoch, validation_split, validation_data, validation_steps, shuffle, distribution_strategy, max_queue_size, workers, use_multiprocessing)
    545         max_queue_size=max_queue_size,
    546         workers=workers,
--> 547         use_multiprocessing=use_multiprocessing)
    548     val_adapter = None
    549     if validation_data:

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/keras/engine/training_v2.py in _process_inputs(model, x, y, batch_size, epochs, sample_weights, class_weights, shuffle, steps, distribution_strategy, max_queue_size, workers, use_multiprocessing)
    604       max_queue_size=max_queue_size,
    605       workers=workers,
--> 606       use_multiprocessing=use_multiprocessing)
    607   # As a fallback for the data type that does not work with
    608   # _standardize_user_data, use the _prepare_model_with_inputs.

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/keras/engine/data_adapter.py in __init__(self, x, y, sample_weights, shuffle, **kwargs)
    611         yield x[i]
    612     dataset = dataset_ops.DatasetV2.from_generator(generator, nested_dtypes,
--> 613                                                    output_shapes=nested_shape)
    614     if shuffle:
    615       dataset = dataset.shuffle(len(x))

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/data/ops/dataset_ops.py in from_generator(generator, output_types, output_shapes, args)
    538     else:
    539       output_shapes = nest.map_structure_up_to(
--> 540           output_types, tensor_shape.as_shape, output_shapes)
    541     if args is None:
    542       args = ()

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/data/util/nest.py in map_structure_up_to(shallow_tree, func, *inputs)
    469                          for input_tree in inputs]
    470 
--> 471   results = [func(*tensors) for tensors in zip(*all_flattened_up_to)]
    472   return pack_sequence_as(structure=shallow_tree, flat_sequence=results)

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/data/util/nest.py in <listcomp>(.0)
    469                          for input_tree in inputs]
    470 
--> 471   results = [func(*tensors) for tensors in zip(*all_flattened_up_to)]
    472   return pack_sequence_as(structure=shallow_tree, flat_sequence=results)

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/framework/tensor_shape.py in as_shape(shape)
   1214     return shape
   1215   else:
-> 1216     return TensorShape(shape)
   1217 
   1218 

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/framework/tensor_shape.py in __init__(self, dims)
    774       else:
    775         # Got a list of dimensions
--> 776         self._dims = [as_dimension(d) for d in dims_iter]
    777 
    778   @property

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/framework/tensor_shape.py in <listcomp>(.0)
    774       else:
    775         # Got a list of dimensions
--> 776         self._dims = [as_dimension(d) for d in dims_iter]
    777 
    778   @property

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/framework/tensor_shape.py in as_dimension(value)
    716     return value
    717   else:
--> 718     return Dimension(value)
    719 
    720 

~/.pyenv/versions/sg/lib/python3.6/site-packages/tensorflow_core/python/framework/tensor_shape.py in __init__(self, value)
    191       raise TypeError("Cannot convert %s to Dimension" % value)
    192     else:
--> 193       self._value = int(value)
    194       if (not isinstance(value, compat.bytes_or_text_types) and
    195           self._value != value):

TypeError: int() argument must be a string, a bytes-like object or a number, not 'tuple'
```

</details>

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
```diff
@@ -0,0 +1,284 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "all_embeddings = emb_model.predict(generator.flow(G.nodes()))\n",
+    "\n",
+    "y = node_subjects.astype(\"category\").cat.codes\n",
```
huonw: A more natural phrasing of this (as in, more applicable for other tasks a user might do with this) might be to leave this as categorical with the human readable names, and do the `cat.codes` below:

```python
y = node_subjects.astype("category")

...

ax.scatter(
    ...
    c=emb_transformed["label"].cat.codes,
    ...
)
```

Alternatively, leave this line as is, but remove the `astype("category")` from the plotting cell, since it's not needed.

tests/mapper/test_node_mappers.py
```diff
@@ -904,3 +904,30 @@ def test_generator_methods(self):
...
+def test_corrupt_full_batch_generator(sparse):
+
+    G, _ = create_graph_features()
+
+    generator = FullBatchNodeGenerator(G, sparse=sparse)
+
+    base_gen = generator.flow(G.nodes())
+    gen = CorruptedNodeSequence(base_gen)
+
+    [shuffled_feats, features, *_], targets = gen.__getitem__(0)
```
huonw:
```suggestion
    [shuffled_feats, features, *_], targets = gen[0]
```

stellargraph/layer/deep_graph_infomax.py
PantelisElinas: I'm not sure if I understood correctly so let's work out the tensor dimensions.

The input `node_features` is (1, N, F) where N is the number of nodes in the graph and F is the dimensionality of the features. The first dimension, 1, is the batch size. In your case it is always 1 so you remove it to get an (N,F) tensor and then average across the rows to get an (F, ) tensor that is `summary`.  Your corrupted `node_features_corrupted` also has the first dimension removed so you have an (N, F) tensor. Your discriminator assumes that you are given an (N, F) and an (F, ) tensors which I think is a strong assumption if you want to support batch sizes greater than 1.

The `GlobalAveragePooling1D` will take as input your (1, N, F) tensor and the return (1, F). At this point, you can remove the first dimension and you have an (F, ) tensor for `summary` and you can continue as before. However, if you implement your `Discriminator` to work with inputs (batch size, N, F) and (batch size, F) then it will work for all batch sizes and you won't have to squeeze and then add the batch size. You should be able (I think) to re-write your `Discriminator` layer to use Keras `batch_dot` operator (https://www.tensorflow.org/api_docs/python/tf/keras/backend/batch_dot) and then you can pass tensors that don't have the batch dimension removed.

PantelisElinas:
> > Now i'm getting 74-79% accurary. I suspect it may have something to do with the logistic regression, the paper reports a ~40% accuracy with logistic regression on raw features while I'm getting ~60% so I'm definitely doing something.

@kieranricardo please check if the paper uses the largest connected component for their experiments and also if they use the fixed, balanced splits. It might make a difference of a few percent.

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
huonw: > but we should at least have an issue about "some models fail with tensorflow 2.0.0" to deal with in some form before the next release

https://github.com/stellargraph/stellargraph/issues/1045

kieranricardo:
@PantelisElinas @huonw I found and implemented a few differences with the paper:

- the paper uses a specific train/test/val split. The training split is balanced, and interestingly the model consistently performs better on the test set compared to the val set. Doing a random split with a class balanced training set yields worse results, so I think this test set is relatively "easy"
- the paper row normalizes the features
- the early stopping the paper implements is slightly different to keras

With these implemented I'm getting an average accuracy of 81%

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,148 @@
...
+            model = Model(inputs=x_in, outputs=x_out)
+            model.compile(loss=tf.nn.sigmoid_cross_entropy_with_logits, ...)
+            ```
+
+        Where `dg_infomax` is an instance of `DeepGraphInfoMax`.
+
+        Returns:
+            input and output layers for use with a keras model
+        """
+        x_inp, node_feats = self.base_model.build(multiplicity=1)
```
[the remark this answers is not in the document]
kieranricardo: good point, I'll add a warning 👍

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,148 @@
...
+        discriminator = Discriminator()
+        scores = discriminator([node_feats, summary])
+        scores_corrupted = discriminator([node_feats_corrupted, summary])
+
+        x_out = tf.stack([scores, scores_corrupted], axis=1)
+
+        x_out = K.expand_dims(x_out, axis=0)
+        return x_corr + x_inp, x_out
+
+    def embedding_model(self, model):
```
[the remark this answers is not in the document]
kieranricardo: That's a good point, one solution would be to add a unique identifier to `DeepGraphInfoMax._NODE_FEATS`:

```
class DeepGraphInfoMax:
    def init(self, ...):
        self._NODE_FEATS = "INFO_MAX_NODE_FEATURES" + ''.join(map(chr, np.random.randint(32, 122, size=8)))
        ...
```

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+        discriminator = Discriminator()
+        scores = discriminator([node_feats, summary])
+        scores_corrupted = discriminator([node_feats_corrupted, summary])
+
+        x_out = tf.stack([scores, scores_corrupted], axis=1)
+
+        x_out = K.expand_dims(x_out, axis=0)
+        return x_corr + x_inp, x_out
```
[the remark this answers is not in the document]
kieranricardo: Actually, it would be just as easy to have the corrupted inputs last. I just chose arbitrarily.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,148 @@
...
+# limitations under the License.
+
+from . import GCN, GAT, APPNP, PPNP
+from ..core.experimental import experimental
+
+from tensorflow.keras.layers import Input, Dense, Lambda, Layer
+import tensorflow as tf
+from tensorflow.keras import backend as K
+
+__all__ = ["DeepGraphInfoMax"]
```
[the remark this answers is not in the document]
kieranricardo: I'll rename it to a more descriptive`BilinearTransform` (fyi pytorch actually has an almost identical layer https://pytorch.org/docs/stable/nn.html#bilinear)

stellargraph/layer/deep_graph_infomax.py
kieranricardo: > Why is squeezing out the batch dimension convenient for those things?

For most downstream tasks its useful to have embeddings in the shape `(number_of_nodes, dimension)` without the batch dimension of 1 e.g. plotting, clustering, training a neural network. So I feel like its more convenient for the user if we squeeze out the batch dim for them.

It also means that the embedding model will behave the same for graphsage type algorithms. In this case the embedding model will return tensors with shape `(head_nodes, dimension)`, so the user won't have to change any downstream code i.e. removing/adding batch dim squeezing code.

Re. it being convenient for the `Discriminator` @PantelisElinas makes a good point that it could re-written to handle both `(1, nodes, feats)` and `(head_nodes, feats)` input tensors quite easily. 

I think the `GlobalAveragePooling1D` won't work for `GraphSAGE` but it makes sense to worry about `GraphSAGE` implementation when I'm actually implementing it! I'll add `GlobalAveragePooling1D`, and maybe still squeeze out the batch dim for the embedding model. Thoughts?

stellargraph/layer/deep_graph_infomax.py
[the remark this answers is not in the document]
huonw: That seems reasonable, although the details might be worth tweaking, e.g.:

```python
import string
import random

self._NODE_FEATS = "INFO_MAX_NODE_FEATURES_" + "".join(random.choices(string.ascii_lowercase, k=8))
```

or just use the ID of the object:

```python
self._NODE_FEATS = f"INFO_MAX_NODE_FEATURES_{id(self)}"
```

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +528,40 @@ def __len__(self):
 
     def __getitem__(self, index):
         return self.inputs, self.targets
+
+
+class CorruptedNodeSequence(Sequence):
+    """
+    Keras compatible data generator that wraps a FullBatchSequence ot SparseFullBatchSequence and provides corrupted
```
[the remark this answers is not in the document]
kieranricardo: I think that's a good point, its kinda confusing to start exposing the `Sequences` here. I think the `CorruptedGenerator` approach sounds good, and can be just as easily extended to `tf.data` generators :+1:

kieranricardo:
Just a side note, the author's implementation takes 2.5s per epoch, while this implementation takes 0.2s per epoch!

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
```diff
@@ -0,0 +1,284 @@
...
+   "metadata": {},
+   "source": [
+    " ## Data Generators\n",
+    " \n",
+    " Now we create the data generators using `CorruptedFullBatchNodeGenerator`. `CorruptedFullBatchNodeGenerator` returns shuffled node features along with the regular node features and we train our model to discriminate between the two. \n",
+    " \n",
+    " Note that:\n",
+    " \n",
+    " - We typically pass all nodes to `generator.flow` because this is an unsupervised task\n",
+    " - We don't pass `targets` to `generator.flow` because these are binary labels (true nodes, false nodes) that are created by `CorruptedFullBatchNodeGenerator`"
```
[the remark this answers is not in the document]
huonw: Reminder:

```suggestion
    " - We don't pass `targets` to `generator.flow` because these are binary labels (true nodes, false nodes) that are created by `CorruptedGenerator`"
```

and the same for all other mentions of `CorruptedFullBatchNodeGenerator`.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,156 @@
...
+            inputs: a list or tuple of tensors with shapes [(N, F), (F,)] containing the node features and summary feature
+                vector.
+        """
+
+        features, summary = inputs
+
+        summary_rank = len(summary.shape)
+        feat_rank = len(features.shape)
+
+        score = tf.tensordot(self.kernel, summary, [[1], [summary_rank - 1]])
```
huonw: (The docs for `tensordot` are ... really confusing, and seem to have a pile of typos. 😦 )

I take it you switched to this more general operator because the tensors now have the batch dimension, and so `tf.matvec` no longer works?

Does it work if `self.kernel` has a batch dimension added?

```suggestion
        score = tf.matvec(tf.expand_dims(self.kernel, axis=0), summary)
```

or even just add a `1` to the `add_weight`: `shape=(1, ...)`.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,156 @@
...
+    This Layer computes the Discriminator function for Deep Graph Infomax (https://arxiv.org/pdf/1809.10341.pdf).
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def build(self, input_shapes):
+
+        self.kernel = self.add_weight(
+            shape=(input_shapes[0][-1], input_shapes[0][-1]),
```
huonw: As a minor code style thing, this could be rephrased to be a bit more self-explanatory:

```python
feature_size = input_shapes[0][-1]
self.kernel = self.add_weight(
    shape=(feature_size, feature_size)
```

Or, alternatively, more general:

```python
first_size = input_shapes[0][-1]
second_size = input_shapes[1][-1]
self.kernel = self.add_weight(
    shape=(feature_size, second_size)
```

and then the `summary` argument doesn't _have_ to match the `features` one (cf. the discussion about the bilinear name above).

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,156 @@
...
+            constraint=None,
+        )
+        self.built = True
+
+    def call(self, inputs):
+        """
+        Applies the layer to the inputs.
+
+        Args:
+            inputs: a list or tuple of tensors with shapes [(N, F), (F,)] containing the node features and summary feature
```
huonw: I think these shapes are now slightly wrong?

```suggestion
            inputs: a list or tuple of tensors with shapes [(batch, N, F), (batch, F,)] containing the node features and summary feature
```

maybe?

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,156 @@
...
+            dg_infomax = DeepGraphInfoMax(...)
+            x_in, x_out = dg_infomax.build()
+            model = Model(inputs=x_in, outputs=x_out)
+            model.compile(loss=tf.nn.sigmoid_cross_entropy_with_logits, ...)
+
+        Returns:
+            input and output layers for use with a keras model
+        """
+
+        if self.base_model.multiplicity != 1:
```
huonw: I think the warning should happen as early as possible, i.e. in `__init__`, so that the user is alerted immediately when they do something incorrect.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,156 @@
...
+            model (keras.Model): the base Deep Graph Infomax model with inputs and outputs created from
+                DeepGraphInfoMax.build()
+        Returns:
+            input and output layers for use with a keras model
+        """
+        x_emb_in = model.inputs[len(self._corruptible_inputs_idxs) :]
+
+        x_emb_out = self._node_feats
+
+        if x_emb_out.shape[0] == 1:
```
huonw: What's this `if` statement needed for?

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,156 @@
...
+
+        Args:
+            model (keras.Model): the base Deep Graph Infomax model with inputs and outputs created from
+                DeepGraphInfoMax.build()
+        Returns:
+            input and output layers for use with a keras model
+        """
+        x_emb_in = model.inputs[len(self._corruptible_inputs_idxs) :]
+
+        x_emb_out = self._node_feats
```
huonw: What happens if `embedding_model` is called before `build`?

stellargraph/mapper/full_batch_generators.py
```diff
@@ -503,3 +505,26 @@ def flow(self, node_ids, targets=None):
...
+    def __init__(self, base_generator):
+
+        if not isinstance(base_generator, FullBatchNodeGenerator,):
+            raise TypeError(
+                f"base_generator: expected FullBatchNodeGenerator, "
+                f"found {type(base_generator).__name__}"
+            )
+        self.base_generator = base_generator
+
+    def flow(self, *args, **kwargs):
```
huonw: This should have docs that point the describe the connection between these args and the `base_generator` ones.

tests/mapper/test_node_mappers.py
huonw: Did you forget?

tests/test_utils/graphs.py
```diff
@@ -198,7 +198,12 @@ def example_graph_1_saliency_maps(feature_size=None):
 
 
 def example_graph_random(
-    feature_size=4, n_edges=20, n_nodes=6, n_isolates=1, is_directed=False
+    feature_size=4,
+    n_edges=20,
+    n_nodes=6,
+    n_isolates=1,
+    is_directed=False,
+    rand_feats=False,
```
huonw: Since this is used rarely, I think it's ok to not be too brief here, and make it easier to understand what it's referring to when reading the code in isolation:

```suggestion
    random_features=False,
```

Another option would be to try always using random features (i.e. replacing the `np.ones` with `np.random.random` in the next line), and see if anything is currently relying on `1` features.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,156 @@
...
+
+from tensorflow.keras.layers import Input, Lambda, Layer, GlobalAveragePooling1D
+import tensorflow as tf
+from tensorflow.keras import backend as K
+import warnings
+
+__all__ = ["DeepGraphInfoMax", "BilinearTransform"]
+
+
+class BilinearTransform(Layer):
```
huonw: This is now a generic name that is a slightly peculiar bilinear transform, because it's not quite fully general like the PyTorch layer: that layer allows for `matrix * Weight * matrix + Bias` with pretty arbitrary dimensions, but this only allows `matrix * Weight * vector + 0` with `matrix.shape[1] == vector.shape[0]`. In addition, the documentation no longer quite matches the name: if it's going to be a bilinear transform, the docs should talk about it from that focus (possibly with a mention of DGI, but not only that).

To keep the scope small and thus keep this PR slightly simpler, I'd personally be happy to call this something very specific, e.g.

```suggestion
class DeepGraphInfoMaxDiscriminator(Layer):
```

or

```suggestion
class DGIDiscriminator(Layer):
```

and then, if we need it for other things in future, we can generalise and rename then (and it should be possible to have a migration path so that users of this layer directly aren't broken).

What do you think?

kieranricardo: I think that makes sense, as you mentioned its not quite a general bilinear transform and we can generalize it later if needed 👍

stellargraph/layer/deep_graph_infomax.py
kieranricardo: `x_emb_out` will be `None` and the user will get a (probably) confusing error message when they try to buildtheir `keras.Model`. Same outcome if they call `DeepGraphInfoMax.embedding_model` with a `model` not created by the right instance of `DeepGraphInfoMax`. 

I'll catch both of these cases here and give an understandable error message

stellargraph/layer/deep_graph_infomax.py
kieranricardo: (More like the docs for `tensorflow` ... 😕 )

Actually I never tested out just using `matvec`, just tried it and it works fine! I had some other issues with keeping the batch dimension and prematurely blamed `matvec` and swapped it out for `tensordot`. I've put `matvec` back in and things are simple again :)

stellargraph/layer/deep_graph_infomax.py
kieranricardo: I put this in here thinking about generalising this method to `GraphSAGE` by only removing the batch dimension of 1 if its present. But I feel like this is bad practice?

tests/mapper/test_node_mappers.py
kieranricardo: ah.. yeah basically. Added this in, broke tests (not related to this change), removed this, and and then forgot to add it back in!

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
```diff
@@ -187,7 +187,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Test classification accuracy: 0.7805578342904019\n"
+      "Test classification accuracy: 0.7358490566037735\n"
```
huonw: This dropped significantly with this run. Is the variance very high? It looks like we're still significantly below the paper, and higher variance than it?

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -127,14 +128,22 @@ def build(self):
...
+        x_out = tf.stack([scores, scores_corrupted], axis=2)
 
         return x_corr + x_inp, x_out
 
+    @property
+    def _unique_id(self):
+        """
+        A unique string id for this object to be used as a keras layer name.
+        """
+        id = str(self).replace(" ", "")[1:-1]
```
huonw: I think all of this can be replaced with just the `id` function, since it's a more complicated way of extracting that information (the only unique bit in the default `str` is the hex number after `at ...` and that is exactly `hex(id(...))`): https://docs.python.org/3/library/functions.html#id

```python
>>> o = object()
>>> hex(id(o))
'0x1033872f0'
>>> str(o)
'<object object at 0x1033872f0>'
```

Could you say why you went with this approach?

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -127,14 +128,22 @@ def build(self):
             GlobalAveragePooling1D()(self._node_feats)
         )
 
-        discriminator = BilinearTransform()
+        discriminator = DGIDiscriminator(name=self._unique_id)
```
huonw: Why set the name on this layer rather than on the layer that is actually used by `embedding_model`?

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -145,12 +154,17 @@ def embedding_model(self, model):
         Returns:
             input and output layers for use with a keras model
         """
-        x_emb_in = model.inputs[len(self._corruptible_inputs_idxs) :]
 
+        if not any(layer.name == self._unique_id for layer in model.layers):
+            raise ValueError(
+                f"model: model must be a keras model with inputs and outputs created "
+                f"by this instance of DeepGraphInfoMax.build() function."
```
huonw: It's not entirely clear to me what this sentence means, because I parse it as "by this instance of (DeepGraphInfoMax.build() function)". I guess you were trying to go for "by (this instance of DeepGraphInfoMax)'s .build() function"? Maybe:

```suggestion
                f"by the build() method of this instance of DeepGraphInfoMax"
```

stellargraph/layer/deep_graph_infomax.py
kieranricardo: unfortunately `_node_feats` is a tensor not a layer, so I don't know of a straightforward way of renaming the relevant layer.

stellargraph/layer/deep_graph_infomax.py
kieranricardo: nice, that reads easier :+1:

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
kieranricardo: Performance variance with train/test/val splits is very high. The paper has a custom cora data set that's pre-split so I was only able to replicate it by using their data loaders, I have a notebook for this if you're interested and running this gets consistent results around 81%.

stellargraph/layer/deep_graph_infomax.py
kieranricardo: I wasn't aware of this function, cheers!

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
huonw: Ah, I didn't realise you were doing that in a separate notebook. If we're not going to include in this notebook (which is fine), we should at least include a comment/piece of text somewhere that mentions the difference.

stellargraph/layer/deep_graph_infomax.py
huonw: Ah, right. So this `DGIDiscriminator` layer basically ends up as a flag marking a model as coming from this `DeepGraphInfoMax` instance, and then, having done that validation, we pull out the actual embedding layer separately?

I'm a little concerned by the mutation with `self._node_feats`: if someone creates two models using a single instance, I think they'll get different models for training, but a very confused embedding model (I think the out tensor will be the one from the second model, always):

```python
dgi = DeepGraphInfoMax(...)
inp1, out1 = dgi.build()
model1 = Model(inpt1, out1)

inp2, out2 = dgi.build()
model2 = Model(inpt2, out2)

emb_inp1, emb_out1 = dgi.embedding_model(model1)
emb_inp2, emb_out2 = dgi.embedding_model(model2)
```

Given this, I suspect `emb_out1 is emb_out2` (i.e. they're exact the same object) and is the embedding layer for `model2`, but `emb_inp1` are the inputs for `model1`. That is, the embedding layer for `model1` has been lost.

Maybe, to address all of this, we could put the marker on the embedding layer by creating a pass-through identity layer and not doing any mutations of the `DeepGraphInfoMax` object, something like:

```python
x_inp, x_out = self.base_model.build(multiplicity=1)

# identity layer so that we can name it to retrieve in `embedding_model`
node_feat = Lambda(lambda x: x, shape=x_out.shape[1:], name=...)(x_out)
```

(Inspired by https://stackoverflow.com/a/47297842/1256624)

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +528,40 @@ def __len__(self):
...
+                f"base_generator: expected FullBatchSequence or SparseFullBatchSequence, "
+                f"found {type(base_generator).__name__}"
+            )
+
+        self.base_generator = base_generator
+        self.targets = np.zeros((1, len(base_generator.target_indices), 2))
+        self.targets[0, :, 0] = 1.0
+
+    def __len__(self):
+        return self.base_generator.__len__()
```
huonw:
```suggestion
        return len(self.base_generator)
```

(Usually, one should never call a dunder method like `__len__` or `__getitem__` if there's an alternative: the alternatives are cleaner.)

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,162 @@
...
+
+        features, summary = inputs
+
+        score = tf.linalg.matvec(features, tf.linalg.matvec(self.kernel, summary),)
+
+        return score
+
+
+@experimental(reason="lack of unit tests", issues=[1003])
+class DeepGraphInfoMax:
```
huonw: Oh, I just realised: Given this is styled as "Deep Graph Infomax" or "DGI" in the paper, maybe it should be called:

```suggestion
class DeepGraphInfomax:
```

tests/layer/test_deep_graph_infomax.py
```diff
@@ -0,0 +1,76 @@
...
+# limitations under the License.
+
+from stellargraph.layer import DeepGraphInfoMax, GCN
+from stellargraph.mapper import FullBatchNodeGenerator, CorruptedGenerator
+from ..test_utils.graphs import example_graph_random
+import tensorflow as tf
+import pytest
+
+
+def test_dgi():
```
huonw: Given we now have a basic unit test, do you think #1003 is basically just "add tests for DGI with GAT, PPNP, APPNP" now?

stellargraph/layer/deep_graph_infomax.py
kieranricardo: hmm yeah that's a good point. I'll wrap `node_feats` in a `Lambda` layer with a randomly generated name

tests/layer/test_deep_graph_infomax.py
kieranricardo: Yeah I think #1003 is adding tests for GAT/PPNP/APPNP along with maybe testing some edge cases. It shouldn't be too much work and I'm happy to add them in there if that would easier review wise.

huonw: Nah, unless you're really enthusiastic, this PR has already been open long enough, let's not add another chunk of new code to review.

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,162 @@
...
+        if base_model.multiplicity != 1:
+            warnings.warn(
+                f"multiplicity: expected the base_model to have a multiplicity of 1, found"
+                f" ({self.base_model.multiplicity}). A multiplicity of 1 will be used to construct the base model."
+            )
+
+        self.base_model = base_model
+
+        self._node_feats = None
+        self._unique_id = hex(id(self))
```
huonw: Just to make it look nice when people investigate or visualise the model, maybe we could include a hint about where this came from, e.g.

```suggestion
        self._unique_id = f"DEEP_GRAPH_INFOMAX_{id(self)}"
```

(Since it's building the string directly, no need to hex-ify it, although it could be e.g.`...{id(self):x}`.)

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -0,0 +1,162 @@
...
+            x_in, x_out = dg_infomax.build()
+            model = Model(inputs=x_in, outputs=x_out)
+            model.compile(loss=tf.nn.sigmoid_cross_entropy_with_logits, ...)
+
+        Returns:
+            input and output layers for use with a keras model
+        """
+
+        x_inp, node_feats = self.base_model.build(multiplicity=1)
+        node_feats = Lambda(lambda x: x, name=self._unique_id)(node_feats)
```
huonw: A brief comment would make it obvious why we're doing this mildly peculiar thing. E.g.

```suggestion
        # identity layer so we can attach a name to the tensor
        node_feats = Lambda(lambda x: x, name=self._unique_id)(node_feats)
```

tests/layer/test_deep_graph_infomax.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # check case when infomax_2.build() has been called
+    model_2 = tf.keras.Model(*infomax_2.build())
+    with pytest.raises(ValueError, match="model: *."):
+        infomax_2.embedding_model(model_1)
+
+    with pytest.raises(ValueError, match="model: *."):
+        infomax_1.embedding_model(model_2)
+
+    emb_model_1 = tf.keras.Model(*infomax_1.embedding_model(model_1))
```
huonw: Are these checking that the embedding model can be correctly constructed? As written, I'm not sure it is testing anything beyond what `test_dgi` does and kinda makes this test seem like it was truncated, without a comment to provide some context.

In particular, I don't think this is going to catch problems like that which we had with old code (with `self._node_feats`), where it was mutating a property on a single `DeepGraphInfomax` instance if `build` was called multiple times.

tests/test_utils/graphs.py
```diff
@@ -210,6 +215,10 @@ def example_graph_random(
     node_ids = range(n_nodes)
     if feature_size is None:
         nodes = pd.DataFrame(index=node_ids)
+    elif random_features:
+        nodes = pd.DataFrame(
+            np.random.random((len(node_ids), feature_size)), index=node_ids
```
huonw: Just to understand, did you try my alternate suggestion of swapping the `ones` for `random` to avoid having to add yet another parameter?

tests/mapper/test_node_mappers.py
huonw: Still not added?!

kieranricardo: what?? I have the change locally, I'l double check with github this time

tests/layer/test_deep_graph_infomax.py
kieranricardo: This test is checking that trying to create an inappropriate embedding model with give an error. Agreed, this test needs more work to test that edge case but I'll leave it for #1003 (it could also have line 75 and 76 deleted, which confuse the purpose of the test).

huonw: My comment was just talking about lines 75 and 76, FWIW.

kieranricardo:
@huonw comment is buried now, but I checked and `inputs, _ = self.base_generator[index]` is on github https://github.com/stellargraph/stellargraph/blob/feature/gcn-deep-graph-infomax/stellargraph/mapper/sequences.py#L561

kieranricardo:
> Still not added?!

@huonw comment is buried now, but I checked and `inputs, _ = self.base_generator[index]` is on github https://github.com/stellargraph/stellargraph/blob/feature/gcn-deep-graph-infomax/stellargraph/mapper/sequences.py#L561

huonw:
That's a different file. The relevant code and comment is visible on the files tab: https://github.com/stellargraph/stellargraph/pull/1001/files#r389416495

tests/test_utils/graphs.py
kieranricardo: No I didn't sorry, tried it out and it works find 👍

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
```diff
@@ -0,0 +1,293 @@
...
+    "gen = corrupted_generator.flow(G.nodes())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Model Creation and Training\n",
+    "\n",
+    "We create an train our `GCNInfoMax` model. Note that the loss used here must always be `tf.nn.sigmoid_cross_entropy_with_logits`."
```
huonw:
```suggestion
    "We create and train our `DeepGraphInfomax` model. Note that the loss used here must always be `tf.nn.sigmoid_cross_entropy_with_logits`."
```

demos/node-classification/deep-graph-info-max/gcn-deep-graph-info-max-cora-example.ipynb
```diff
@@ -0,0 +1,293 @@
+{
```
huonw: Oh, another minor thing in addition to the spelling: this should be named `demos/node-classification/deep-graph-infomax/gcn-deep-graph-infomax-cora-example.ipynb`.

That is, replace both `info-max`s with `infomax`, matching the paper and the `InfoMax` -> `Infomax` renaming.

PR #1005, opened 2020-03-03
Remove legacy construction from mappers

huonw:
See: #717


huonw:
Folding this into #1000

Issue #933, opened 2020-02-25
Review and clean up tensorflow warnings


timpitman:
### Description

We've got many warnings on CI from Tensorflow when running demo notebooks - while these are mostly stripped from the published outputs (except #826), Demos should ideally run without warnings.

We should go through and address as many warnings as possible (eg fix deprecated functions such as already done for #803 ) - in some cases raising new issues if there's significant work.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

kieranricardo:
Warnings I've found:

- `ensemble-link-prediction-example`: `WARNING:tensorflow:Early stopping conditioned on metric `val_weighted_acc` which is not available. Available metrics are: loss,acc,val_loss,val_acc`
- `hinsage/movielens-recommender.ipynb`: `WARNING:tensorflow:multiprocessing can interact badly with TensorFlow, causing nondeterministic deadlocks. For high performance data pipelines tf.data is recommended



WARNING:tensorflow:multiprocessing can interact badly with TensorFlow, causing nondeterministic deadlocks. For high performance data pipelines tf.data is recommended.
 





WARNING:tensorflow:multiprocessing can interact badly with TensorFlow, causing nondeterministic deadlocks. For high performance data pipelines tf.data is recommended.

Issue #826, opened 2020-02-10
incomplete removal of tensorflow warnings from format-notebooks.py


timpitman:
### Describe the bug

`format-notebooks.py` doesn't fully remove multiline tensorflow warnings from cell output.

Found in #816: https://github.com/stellargraph/stellargraph/pull/816#discussion_r376823876

### To Reproduce

1. run a notebook, with reformatting:
```
python ./scripts/format_notebooks.py --default -e --overwrite demos/link-prediction/gcn/cora-gcn-links-example.ipynb
```

### Observed behaviour

see the last cell (24), in the top of the output is spurious:
```
  ['...']
  ['...']
```

note the original output (before being stripped):
```
WARNING:tensorflow:sample_weight modes were coerced from
  ...
    to  
  ['...']
WARNING:tensorflow:sample_weight modes were coerced from
  ...
    to  
  ['...']
```

### Expected behaviour

Ideally, the full warning text should be removed.  If this isn't possible, another approach (eg disabling TF warnings prior to running notebooks) should be considered.

PR #1008, opened 2020-03-03
Fix some tensorflow warnings in notebooks

kieranricardo:
This PR fixes some of the warnings in the CI tested demo notebooks

**Tensorflow warnings fixed:**

- ensemble-link-prediction-example has a warning due to changing of names of metrics in tensorflow.

         `WARNING:tensorflow:Early stopping conditioned on metric `val_weighted_acc` which is not 
          available. Available metrics are: loss,acc,val_loss,val_acc`

**Other warnings fixed:**

- sklearn warning about logistic regression in attri2vec-citeseer-node-classification-example.ipynb. The logistic regression model doesn't converge here. This is fixed by increasing the number of iters

       `ConvergenceWarning: lbfgs failed to converge (status=1):`

- pandas warning in cluster-gcn-node-classification.ipynb :

      `SettingWithCopyWarning: A value is trying to be set on a copy of a slice from a DataFrame`

See https://github.com/stellargraph/stellargraph/issues/933 for more info


huonw:
FWIW, from a developer/review/future perspective, it's much better to be specific about what a patch fixes in the PR description, especially in this case because the comment in #933 discusses more than just what this PR touches, and, the actual changes here are disguised by all the noise from rerunning the notebooks.

The PR description would then go into the squash-and-merge commit message to make it easy to find and useful for someone in future using `git blame`. (For example, in https://github.com/stellargraph/stellargraph/commit/be92b67a42f6a9c5dfcafa4f36015920c1d50ee8 I deleted the auto-generated GitHub message which is noisy/useless, and copied in the description from the PR https://github.com/stellargraph/stellargraph/pull/865 .)

demos/ensembles/ensemble-link-prediction-example.ipynb
```diff
@@ -321,6 +321,7 @@
       "\n",
```
kjun9: <p>Are we keeping the <code>['...']</code>s  for now?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1008/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='50'/>

demos/node-classification/attri2vec/attri2vec-citeseer-node-classification-example.ipynb
```diff
@@ -104,6 +104,7 @@
       "\n",
```
kjun9: <p>Wow, so much faster!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1008/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

kjun9: <p>Comment should be updated to say 20% and 80% if we require 20% to avoid warnings</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1008/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

demos/node-classification/cluster-gcn/cluster-gcn-node-classification.ipynb
```diff
@@ -358,12 +358,12 @@
      "data": {
```
kjun9: <p>Why is the previous cell removed?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1008/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

kieranricardo: the previous cell wasn't actually affecting anything downstream and was editing copies of a DataFrame causing the warnings! So I just removed it

demos/node-classification/attri2vec/attri2vec-citeseer-node-classification-example.ipynb
kieranricardo: oh I didn't notice; is that from the new stellar class?

demos/node-classification/cluster-gcn/cluster-gcn-node-classification.ipynb
kjun9: Hmm I wonder if it's considered bad data science practice to not separate the ground truth labels from the data though?  Is there a way we can still do that without producing warnings?

demos/node-classification/attri2vec/attri2vec-citeseer-node-classification-example.ipynb
kjun9: I'm guessing so, each epoch has dropped to 4s from 30s!

demos/ensembles/ensemble-link-prediction-example.ipynb
kieranricardo: oh I thought we had fixed the stripping of these warnings?

Yeah keeping them for now because it seems like an issue with tensorflow, e.g. tensorflow's official keras train example has these warnings (https://www.tensorflow.org/guide/keras/train_and_evaluate). I saw someone post a work around of SO but it doesn't worth the effort to implement for all the demos when hopefully this will be fixed in a future release.

demos/node-classification/attri2vec/attri2vec-citeseer-node-classification-example.ipynb
kieranricardo: wow, thats amazing! nice work @kjun9 @huonw

kieranricardo: nice catch!

demos/node-classification/cluster-gcn/cluster-gcn-node-classification.ipynb
kieranricardo: fair point, it should be pretty easy to to do. i'll give it a go

kieranricardo: done! I added the cell back in but used `train_data = train_data.drop(..., inplace=False, ...)`

demos/node-classification/attri2vec/attri2vec-citeseer-node-classification-example.ipynb
huonw: <p>Interesting. I'll investigate to verify, because if it is that much faster, that's a good-news-story we can tell!</p>

demos/ensembles/ensemble-link-prediction-example.ipynb
kjun9: <p>I'm missing a bit of context for why this had <code>val_weighted_acc</code>  in the first place, when it apparently doesn't exist. Possibly worth checking with @PantelisElinas?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1008/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='52'/>

demos/node-classification/cluster-gcn/cluster-gcn-node-classification.ipynb
```diff
@@ -357,13 +357,13 @@
     {
```
kjun9: <p>Missing a plot here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1008/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='79'/>

kieranricardo: ah.. dunno how that happened but nice catch!

demos/ensembles/ensemble-link-prediction-example.ipynb
kieranricardo: yeah I'm not sure but it seems like a typo and I think this warning is new so maybe it was never picked up. Either way I think using the `val_acc` is reasonable here

kjun9: okay sure 👍

kjun9: Hmm I don't see these early stopping parameters in the tensorflow docs for `Model` at all actually 🤔 https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit_generator
or even in the 1.15 docs https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/keras/Model#fit_generator

Does it seem like it's actually doing anything currently...?

The early stopping docs for tensorflow looks very different: https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/EarlyStopping

kieranricardo: okay checked this out and just realised that `model` here is an instance of our custom `BaggingEnsemble` not `tf.keras.Model`, and `BaggingEnsemble.fit_generator` supports this argument

kjun9: Oops thanks for figuring that out! 

I do feel like this is potentially a bug though - in an earlier cell it looks like when we compile the model, we use
```python
model.compile(
    optimizer=keras.optimizers.Adam(lr=1e-3),
    loss=keras.losses.binary_crossentropy,
    weighted_metrics=["acc"],
)
```
so `val_weighted_acc` should be available, but it's not? And if we use `val_acc` instead, are we supposed to be passing in `metrics=["acc"]` when we compile the model?

The existing warning message said:
```
WARNING:tensorflow:Early stopping conditioned on metric `val_weighted_acc` which is not available. Available metrics are: loss,acc,val_loss,val_acc
```

so it looks to me like something strange seems to be going on, where the `weighted_metrics` parameter ends up creating a normal metric maybe..? I'm happy for it to be a separate issue though, since the warnings have been removed.

huonw: Why have all the leftover `['...']`s appeared now? Those left-over pieces from the warning removal (#826) weren't there before.

> hopefully this will be fixed in a future release

At the moment, this seems more hopeful than not, as it doesn't seem anyone has filed an issue against tensorflow https://github.com/tensorflow/tensorflow/issues?q=is%3Aissue+is%3Aopen+sample_weight+modes+were+coerced+from . Maybe we can do it; it should be ~easy, since the example in their documentation serves as a reproducer.

kieranricardo: Seems like `tf` is still using weighted accuracy but just isn't adding `"weighted_"` to the string.

kieranricardo: My guess is that these warnings appeared here because this is the first time this notebook has been run with `tensorflow` 2.1.0. 

I'll file an issue with tensorflow and see what they say

demos/node-classification/attri2vec/attri2vec-citeseer-node-classification-example.ipynb
huonw: Ah, unfortunately, I cannot reproduce this: I tried running on the latest develop (69fc13bd) and observed epochs taking ~26s, which is the same as what it took on the commit that switched over to the new StellarGraph (2dfff02d701922ed1f2164945dbd75651f9ab6a8), and on the commit before that (172edad0).

Maybe you're running this on a GPU, @kieranricardo?

huonw: > Maybe you're running this on a GPU, @kieranricardo? Or a particularly powerful CPU? (Or some other reason for the difference?)
 
We suspect it may be because he installed `tensorflow-mkl`, and is thus getting the most out of his (Intel) CPU.

demos/ensembles/ensemble-link-prediction-example.ipynb
kieranricardo: filed an issue with tensorflow: https://github.com/tensorflow/tensorflow/issues/37500

wow they have a lot of open issues!

kjun9: Yeah looks like it's probably a "feature"
```
>>> import tensorflow as tf
>>> x_in = tf.keras.layers.Input(1)
>>> x_out = tf.keras.layers.Dense(1)(x_in)
>>> model = tf.keras.Model(inputs=x_in, outputs=x_out)
>>> model.compile(
...     optimizer='adam',
...     loss='categorical_crossentropy',
...     weighted_metrics=["acc"],
...     metrics=["acc"]
... )
>>> print(model.metrics_names)
['loss', 'acc', 'acc_1']
```

kieranricardo: @huonw @kjun9 I got a response. Turns out the warning has already been fixed and the `tf-nightly` version doesn't have the warning

huonw: Great. I wonder if that fix is included in the 2.2.0 release that seems like it'll happen soon (https://github.com/tensorflow/tensorflow/releases/tag/v2.2.0-rc0).

Issue #896, opened 2020-02-19
Demo directory for link prediction via random walks contains a very large amount of code


kjun9:
### Description

[`demos/link-prediction/random-walks`](https://github.com/stellargraph/stellargraph/tree/4070ccf24d67802a08a9f70ca3e7567a0a53a7fc/demos/link-prediction/random-walks/utils) contains a `utils` module which contains a substantial amount of code. 

We should move as much of this into the core library as possible, since moving this into the library means:
* the code is much easier to maintain, and can be tested independently from the demos
* it's easier to reuse it across more notebooks
* a user can follow/copy the demo notebook without having to dig into / copy a whole mini module shipped alongside the demo

### User Story

**As** a data scientist

**I want to** use stellargraph functions to help me do link prediction instead of more code shipped with a demo

**So that** my job is easier

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
Not only does this contain a large amount of code, but it seems to contain its own implementation of node2vec https://github.com/stellargraph/stellargraph/tree/develop/demos/link-prediction/random-walks/utils/node2vec (taken seemingly directly from https://github.com/aditya-grover/node2vec) rather than using the one in the library.

Issue #827, opened 2020-02-10
Add temporal walk demo notebook


kjun9:
### Description

We need a demo notebook to show that we can reproduce the findings in the reference paper for temporal walks

### User Story

**As** a data scientist

**I want to** run a demo notebook for temporal walks

**So that** I can see its usefulness

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

kjun9:
Current version of this isn't producing results that match the paper, but the performance still looks good - to move forward, we could just make a note of the observed differences for now.

PR #1020, opened 2020-03-05
Introduce `LinkPredictionClassifier` in the library

kjun9:
This begins tackling #896 enough to make #827 easier and less hacky to put together.

The new class `LinkPredictionClassifier` is designed to be used on top of an existing feature learner (e.g. Node2Vec, CTDNE) - it's slightly more complex than just having a logistic regression layer on top so I think it's worth having a class for it to handle things like:
* keep the node feature learner used to train the classifier in the class, while allowing the use of another feature learner for prediction/evaluation
* use and evaluation of different binary operators to create edge features
* correctly calculating the roc auc score based on the classes in the classifier

This doesn't remove any code from `demos/link-prediction/random-walks/utils` since everything is still needed to run `demos/link-prediction/random-walks/main.py`

Since some of the details that were hidden away in the link prediction demo are now shown as part of the notebook, I added a bit more markdown to explain what's going on. Specifically:
* the various splits of train/test set and train/test graph and how they are used, is shown more explicitly
* `Node2VecFeatureLearning` object is exposed - this needs to be replaced with something in the library at some point too - possibly with #536


stellargraph/link_prediction.py
```diff
@@ -0,0 +1,185 @@
+# -*- coding: utf-8 -*-
```
kjun9: This is currently at top-level, but happy to move it to wherever seems more appropriate

stellargraph/link_prediction.py
```diff
@@ -0,0 +1,185 @@
...
+        raise ValueError(f"Unexpected binary operator: {binary_operator}")
+
+
+class LinkPredictionClassifier:
+    """
+    Link Prediction Classifier designed to be used with an existing node representation learner.
+
+    Args:
+        transform_node (callable): A function to transform node IDs into node features.
+        binary_operators (list of str, optional): Binary operators applied on node features to
```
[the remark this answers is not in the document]
kjun9: I got the impression from https://github.com/stellargraph/stellargraph/blob/develop/demos/link-prediction/random-walks/utils/predictors.py that this kind of support for evaluating each of the binary operators against each other is a core part of what we see as the correct way of doing link prediction. 

We want to make it easy to write new link prediction demos, and so if we expect all of them to replicate the same sort of binary operator evaluation workflow, I thought it was worth just baking it into a single class? Maybe you could say it's less natural but more opinionated approach than the alternative you've mentioned.

The Node2Vec paper (and CTDNE, and maybe others?) has their setup like this so I guess that's where the original inspiration comes from.

stellargraph/link_prediction.py
```diff
@@ -0,0 +1,185 @@
...
+
+def _operator_l1(u, v):
+    return np.abs(u - v)
+
+
+def _operator_l2(u, v):
+    return (u - v) ** 2
+
+
+def _create_link_feature(pair, binary_operator):
```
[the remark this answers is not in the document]
kjun9: Makes sense 👍 I'll update it

stellargraph/link_prediction.py
```diff
@@ -0,0 +1,185 @@
...
+            self._binary_operators = binary_operators
+        else:
+            self._binary_operators = _BINARY_OPERATORS
+        self._classifiers = dict()
+
+    def _transform_node_pairs(self, link_examples, transform_node=None):
+        if transform_node is None:
+            transform_node = self.transform_node
+        return [
+            (transform_node(src), transform_node(dst)) for src, dst in link_examples
```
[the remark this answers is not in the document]
kjun9: Hmm, I guess then we have to enforce that the `transform_node` function being passed in by the user, supports the vectorised way of computing node features?

Maybe this makes more sense to do once we have node2vec implemented in our library, rather than using `gensim`'s word2vec, since the `transform_node` function will be coming from a trained word2vec model. I'd imagine currently this would just force the user to write a wrapper around the word2vec model to "appear" like it can do vectorised operations.

stellargraph/link_prediction.py
```diff
@@ -0,0 +1,185 @@
...
+        binary_operators (list of str, optional): Binary operators applied on node features to
+            produce the corresponding edge feature. Each binary operator will be used to train and
+            predict a separate classifier. By default, all binary operators are used.
+
+    """
+
+    def __init__(self, transform_node, binary_operators=None):
+        self.transform_node = transform_node
+        if binary_operators is not None:
+            self._binary_operators = binary_operators
```
huonw: This doesn't validate that these are valid; why did you take this approach?

stellargraph/link_prediction.py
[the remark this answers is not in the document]
huonw: Ah; I see. I guess it's fine as is, then (FWIW, vectorising an operation is fairly easy: `np.vectorize(some_func)`).

stellargraph/link_prediction.py
```diff
@@ -0,0 +1,185 @@
...
+            max_iter (int): Maximum number of training iterations for the classifier.
+
+        """
+        node_feature_pairs = self._transform_node_pairs(link_examples)
+        for binary_operator in self._binary_operators:
+            operator_func = _get_operator(binary_operator)
+            link_features = [operator_func(src, dst) for src, dst in node_feature_pairs]
+            clf = Pipeline(
+                steps=[
+                    ("sc", StandardScaler()),
```
huonw: Why is this "feature engineering" done deep in the model here?

stellargraph/link_prediction.py
```diff
@@ -0,0 +1,185 @@
...
+                    (
+                        "clf",
+                        LogisticRegressionCV(
+                            Cs=10,
```
huonw: Where do these magic numbers come from?

stellargraph/link_prediction.py
```diff
@@ -0,0 +1,185 @@
...
+                operators regardless of which classifiers have been trained. By default, all trained
+                classifiers are used.
+
+        Returns:
+            Dict of Scores grouped by binary operator
+        """
+        predictions = self.predict(link_examples, transform_node, binary_operators)
+        scores = dict()
+        for binary_operator, predicted in predictions.items():
+            if self._classifiers[binary_operator].classes_[0] == 1:
```
huonw: This `if` statement seems a bit peculiar, what's special about the first class having value `1`?

stellargraph/link_prediction.py
kjun9: Ah hmm sorry I missed that. I did still want to keep the original strings in order to create entries in the other dicts later e.g. `self._classifiers`. I guess this could be
```python
if binary_operators is not None:
    self._binary_operators = {op: _get_operator(op) for op in binary_operators}
else:
    self._binary_operators = _known_operators
```

huonw:
If we're going to reduce it that much, it seems like it's much less useful? As in, there's almost more code in each demo than in the library?

To get a better sense of the consequences of this, is there another demo we can use as an example of using this new functionality, so we're looking at two and thus hopefully get a better sense of what the an appropriate general solution is?

kjun9:
@huonw 

Currently only the Node2vec and Metapath2vec demos use this, and they're written as options of one main script, and look mostly the same.

The CTDNE (temporal walk) demo I'm working on, if using this `LinkPredictionClassifier` as it stands, looks something like:
```python
temporal_model = Word2Vec(temporal_walks, ...)
clf = LinkPredictionClassifier(lambda node: temporal_model.wv[node])
clf.fit(link_examples, link_labels)
scores = clf.evaluate_roc_auc(link_examples_test, link_labels_test)
print(scores)
```

There's no reason to deviate from the standard/reference approach in any of these demos though, so I'm not sure if it says much about whether it's a good general solution.

Other link prediction models we have are all keras models that train end-to-end, instead of training this separate layer of classifier on top of the embeddings.

PantelisElinas:
@kjun9 your implementation looks reasonable but I have to admit that I am not convinced that this extra layer of abstraction is necessary. 

You might want to have a look at the following notebook that does not use the code from utils and it achieves the same result for a single operator combining node representations to obtain edge representations. The core idea is to have a method that combines node representations and then the user is free to train their binary classifier(s) as they please. It might be better for us to write a general tutorial on how to do link prediction given node representations obtained using any of the methods available in `StellarGraph` and point users to it if they ask about the link prediction use-case. 

See https://github.com/stellargraph/stellar-practical-ml-on-graphs/blob/master/notebooks/node2vec-link-prediction.ipynb 

P.

stellargraph/link_prediction.py
PantelisElinas: You need to know which column indicates the predictions for the edge label, i.e., the label you are actually predicting, indicated by the label 1. Then you can pass the correct `y_score` data to the `roc_auc_score` method. One could train the classifier to predict no-edges instead but that would be indicated by label 0 and calculating the AUC ROC score would require picking  the other column.

huonw: Ah, right, so this `if` block is actually a manual version of:

```python
positive_column = self._classifiers[binary_operator].classes_.index(1)
scores[binary_operator] = roc_auc_score(link_labels, predicted[:, positive_column])
```

kjun9:
Closing this based on feedback - it seems like it looks fine to just perform this kind of workflow in individual notebooks instead of having something in the library for it.

I think the original issue #896 can instead be tackled by moving things out of `utils` and into the notebooks

Issue #1012, opened 2020-03-04
Flaky test: test_GCN_Aadj_feats_op in tests.core.test_utils


huonw:
### Describe the bug

The `test_GCN_Aadj_feats_op` test in `tests.core.test_utils` sometimes fails. E.g .https://buildkite.com/stellar/stellargraph-public/builds/1972#annotation-junit

### To Reproduce

1. Run the test many times?

### Observed behavior

```
example_graph = 

    def test_GCN_Aadj_feats_op(example_graph):
        node_list = list(example_graph.nodes())
        Aadj = example_graph.to_adjacency_matrix()
        features = example_graph.node_features(node_list)
    
        features_, Aadj_ = GCN_Aadj_feats_op(features=features, A=Aadj, method="gcn")
        assert np.array_equal(features, features_)
        assert 6 == pytest.approx(Aadj_.todense().sum(), 0.1)
    
        features_, Aadj_ = GCN_Aadj_feats_op(
            features=features, A=Aadj, method="chebyshev", k=2
        )
        assert len(features_) == 4
        assert np.array_equal(features_[0], features_[0])
        assert np.array_equal(features_[1].todense(), sp.eye(Aadj.shape[0]).todense())
        assert features_[2].max() < 1
>       assert 5 == pytest.approx(features_[3].todense()[:5, :5].sum(), 0.1)
E       assert 5 == 4.4294434742517765 ± 4.4e-01
E        +  where 4.4294434742517765 ± 4.4e-01 = (4.4294434742517765, 0.1)
E        +    where  = pytest.approx
E        +    and   4.4294434742517765 = ()
E        +      where  = matrix([[ 0.6466667 ,  0.31026873,  0.16970565,  0.51854499, -0.3002221 ],\n        [ 0.31026873, -0.64000003,  0.13856...07,  0.40000004, -0.50666664,  0.22861906],\n        [-0.3002221 ,  0.16970564,  0.58787755,  0.22861906,  0.60000004]]).sum

tests/core/test_utils.py:101: AssertionError
```

### Expected behavior

The test never fails.

### Environment

Operating system: CI

Python version: 3.6

Package versions: StellarGraph commit fb20983

<details>

```
absl-py==0.9.0
ansiwrap==0.8.4
appdirs==1.4.3
astor==0.8.1
attrs==19.3.0
backcall==0.1.0
black==19.10b0
bleach==3.1.1
boto==2.49.0
boto3==1.12.12
botocore==1.15.12
[...]

Issue #621, opened 2020-01-08
Refactor GCN and RGCN


kjun9:
### Description

Related to discussion in #558 - variations of GCN (e.g. using Chebyshev polynomials) can use multiple adjacency matrices, much like RGCN, so it makes sense to combine the two classes. 

It might also make sense to do similar refactoring for other homogeneous models with heterogeneous variants in general.

### User Story

**As** a data scientist

**I want to** use a single GCN class for different GCN variants

**So that** I am not confused about which class to use.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1028, opened 2020-03-05
Remove chebyshev filter option

kjun9:
This is currently not implemented correctly and is causing problems
due to tests being flaky #1012 #1013 

The existing code is probably not going to be very useful for
eventually making the option available again, so it is being
removed. Instead, it'll largely depend on refactoring GCN and RGCN,
covered by #621.

### DIFF
--- a/stellargraph/core/utils.py
+++ b/stellargraph/core/utils.py
@@ -159,7 +159,9 @@ def GCN_Aadj_feats_op(features, A, k=1, method="gcn"):
         print("Using GCN (local pooling) filters...")
         A = preprocess_adj(A)
     elif method == "chebyshev":
-        raise ValueError("method 'chebyshev' did not behave correctly and has been removed")
+        raise ValueError(
+            "method 'chebyshev' did not behave correctly and has been removed"
+        )
     elif method == "sgc":
         # Smoothing filter (Simplifying Graph Convolutional Networks)
         if isinstance(k, int) and k > 0:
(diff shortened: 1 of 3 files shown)


stellargraph/core/utils.py
```diff
@@ -192,20 +159,6 @@ def preprocess_adj(adj, symmetric=True):
         print("Using GCN (local pooling) filters...")
         A = preprocess_adj(A)
 
-    elif method == "chebyshev":
```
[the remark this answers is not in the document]
kjun9: Good idea 👍

stellargraph/core/utils.py
```diff
@@ -218,5 +170,7 @@ def preprocess_adj(adj, symmetric=True):
                     type(k), k
                 )
             )
+    elif method == "chebyshev":
+        raise ValueError("method 'chebyhev' did not behave correctly and has been removed")
```
huonw: Err oops:

```suggestion
        raise ValueError("method 'chebyshev' did not behave correctly and has been removed")
```

Also, the overall diff will be slightly cleaner if you move this above `sgc`, so that it's where it was before.

stellargraph/mapper/full_batch_generators.py
```diff
@@ -123,7 +123,7 @@ def __init__(
             else:
                 raise ValueError("argument 'transform' must be a callable.")
 
-        elif self.method in ["gcn", "chebyshev", "sgc"]:
+        elif self.method in ["gcn", "sgc"]:
```
huonw: I assume that it's impossible for anyone to have actually used `chebyshev` here? As in, if they passed in `"chebyshev"` their code would've hit an error like #411, and so it's not worth doing a special case like `GCN_Aadj_feats_op`?

kjun9: It may have been possible to create the generator object, but it wouldn't have been useable in any way, since the shape of `generator.features` would've been wrong for anything downstream. Here we do already have an `else: raise` at the bottom too so it's probably fine?

PR #1029, opened 2020-03-05
Enhancement/804 wys non experimental

kieranricardo:
This PR adds unit tests for the Watch Your Step aglorithm.

Note there's some issues with the demo fixed in https://github.com/stellargraph/stellargraph/pull/1031


tests/mapper/test_adjacency_generators.py
```diff
@@ -0,0 +1,121 @@
...
+
+def test_flow(barbell):
+
+    generator = AdjacencyPowerGenerator(barbell, num_powers=5)
+    for i, x in enumerate(generator.flow(batch_size=1)):
+
+        if i > 5:
+            break
+
+    assert i > 5
```
kjun9: What's this test trying to catch? Maybe a way to rephrase it would be to just check the length of the sequence object matches expectation, if that was the intention.

tests/layer/test_watch_your_step.py
```diff
@@ -0,0 +1,90 @@
...
+    gen = generator.flow(batch_size=4)
+    wys = WatchYourStep(generator)
+
+    x_in, x_out = wys.build()
+
+    model = Model(inputs=x_in, outputs=x_out)
+    model.compile(optimizer="adam", loss=graph_log_likelihood)
+    model.fit(gen, epochs=1, steps_per_epoch=int(len(barbell.nodes()) // 4))
+
+    embs = get_embeddings(model)
```
kjun9: Do you think it's possible to add a small test for the `get_embeddings` method specifically with a model where we set all the weights to be 1s or something? Currently this test just checks the shape, but it would be good to test whatever we can test for actual values easily enough

tests/mapper/test_adjacency_generators.py
```diff
@@ -0,0 +1,121 @@
...
+    Aadj = normalize_adj(barbell.to_adjacency_matrix(), symmetric=False).todense()
+    actual_powers = [Aadj]
+    for _ in range(num_powers - 1):
+        actual_powers.append(actual_powers[-1].dot(Aadj))
+
+    generator = AdjacencyPowerGenerator(barbell, num_powers=num_powers)
+    for i, (x, y) in enumerate(generator.flow(batch_size=1)):
+
+        if i == len(barbell.nodes()):
+            break
```
kjun9: Why is this break necessary? Doesn't `enumerate` only generate from `0`th to `n-1`th items?

tests/mapper/test_adjacency_generators.py
```diff
@@ -0,0 +1,121 @@
...
+
+    num_powers = 5
+    generator = AdjacencyPowerGenerator(barbell, num_powers=num_powers)
+    for x, y in generator.flow(batch_size=batch_size):
+
+        assert x[0].shape == (batch_size,)
+        assert y.shape == (batch_size, 1, len(barbell.nodes()))
+        assert x[1].shape == (batch_size, num_powers, len(barbell.nodes()))
+
+        break
```
kjun9: Why do we want to break here after the first iteration? If we do, could we just get the first batch explicitly
```python
x, y = generator.flow(batch_size=batch_size)[0]
```
similarly for 
- [ ] line 101

tests/mapper/test_adjacency_generators.py
kieranricardo: this is trying to test whether the `Dataset` loops infinitely. The graph only has 5 nodes so if `i > 5` it means that the `Dataset` is repeated. 

I looked into it and there's a `tf.data.experimental.cardinality` function that I'll use instead of looping through 👍

tests/mapper/test_adjacency_generators.py
kieranricardo: The dataset loops infinitely, so this breaks the loop once all the powers have been tested

tests/mapper/test_adjacency_generators.py
kieranricardo: indexing isn't supported for `tf.data.Dataset` objects. But I can do:

```for x, y in generator.flow(batch_size=batch_size).take(1):
    ...
```

To avoid having that ugly break in there.

tests/layer/test_watch_your_step.py
kieranricardo: Currently there's not a an `embeddings_initializer` param for `WatchYourStep`. But I think one would be useful, I'll add that param in there and use it to test the embeddings

tests/mapper/test_adjacency_generators.py
kieranricardo: But I can get rid of the `break` by using `generator.flow(batch_size=1).take(len(barbell.nodes()))`

tests/layer/test_watch_your_step.py
```diff
@@ -0,0 +1,102 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from stellargraph.layer import AttentiveWalk, WatchYourStep, get_embeddings
+import numpy as np
+from ..mapper.test_graphwave_generator import barbell
```
kjun9: Hmm should this import be updated
```suggestion
from ..test_utils.graphs import barbell
```

PR #1031, opened 2020-03-06
WYS demo reproducability

kieranricardo:
This PR makes the WYS demo reproducible and increases the training epochs in this demo. Training for more epochs caused nan loss values, so this PR also fixes those numerical stability issues by combing the `log` and `sigmoid` operations.

See: #1024


stellargraph/layer/watch_your_step.py
```diff
@@ -202,7 +202,7 @@ def build(self):
             name="WATCH_YOUR_STEP_RIGHT_EMBEDDINGS",
         )(vectors_left)
 
-        sigmoids = tf.keras.activations.sigmoid(outer_product)
+        # sigmoids = tf.keras.activations.sigmoid(outer_product)
```
huonw: Maybe instead of leaving commented out code, we could delete it, or at least explain what's going on.

stellargraph/losses.py
```diff
@@ -26,15 +26,20 @@ def graph_log_likelihood(y_true, y_pred):
...
-    sigmoids = tf.gather(y_pred, [1], axis=1)
+    scores = tf.gather(y_pred, [1], axis=1)
 
     adj_mask = tf.cast((batch_adj == 0), "float32")
 
     loss = tf.math.reduce_sum(
         tf.abs(
-            -(expected_walks * tf.math.log(sigmoids + 1e-9))
-            - adj_mask * tf.math.log(1 - sigmoids + 1e-9)
+            -(expected_walks * _log_sigmoid(scores))
```
huonw: As far as I can tell, this could use https://www.tensorflow.org/api_docs/python/tf/math/log_sigmoid

```suggestion
            -(expected_walks * tf.math.log_sigmoid(scores))
```

Or maybe it has to be:

```suggestion
            (expected_walks * tf.math.log_sigmoid(scores))
```

stellargraph/layer/watch_your_step.py
kieranricardo: missed this, will delete

stellargraph/losses.py
kieranricardo: I've added this in 👍

stellargraph/losses.py
```diff
@@ -19,21 +19,30 @@
 from .core.experimental import experimental
 
 
-@experimental(reason="lack of unit tests", issues=[804])
-def graph_log_likelihood(y_true, y_pred):
+def graph_log_likelihood(batch_adj, wys_output):
```
huonw: This is a bit mind-bending to me: the "y_true" are actually just rows from the adjacency matrix, and the "y_pred" are arbitrary scores and the random walk slices, and aren't actually predicting the y_trues at all? Maybe this could be explained a bit more in the doc string?

demos/embeddings/watch-your-step-cora-demo.ipynb
```diff
@@ -11,9 +11,11 @@
     "from stellargraph.layer.watch_your_step import WatchYourStep, get_embeddings\n",
```
huonw: <p>We don't do this in any other notebook. Is it important to do it here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1031/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

huonw: <p>Not related to this PR specifically, but why does this need <code>steps_per_epoch</code> ? Isn't that basically implicitly controlled by the length of <code>train_gen</code> ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1031/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

stellargraph/layer/watch_your_step.py
```diff
@@ -223,12 +223,11 @@ def build(self):
...
         outer_product = self._right_embedding(vectors_left)
 
-        sigmoids = tf.keras.activations.sigmoid(outer_product)
         expected_walk = self.num_walks * self._attentive_walk(input_powers)
 
         # layer to add batch dimension of 1 to output
         expander = Lambda(lambda x: K.expand_dims(x, axis=1))
 
-        output = Concatenate(axis=1)([expander(expected_walk), expander(sigmoids)])
+        output = Concatenate(axis=1)([expander(expected_walk), expander(outer_product)])
```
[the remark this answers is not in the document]
kieranricardo: nah unfortunately keras does weird stuff with multiple outputs - each output has to go to a different loss function that's why im doing the strange concat here

stellargraph/losses.py
```diff
@@ -19,21 +19,30 @@
...
+            scores (with shape (batch_rows, 1, num_nodes) and attentive expected random walk
+            (with shape (batch_rows, 1, num_nodes) concatenated.
+    Returns:
+        the graph log likelihood loss for the batch
     """
-    batch_adj = tf.gather(y_true, [0], axis=1)
 
-    expected_walks = tf.gather(y_pred, [0], axis=1)
-    sigmoids = tf.gather(y_pred, [1], axis=1)
+    batch_adj = tf.squeeze(batch_adj, axis=1)
```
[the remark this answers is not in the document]
kieranricardo: Before I was gathering on `axis` of the `(batch_size, 1, num_nodes)` tensor to get a `(batch_size, num_nodes)` tensor.

so in this case its equivalent to the `gather` operation,  just slightly more optimal because as I understand it `squeeze` doesnt change the underlying data structure.

stellargraph/losses.py
```diff
@@ -19,21 +19,30 @@
...
 
 
-@experimental(reason="lack of unit tests", issues=[804])
-def graph_log_likelihood(y_true, y_pred):
+def graph_log_likelihood(batch_adj, wys_output):
     """
+    Computes the graph log likelihood loss function as in https://arxiv.org/abs/1710.09599
+
+    Args:
+        batch_adj: tensor with shape (batch_rows, 1, num_nodes) containing rows of the adjacency matrix
```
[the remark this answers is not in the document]
kieranricardo: I used `batch_rows` to emphasise that these are whole rows of the adjacency matrix being passed in - happy to change if its unclear

demos/embeddings/watch-your-step-cora-demo.ipynb
kieranricardo: its a repeated tensorflow dataset so it needs `steps_per_epoch` specified, otherwise you get:

````ValueError: When passing an infinitely repeating dataset, you must specify the `steps_per_epoch` argument.```

Comment (kieranricardo) [reply demos/embeddings/watch-your-step-cora-demo.ipynb]: I think that's left over code from development when I messed up the imports, I'll change it to `from stellargraph.layer import WatchYourStep, get_embeddings`

File: stellargraph/losses.py [hunk 8]
```diff
@@ -19,21 +19,30 @@
 from .core.experimental import experimental
 
 
-@experimental(reason="lack of unit tests", issues=[804])
```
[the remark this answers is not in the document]
kieranricardo: ah good point, I'll chuck in a unit test in another PR and put this back.

stellargraph/losses.py
kieranricardo: so far I've got: 

```Computes the graph log likelihood loss function as in https://arxiv.org/abs/1710.09599. This is different
    to most keras loss functions in that it doesn't directly compare predicted values to expected values. It uses
    `wys_output` which contains the dot products of embeddings and expected random walks, and part of the adjacency
    matrix `batch_adj` to calculate how well the node embeddings capture the graph structure in some sense.```

File: demos/embeddings/watch-your-step-cora-demo.ipynb (hunk 4 above)
Comment (huonw) [reply demos/embeddings/watch-your-step-cora-demo.ipynb]: I think you're looking at the wrong place. ReviewNB always attaches the comment to the first line of the diff, not the actual cell that it was commented on. You have to look on ReviewNB itself to see the right place. In this case, it's attached to the `tf.random.set_seed` cell.

File: stellargraph/losses.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (huonw) [reply stellargraph/losses.py]: Either one, just as long as there's a reason.

File: demos/embeddings/watch-your-step-cora-demo.ipynb (hunk 4 above)
Comment (kieranricardo) [reply demos/embeddings/watch-your-step-cora-demo.ipynb]: ah gotcha, just thought it would be nice for the notebook to reproducible but its not important here.

Comment (huonw) [reply demos/embeddings/watch-your-step-cora-demo.ipynb]: Yeah, we seemingly don't care about the notebooks being exactly reproducible, just that rerunning them gives similar results.

File: stellargraph/losses.py [hunk 9]
```diff
@@ -19,21 +19,30 @@
...
-def graph_log_likelihood(y_true, y_pred):
+def graph_log_likelihood(batch_adj, wys_output):
     """
+    Computes the graph log likelihood loss function as in https://arxiv.org/abs/1710.09599
+
+    Args:
+        batch_adj: tensor with shape (batch_rows, 1, num_nodes) containing rows of the adjacency matrix
+        wys_output: tensor with shape (batch_rows, 2, num_nodes) containing the embedding outer product
+            scores (with shape (batch_rows, 1, num_nodes) and attentive expected random walk
```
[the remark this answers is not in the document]
kieranricardo: I went for the second option :)

stellargraph/losses.py
```diff
@@ -20,21 +20,32 @@
 
 
 @experimental(reason="lack of unit tests", issues=[804])
-def graph_log_likelihood(y_true, y_pred):
+def graph_log_likelihood(batch_adj, wys_output):
     """
+    Computes the graph log likelihood loss function as in https://arxiv.org/abs/1710.09599. This is different
```
huonw: Per https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings , the python convention that tooling will obey is to have a one line summary as its own paragraph:

```python
    """
    Computes the graph log likelihood loss function as in https://arxiv.org/abs/1710.09599.

    This is different ...

    ...
    """
```

Issue #934, opened 2020-02-25
Create a demo notebook to replace script for random walk link prediction


timpitman:
### Description

The demo script for random walk link prediction was restored for #895 as it does significantly more than the Jupyter demo notebook.   As part of our plan to replace scripts with notebooks, we should remove `demos/link-prediction/random-walks/main.py` and replace it with a suitable demo (eg heterogenous with `BlogCatalog3` dataset?)

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

kjun9:
Hmm actually this is currently a bit difficult to do due to https://github.com/stellargraph/stellargraph/issues/907 - that site's been down for a while now, probably need to just rewrite the demo with a different dataset?

timpitman:
I think we should probably just upload that dataset to figshare?

kjun9:
Seems like my laptop can't handle the current metapath2vec demo with blog catalog - I've reduced some of the random walk parameters to:
```
num_walks = 1  # previously 10
walk_length = 20  # previously 80
```
but the graph is large enough that the random walks have been running for a few hours without any visible progress, will try reduce the parameters further..

kjun9:
I think it might be too much of a rush to try get this in with https://github.com/stellargraph/stellargraph/pull/1399 and https://github.com/stellargraph/stellargraph/pull/1392 so I'm thinking we could just remove the `utils` directory for 1.0: https://github.com/stellargraph/stellargraph/pull/1462

Seems unlikely that the script is being run or is able to be run, and it would clean up our demos for the release. Unsupervised Metapath2Vec is being demo'd via `demos/embeddings/stellargraph-metapath2vec.ipynb` which does hint at the bottom of the notebook:
```
## Downstream Tasks

The node embeddings calculated using Metapath2Vec can be used as feature vectors in a downstream task such as node attribute inference (e.g., inferring the gender or age attribute of 'user' nodes), community detection (e.g., clustering of 'user' nodes based on the similarity of their embedding vectors), and link prediction (e.g., prediction of friendship relation between 'user' nodes).
```
So that's probably more helpful than the existing scripts

Issue #810, opened 2020-02-06
Update demos/link-prediction/random-walks/main.py to use new datasets module


timpitman:
### Description

This was split from the original task #610, since some larger changes are needed for this script (eg it requires manual steps of pickling and saving one of the datasets, which should be incorporated into the script)

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

timpitman:
duplicate of #758

PR #1032, opened 2020-03-06
Support StellarGraph input and output in EdgeSplitter

huonw:
This adjust the `stellargraph.data.EdgeSplitter` class so that it can consume and emit a `StellarGraph` object: if it is constructed like `EdgeSplitter(some_stellargraph)`, then `train_test_split` will return a `StellarGraph` object. For backwards-compatibility, `EdgeSplitter(some_networkx)` still works, and returns a NetworkX object; that is, existing code will not be affected.

For now, this compatibility is useful even just for our own code as there's still quite a few of our demos that would require a lot more code to switch to use this new form, and so I have not switched them here:

- `demos/link-prediction/gcn/cora-gcn-links-example.ipynb`, `demos/link-prediction/graphsage/cora-links-example.ipynb`: unlike other demos that use the Cora dataset, these use the subject as a feature (rather than as a target), and so the graph returned by `Cora().load()` can't be used directly (#1039)
- `demos/link-prediction/random-walks/main.py`, `demos/link-prediction/random-walks-cora-lp-demo.ipynb`: these seem to have their own implementation of node2vec and metapath2vec built on top of NetworkX and so aren't using StellarGraph objects anyway (or, it seems, anything from the `stellargraph` library other than `EdgeSplitter`) (related to #810, #896 and #934)

However, some notebooks can be updated without too much code, in particular `demos/calibration/calibration-pubmed-link-prediction.ipynb` and `demos/ensembles/ensemble-link-prediction-example.ipynb` can move over to using `Cora().load()` (per #812 and #717) by passing the returned `StellarGraph` object directly into `EdgeSplitter`, with not many other code changes required.

See: #174


huonw:
That sort of work is definitely appropriate for a separate ticket. In that particular case, it's #933, which is fixed in #1008.

PR #1046, opened 2020-03-09
Test neo4j notebooks on CI

huonw:
This uses docker-compose to add a Neo4j instance running on CI. The neo4j container is created by building a docker image from the official neo4j images that includes the ["APOC"](https://neo4j.com/developer/neo4j-apoc/) plugin that we need. Building and running this image is easy and fast, only a few seconds if the neo4j image is already available on the agent; as such, for simplicity, this unconditionally runs the image neo4j for every step that uses one of the 3.x runners, even though most do not use neo4j.

The notebooks are tested as a single step, that runs them in series, because the directed GraphSAGE notebooks rely on having Cora already available in the database, and thus the `load-cora-into-neo4j.ipynb` notebook needs to run first.

This makes some minor edits to the notebooks so that they can work on CI, with the major one restructuring how the connection to neo4j works, by using Neo4j/py2neo's defaults for everything except the host, which is read from the `STELLARGRAPH_NEO4J_HOST` environment variable if it exists. This is used on CI to tell to the code to connect to the neo4j container which is available at `neo4j`, not `localhost` since it's a separate container. An alternative would be to specify `network_mode: "service:runner-3_6"` in `neo4j`, so that it's available under the runner container's `localhost` and thus need no custom code in the notebook at all. Thoughts?

This tests on Neo4j 3.5, which is not the latest release, but (a) it's probably good to support a bit of a range, and (b) the notebooks fail on Neo4j 4.0 (https://github.com/stellargraph/stellargraph/issues/1055). Once that issue is fixed, I believe we should be able to extend this to test against multiple versions of Neo4j using [build args](https://docs.docker.com/engine/reference/builder/#arg).

See: #849


demos/connector/neo4j/load-cora-into-neo4j.ipynb
```diff
@@ -306,8 +306,10 @@
    "source": [
```
timpitman: <p>Will this now fail if that environment variable isn't set? It should really default to localhost; could we also use papermill to inject our CI settings?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1046/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

demos/connector/neo4j/undirected-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -47,8 +47,10 @@
    "source": [
```
timpitman: <p>not sure where this "['...']" output comes from?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1046/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='37'/>

demos/connector/neo4j/load-cora-into-neo4j.ipynb
huonw: <p>No,  I've been running them locally as is (against a <code>localhost</code>  Neo4j instance). The <code>get</code>  means it returns <code>None</code>,  and this API accepts <code>host=None</code>  as referring to the default.</p>

demos/connector/neo4j/undirected-graphsage-on-cora-neo4j-example.ipynb
huonw: <p>https://github.com/stellargraph/stellargraph/issues/826</p>

.buildkite/steps/test-neo4j-notebooks.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -xeo pipefail
+
+echo "--- :python: installing stellargraph"
+# install stellargraph itself and the dependencies necessary for neo4j, which (hopefully) won't
+# install any other ones. (The neo4j deps aren't installed into the docker image, so the main tests
```
kjun9: I'm not sure if I understand this comment - which "other ones" are you referring to that we shouldn't be installing?

huonw: Other dependencies, because they should be in the docker image already. I've rephrased the comment.

huonw:
> The readme for the demos could probably be updated too, and direct users to use pip install stellargraph[neo4j] but that could be done as a separate task to redo the readme in general

That's a good idea, but I think we shouldn't do it quite yet: `pip install stellargraph[neo4j]` won't currently work, since the 0.10.0 release didn't have the `neo4j` extra.

.buildkite/steps/test-neo4j-notebooks.sh
kjun9: Thanks, makes sense 👍

docker/stellargraph-neo4j/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM neo4j:3.5
```
TheDen: Do we want to build this docker image in `stellar-devops` and push our own neon4j image and, or keep it here? You mentioned building the image is fast, but it'd offload the step

docker/stellargraph-neo4j/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+
+SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
+
+ENV APOC_VERSION="3.5.0.9" APOC_SHA="7f630b1ba3dccd0b59fa609a3f488820059044b3075d9604a79741fea379544c"
+RUN echo "--- downloading apoc jar" && \
+    wget --no-verbose "https://github.com/neo4j-contrib/neo4j-apoc-procedures/releases/download/${APOC_VERSION}/apoc-${APOC_VERSION}-all.jar" --directory-prefix plugins/ && \
+    echo "${APOC_SHA}  ${PWD}/plugins/apoc-${APOC_VERSION}-all.jar" | sha256sum -c
```
TheDen: Very minor but we use `sha512sum` in other images (nothing critical so you can ignore this)

huonw: Ah, I'd got a mistaken impression from stellar-devops (`sha256sum` was the first one I grepped for and I got some results, so I just went with it, but it does seem like `sha512sum` is the more common one).

docker/stellargraph-neo4j/Dockerfile
huonw: I think it would be better to keep it here: it's not currently published anywhere, just built transiently, and, even if we did push it (which we could do with prebuilding and caching here), it's seems slightly nicer for the original Dockerfile be part of the open source library, rather than just published publicly from a private repo. What do you think?

(We can always change our approach in future, too.)

TheDen: Yeah sounds like a good plan 👍

Issue #1045, opened 2020-03-09
Set tensorflow minimum version to 2.1


huonw:
### Description

Some things work better on tensorflow 2.1:

- the `_generator` method we use a lot of are deprecated in 2.1 and so it'd be best not to use them, but doing the switch in 2.0 may not work properly, so requiring 2.1 means there's no balancing act required (#803 and #851)
- https://github.com/stellargraph/stellargraph/pull/1001#discussion_r389416173

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #851, opened 2020-02-12
Rewrite Keras Model.*_generator methods to Model.*

huonw:
See: #803

PR #1048, opened 2020-03-10
Set tensorflow minimum version to 2.1

kjun9:
This should help #851 and #546 

See: #1045

### DIFF
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -40,12 +40,10 @@ services:
       - DOCKER_CLIENT_TIMEOUT=180
       - COMPOSE_HTTP_TIMEOUT=180
 
+
   runner-3_7:
     <<: *runner-config
     build:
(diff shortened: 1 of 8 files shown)


.buildkite/docker-compose.yml
```diff
@@ -37,6 +37,8 @@ services:
       - BUILDKITE_COMMIT
       - BUILDKITE_TAG
       - CODECOV_TOKEN
+      - DOCKER_CLIENT_TIMEOUT=360
```
huonw: If we were to land this, it would be good to have a comment here and in the other one:

```suggestion
      # the images are large, so pushing with the defaults can timeout
      - DOCKER_CLIENT_TIMEOUT=360
```

kjun9:
@huonw Thanks, yeah that makes sense 👍 Okay I'll remove the docker timeout flags from here

Issue #1006, opened 2020-03-03
Investigate `WARNING:tensorflow:multiprocessing can interact badly with TensorFlow`


kieranricardo:
`hinsage/movielens-recommender.ipynb` has the following warning:

```
WARNING:tensorflow:multiprocessing can interact badly with TensorFlow, causing 
  nondeterministic deadlocks. For high performance data pipelines tf.data is recommended
```

It seems like this warning will pop up whenever someones uses `use_multiprocessing=True` with any of our keras generators. Replacing our generators with `tf.data` will fix this warning however this will take a significant amount of work. This issue is to explore whether its worth replacing our generators with `tf.data` objects.

timpitman:
Potentially we can resolve by simply not using multiprocessing

timpitman:
issue kept to investigate the underlying issue.

kechan:
I received this error when my generator is from tf.keras.utils.Sequence. And for at least once, my training did freeze...

timpitman:
Hi @kechan, thanks for reporting your issue, it would be great to get some further details - are you able refer us to any sample code on Github that shows your exact problem?

We're working on #1206 which should resolve issues with multiprocessing too.

Rekha5649:
I have used this  'use_multiprocessing=True'  with my Datagenerator so I can have data ready before training but getting the same error.

JasonCZH4:
I use keras.fit_generator and multiprocessing = True.I have the same error too

PR #1051, opened 2020-03-10
Remove the use of multiprocessing

kjun9:
This sets `use_multiprocessing=False` in demos in order to avoid tensorflow warnings about using multiprocessing with generators.

See #1006 #1073


demos/link-prediction/hinsage/movielens-recommender.ipynb
```diff
@@ -469,17 +469,18 @@
...
+      "\tloss: 11.8834\n",
+      "\troot_mean_square_error: 3.4467\n",
+      "\tmean_absolute_error: 3.2562\n"
      ]
     }
    ],
    "source": [
     "test_metrics = model.evaluate_generator(\n",
-    "    test_gen, verbose=1, use_multiprocessing=True, workers=num_workers\n",
+    "    test_gen, verbose=1, use_multiprocessing=False, workers=num_workers\n",
```
kieranricardo: We can get rid of the `workers` param here 'cause we're not using multiprocessing.

```suggestion
    "    test_gen, verbose=1, use_multiprocessing=False\n",
```

kjun9: I think the `workers` param should get used to do multithreading instead - see https://github.com/stellargraph/stellargraph/pull/925#discussion_r383086045

kjun9:
> I think this doesn't quite address #1006 which is more a performance investigation/seeing whether we should ignore this warning or look at replacing our Sequence objects with tf.data.Datasets. Could you make another issue and connect it instead? I don't want this to get lost!

Yep makes sense 👍 I'll make another issue, but I think we decided to resolve the warnings like this for 1.0, so we might want to remove #1006 from 1.0 then (cc @timpitman )

demos/link-prediction/hinsage/movielens-recommender.ipynb
kieranricardo: oh good point, missed that in the docs 👍

Issue #1043, opened 2020-03-09
Reproducibility for Graph Classification


PantelisElinas:
### Description

Implement reproducibility for the `GraphClassification` and related classes as well as corresponding unit tests.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed

PantelisElinas:
From https://github.com/stellargraph/stellargraph/pull/1053#discussion_r392765046
> This and the random.shuffle could be a single more declarative line using indexes = np.random.permutation(len(self.graphs)) which expresses the intent more directly, I think. (It's also more efficient since it gets to stay within the NumPy/native-code world more, but that doesn't matter for this particular code.)

Keep this in mind when working on this ticket.

huonw:
@kjun9 I think this may be mostly fixed by the seeding and shuffling in https://github.com/stellargraph/stellargraph/pull/1175 ?

Issue #1042, opened 2020-03-09
Unit tests for Graph Classification


PantelisElinas:
### Description

We need unit tests for the `GraphClassification` and related classes so it can move from experimental to production.

### Done Checklist

- [ ] Tests written and coverage checked
- [ ] Code review performed

Issue #928, opened 2020-02-24
Graph generator


PantelisElinas:
### Description

For GNN-based graph classification, we need a StellarGraph graph generator to feed graphs with node attributes and graph-level labels to Keras for model training as well as evaluation and prediction.

This ticket is for implementing such a generator.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] CHANGELOG.md updated

Issue #1038, opened 2020-03-07
GAT input shape error while testing


ocesp98:
I used GAT to build a model on the train and validation set. So first, I built the network based on the train and validation set (=9000 rows). The node_features added to the first Stellargraph hence contain also 9000 rows and 5 attributes. 

The problem occurs when I'm predicting on the test set. Given that I want to use the inductive property of GAT, I held the test set separate from the training and validation set. To test, I built a second network where I added the test nodes to the first network (consisting of the train and validation set). This network hence contains 10000 nodes (9000 already existing and 1000 from the test set). The node_features added to this second stellargraph object have a dimension of 10000 rows and 5 attributes. However, when I want to predict on the test generator, I get following error:

> ValueError: Error when checking input: expected input_1 to have shape (9000, 5) but got array with shape (10000, 5)

I used following code:

`stelgr = sg.StellarGraph(H, node_features=node_features) #this is for training`
`generator = FullBatchNodeGenerator(stelgr, method="gat")`
`train_gen = generator.flow(train_data.index, train_targets)`
`gat = GAT(layer_sizes=[32, 32, 1],
    activations=["relu", "relu", "sigmoid"],
    attn_heads=8,
    generator=generator,
    in_dropout=0.4,
    attn_dropout=0.3)`

Then I trained the model. To test, I rebuilt the network:

`stelt = sg.StellarGraph(Ht, node_features = node_features2)`
`generator = FullBatchNodeGenerator(stelt, method="gat")`
`test_gen = generator.flow(test_data.index, test_targets)`

So Ht is the full network with the test nodes added. Node_features2 contains all features, including the features for the test set and has shape (10000,5). Node_features only contains the features for the train and validation set and has shape (9000,5).

The full error is as follows:

>  ValueError                                Traceback (most recent call last)
> <ipython-input-84-c4c1e4285085> in <module>
>       1 #t_nodes = test_data.index
>       2 #t_mapper = generator.flow(t_nodes)
> ----> 3 t_predictions = model.predict_generator(test_gen)
> 
> ~\Anaconda3\envs\StellarGraph2\lib\site-packages\tensorflow\python\keras\engine\training.py in predict_generator(self, generator, steps, callbacks, max_queue_size, workers, use_multiprocessing, verbose)
>    1553         use_multiprocessing=use_multiprocessing,
>    1554         verbose=verbose,
> -> 1555         callbacks=callbacks)
>    1556 
[...]

PR #1053, opened 2020-03-10
Supervised graph classification with GCN

PantelisElinas:
Hi,

this is an implementation of a supervised graph classification model based on GCN layers. It is based on the model presented in [1], with the main exception being that the model in [1] uses attention whereas the one here does not. For your reference, the demo notebook includes a picture of the architecture. This work is not an attempt at replicating [1] but a similar model that we can use as baseline moving forward.

Everything is experimental as unit tests are missing and the demo notebook is a basic demonstration of how to perform supervised graph classification. 

**Note:** Minimum requirements for this are tensorflow > 2.0.1. #1045 will tackle this problem but for now please be aware that anything less and this code will not run.

Other related tickets: #1042 and #1043.

Also, this pull request covers #928.

P.

**References:**

1. Fake News Detection on Social Media using Geometric Deep Learning, F. Monti, F. Frasca, D. Eynard, and D. Mannion, ICLR 2019, https://arxiv.org/pdf/1902.06673.pdf


stellargraph/mapper/sequences.py
```diff
@@ -527,3 +530,128 @@ def __len__(self):
...
+                )
+
+            self.targets = np.asanyarray(targets)
+
+        self.on_epoch_end()
+
+    def __len__(self):
+        return int(np.ceil(len(self.graphs) / self.batch_size))
+
+    def __normalize_adj(self, adj):
```
huonw: This looks quite similar to the `normalize_adj` function in `core.utils`:

https://github.com/stellargraph/stellargraph/blob/cbbe37b67f11d28a265165d19644ad4789a2a8c2/stellargraph/core/utils.py#L35-L53

Could you explain the reason for reimplementing it?

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +530,128 @@ def __len__(self):
...
+        degree_matrix.setdiag(degree_matrix_diag)
+        degree_matrix = degree_matrix.tocsr()
+        adj = degree_matrix @ adj @ degree_matrix
+        return adj
+
+    def __getitem__(self, index):
+        graphs = self.graphs[
+            index * self.batch_size : (index * self.batch_size) + self.batch_size
+        ]
+        adj_graphs = [graph.to_adjacency_matrix() for graph in graphs]
```
huonw: This has to recompute the adjacency matrices for each matrix in every epoch. What do you think about computing them once, earlier?

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +530,128 @@ def __len__(self):
...
+        # feature matrices of the largest graph.
+        max_nodes = max([graph.number_of_nodes() for graph in graphs])
+
+        # The operations to normalize the adjacency matrix are a bit slow.
+        # FIXME: Either optimize this or implement as a layer.
+        if self.normalize_adj:
+            adj_graphs = [self.__normalize_adj(adj).toarray() for adj in adj_graphs]
+
+        graph_targets = None
+        #
```
huonw: Missing comment?

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +530,128 @@ def __len__(self):
...
+
+        # The number of nodes for the largest graph in the batch. We are going to pad with 0 rows and columns
+        # the adjacency and node feature matrices (only the rows in this case) to equal in size the adjacency and
+        # feature matrices of the largest graph.
+        max_nodes = max([graph.number_of_nodes() for graph in graphs])
+
+        # The operations to normalize the adjacency matrix are a bit slow.
+        # FIXME: Either optimize this or implement as a layer.
+        if self.normalize_adj:
+            adj_graphs = [self.__normalize_adj(adj).toarray() for adj in adj_graphs]
```
huonw: I think we can do this densification better: at the the moment we start with sparse (normalised) adjacency matrices, convert these to dense, and then fill out the rest of the each one with zeros to create new dense matrices. However, sparse matrices are really good at representing zeros cheaply, and so adding extra zero rows and columns can be done very efficiently by just playing with their shapes.

```python
def _resize_csr_square(m, new_dim):
    return sparse.csr_matrix((m.data, m.indices, m.indptr), shape=(new_dim, new_dim))

    ...

            adj_graphs = [self.__normalize_adj(adj) for adj in adj_graphs]

    ...

        adj_graph = np.stack([_resize_csr_square(adj, max_nodes).toarray() for adj in adj_graphs])
```

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +530,128 @@ def __len__(self):
...
+            ]
+
+        # pad adjacency and feature matrices to equal the size of those from the largest graph
+        features = [
+            np.pad(
+                graph.node_features(graph.nodes()),
+                pad_width=(0, max_nodes - graph.number_of_nodes()),
+                mode="constant",
+                constant_values=0.0,
+            )[:, : self.node_features_size]
```
huonw: I think this pads the second axis and then immediately strips it off?

Based on the `np.pad` docs, it looks like passing `(0, 0)` into `pad_width` should avoid doing any padding on that axis: `pad_width=((0, max_nodes - graph.number_of_nodes()), (0, 0)),` plus

```suggestion
            )
```

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +530,128 @@ def __len__(self):
...
+            graph_targets = self.targets[
+                index * self.batch_size : (index * self.batch_size) + self.batch_size
+            ]
+
+        # pad adjacency and feature matrices to equal the size of those from the largest graph
+        features = [
+            np.pad(
+                graph.node_features(graph.nodes()),
+                pad_width=(0, max_nodes - graph.number_of_nodes()),
+                mode="constant",
```
huonw: Both this and and `constant_values=0.0` are the default values, so this could potentially be reduced to `np.pad(graph.node_features(...), pad_width=...)`.

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +530,128 @@ def __len__(self):
...
+        #      batch size x number of classes
+        return [features, masks, adj_graphs], graph_targets
+
+    def on_epoch_end(self):
+        """
+         Shuffle all graphs at the end of each epoch
+        """
+        indexes = list(range(len(self.graphs)))
+        random.shuffle(indexes)
+        self.graphs = [self.graphs[i] for i in indexes]
```
huonw: If you wanted to, `graphs` could be stored as a numpy array (e.g. `self.graphs = np.asanyarray(graphs)` in `__init__`) which will have `dtype=object`. This then allows this code to be:

```python
indexes = np.random.permutation(len(self.graphs))
self.graphs = self.graphs[indexes]
self.targets = self.targets[indexes]
```

stellargraph/mapper/sequences.py
```diff
@@ -527,3 +530,128 @@ def __len__(self):
 
     def __getitem__(self, index):
         return self.inputs, self.targets
+
+
+@experimental(reason="Missing unit tests.")
```
huonw: A user doesn't construct these directly, so to avoid getting extra warnings (i.e. a warning from both `GraphGenerator(...)` and `gen.flow(...)`), it may not be necessary to annotate this one:

```suggestion
```

stellargraph/mapper/graph_generator.py
```diff
@@ -0,0 +1,102 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from ..core.graph import StellarGraph
+from ..core.utils import is_real_iterable
+from .sequences import GraphSequence
+
+
+class GraphGenerator:
```
huonw:
```suggestion
@experimental(reason="Missing unit tests", issues=[1042])
class GraphGenerator:
```

stellargraph/layer/graph_classification.py
```diff
@@ -0,0 +1,236 @@
...
+
+        # Add optional bias & apply activation
+        if self.bias is not None:
+            output += self.bias
+        output = self.activation(output)
+
+        return output
+
+
+@experimental(reason="Missing unit tests")
```
huonw:
```suggestion
@experimental(reason="Missing unit tests", issues=[1042])
```

stellargraph/layer/graph_classification.py
```diff
@@ -0,0 +1,236 @@
...
+class GraphClassification:
```
huonw: This seems to be a very general name for what seems to be a specific solution to graph classification (although clearly one that has some flexibility), could you talk a bit about why this is the right name?

stellargraph/layer/graph_classification.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+
+    def __init__(
+        self, layer_sizes, activations, generator, bias=True, dropout=0.0, **kwargs
+    ):
+        if not isinstance(generator, GraphGenerator):
+            raise TypeError("Generator should be a instance of GraphGenerator")
+
+        if len(layer_sizes) != len(activations):
+            raise AssertionError(
```
huonw:
```suggestion
            raise ValueError(
```

stellargraph/layer/graph_classification.py
```diff
@@ -0,0 +1,236 @@
...
+        activations (list of str): list of activations applied to each GCN layer's output.
+        generator (GraphGenerator): an instance of :class:`GraphGenerator` class constructed on the graphs used for
+            training.
+        bias (bool, optional): toggles an optional bias in graph convolutional layers.
+        dropout (float, optional): dropout rate applied to input features of each GCN layer.
+        kernel_regularizer (str): normalization applied to the GCN layer kernels.
+    """
+
+    def __init__(
+        self, layer_sizes, activations, generator, bias=True, dropout=0.0, **kwargs
```
huonw: Instead of adding new examples for #801, maybe this could list out the arguments that are computed in `_get_regularisers_from_keywords`?

stellargraph/layer/graph_classification.py
```diff
@@ -0,0 +1,236 @@
...
+    GlobalAveragePooling1D,
+)
+
+from ..mapper import GraphGenerator
+from .cluster_gcn import ClusterGraphConvolution
+from ..core.experimental import experimental
+
+
+@experimental(reason="Missing unit tests")
+class GraphClassificationConvolution(ClusterGraphConvolution):
```
huonw: I'm a bit confused by this class: its docs talk about being based on a GCN layer, but... its actually based on a Cluster-GCN one. In addition (and more importantly), it has a custom `call` function. Could you describe a bit about why it needs to differ to the base layer?

CHANGELOG.md
```diff
@@ -17,7 +17,7 @@
 Some new algorithms and features are still under active development, and are available as an experimental preview. However, they may not be easy to use: their documentation or testing may be incomplete, and they may change dramatically from release to release. The experimental status is noted in the documentation and at runtime via prominent warnings.
 
 - DistMult: computes embeddings for nodes and edge types in knowledge graphs, and use these to perform link prediction [\#755](https://github.com/stellargraph/stellargraph/issues/755). The implementation hasn't been validated to match the paper.
-
+- GraphClassification: supervised graph classification model based on Graph Convolutional layers (GCN) [\#929](https://github.com/stellargraph/stellargraph/issues/929).
```
huonw: The raw file is slightly nicer to read/edit if the blank line after the list is retained:

```suggestion
- GraphClassification: supervised graph classification model based on Graph Convolutional layers (GCN) [\#929](https://github.com/stellargraph/stellargraph/issues/929).

```

and similarly for the blank line after the table in the `README.md`.

stellargraph/layer/graph_classification.py
huonw: Relatedly to the `kwargs`, but somewhat different (and hence a separate thread), this class is currently hard-coded to support only a single version of GCN, and thus has to pass through all the right arguments. I wonder if there's a more general form, similar to how #1001 generalised `GCNInfoMax` which was hard-coded to GCN to `DeepGraphInfoMax` that supports many kinds of layers (and can be generalised further). For instance, one could use GraphSAGE or some other inductive algorithms. What do you think?

Hypothetically, one could imagine having this be used something like:

```python
cluster_gcn = ClusterGCN(...)
graph_classification = GraphClassification(cluster_gcn)
```

Getting the generator support might be a little awkward, but I feel that we can get something working here. I could imagine something like `GraphGenerator(graphs, ClusterNodeGenerator, clusters=1, lam=3)` where basically the `GraphGenerator` manages a base generator for each graph or something.

We could still offer convenient ways to get a useful default approach.

Do you have any thoughts on this?

huonw:
Quite a few of the small things I noted should only take a minute or two to fix, but if they're more complicated or more awkward than it looks on the surface, it's ok to leave them for later and they can be cleaned-up separately as part of the final de-`@experimental`ising.

stellargraph/mapper/graph_generator.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(self, graphs, name=None):
+
+        for graph in graphs:
+            if not isinstance(graph, StellarGraph):
+                raise TypeError("All graphs must be StellarGraph objects.")
+
+        self.graphs = graphs
+        self.name = name
+        # we assume that all graphs have node features of the same dimensionality
```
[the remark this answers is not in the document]
PantelisElinas: I hadn't thought about someone trying to pass a heterogeneous graph. I'll update as above.

stellargraph/mapper/sequences.py
PantelisElinas: I didn't know we had this in utils. I'll check if it is the same op and replace it.

stellargraph/mapper/sequences.py
PantelisElinas: The usual time vs space tradeoff. I think in this case, we should expect each graph to be small in the order of hundreds of nodes so pre-computing should be fine. Memory usage might start to become an issue if the number of graphs is very large which could happen if each is a subgraph of a very large graph, i.e., financial transaction network. 

I'll update it to pre-compute for now. We can come up with a heuristic later on to determine if pre-computing might use too much memory and not do it. I think it might be best that we go for speed now since all benchmark datasets tend to me small.

Ultimately, we want this normalisation done in a layer but the normalisation layers we have at the moment are not suitable because of how they handle the batch dimension.

stellargraph/layer/graph_classification.py
PantelisElinas: The `ClusterGraphConvolution` layer is a GCN layer. It exists because of limitations/restrictions on our `GraphConvolution`, i.e., GCN, and `GraphAttention`, i.e, GAT, layers. The GAT and GCN layers require that graph size is fixed (see #1038) and also expect that batch size = 1 (and basically treat it as a nuisance that must be removed and added it to keep Keras happy). Further, one of the inputs indicates node indices, `out_indices`, that the last layer uses to `gather` the outputs just for the nodes of interest, i.e., those nodes that have targets (for training and evaluation) or the nodes we want predictions for; before `gather` was introduced our GCN and GAT models made predictions for all nodes in the graph regardless of the values passed to `generator.flow()` making the API inconsistent with methods such as GraphSAGE. 

When I implemented ClusterGCN, I wanted to use our existing GCN layer but the above issues that we did not have a solution for at the time forced me to create the new `ClusterGraphConvolution` layer which is essentially a more flexible GCN layer that does not support sparse inputs but can handle graphs of different sizes.

For graph classification, I wanted to re-use our GCN and GAT layers but the above problems once again made it difficult. The closest I had to a re-usable layer was `ClusterGraphConvolution` but it wasn't exactly right. The problem is that the input for graph classification does not include node indices, i.e., `out_indices`, since we need the output for all nodes for mean pooling. In graph classification `out_indices` has been replaced by `mask` to handle different size graphs. Also, even though Cluster-GCN is a mini-batch method, its definition of a batch is a bit different than in the graph case. Basically, the input for Cluster-GCN is still a single graph hence batch size = 1 and so the layer treats it like a nuisance to remove and add before and after calculation. The `GraphClassificationConvolution` layer receives data with batch size >= 1 and handles them correctly with a call to `keras.batch_dot()`.  The change to make the latter work was only necessary in the `call` method so the least I had to do to make it work is sub-class `ClusterGraphConvolution` and overwrite the `call` method. 

If/when we manage to fix our basic GCN and GAT layers, we can replace `GraphClassificationConvolution` with one of them and changes should be minimal. If #1038 is resolved, then we are going to be half way there; we would still have to deal with the batch dimension correctly; I think we might be able to fix it using `batch_dot` but we also have to check that `gather` works with a batch size dimension which is not clear from the spartan documentation at https://www.tensorflow.org/api_docs/python/tf/keras/backend/gather.

stellargraph/layer/graph_classification.py
PantelisElinas: Ultimately, this would be the best way to build a general graph classification model where the user has control of the type of graph convolution used. However, as I described in my comment about the need for the custom `GraphClassificationConvolution` layer, there are many implementation details for our existing layers and models that make it difficult to use as you suggest. The treatment of the batch dimension, the `out_indices` input tensor and the notion of a last layer for the existing models make the above solution a bit difficult to implement. I think if we resolve these issues then we can re-visit this.

stellargraph/mapper/sequences.py
PantelisElinas: I didn't know this. I'll update as suggested.

stellargraph/mapper/sequences.py
PantelisElinas: I must have misunderstood the documentation because I could not get it to pad rows only. I'll have another look.

stellargraph/mapper/sequences.py
PantelisElinas: I usually prefer to be explicit to avoid any future surprises. But I can see how being too verbose can reduce readability. I'll just remove them.

stellargraph/mapper/sequences.py
PantelisElinas: Removed as suggested although in theory anyone could implement a custom generator that returns a `GraphSequence`. I would guess that the likelihood of that happening before we have unit tests is rather low so best to avoid the clutter of all the warnings as suggested.

huonw: Yeah, it's unlikely, and, we don't seem to encourage or support users using our `Sequence`s, e.g. they're not in our documentation at all.

stellargraph/layer/graph_classification.py
PantelisElinas: I've updated with regards to kwargs but I want to keep the example because this model cannot be used unless at least another layer is added to its output, e.g., a Dense layer. For other models such as GCN, the last convolutional layer can also be the classification layer if it has the correct activation applied but not for the graph classification model.

huonw: > I want to keep the example because this model cannot be used unless at least another layer is added to its output, e.g., a Dense layer

I'm not sure how that's relevant to #801? Maybe I was confusing, I meant: "Instead of adding a new class that has the problem noted in #801, maybe this could ...".

PantelisElinas: I thought you meant the example of how to use the model not the list for #801. All good,  I'll just close this thread.

stellargraph/mapper/sequences.py
PantelisElinas: Resizing the csr matrix as suggested does not work. The following program,

```
import scipy.sparse as sps
a = sps.rand(10, 10, density=0.01, format='csr')
sps.csr_matrix((a.data, a.indices, a.indptr), shape=(12,12))

```
raises the following exception,

`ValueError: index pointer size (11) should be (13)`

I'm using scipy version 1.4.0. Something must have changed in scipy that this no longer works as I see this has been suggested as a solution at https://stackoverflow.com/questions/26462617/adding-a-column-of-zeroes-to-a-csr-matrix as well.

However, calling `a.resize((12,12))` seems to work as expected so I might give this one a try.

huonw: Ah, sorry for the misleading advice; that answer is indeed what I was working off. I looked for a resizing function like that but didn't notice it; oops. Looking at the implementation it does seem like it's a moderately complicated operation for `csr` matrices, so definitely good to use that 👍

stellargraph/layer/graph_classification.py
PantelisElinas: Essentially, a general name for a very generic algorithm. Most other models we implement already have names as published, e.g., GCN, GAT, GraphWave, DDGK, DGCNN, etc. but this one is so generic that it is hard to come up with a good name; any suggestions are welcome.

That said, I will rename it `SupervisedGraphClassification` because it is more specific.

stellargraph/layer/graph_classification.py
huonw: > The problem is that the input for graph classification does not include node indices, i.e., out_indices, since we need the output for all nodes for mean pooling. In graph classification out_indices has been replaced by mask to handle different size graphs

Other than the gather difficulties, _could_ we use `out_indices` rather than `mask`, and pass all of the nodes as indices?

> I think we might be able to fix it using batch_dot but we also have to check that gather works with a batch size dimension which is not clear from the spartan documentation at https://www.tensorflow.org/api_docs/python/tf/keras/backend/gather.

That is rather poor, especially because, based [on the source](https://github.com/tensorflow/tensorflow/blob/988ba54e96aa1c0254c00823049cf46109d4335c/tensorflow/python/keras/backend.py#L1936-L1965), it just defers to `tf.gather`, which has much better docs: https://www.tensorflow.org/api_docs/python/tf/gather

Although, given how few arguments the Keras one has, we may have to call `tf.gather` directly, if `axis`'s default isn't what we need.

> If #1038 is resolved

Minor, but that issue is now closed, without us having actually fixed GAT/GCN, so if we're going to want to do that work, we'll need a another issue to track it.

PantelisElinas: Initially, I did consider using `out_indices` instead of `masks` but with having to pad the data matrices it is not clear to me what to pad the `out_indices` with since we need `out_indices` to have `batch size x max nodes` size or we can't stack them together for the batch. Looking at the documentation for `tf.gather`  and if I understand correctly, repeating an index makes extra copies of it which is not what we want. 

We are going to open a new issue to tackle the problems with our GAT (and I think also GCN) layer.

stellargraph/mapper/graph_generator.py
```diff
@@ -0,0 +1,105 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from ..core.graph import StellarGraph
+from ..core.utils import is_real_iterable
+from .sequences import GraphSequence
+from ..core.experimental import experimental
+
+
+@experimental(reason="Missing unit tests.", issues=[1042])
+class GraphGenerator:
```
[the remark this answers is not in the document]
PantelisElinas: On this one I would have to disagree. This generator is not specific and certainly it is not full batch. It generates batches of graphs which as you point out are represented by their adjacency and node feature matrices but it can be no other way since the input to neural networks has to be in tensor form. I specifically called this `GraphGenerator` because I hope we will be able to re-use in the future with other graph classification algorithms. For example, we should be able to re-use it as is for #923.

stellargraph/layer/graph_classification.py
PantelisElinas: I don't want to argue too much about what to name things, but as I said before many models and layers have distinct names in the literature so those will be easy to name. This model, however, does not so we can name it whatever we want. I don't particularly like very long names so I'll just suggest we go with `SupervisedGCNGraphClassification`, that is drop mention of node pooling in the name and just have it explained in the documentation. This way, if we want to use a different pooling layer, we won't have to rename the model class as well. I had a preference for just `SupervisedGraphClassification` because then we could have extended both the model and the pooling method without renaming the model class but happy to go with the more specific `SupervisedGCNGraphClassification` or maybe `GCNSupervisedGraphClassification`. Just let me know which name your prefer of all the ones suggested in this thread.

stellargraph/mapper/sequences.py
PantelisElinas: I have updated as suggested.

stellargraph/layer/graph_classification.py
huonw: I'm fine either `SupervisedGCNGraphClassification` or `GCNSupervisedGraphClassification `, with a mild preference for the latter, because it potentially makes the connection to a more general `SupervisedGraphClassification` class clearer 👍 



> I had a preference for just SupervisedGraphClassification because then we could have extended both the model and the pooling method without renaming the model class 

My concern is that we won't be able to do this in a great way. To do an extension with the class directly, we'll likely need to add even more arguments to have all the options for all of the supported algorithms (e.g. `attn_heads` and `attn_heads_reduction` for GAT, `teleport_probability` for APPNP) as well as some custom way for the user to tell the graph classification code which to use, even though we've already got most of the pieces of that: the individual model classes (each takes the appropriate arguments, and "which to use" is the user choosing the class they want).

I thus suspect we may want a noticeably different API for a general form, such as something more similar to `DeepGraphInfomax` where we can reuse those existing model-defining classes (this may require some careful adjustments to those classes too):

```python
class SupervisedGraphClassification:
    def __init__(self, base_model, generator, pooling="mean"):
        ...

    ...
```

The existing `GCNSupervisedGraphClassification` can then be implemented in terms of it, without us having to do any complicated argument overloading tricks:

```python
class GCNSupervisedGraphClassification(SupervisedGraphClassification):
    def __init__(self, layer_sizes, activations, generator, ..., pooling="mean"):
        base_model = GCN(layer_sizes, activations, ...)
        super().__init__(self, base_model, generator, pooling)
```

(We can then deprecate this specialised class if we wish, or leave it as a convenience. Also, this is all predicated on GCN/GAT/... working for graph classification.)

Do you have a different idea for how you were planning to generalise, that I'm not understanding?

> This way, if we want to use a different pooling layer, we won't have to rename the model class as well

Yeah, I guess it's easier to be able to swap a single layer simple layer like this in or out (as my hypothetical `pooling` argument above demonstrates), and so there's not so many subtle future generalisation/migration concerns. 👍

stellargraph/layer/graph_classification.py
huonw: > We are going to open a new issue to tackle the problems with our GAT (and I think also GCN) layer.

Just so I'm not making an assumption about the responsibilities and the task doesn't get dropped accidentally, "we" = you?

PantelisElinas: I had a chat with Kevin with regards to whether we should re-open #1038 or create a new issue. We agreed to open a new ticket and Kevin did so last week #1076.

stellargraph/layer/graph_classification.py
PantelisElinas: I hope we can agree that for now, we can keep it as is and after we have implemented a couple more graph classification algorithms, we can revisit and re-factor as necessary. We should probably discuss bigger picture design issues either in a different ticket or offline.

huonw: I was explaining my reasoning to make sure we were on the same page since your mention of extending this class made me wonder if you had some other plan in mind that allowed us to have the good name right now. In addition, I was careful to the code examples tiny to just cover the core idea and only explore the design space necessary to guide the names in PR.

(Also, since there was no comment like "I went with `GCNSupervisedGraphClassification`", I didn't realise that you'd decided upon, edited and pushed a new name.)

PantelisElinas:
@huonw I have updated the implementation based on your last review comments so I have re-requested a review.

P.

stellargraph/mapper/graph_generator.py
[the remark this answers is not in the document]
huonw: My reasoning with calling it "specific" was that this generator yields the graphs in a specific format, not that it relates to a specific algorithm or can't be reused (neither of my suggested names hint at that, just like the `FullBatchNodeGenerator` type doesn't talk about a specific algorithm, but does hint at *how* it yields the graph). In particular, it pads all matrices to all have the same shape (given this, another name could be `PaddedGraphGenerator`).

My reasoning with the "full batch" name is that it yields the full information about a graph, and so is, in some sense, a generalisation of `FullBatchNodeGenerator` to have a batch size > 1. This could potentially be another (maybe long term) approach here, unify this with `FullBatch*Generator` by allowing passing multiple graphs into those types.

Other options I could see for types that could be called a "graph generator":

- a generator that only yields a single graph at a time (although I guess this would just be an optimised form of `.flow(..., batch_size=1)`)
- something using [ragged tensors](https://www.tensorflow.org/guide/ragged_tensor) to not have to do any padding

However, I guess this is a perfectly reasonable default behaviour one might expect from something called a "graph generator".

stellargraph/layer/graph_classification.py
```diff
@@ -0,0 +1,255 @@
...
+
+    The model minimally requires specification of the GCN layer sizes as a list of ints
+    corresponding to the feature dimensions for each hidden layer,
+    activation functions for each hidden layers, and a generator object.
+
+    To use this class as a Keras model, the features and pre-processed adjacency matrix
+    should be supplied using the :class:`GraphGenerator` class.
+
+    Notes:
+      - The inputs are tensors provided by the :class:`GraphGenerator` object.
```
huonw: I think this note and the paragraph immediately before it are saying essentially the same thing. Maybe they could be combined?

demos/graph-classification/supervised-graph-classification.ipynb
```diff
@@ -0,0 +1,559 @@
+{
```
huonw: <p>There seems to be an image here that neither ReviewNB nor Github's rendering can show https://github.com/stellargraph/stellargraph/blob/6c7d2e09681ccdbd3afc1e56a872680cc6fc5523/demos/graph-classification/supervised-graph-classification.ipynb, although nbviewer.jupyter.org can: https://nbviewer.jupyter.org/github/stellargraph/stellargraph/blob/6c7d2e09681ccdbd3afc1e56a872680cc6fc5523/demos/graph-classification/supervised-graph-classification.ipynb</p><p> </p><p>I wonder if we can find a different way to do this that Github can handle?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1053/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

huonw: <p>The <code>for</code>  loop could be written using <code>index</code>  and indexing:</p><pre class="ql-syntax" spellcheck="false">test_metrics[model.metrics_names.index("acc")]
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1053/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

huonw: <p>The code in the nested loops is very dense; maybe there could be a little more whitespace  (e.g. before the inner <code>for</code>  and each of the <code>if</code>  statements.)</p><p> </p><p>Also, why clone model  when assigning to <code>best_model</code>? I don't think the <code>model</code> variable gets reused?</p><p> </p><p>Finally, if a user was trying to use this graph classification algorithm, I wonder if they'd struggle to translate this nested loop into a straight-line (non-loop) form that they'd actually use for creating and running the model. One option would be to a section demonstrating the conventional/basic code, and then go into the 10-fold validation code. What do you think?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1053/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

huonw: <p>If you were so inclined, you could use python's native rounding-formatting:</p><p> </p><pre class="ql-syntax" spellcheck="false">print(
&nbsp; &nbsp; f"Accuracy over all folds mean: {np.mean(test_accs)*100:.1} and std: {np.std(test_accs)*100:.1}"
)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1053/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

huonw: <p>The validation accuracy is quite quantised here, which might be surprising/disconcerting to some readers (I certainly was a bit wary of it, until I worked out what was going on). Is it worth mentioning the quantisation is because the validation/test set for each fold has only ~19 (10% of 188) elements?</p><p> </p><p>Also, it seems that training barely does anything for the metrics on the train set, especially given how much noise there is?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1053/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

stellargraph/mapper/graph_generator.py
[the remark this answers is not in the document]
PantelisElinas: Your definition of "full batch" is not standard. The data here is not a single graph but a collection of graphs. It would be full batch if we required that all graphs are processed at once for each epoch. This is not the case. For now, this generator is correctly named. We can revisit in the future if the need arises. I am not familiar with ragged tensors and trying to re-write the generator now to use them would not be the best use of my time considering that there is no guarantee that it will work as it is now.

stellargraph/mapper/sequences.py
```diff
@@ -530,6 +533,126 @@ def __getitem__(self, index):
...
+        ]
+
+        adj_graphs = self.normalized_adjs[
+            index * self.batch_size : (index * self.batch_size) + self.batch_size
+        ]
+
+        # The number of nodes for the largest graph in the batch. We are going to pad with 0 rows and columns
+        # the adjacency and node feature matrices (only the rows in this case) to equal in size the adjacency and
+        # feature matrices of the largest graph.
+        max_nodes = max([graph.number_of_nodes() for graph in graphs])
```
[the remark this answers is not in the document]
PantelisElinas: I think it is fine as it is. We don't know if some future dataset will a graph that is considerably larger than all others and we might be wasting space to avoid an operation that doesn't seem to be that costly. I don't see any reason to over-optimise at this stage. I want to get some more datasets into the library, some of which are larger than MUTAG, to get a better idea of performance before spending more time on optimisations.

stellargraph/mapper/graph_generator.py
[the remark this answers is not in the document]
huonw: > Your definition of "full batch" is not standard. The data here is not a single graph but a collection of graphs. It would be full batch if we required that all graphs are processed at once for each epoch

You're right, that full-batch is a misleading name here (and hence it was only one of several that I provided), but I don't think any of your reply addresses the fact that what we call `FullBatchNodeGenerator` and what we call `GraphGenerator` are doing essentially the same thing, just the latter allows more than one graph as input. Their output is even passed into essentially the same type of model, for essentially the same task (at least, `FullBatchNodeGenerator` can be used for computing node embeddings).

Finding generalisations like this is a core part of both mathematics and programming...

> For now, this generator is correctly named. We can revisit in the future if the need arises

I'm not saying it is incorrectly named. I'm saying that it _may_ be _sub-optimally_ named, because it takes up a really basic name in "API space" and thus makes future generalisations difficult.

Like with the name of `GraphClassification`, revisiting in future is hard: if users already have a pile of code using `GraphGenerator`, it is extremely poor of us to rename that without some sort of migration assistance. Having that migration assistance likely means we cannot use `GraphGenerator` for something else, like a superclass for all graph generators.

> I am not familiar with ragged tensors and trying to re-write the generator now to use them would not be the best use of my time considering that there is no guarantee that it will work as it is now.

I don't think I was even hinting at doing any rewrites or changing the code.

I'm trying to think of options that we _might_ want, to explore whether we're going to have other forms of "graph generators". If we're definitely not going to have any other forms, then calling this singular one (that's the only one we'll ever have) `GraphGenerator` is fine. If there may be other forms, then we may want to think about the name. My list was a list of possibilities for other forms.

I'm trying to have a discussion about future possibilities here, to make sure we're not boxing ourselves into a suboptimal corner. If you're ok with the risk that we'll be stuck with `GraphGenerator` referring to this specific style of generating graphs, I guess we'll just make do with it as is.

demos/graph-classification/supervised-graph-classification.ipynb
PantelisElinas: I don't know how github displays images or why it fails on this one specifically.

We can just remove it since the architecture is clearly explained in the text and easily understood from reading `create_graph_classification_model`. 

That said, I'll have a look if the problem is with the image format.

stellargraph/mapper/graph_generator.py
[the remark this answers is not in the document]
huonw: I'm happy to stop trying to think about this for now and just land it as is. We can have a longer term brainstorm, until the point we de-`@experimental`-ise it.

demos/graph-classification/supervised-graph-classification.ipynb
PantelisElinas: I think the accuracy plot is not surprising given, as you said, the small amount of data. Also, the plot is for the best found model out of 50 checked models so we don't know exactly how the losses for all the other models behave although I suspect similarly. The dataset is also not balanced; there are twice as many graphs labeled "1" than "-1".

Generally, the loss is the more interesting quantity here although this being the best model, it looks like it got really lucky because the validation loss is lower than the training loss. It could be that this batch is easy to predict. The training loss look like it behaves correctly although it is noisy. We can make it look smoother by reducing the learning rate. The value I currently use was a good compromise between learning fast enough and producing reasonably looking loss curves.

Anyway, these curves don't always look like the ideal one we see in textbooks. If we could run this on a bigger dataset with some more interesting and informative features, and well balanced, I suspect the curves would look more textbook-like.

I'll mention a couple of the above points in the notebook.

PantelisElinas: I'll update to make the loop easier to read (I hope.)

The copying of the model was added because I initially thought I can create one model and then re-initialise the weights at each iteration of the loop instead of creating a new one. Unfortunately, I could not find a way to reset a model's weights and was forced to create a new one every time. As you correctly point out, there is not need to copy the model anymore, and I will remove.

The training loop is rather common for k-fold training. Hopefully a small re-factoring will improve clarity. I don't want to assume that our users cannot un-roll a simple loop like this and I prefer not to make this notebook grow too long. That said, I had initially coded this using a single split of the data but because the data is small, every run would produce widely different performance and I thought not worth having it because it unfairly makes the method look poor.

PantelisElinas: I added the image to the repo and it looks like it is rendering correctly now.

stellargraph/mapper/graph_generator.py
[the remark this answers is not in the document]
PantelisElinas: I suggest that instead of discussing what-if scenarios and library-wide API design in any single ticket, we should, as you suggest, have a proper design discussion offline. There are many small differences between all the different generators, models, and layers that force us to replicate code. You are correct that `FullBatchGenerator` and `GraphGenerator` are similar and in fact, the latter is more general and could be used for training node representation models, if it wasn't for our "need" to output predictions only for a subset of the nodes, hence the need for `out_indices` and the call to `gather` in the end. So the way I see it, it is `FullBatchGenerator` that will eventually be deprecated for not being very flexible, not `GraphGenerator`.

Anyway, I think we agree to just table this discussion for now.

stellargraph/mapper/sequences.py
```diff
@@ -530,6 +533,126 @@ def __getitem__(self, index):
...
+        #      batch size x max number of nodes x max number of nodes
+        # graph_targets is array of dimensionality
+        #      batch size x number of classes
+        return [features, masks, adj_graphs], graph_targets
+
+    def on_epoch_end(self):
+        """
+         Shuffle all graphs at the end of each epoch
+        """
+        indexes = list(range(len(self.graphs)))
```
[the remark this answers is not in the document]
PantelisElinas: No objection about the suggested change, but I would prefer to tackle it as part of #1043 (see https://github.com/stellargraph/stellargraph/issues/1043#issuecomment-599840069) since this bit of code will very likely have to be updated for that ticket.

PantelisElinas:
@huonw I have done another round of updates that you can review. However, the notebook fails on CI and I'm not sure why. It doesn't finish training so I suspect that it runs out of time? I don't know how the CI is configured so let me know if you have any suggestions.

P.

huonw:
> It doesn't finish training so I suspect that it runs out of time?

Yep; the training cell takes too long (> 600s).

https://buildkite.com/stellar/stellargraph-public/builds/2350#e1c064fc-78be-4a5e-a235-78f6dfabc199/429-1353

> I don't know how the CI is configured so let me know if you have any suggestions.

The best way to handle this is to use our [papermill-based parameterisation](https://papermill.readthedocs.io/en/latest/usage-parameterize.html). This involves reducing some of the (hyper)parameters that make the notebook slow to run by:

- choosing which parameters will be reduced; in this case the number of epochs, number of restarts and/or even the number of folds seem like good candidates?
- adding a cell that defines them (with their real values), and adding the `parameters` tag to that cell (see the linked docs above for how to do it in the Jupyter Notebook and Jupyter Lab interfaces). For instance, if we go with the first two it might be placed just before `train_fold` is defined and contain
  ```python
  epochs = 500

  # For each fold, ...
  num_restarts = 5
  ```
- updating [`.buildkite/notebook-parameters.yml`](https://github.com/stellargraph/stellargraph/blob/develop/.buildkite/notebook-parameters.yml) with any new parameters and their reduced amount, e.g. for that code above, we'd have to add something like the following to the end of the file
  ```yml
  # fewer models per fold for demos/graph-classification/supervised-graph-classification.ipynb
  num_restarts: 2
  ```

When running on CI, Papermill will insert an extra cell that overrides the default values, and so hopefully make the notebook run quickly.

The important thing with these is to make sure all code paths are executed (hence `num_restarts != 1` to ensure execution of implicit code path of running the restart loop multiple times), without needing to get good model performance, since it's just run transiently on CI.

demos/graph-classification/supervised-graph-classification.ipynb
huonw: Hm, unfortunately I don't see it: https://github.com/stellargraph/stellargraph/blob/ca295c3e9a8dcc61fe7570cb765a7a7f543a85c0/demos/graph-classification/supervised-graph-classification.ipynb

image

One solution would be to include the image as you originally had it, but tweak the text to not rely on it so that it's fine if it doesn't render (I think this may be as simple as removing the "The below image shows the architecture we are going to use." sentence, since, as you say, the architecture is explained clearly in the text).

huonw: Looks great.

huonw: It's not a matter of "can" or "cannot", it's a matter of "how can we make it easy for them?". I know I personally appreciate things being spelled out when I'm learning them, even if I can work out the more complicated form. This is especially true because there's several layers here, not just loops: to understand the "core" concepts of doing graph classification, the user currently has to string together the definitions of `create_graph_classification_model`, `train_fold` and `get_generators` as well as detangle the 10-fold-related code, which is a lot more complicated than understanding the straight-line no-function form.

I mocked up a quick possible example of what an addition might look like https://nbviewer.jupyter.org/gist/huonw/2933706bc39adc2ba2b85b829b8513b2 (it can be downloaded in the top right hand corner).

I agree with your concern about the length, though, and it's unfortunate that every `fit` (and `evaluate`) call seems to be hitting #826, causing the amount of output to inflate. That said, along those lines, maybe we could have `verbose=0` for the evaluate call too to minimise the `1/1 [==============================] - 0s 8ms/step - loss: 0.4701 - acc: 0.7895` lines?

PantelisElinas:
I have created #1090 for a tutorial on graph classification that is more basic that the demo notebook associated with this ticket.

PantelisElinas:
> Looks fine, with the main remaining concern being the image in the notebook still not behaving quite right ([#1053 (comment)](https://github.com/stellargraph/stellargraph/pull/1053#discussion_r393434502)).

It renders correctly in github (checked on a couple of different machines and web browsers) and I have updated the text as suggested. Hopefully, it will not confuse anyone if not rendered.

huonw:
Ah, yeah, it does for me on develop too 👍
I don't know why I couldn't (and still can't) see it in the link in that comment, even though the linked commit does seem to have the notebook using the `` markdown form with the image adjacent 🤷‍♂

PR #1033, opened 2020-03-06
Enhancement/894 no node error handling

kieranricardo:
This PR adds a check in the `StellarGraph` constructor to check that all nodes referred to in `edges` are contained in `nodes`. It also adds functionality to infer the nodes when an edgelist is passed and `nodes=None`.

See: #894

PR #1056, opened 2020-03-10
Add CTDNE (time-based random walk) demo

kjun9:
See #827 

This running on CI with reduced numbers for:
```
walk_length: 5  # random walk length
cw_size: 5  # context window size
```

@habiba-h I kept your approach of comparing biased random walks to temporal walks - there's barely any difference in the performance though 🤷‍♂ 

Also added the link prediction utilities directly in the demo - it ended up looking fine I think, so #1020 is closed


huonw:
> I can do that here or in a separate PR depending on which is preferred for reviewers, but I think the notebook can be reviewed for its general content/workflow regardless

It'd be good to land this being checked by CI rather than have to add another FIXME. I'd be happy to see the dataset added here or in a separate PR (that lands first).

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "# split recent edges further to train and test sets\n",
+    "test_mask = np.random.rand(len(edges_other)) < test_subset\n",
+    "edges_train = edges_other[~test_mask]\n",
+    "edges_test = edges_other[test_mask]\n",
+    "\n",
+    "print(\n",
+    "    f\"Number of edges in graph: {len(edges_graph)}\\n\"\n",
+    "    f\"Number of edges in training set: {len(edges_train)}\\n\"\n",
+    "    f\"Number Of edges in test set: {len(edges_test)}\"\n",
+    ")"
```
huonw: I think this cell (and the `create_link_examples` one too) can theoretically use [`EdgeSplitter`](https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.data.EdgeSplitter), and, even if not, can use [`sklearn.model_selection.train_test_split`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html) to be a bit more declarative. What do you think?

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def create_link_examples(g, edges):\n",
+    "    pos = [(src, tgt) for (src, tgt) in edges[[\"source\", \"target\"]].values]\n",
+    "    neg = sample_negative_examples(g, pos)\n",
+    "    return pos, neg\n",
+    "\n",
+    "\n",
+    "def sample_negative_examples(g, positive_examples):\n",
```
huonw: I'm sure you realise this, but it's lucky this is a graph with only 151 edges, since this function is O(n^2) :smile:

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+   ],
+   "source": [
+    "from stellargraph.data import TemporalRandomWalk, BiasedRandomWalk\n",
+    "\n",
+    "num_walks_per_node = 10\n",
+    "walk_length = 80\n",
+    "cw_size = 10\n",
+    "num_cw = (\n",
+    "    len(graph.nodes()) * num_walks_per_node * (walk_length - cw_size)\n",
+    ")  # as recommended in reference paper\n",
```
huonw: This line has a lot of moving parts, so what specifically was recommended in the reference paper? Maybe this could refer to a specific section?

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Running random walks\n",
+    "\n",
+    "We try to keep the setup comparable between the use of temporal and biased (static) random walks. For temporal walks, the input parameter is defined in terms of the total number of context windows you are interested in obtaining, which differs from the traditional approach of specifying the number of walks to run per node in the graph. We calculate the number of context windows we need in terms of the traditional parameters as:\n",
+    "```\n",
+    "num_cw = num_nodes * num_walks_per_node * (walk_length - cw_size)\n",
```
huonw: What's the motivation for repeating the code here? If you think this is clearer, maybe the code could be tweaked to allow writing this directly? E.g.

> Define the parameters we use ...
> ```python
> num_walks_per_node = 10
> walk_length = 80
> cw_size = 10
> ```
> We try to keep the setup comparable [...] We calculate the number of context windows we need in terms of the traditional parameters as:
>
> ```python
> num_cw = graph.number_of_nodes() * num_walks_per_node * (walk_length - cw_size)
> ```
>
> We're now ready to do the walks:
>
> ```python
> temporal_rw = TemporalRandomWalk(graph)
> ...
> ```

The discussion of what was recommend in the reference paper could then move into the discussion of the `num_cw` cell, too.

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "\n",
+    "static_rw = BiasedRandomWalk(graph)\n",
+    "static_walks = static_rw.run(\n",
+    "    nodes=graph.nodes(), n=num_walks_per_node, length=walk_length\n",
+    ")\n",
+    "\n",
+    "\n",
+    "print(\"Number of temporal random walks: {}\".format(len(temporal_walks)))\n",
+    "print(\"Number of static random walks: {}\".format(len(static_walks)))"
+   ]
```
huonw: How long does the walking in this cell take? Is it worth splitting it up into a `TemporalRandomWalk` cell followed by a `BiasedRandomWalk` one so that changing the parameters for one doesn't require re-running the other one too?

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    ")\n",
+    "static_model = Word2Vec(\n",
+    "    static_walks, size=128, window=10, min_count=0, sg=1, workers=2, iter=1\n",
+    ")\n",
+    "\n",
+    "# using the trained word2vec model, create a function we can use to generate embeddings\n",
+    "def temporal_embedding(u):\n",
+    "    try:\n",
+    "        return temporal_model.wv[u]\n",
+    "    except KeyError:\n",
```
huonw: Is this `KeyError` handling any nodes that don't have an embedding?

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "        return np.zeros(128)\n",
```
huonw: Instead of hardcoding the 128 here, maybe this could be either computed from `temporal_model` (maybe there's a `temporal_model.size` property?) or pulled out into a parameter?

```python
from genesis.models import Word2Vec

embedding_size = 128
temporal_model = Word2Vec(...., size=embedding_size, ...)
static_model = Word2Vec(...., size=embedding_size, ...)

unseen_node_embedding = np.zeros(embedding_size)

def temporal_embedding(u):
    try:
        return temporal_model.wv[u]
    except KeyError:
        return unseen_node_embedding
```

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "temporal_node_embeddings = temporal_model.wv.vectors\n",
+    "static_node_embeddings = static_model.wv.vectors\n",
+    "plot_tsne(\"TSNE visualisation of temporal node embeddings\", temporal_node_embeddings)\n",
```
huonw: There's no labels we can use for these to make the plots coloured, and thus more interpretable (rather than just "oh, yeah, seem grouped")?

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+   "metadata": {},
+   "source": [
+    "## Link Prediction using Node Embeddings\n",
+    "\n",
+    "The node embeddings we've generated can now be used as input for a link prediction task. The reported AUC scores from the reference paper for this dataset are:\n",
+    "\n",
+    "* Node2Vec: 0.759 \n",
+    "* CTDNE: 0.777\n",
+    "\n",
+    "Below are a set of helper functions we can use to train and evaluate a link prediction classifier. The rest of the notebook will use the binary operator defined in the cell below. Available operators in the demo are:\n",
```
huonw: I'm not personally a huge fan of adding unused code to notebooks: it's easy for the code to just be a distraction from the teaching, and it's not tested and so may break. We could delete it, but that's a bit less nice than exploring multiple styles, especially if they give very different results.

Do you think we could at least run the temporal embeddings ROC AUC evaluation with all of them, but maybe still use only L2 for static and the t-SNE plot?

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "        operator_func[binary_operator]\n",
+    "        if isinstance(binary_operator, str)\n",
+    "        else binary_operator\n",
+    "    )\n",
+    "    return [\n",
+    "        op_func(transform_node(src), transform_node(dst)) for src, dst in link_examples\n",
+    "    ]\n",
+    "\n",
+    "\n",
+    "def link_prediction_clf(max_iter=2000):\n",
```
huonw: What's CLF stand for? I see it's used in the sklearn docs too, but... it's not obvious to me.

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "        ]\n",
+    "    )\n",
+    "\n",
+    "\n",
+    "def evaluate_roc_auc(clf, link_features, link_labels):\n",
+    "    predicted = clf.predict_proba(link_features)\n",
+    "\n",
+    "    # check which class corresponds to positive links\n",
+    "    if clf.classes_[0] == 1:\n",
+    "        return roc_auc_score(link_labels, predicted[:, 0])\n",
```
huonw: This could be:

```python
positive_column = clf.classes_.index(1)
return roc_auc_score(link_labels, predicted[:, positive_column])
```

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def create_link_examples(positive_examples, negative_examples):\n",
+    "    return (\n",
+    "        positive_examples + negative_examples,\n",
+    "        np.array([1] * len(positive_examples) + [0] * len(negative_examples)),\n",
```
huonw:
```suggestion
    "        np.repeat([1, 0], [len(positive_examples), len(negative_examples)])\n",
```

Or:

```suggestion
    "        np.concatenate([np.ones_like(positive_examples), np.zeros_like(negative_examples)]),\n",
```

kjun9:
> It'd be good to land this being checked by CI rather than have to add another FIXME. I'd be happy to see the dataset added here or in a separate PR (that lands first).

@huonw I've opened https://github.com/stellargraph/stellargraph/pull/1058 if you're interested in reviewing!

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: Hmm right, I think the initial split still needs to be done via grabbing slices of the pandas dataframe since we want to split the old vs new edges. I'll switch the train/test split to use one of those though

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: What I initially wanted to do was to show the equation from the paper in the markdown, and then write a comment on that line of the code to indicate that that expression matches the equation, but yeah it looks a bit odd that it's just repeating the same code.

I'll split into multiple cells like you've suggested

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: Ah yep, I'll include an explanation

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,680 @@
+{
```
habiba-h: <p>I would suggest to split this into three steps.</p><ol><li>Hyperparameters setting.</li><li>Temporal Random Walks</li><li>Static Random Walks</li></ol><p>Too much going on in one place.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1056/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: Yeah I guess I could omit this part, I found it slightly useful to see whether the embeddings "looked informative" on the surface while I was working on it

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: I assumed "classifier"?

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
habiba-h: <p>Interestingly, the AUC for the temporal and static method are almost the same  over and above being better than the paper's results.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1056/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

kjun9: Yeah I tried to figure out for a while whether the way I've set up is cheating in some way but decided to move forward eventually... so let me know if you spot something!

habiba-h: <p>I think it should be mentioned here that CTDNE method can potentially miss certain nodes in the walks and hence there won't be any embeddings calculated for them. This issue is being  fixed here by assigning embeddings vectors of 0's for the nodes not visited in any of the walks. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1056/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: Some of the operators ended up causing convergence warnings while training the classifier (although the scores still looked fine) so that's partly why I didn't go through all of them - should I just remove them?

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
huonw: I think it's reasonable to include, since as you say, it gives a hint about whether the embeddings are picking up signal in some form, I'm just wondering if there's a way we can augment them further... it seems not easily, from what's in the dataset, though.

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: Yeah they do take a little while to run (maybe a few minutes?) I'll split them

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: Theoretically, the temporal embeddings might cluster together nodes that have edges with similar time values, and separate those with very different time values. I can try to assign colours to nodes based on something like the average time across all its incident edges, but no guarantee that it will actually look meaningful

huonw: Ah, yeah. Interesting idea, although maybe too much code to be worth it, unless you think it is going to be enough of an improvement?

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: I've split this up now 👍

kjun9: Thanks, I've added a bit of an explanation

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: The code wasn't too much to add in the end, but I'm not sure if the visual is all that convincing:
image

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: I guess an alternative would be to just do the random sampling in a while loop and each time check that it doesn't exist in the graph - do you think that would be better?

Looks like our `EdgeSplitter._sample_negative_examples_global` has a slightly different approach too, but I thought it was simple/clear enough to just do it in the notebook instead.

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,725 @@
...
+    "def sample_negative_examples(g, positive_examples):\n",
+    "    possible_edges_all = {\n",
+    "        (src, tgt) for src in g.nodes() for tgt in g.nodes() if src != tgt\n",
+    "    }\n",
+    "    possible_edges_neg = list(\n",
+    "        possible_edges_all.difference(\n",
+    "            {e for (src, tgt) in positive_examples for e in [(src, tgt), (tgt, src)]}\n",
+    "        )\n",
+    "    )\n",
+    "    return random.choices(possible_edges_neg, k=len(positive_examples))\n",
```
huonw: Do you mean to sample with replacement?

Currently this is extremely likely to have overlaps. There's 151&times;150 = 22650 non-self loop edges in a graph with 151 nodes, and out of the 50572 (total edges) &times; 0.25 (graph vs. train/test split) &times; 0.75 (train split) = 9482 positive train edges, it seems there's usually ~1300 unique undirected edges. This means this `choices` is sampling with replacement N=9482 examples from K=~21350 possible edges, which makes a multinomial distribution with probability vector (1/K, 1/K, ..., 1/K) (length K) and sample size N. We can work out the expected number of the K unique values that are sampled, by subtracting the number that *aren't* sampled from the total number K: `K - K * (1 - 1/K)**N = 7658.1` (https://stats.stackexchange.com/questions/420643/expected-number-of-zeros-in-multinomial-vector). That is, this is, on average, getting only 7658 unique negative edges across the 9482 elements of list here. (That theoretical average matches pretty closely to values I see in practice, across a few runs: 7658, 7643, 7614, 7700.) The probability of no duplicates is rather small: ~`(1 - 1/K)**(N*(N-1)/2)` &approx; 4&times;10<sup>-913</sup> (https://en.wikipedia.org/wiki/Birthday_problem#Cast_as_a_collision_problem).

I suspect this may be incorrect, which case, maybe this should be done without replacement, e.g. via `random.sample`?

```suggestion
    "    return random.sample(possible_edges_neg, k=len(positive_examples))\n",
```

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: In section 3 of the paper:

> The remaining 25% are considered as positive links and
we sample an equal number of negative edges randomly

is the only reference to how this is done. I would expect the scores to improve further if negative edges were forced to not overlap with any of the positives at all? So in this case I'm not sure it explains the improvement in our code.

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,725 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def plot_tsne(title, x, y=None):\n",
+    "    tsne = TSNE(n_components=2)\n",
+    "    x_t = tsne.fit_transform(x)\n",
+    "\n",
+    "    plt.figure(figsize=(7, 7))\n",
+    "    plt.title(title)\n",
+    "    if y is not None:\n",
```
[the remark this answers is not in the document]
kjun9: Ah thanks for providing those links! `jet` looks very monochrome-y to me.. maybe I could use the colours being used in the `plot_history` util?
```python
    colors = plt.rcParams["axes.prop_cycle"].by_key()["color"]
    color_train = colors[0]
    color_validation = colors[1]
```

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,725 @@
...
+    ")\n",
+    "temporal_link_features_test = link_examples_to_features(\n",
+    "    link_examples_test, temporal_embedding, binary_operator\n",
+    ")\n",
+    "temporal_clf.fit(temporal_link_features, link_labels)\n",
+    "temporal_score = evaluate_roc_auc(\n",
+    "    temporal_clf, temporal_link_features_test, link_labels_test\n",
+    ")\n",
+    "\n",
+    "print(\"\\n  **** Score on test set of links ****\\n\")\n",
```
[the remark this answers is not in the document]
kjun9: Yeah I'll remove them 👍

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: Thanks, yes I think without replacement makes sense, and perhaps explains the high scores, I'll give this a go

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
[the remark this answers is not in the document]
huonw: That's an option, but I don't know of a good way to specify that conveniently, without a chunk of manual code. I personally think the priority here should be obvious code, meaning we shouldn't have too much user-facing tweaking code, rather than perfect "report ready" output.

If the manual code isn't too bad, or there's another way to get better colours (like using a non-`jet` colormap), I'm all for it, though.

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
kjun9: Looks like it would be `l1` and `l2` - I'll switch to just using both of those unconditionally

kjun9: The unconditional switch ended up looking a lot more verbose and I wasn't such a fan of introducing even more helper functions in the notebook.. so I ended up just going with `l2`. The scores end up looking pretty much identical for `l1` and `l2` too

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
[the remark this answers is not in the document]
kjun9: I've just gone with `jet` for now, but it also meant a legend is more necessary since the colours don't match the general expectation (?) of green being true and red being false, so it does end up with slightly more code

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,725 @@
...
+    "operator_func = {\n",
+    "    \"hadamard\": operator_hadamard,\n",
+    "    \"avg\": operator_avg,\n",
+    "    \"l1\": operator_l1,\n",
+    "    \"l2\": operator_l2,\n",
+    "}\n",
+    "\n",
+    "\n",
+    "def link_examples_to_features(link_examples, transform_node, binary_operator):\n",
+    "    op_func = (\n",
```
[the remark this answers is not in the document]
kjun9: Yep I've removed the dictionary of string to operator thanks

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,725 @@
...
+    "\n",
+    "operator_func = {\n",
+    "    \"hadamard\": operator_hadamard,\n",
+    "    \"avg\": operator_avg,\n",
+    "    \"l1\": operator_l1,\n",
+    "    \"l2\": operator_l2,\n",
+    "}\n",
+    "\n",
+    "\n",
+    "def link_examples_to_features(link_examples, transform_node, binary_operator):\n",
```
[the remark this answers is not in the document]
kjun9: I've gone with one operator and removed the argument and that part of the printing

demos/link-prediction/random-walks/ctdne-link-prediction.ipynb
```diff
@@ -0,0 +1,686 @@
+{
```
huonw: <p>Now that https://github.com/stellargraph/stellargraph/pull/1033 is merged, this can be slightly simplified potentially, although maybe we want to be sure we've got all the nodes using the main graph? (Also the <code>source_column</code> and <code>target_column</code> parameters are being set to their default and so don't seem necessary?)</p><p> </p><pre class="ql-syntax" spellcheck="false">graph = StellarGraph(edges=edges_graph, edge_weight_column="time")
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1056/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

huonw: <p>Extra quote at the end of the NOTE line?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1056/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

huonw: <p>Given we way out-perform the paper, and the static and temporal walks are very similar, maybe it's worth at least acknowledging both of these (maybe something like "It turns out both the Node2Vec and CTDNE in this notebook perform much better than the paper. We may not have exactly reproduced the conditions under which the paper was testing."), so the user isn't left with a "did they even notice the difference?" question.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1056/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

huonw: <p>Is it worth adding some text to give a bit more context about what we're looking at here? At least something like</p><p> </p><blockquote>As with the nodes, we can visualise the embeddings for each link to see how the positive links (with class <code>1</code> ) are reasonably well clustered and similarly for the false links (with class <code>0</code> ).</blockquote>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1056/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

kjun9: Hmm it seems like the reduced graph has 2 less nodes than the full graph actually:
```
Number of nodes in full graph: 151
Number of nodes in training graph: 149
```
this is handled already for temporal walks by the function that grabs the embedding with `try: ... except KeyError`, but I'd have to add the same logic for static walks too. Alternatively:
* keep passing in `nodes=pd.DataFrame(index=full_graph.nodes())` when constructing the graph
* update the dataset loader to make sure every node still exists in the reduced graph

habiba-h: @kjun9  whats the reduced graph? I might have missed something.

kjun9: The reduced graph is from taking the first 75% of edges from the full dataset as the "graph", and the remaining 25% edges for training and test examples

habiba-h: Is this how they are doing it in the paper?

habiba-h: But yeah this would make sense that some nodes might be left out with the edge splitting.

kjun9: Yep, from the paper:
> To generate a set of labeled examples for link prediction, we first sort the edges in each graph by time (ascending) and use the first 75% for representation learning. The remaining 25% are considered as positive links and we sample an equal number of negative edges randomly

habiba-h: Another detail that the paper brushed under the rug...
Your alternative sounds more proper to me. With the note in the demo that the original method does not explain in their experimental section how they handle the missing nodes after the split and this is how you are taking care of it.

huonw: > keep passing in nodes=pd.DataFrame(index=full_graph.nodes()) when constructing the graph

Given reducing the code would likely break it, going with this option and keeping it as is seems fine to me.

kjun9:
Thanks everyone for the reviews! Will merge once CI passes

PR #1058, opened 2020-03-11
Add enron employees dataset

kjun9:
I've had to introduce a `url_contains_directory` parameter to `DatasetLoader` slightly to allow for datasets that are zipped without the directory - not sure what the best name for it is though.

The dataset comes from http://networkrepository.com/ia-enron-employees.php and is supposed to be cited with:
```
@inproceedings{nr,
     title={The Network Data Repository with Interactive Graph Analytics and Visualization},
     author={Ryan A. Rossi and Nesreen K. Ahmed},
     booktitle={AAAI},
     url={http://networkrepository.com},
     year={2015}
}
```

Maybe we could add some mechanism in `DatasetLoader` to hold onto the citation information? Otherwise any notebook that uses this dataset would have to paste something like that

### DIFF
--- a/stellargraph/datasets/dataset_loader.py
+++ b/stellargraph/datasets/dataset_loader.py
@@ -45,24 +45,25 @@ class DatasetLoader:
     directory_name = ""
     url = ""
     url_archive_format: Optional[str] = None
+    url_archive_contains_directory: bool = True
     expected_files: List[str] = []
     description = ""
     source = ""
     data_subdirectory_name: Optional[str] = None
-    url_archive_contains_directory: bool = True
 
     @classmethod
     def __init_subclass__(
         cls,
+        *,
         name: str,
         directory_name: str,
         url: str,
         url_archive_format: Optional[str],
+        url_archive_contains_directory: bool = True,
         expected_files: List[str],
         description: str,
         source: str,
         data_subdirectory_name: Optional[str] = None,
-        url_archive_contains_directory: bool = True,
         **kwargs: Any,
     ) -> None:
         """Used to set class variables during the class definition of derived classes and generate customised docs.
@@ -71,11 +72,11 @@ class DatasetLoader:
         cls.directory_name = directory_name
         cls.url = url
         cls.url_archive_format = url_archive_format
+        cls.url_archive_contains_directory = url_archive_contains_directory
         cls.expected_files = expected_files
         cls.description = description
         cls.source = source
         cls.data_subdirectory_name = data_subdirectory_name
-        cls.url_archive_contains_directory = url_archive_contains_directory
 
         if url_archive_format is None and len(expected_files) != 1:
             raise ValueError(
--- a/stellargraph/datasets/dataset_loader.py
+++ b/stellargraph/datasets/dataset_loader.py
@@ -49,6 +49,7 @@ class DatasetLoader:
     description = ""
     source = ""
     data_subdirectory_name: Optional[str] = None
+    create_directory: bool = False
 
     @classmethod
     def __init_subclass__(
@@ -61,6 +62,7 @@ class DatasetLoader:
         description: str,
         source: str,
         data_subdirectory_name: Optional[str] = None,
+        create_directory: bool = False,
         **kwargs: Any,
     ) -> None:
         """Used to set class variables during the class definition of derived classes and generate customised docs.
@@ -73,6 +75,7 @@ class DatasetLoader:
         cls.description = description
         cls.source = source
         cls.data_subdirectory_name = data_subdirectory_name
+        cls.create_directory = create_directory
 
         if url_archive_format is None and len(expected_files) != 1:
             raise ValueError(
@@ -125,6 +128,12 @@ class DatasetLoader:
         """Convert dataset relative file names to their full path on filesystem"""
         return os.path.join(self.base_directory, filename)
 
+    def _resolve_unpack_path(self):
+        if self.create_directory:
+            return self.base_directory
+        else:
+            return self._all_datasets_directory()
+
     def _missing_files(self) -> List[str]:
         """Returns a list of files that are missing"""
         return [
@@ -175,7 +184,7 @@ class DatasetLoader:
                 # directory_name for this dataset must match the directory name inside the archive file
                 unpack_archive(
                     temporary_filename,
-                    self._all_datasets_directory(),
+                    self._resolve_unpack_path(),
                     self.url_archive_format,
                 )
             # verify the download
--- a/stellargraph/datasets/dataset_loader.py
+++ b/stellargraph/datasets/dataset_loader.py
@@ -49,7 +49,7 @@ class DatasetLoader:
     description = ""
     source = ""
     data_subdirectory_name: Optional[str] = None
-    url_contains_directory: bool = True
+    url_archive_contains_directory: bool = True
 
     @classmethod
     def __init_subclass__(
@@ -62,7 +62,7 @@ class DatasetLoader:
         description: str,
         source: str,
         data_subdirectory_name: Optional[str] = None,
-        url_contains_directory: bool = True,
+        url_archive_contains_directory: bool = True,
         **kwargs: Any,
     ) -> None:
         """Used to set class variables during the class definition of derived classes and generate customised docs.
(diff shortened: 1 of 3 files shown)


stellargraph/datasets/datasets.py
```diff
@@ -707,3 +707,42 @@ def load(self):
...
+class IAEnronEmployees(
+    DatasetLoader,
+    name="IAEnronEmployees",
+    directory_name="ia-enron-employees",
+    url="http://nrvis.com/download/data/dynamic/ia-enron-employees.zip",
+    url_archive_format="zip",
+    expected_files=["ia-enron-employees.edges", "readme.html"],
+    description="A dataset of edges that represent emails sent from one employee to another."
+    "There are 50572 edges, and each of them contains timestamp information. "
+    "Edges refer to 151 unique node IDs in total.",
```
huonw: > Maybe we could add some mechanism in DatasetLoader to hold onto the citation information? Otherwise any notebook that uses this dataset would have to paste something like that

Instead of forcing the notebooks to paste in the citation manually, we can include it here, like some of the other datasets too, e.g. add an extra line

```python
"Ryan A. Rossi and Nesreen K. Ahmed “The Network Data Repository with Interactive Graph Analytics and Visualization” (2015)"
```

If the notebook then does `display(HTML(dataset.description))` like most do, it'll include this info, as will our documentation.

Maybe it'd be good to expand this to have an actual "citation" field too.

stellargraph/datasets/datasets.py
```diff
@@ -707,3 +707,42 @@ def load(self):
...
+    source="http://networkrepository.com/ia-enron-employees.php",
+    url_contains_directory=False,
+):
+    def load(self):
+        """
+        Load this data into a set of nodes and edges
+
+        Returns:
+            A tuple ``(nodes, edges)`` where each element is a Pandas DataFrame. Edges are kept
```
huonw: I think it would be good to return a `StellarGraph`, as well as potentially the `nodes` and `edges` DataFrames (or maybe just the `edges` one). `StellarGraph` is our native graph format, and so its good to try to phrase things in terms of that. What's the motivation for returning 'raw' dataframes?

stellargraph/datasets/datasets.py
```diff
@@ -707,3 +707,42 @@ def load(self):
...
+            sorted in ascending order of time as per the original dataset.
+        """
+        self.download()
+
+        edges_path = self._resolve_path("ia-enron-employees.edges")
+        edges = pd.read_csv(edges_path, sep=" ", header=None)
+
+        # clean up columns
+        edges.columns = ["source", "target", "x", "time"]
+        edges[["source", "target"]] = edges[["source", "target"]].astype(str)
```
huonw: Why convert these to strings?

stellargraph/datasets/datasets.py
```diff
@@ -707,3 +707,42 @@ def load(self):
...
+        Load this data into a set of nodes and edges
+
+        Returns:
+            A tuple ``(nodes, edges)`` where each element is a Pandas DataFrame. Edges are kept
+            sorted in ascending order of time as per the original dataset.
+        """
+        self.download()
+
+        edges_path = self._resolve_path("ia-enron-employees.edges")
+        edges = pd.read_csv(edges_path, sep=" ", header=None)
```
huonw: Instead of specifying the columns via a mutation, they could be specified directly:

```suggestion
        edges = pd.read_csv(edges_path, sep=" ", header=None, names=["source", "target", "x", "time"])
```

and even the `x` one ignored immediately:

```suggestion
        edges = pd.read_csv(edges_path, sep=" ", header=None, names=["source", "target", "x", "time"], usecols=["source", "target", "time"])
```

stellargraph/datasets/datasets.py
```diff
@@ -707,3 +707,42 @@ def load(self):
...
+
+        # clean up columns
+        edges.columns = ["source", "target", "x", "time"]
+        edges[["source", "target"]] = edges[["source", "target"]].astype(str)
+        edges = edges.drop(columns=["x"])  # unused column
```
huonw: Do you know what this column is meant to be? It seems to always be 1?

stellargraph/datasets/datasets.py
```diff
@@ -707,3 +707,42 @@ def load(self):
...
+
+        nodes = pd.DataFrame(
```
huonw: I don't think the `id` name is too important, and thus this could be done by setting the index directly rather than making a column:

```python
nodes = pd.DataFrame(index=np.unique(...))
```

(If the name is important, it can be set with `nodes.index.name = "id"`)

stellargraph/datasets/dataset_loader.py
```diff
@@ -49,6 +49,7 @@ class DatasetLoader:
     description = ""
     source = ""
     data_subdirectory_name: Optional[str] = None
+    url_contains_directory: bool = True
```
huonw: Maybe this could be specified just after the other `url_...` parameters, just to keep the "idea" close together (and the same move in `IAEnronEmployees`)? I think everything uses keyword arguments here, so shuffling/shifting the existing arguments shouldn't break any other code)

Also, since it only refers to archives, maybe it could be:

```suggestion
    url_archive_contains_directory: bool = True
```

It could potentially be `url_archive_contains_base_directory` or `url_archive_contains_top_level_directory` too, but I don't think that matters so much.

stellargraph/datasets/datasets.py
```diff
@@ -707,3 +707,42 @@ def load(self):
             ``graph``.
         """
         return _load_tsv_knowledge_graph(self)
+
+
+class IAEnronEmployees(
+    DatasetLoader,
+    name="IAEnronEmployees",
```
huonw: The website styles this dataset as 

```suggestion
    name="ia-enron-employees",
```

but I don't know if that's the canonical name by which people usually refer to it.

stellargraph/datasets/datasets.py
kjun9: For now, this dataset is only used for the CTDNE demo, which needs to preprocess the edgelist further to create the graph and train/test examples. I noticed other dataset loaders actually do this also in the `load` method, but I thought it might be good to show how the edges are split in the notebook. What do you think?

stellargraph/datasets/datasets.py
kjun9: It's due to the fact that we use the gensim `Word2Vec` model in the demo, which requires string IDs. I can do the conversion in the notebook instead if that seems better

stellargraph/datasets/datasets.py
kjun9: Network repository seems to only describe the graph properties of the data, I couldn't find explanation for any of the columns, and there aren't any column names in the dataset. Maybe it was a column relevant for a larger version of the dataset? But as you say, it always seems to be 1 here.

stellargraph/datasets/dataset_loader.py
kjun9: The parameters that precede it don't have default values though, whereas I think this one probably makes sense to have a default value of `True`, so it can't be moved in front of them

kjun9: I can still move where we refer to it in `IAEnronEmployees` class to keep the `url` related things together though, since all parameters are referred to via keyword arguments

huonw: Ah, interesting. We could potentially make all the arguments keyword-only, which would allow one with a default to be before one without, e.g.: `def f(x, *, y=1, z): pass` is accepted. That is,

```python
 def __init_subclass__(
    self,
    *,
    name: str,
    ...
```

stellargraph/datasets/datasets.py
huonw: Yeah, it's probably relevant to show the splitting details in the notebook, and I don't think we have sufficient functionality on `StellarGraph` to do that splitting directly. Maybe we could just add comments for the various bits and pieces here, like the string conversion below and that fact that it doesn't return a `StellarGraph` object, so we don't completely forget the reasoning.

Another option for this would be to return `(graph, edges)` but basically ignore the graph (other than `graph.nodes()`) in the current notebook. This might be good because it's more future compatible: switching `load` to returning `(graph, edges)` (or `(nodes, edges, graph)` or something) would be a breaking change (theoretically), but starting with `(graph, edges)` and converting notebook to use `graph` more would once the `StellarGraph` APIs are available would not be. (That said, I imagine not too many users would use this in their own code.)

stellargraph/datasets/dataset_loader.py
kjun9: okay I can do that here 👍

stellargraph/datasets/datasets.py
kjun9: Returning `(graph, edges)` sounds good. I'll add a comment about the string IDs.

Issue #866, opened 2020-02-13
Poor formatting in several places in documentation


huonw:
### Describe the bug

Various places in our docs https://stellargraph.readthedocs.io/en/latest/api.html have poor formatting:

- [x] [`SampledBreadthFirstWalk.run`](https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.data.SampledBreadthFirstWalk.run) "with replacement ..."
   ![image](https://user-images.githubusercontent.com/1203825/74402045-3aa22100-4e77-11ea-9371-232dd4bccc2f.png)
- [x] [`TemporalRandomWalk.run`](https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.data.TemporalRandomWalk.run) `initial_edge_bias` and `walk_bias`
  ![image](https://user-images.githubusercontent.com/1203825/74402098-62918480-4e77-11ea-9896-1055cf08f240.png)
- [x] [`EdgeSplitter`](https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.data.EdgeSplitter). All the parameter types are specified as `<...>`, which Sphinx doesn't pick up and format like the `(...)` ones elsewhere, e.g.
  ![image](https://user-images.githubusercontent.com/1203825/74402181-9a003100-4e77-11ea-9780-ec6448f2077d.png)
- [x] [`load_dataset_BlogCatalog3`](https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.data.load_dataset_BlogCatalog3) `location`'s type:
  ![image](https://user-images.githubusercontent.com/1203825/74402219-b2704b80-4e77-11ea-9e61-ca53eb1768e3.png)
- [x] [`FullBatchNodeGenerator`](https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.mapper.FullBatchNodeGenerator),  [`FullBatchLinkGenerator`](https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.mapper.FullBatchLinkGenerator) [`ClusterNodeGenerator`](https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.mapper.ClusterNodeGenerator). "For more information ..." has ended up as a "definition list" (`<dl>` in HTML)
  ![image](https://user-images.githubusercontent.com/1203825/74402257-cc119300-4e77-11ea-9abb-3db813dca170.png)
[...]

PR #1061, opened 2020-03-11
Fix formatting in explorer classes

kjun9:
This fixes poorly formatted bullet points in two explorer classes.

Now it looks like
## sampled breadth first walk
image

## temporal walk
image

Compared to before - see #866

### DIFF
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -507,12 +507,13 @@ class SampledBreadthFirstWalk(GraphWalk):
         Performs a sampled breadth-first walk starting from the root nodes.
 
         Args:
-            nodes (list): A list of root node ids such that from each node n BFWs will be generated up to the
+            nodes (list): A list of root node ids such that from each node a BFWs will be generated up to the
                 given depth d.
-            n_size (int): The number of neighbouring nodes to expand at each depth of the walk. Sampling of
-            n (int, default 1): Number of walks per node id. Neighbours with replacement is always used
-                regardless of the node degree and number of neighbours requested.
-            seed (int, optional): Random number generator seed; default is None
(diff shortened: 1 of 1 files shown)


stellargraph/data/explorer.py
```diff
@@ -818,13 +818,16 @@ def run(
                 than or equal to the context window size.
             initial_edge_bias (str, optional): distribution to use when choosing a random
```
habiba-h: Not that it makes any meaningful difference  but just for consistency's sake :-)

The descriptions of `num_cw (int)`, `cw_size (int)`, and `max_walk_length (int)`  start with  capital letters whereas, `initial_edge_bias (str, optional)` and `walk_bias (str, optional)` start with small letters. Also, full stops missing at a couple of places.

habiba-h:
@kjun9  other than the couple of suggestions above for some minor changes, the description for the `TemporalRandomWalk` looks succinct.

kjun9:
I might just fix both explorer classes here, will be ready again in a minute!

stellargraph/data/explorer.py
kjun9: Thanks, will update

stellargraph/data/explorer.py
```diff
@@ -507,13 +507,13 @@ def run(self, nodes, n_size, n=1, seed=None):
         Performs a sampled breadth-first walk starting from the root nodes.
 
         Args:
-            nodes (list): A list of root node ids such that from each node n BFWs will be generated up to the
```
habiba-h: there is no `d` depth parameter. How is the depth specified?

kjun9: The depth is inferred by the length of the list `n_size` - I'll clarify the docs

Issue #1039, opened 2020-03-08
New StellarGraph: remove use of networkx from Cora link prediction examples


huonw:
### Description

In #1032, `EdgeSplitter` is updated to support the new `StellarGraph` class which allowed removing the use of `networkx` from some link prediction notebooks. It is not used in `demos/link-prediction/gcn/cora-gcn-links-example.ipynb` or `demos/link-prediction/graphsage/cora-links-example.ipynb` because, unlike other demos that use the Cora dataset, these use the node subject as a feature (rather than as a target), and so the `StellarGraph` returned by `Cora().load()` cannot be used directly and needs to be augmented, which is done with NetworkX at the moment. 

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1062, opened 2020-03-11
Add subject_as_feature=True flag to Cora.load

huonw:
This adds a flag to include the paper subject as one-hot encoded feature elements when loading the Cora dataset into a graph: `Cora().load(subject_as_feature=True)`. This can then be used in the link predictions demos.

See: #1039


huonw:
That's true, the change here was just reproducing the logic from the notebook, which was included when it was first added (https://github.com/stellargraph/stellargraph/pull/543). I'll try running the notebook without that, and see if the results change significantly. If not, we can potentially not add this extra complexity.

huonw:
The accuracy seems to drop from consistently above 0.7 (usually around 0.75) to consistently below for the GCN example, so I'll leave it as it was with the subject included.

Issue #1040, opened 2020-03-09
New StellarGraph: remove use of NetworkX from weighted Node2vec notebook


huonw:
### Description

The `demos/node-classification/node2vec/stellargraph-node2vec-weighted-random-walks.ipynb` notebook has to add synthetic weights to the Cora graph. This isn't currently easy with the new StellarGraph and so it is done via NetworkX. We should switch it to use the new StellarGraph only.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1063, opened 2020-03-11
Add the ability to compute edge weights in Cora for weighted node2vec

huonw:
This adds yet more arguments to `Cora.load`: this time, an `edge_weights` function that can be used to compute edge weights from the source and target node information. This can then be used in the weighted node2vec demo, which computes the jaccard distance between those features.

See: #1040


stellargraph/datasets/datasets.py
```diff
@@ -102,13 +110,22 @@ def load(self, directed=False, largest_connected_component_only=False):
             directed (bool): if True, return a directed graph, otherwise return an undirected one.
             largest_connected_component_only (bool): if True, returns only the largest connected
                 component, not the whole graph.
+            edge_weights (callable, optional): a function that accepts two Pandas DataFrames; the
```
kjun9: Did you consider having `edge_weights` be a function that just takes two numbers and returns a number instead (i.e. edge => weight), but went with this approach because it's likely to be more efficient?

huonw: As in it takes the node IDs of the edge endpoints and returns a weight? I don't think that works the best with how this function works, because the caller doesn't have access to the node data directly; they'd have to separately load the CSV file containing the node info. I suppose another way would be to pass in the node data DataFrame.

In any case, yes, the bulk operations will be more efficient. Even just getting the information for the source vertex of each edge is significantly faster with a bulk operation:

```python
import pandas as pd
import numpy as np
import os

cites = pd.read_csv(os.path.expanduser("~/stellargraph-datasets/cora/cora.cites"), sep="\t", header=None, names=["target", "source"])
feature_names = ["w_{}".format(ii) for ii in range(1433)]
subject = "subject"
column_names = feature_names + [subject]
content = pd.read_csv(os.path.expanduser("~/stellargraph-datasets/cora/cora.content"), sep="\t", header=None, names=column_names)

%time a = content.loc[cites.source]
# CPU times: user 14.4 ms, sys: 910 µs, total: 15.3 ms
# Wall time: 14.4 ms

%time a = [content.loc[x] for x in cites.source]
# CPU times: user 4.84 s, sys: 97.9 ms, total: 4.93 s
# Wall time: 4.91 s

# numpy slicing with manual index lookup
%time content_numpy = content.iloc[:-1, :].to_numpy(); a = [content_numpy[content.index.get_loc(x), :] for x in cites.source]
# CPU times: user 64.5 ms, sys: 6.27 ms, total: 70.7 ms
# Wall time: 69.2 ms
```

That is, the bulk look-up is 350&times; faster than the simple individual look-ups, and the actual computation on the result is likely to be much faster too, since there's no Python looping overhead.

kjun9: > caller doesn't have access to the node data directly

Ah yep, sorry I think I didn't quite read properly how it was used.. Looking at the notebook, so I guess to write the `jaccard_weights` function, it requires knowing what the raw data looks like, whereas I feel like the dataset loaders are trying to abstract that all away? Maybe an alternative would be like `edge_weights(graph, source_ids, target_ids)` or even `edge_weights(graph)` (but maybe only relevant if we expect any users to use the dataset loaders and define their own edge weight scheme). What do you think?

huonw: > it requires knowing what the raw data looks like, whereas I feel like the dataset loaders are trying to abstract that all away?

I don't think they're quite trying to abstract it away, just make it more convenient, but yeah, if there's something we can do that is a less leaky abstraction, that's probably better.

> Maybe an alternative would be like edge_weights(graph, source_ids, target_ids)

That's not a bad idea, however, it means we'll need to create two `StellarGraph`s: one without edge weights and then one with the edge weights filled in, since we don't currently support "mutations" (neither in-place or not). I'll try it with `edge_weights(graph, node_subjects, source_ids, target_ids)` and see if it's worth it. 👍 

> even edge_weights(graph)

That would be convenient, but I think this would be relatively hard to use in a bulk way, to get edge weights that match the edges.

> but maybe only relevant if we expect any users to use the dataset loaders and define their own edge weight scheme

My thinking is that this is relatively rare; I think we should be optimising these datasets to power our own demos, rather than making sure they're completely general-purpose (most data scientists will be interested in their own data, rather than random public ones).

huonw: > see if it's worth it

It looks good. The approximate time to run goes from maybe ~900ms with the old form to ~950ms with one that builds an unweighted StellarGraph followed by a weighted one.

kjun9:
Not sure what's causing it, but the plots and accuracy scores on the notebook don't look correct?

huonw:
Ah, good catch. I suspect I haven't rerun it with the proper parameters (I'd reduced them for development, since it takes so long). Rerunning it now...

huonw:
Hm, I just reran it, and it gave the same poor results... I'll try re-run it on develop, to be sure that it's something wrong with this PR.

huonw:
Running on develop gives good results, so it is a problem with this PR.

huonw:
Hm, the plots are broadly similar, except for the labels. I wonder if the labels are just getting mixed up somehow :thinking:

kjun9:
> Hm, the plots are broadly similar, except for the labels. I wonder if the labels are just getting mixed up somehow 🤔

maybe due to `subjects` now being a dataframe with an index, instead of a list?

huonw:
It's a mismatch between the order of nodes in the `Word2Vec` model and the one in the `StellarGraph`, so it just requires doing a reindexing.

PR #1070, opened 2020-03-12
Fix edge splitter docstrings

kjun9:
This fixes poorly formatted docstrings in the EdgeSplitter class

Part of #866 

Link: https://stellargraph--1070.org.readthedocs.build/en/1070/api.html?highlight=edgesplitter#stellargraph.data.EdgeSplitter

### DIFF
--- a/stellargraph/data/edge_splitter.py
+++ b/stellargraph/data/edge_splitter.py
@@ -53,8 +53,8 @@ class EdgeSplitter(object):
     (which is not calculated in this case).
 
     Args:
-        g: <StellarGraph or networkx object> The graph to sample edges from.
-        g_master: <StellarGraph or networkx object> The graph representing the original dataset and a superset of the
+        g (StellarGraph or networkx object): The graph to sample edges from.
+        g_master (StellarGraph or networkx object): The graph representing the original dataset and a superset of the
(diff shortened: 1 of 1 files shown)


stellargraph/data/edge_splitter.py
```diff
@@ -480,11 +480,11 @@ def _reduce_graph_by_edge_type_and_attribute(
...
-            minedges: <list> Spanning tree edges that cannot be removed.
-            p: <float> Factor by which to reduce the size of the graph.
-            edge_label: <str> The edge type to consider.
-            edge_attribute_label: <str> The edge attribute to consider.
-            edge_attribute_threshold: <str> The threshold value; only edges with attribute value larger than the
+            minedges (list): Spanning tree edges that cannot be removed.
+            p (float): Factor by which to reduce the size of the graph.
+            edge_label (str): The edge type to consider.
+            edge_attribute_label (str): The edge attribute to consider.
+            edge_attribute_threshold (str): The threshold value; only edges with attribute value larger than the
```
kjun9: Although these are "optional", they're also not, since an error is thrown if they're passed in with the default value of `None` so I chose to not document as optional.

PR #1071, opened 2020-03-12
Fix docstrings using "note::"

kjun9:
Part of #866 

Links:

* graphsage (see blue "Note" box) https://stellargraph--1071.org.readthedocs.build/en/1071/api.html#stellargraph.layer.graphsage.GraphSAGE
* hinsage (see blue "Note" box) https://stellargraph--1071.org.readthedocs.build/en/1071/api.html#stellargraph.layer.hinsage.HinSAGE

Note that these "Note" boxes are used to list a few more additional optional parameters for specific situations - an alternative approach would be to keep these parameters in the normal list of parameters, and the "Note" box just mentions these parameters to say that they are only used for specific situations - I can switch to this if that seems better


habiba-h:
@kjun9 , 

> Note that these "Note" boxes are used to list a few more additional optional parameters for specific situations - an alternative approach would be to keep these parameters in the normal list of parameters, and the "Note" box just mentions these parameters to say that they are only used for specific situations - I can switch to this if that seems better

I would vote for this. The parameters should be all together in one list. In the Note should me mentioned when they are needed. 

All this: 

> If a generator is not specified, then additional keyword arguments must be supplied:
> n_samples (list): The number of samples per layer in the model.
> input_dim (int): The dimensions of the node features used as input to the model.
> multiplicity (int): The number of nodes to process at a time. This is 1 for a node inference and 2 for link inference (currently no others are supported).

can just be a continuation of the list.  Right now, these alternative parameters don't seem like part of the parameter list.

Issue #828, opened 2020-02-10
Add more unit tests for temporal walks


kjun9:
### Description

We need to flesh out more of the unit tests for temporal walks to ensure the implementation is correct and necessarily checks are being performed on the input graph and parameters.

### User Story

**As** a data scientist

**I want to** use temporal walks without running into problems

**So that** I can do my work

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1072, opened 2020-03-12
Validate times when constructing TemporalRandomWalk

kjun9:
This adds validation of edge weights (time) for temporal random walks.

The construction of edges and times to sample from I've moved to the constructor, since I thought it made sense to do the time validation when the user creates a `TemporalRandomWalk` object, instead of running this validation every time `run` gets called.

Part of #828


tests/data/test_temporal_random_walker.py
```diff
@@ -17,21 +17,32 @@
...
+@pytest.fixture()
+def temporal_graph():
+    nodes = [1, 2, 3, 4, 5, 6]
+    edges = [(1, 2, 5), (2, 3, 2), (2, 4, 10), (4, 5, 3), (4, 6, 12)]
+    return create_temporal_graph(nodes, edges)
+
+
+def temporal_graph_negative_times(num_edges):
+    nodes = [1, 2, 3, 4, 5, 6]
+    edges = [
```
kieranricardo: could this be something like:

```suggestion
    edges = np.vstack([
        np.random.choice(nodes, size=num_edges), 
        np.random.choice(nodes, size=num_edges), 
        -np.ones(num_edges),
    ]) 
```

kjun9: Thanks! Actually I think this should also work
```python
edges = np.hstack([np.random.choice(nodes, size=(num_edges, 2)), -np.ones((5,1))])
```

stellargraph/data/explorer.py
```diff
@@ -796,6 +796,29 @@ class TemporalRandomWalk(GraphWalk):
     for the algorithm, only the relative differences (e.g. seconds, days, etc).
     """
 
+    def __init__(self, graph, graph_schema=None, seed=None):
+        super().__init__(graph, graph_schema=graph_schema, seed=seed)
+        self._edges, self._times = self.graph.edges(include_edge_weight=True)
+        self._validate_times()
```
huonw: Instead of constructing the object and then validating, meaning it may be in an invalid state, what do you think about validating first?

```python
edges, times = self.graph.edges(include_edge_weight=True)

(neg_time_locs,) = np.where(times < 0)
if len(neg_time_locs) > 0:
    ...

# all good
self._edges = edges
self._times = times
```

Totally minor, though.

stellargraph/data/explorer.py
```diff
@@ -796,6 +796,29 @@ class TemporalRandomWalk(GraphWalk):
...
+            neg_time_edges_formatted = ", ".join(
+                [
+                    str((self._edges[loc], self._times[loc]))
+                    for loc in neg_time_locs[:max_edges_shown]
+                ]
+            )
+            if num_neg_times > max_edges_shown:
+                neg_time_edges_formatted += " ..."
+            raise ValueError(
+                f"All edge times must be non-negative. Found {num_neg_times} negatives: "
```
huonw: This message is good, but it doesn't quite match the style we're kinda sorta using elsewhere (at least, I and sometimes @kieranricardo do), maybe it could be:

```
graph: expected all edge types to be non-negative, found ...
```

In particular this tells the user which parameter the problem is with. Although maybe that's obvious in this case and so is pointless?

tests/data/test_temporal_random_walker.py
```diff
@@ -81,3 +92,14 @@ def test_exp_biases_extreme(temporal_graph):
...
     assert sum(biases) == pytest.approx(1)
+
+
+@pytest.mark.parametrize("num_edges", [1, 10, 20])
+def test_validate_times(num_edges):
+    with pytest.raises(ValueError, match=r".*edge times must be non-negative.*") as err:
+        TemporalRandomWalk(temporal_graph_negative_times(num_edges))
+
+    err_msg = err.value.args[0]
+    assert str(num_edges) in err_msg
```
huonw: As alternative to this manual checking, we could dynamically construct the appropriate regex:

```python
trailing = r"\.\.\." if num_edges > 10 else ""
matcher = fr".*edge times must be non-negative.*{num_edges}.*{trailing}$"

with pytest.raises(ValueError, match=matcher): 
    ...
```

tests/data/test_temporal_random_walker.py
```diff
@@ -17,19 +17,30 @@
...
-    edge_cols = ["source", "target", "weight"]
-    return StellarGraph(
-        nodes=pd.DataFrame(index=nodes), edges=pd.DataFrame(edges, columns=edge_cols),
+    return create_temporal_graph(nodes, edges)
+
+
+def temporal_graph_negative_times(num_edges):
+    nodes = [1, 2, 3, 4, 5, 6]
+    edges = np.hstack(
+        [np.random.choice(nodes, size=(num_edges, 2)), -np.ones((num_edges, 1))]
```
huonw: Is it worth having some non-negative ones here, to ensure we don't accidentally refactor to something like `np.all(times < 0)` instead of `np.any(times < 0)`/`np.where(times < 0)`? E.g. 

```suggestion
        [np.random.choice(nodes, size=(num_edges, 2)), -np.repeat([-1, 1], num_edges / 2)]
```

Plus adjusting `1` to `2` in the test parameterisation, and the `num_edges > 10` threshold too.

tests/data/test_temporal_random_walker.py
kjun9: Nice 👍

stellargraph/data/explorer.py
kjun9: Ah yeah that style was what I was going for, so I'll switch to that

stellargraph/data/explorer.py
kjun9: Is your suggestion based more on readability or do you mean that if we assign `self._edges` and `self._times` first before running validation, it's somehow possible for the object to be instantiated in an invalid state?

tests/data/test_temporal_random_walker.py
kieranricardo: nice!

tests/data/test_temporal_random_walker.py
kjun9: I might go with `temporal_graph_negative_times(num_neg_edges)` then this becomes
```python
[np.random.choice(nodes, size(num_neg_edges * 2, 2)), -np.repeat([-1, 1], num_neg_edges)]
```

stellargraph/data/explorer.py
huonw: A bit of both, but I guess this class is simple enough that the invalid state probably isn't observable (I think the `self` value would have to have been saved somewhere externally accessible inside `__init__`, and then after the validation error is thrown, the saved instance accessed and used: I don't think this class does that saving anywhere, and thus we shouldn't worry).

tests/data/test_temporal_random_walker.py
huonw: That's better :+1: (also, I noticed that my suggestion and your replacement accidentally retain the `-` in front of the `np.repeat`, which is probably unnecessary)

stellargraph/data/explorer.py
kjun9: Hmm I see, I hadn't considered that! And didn't realise that would result in that externally accessible thing actually pointing to the semi-constructed object.. I've taken your suggestion just to be safe

stellargraph/data/explorer.py
```diff
@@ -796,6 +797,25 @@ class TemporalRandomWalk(GraphWalk):
...
+
+        # validate edge times
+        (neg_time_locs,) = np.where(times < 0)
+        num_neg_times = len(neg_time_locs)
+        if num_neg_times > 0:
+            neg_time_edges_formatted = comma_sep(
+                neg_time_locs, stringify=lambda loc: str((edges[loc], times[loc])),
+            )
+            raise ValueError(
+                f"graph: expected edge times to be non-negative, found {num_neg_times}: "
```
```suggestion
                f"graph: expected edge times to be non-negative, found {num_neg_times} negative times: "
```

tests/data/test_temporal_random_walker.py
```diff
@@ -17,19 +17,33 @@
...
-        nodes=pd.DataFrame(index=nodes), edges=pd.DataFrame(edges, columns=edge_cols),
+    return create_temporal_graph(nodes, edges)
+
+
+def temporal_graph_negative_times(num_neg_edges):
+    nodes = [1, 2, 3, 4, 5, 6]
+    edges = np.hstack(
+        [
+            np.random.choice(nodes, size=(num_neg_edges * 2, 2)),
+            np.repeat([[-1], [1]], num_neg_edges, axis=0),
```
huonw: Totally minor and potentially not an improvement, but this could be:

```suggestion
            np.repeat([-1, 1], num_neg_edges)[:, np.newaxis],
```

if you were so inclined.

kjun9: I like that it avoids defining the nested lists so I'll use your suggestion

huonw:
> I guess I should really just be validating that they are numerical instead? I had incorrectly assumed that weights are already forced to be numerical upon constructing a stellargraph

Uhh, I'd (implicitly) assumed this too. Maybe we should validate it when constructing the `StellarGraph` instead?

kjun9:
> Maybe we should validate it when constructing the StellarGraph instead?

Yeah I guess they should actually always be numerical since they're supposed to be used to create adjacency matrices, etc. I can close this and do that instead if we agree that's a good idea

kjun9:
Closing this in favour of #1118

PR #1077, opened 2020-03-13
Test against tensorflow 2.2 release candidate 0

huonw:
Validating whether we're likely to be broken by the upcoming release.


huonw:
It does break: #1078

Issue #1060, opened 2020-03-11
ComplEx/DistMult: reproduce the 'filtered' metrics from the paper


huonw:
### Description

In #901, we add a notebook for ComplEx that reproduces the "raw" metrics (MRR, Hits at 1, Hits at 3 and Hits at 10), but not the "filtered" metrics. We've been unable to reproduce those ones, seeing only fractional improvements over the raw numbers (e.g. 0.571 -> 0.578 for MRR on WN18, vs. the paper's 0.587 -> 0.941).

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #930, opened 2020-02-24
Add CHANGELOG entries for new work since 0.9

huonw:
This is created by going over the list of changes since 0.9 and added changelog entries for the interesting ones. It also adds a table with a basic comparison of the memory use and construction time of the new `StellarGraph` object.

PR #1080, opened 2020-03-13
Compute filtered ranks for evaluating ComplEx

huonw:
This adds the final piece of the evaluation required to make `ComplEx` non-`@experimental`. it extends the ranking procedure performed in #901 to also compute the "filtered" ranks. This gives the rest of the metrics in Table 2 of the ComplEx paper (http://jmlr.org/proceedings/papers/v48/trouillon16.pdf):

image

As a reminder from #901, the knowledge graph link prediction metrics for a test edge `E = (s, r, o)` connecting nodes `s` and `o` are calculated by ranking the prediction for that edge against all modified-source `E' = (n, r, o)` and modified-object `E'' = (s, r, n)` edges (for all nodes `n` in the graph). The "raw" ranks are just the rank of `E` against the `E'` and against the `E''`.

The "filtered" ranks exclude the modified edges `E'` and `E''` that are known, i.e. are in the train, validation or test sets. For instance, if `E = (A, x, B)` has score 1, but the modified edges `(A, x, C)` and `(A, x, D)` have scores 1.3 and 1.5 respectively, `E` has raw modified-object rank 3. If `(A, x, D)` is in the train set (or validation or test) but `(A, x, C)` is not, it is excluded from the filtered ranking, and so `E` has filtered modified-object rank 2.

This has been a struggle to implement correctly, because it has been difficult to correctly use the right node nodes in the right place of the ranking procedure. For modified-object ranking, with `E` and `E''` as above, calculating the score of `E` in the column of scores of every modified-object edge `E''` needs to use `o`, but calculating the known edges similar to `E` needs to use `(s, r, _)`, not `(o, r, _)` (the latter is meaningless). (And similarly for modified-subject ranking.) It sounds obvious when written out like this, but it's somewhat difficult to keep track of which entity needs to go where in practice. (@kieranricardo had this key insight.)

The implementation works by: start with the raw `greater` matrix, where each column represents the test edges `E` with a row for every node in the graph (i.e. row `n` represents swapping node `n` into the subject or object) and the elements of a column are `True` if the score of that modified edge is greater than the score of `E`. For each edge/column, compute the indices of the similar known edges and set those indices to false, leaving only unknown edges with scores greater than `E`.

See: #1060


stellargraph/layer/knowledge_graph.py
```diff
@@ -175,27 +175,93 @@ def rank_edges_against_all_nodes(model, test_data, known_edges_graph):
...
             raw_rank = 1 + greater.sum(axis=0)
-            assert raw_rank.shape == (batch_size,)
 
-            return raw_rank
+            # FIXME(#870): this would be better without external IDs <-> ilocs translation
+            true_nodes = known_edges_graph._nodes.ids.from_iloc(true_node_ilocs)
+            true_rels = known_edges_graph._edges.types.from_iloc(true_rel_ilocs)
+            for i, (node_id, rel) in enumerate(zip(true_nodes, true_rels)):
+                if modified_object:
+                    neighs = known_edges_graph.out_nodes(node_id, edge_types=[rel])
```
kieranricardo: previously, if `modified_object` (or `true_is_source`) was `True`, I think it was the in edges being calculated:

```
real_edges = [
    (s, i)
    for i, (obj, rel) in enumerate(zip(true_ilocs, rels))
    for s, o in edges_by_type[rel] if o == obj
]
```
because we have `o`, and we're finding `s->o`. but... I think its a bug here that we're passing the `object` indices, 

When `modified_object` is True:

- We *should* be passing the `subject` indices and finding the out edges
- My code was passing the `object` indices and finding the in edges
- Your code is passing the `object` indices and finding the out edges

If that makes sense?

stellargraph/layer/knowledge_graph.py
```diff
@@ -175,27 +172,66 @@ def rank_edges_against_all_nodes(model, test_data, known_edges_graph):
 
         num_nodes = known_edges_graph.number_of_nodes()
 
-        def ranks(pred, true_ilocs):
-            batch_size = len(true_ilocs)
+        def ranks(
```
kieranricardo: Could you add a docstring for `ranks`?

Also I think `ranks` can pulled out of `ComplEx` now - it doesn't seem have anything `ComplEx` specific and I think the only outer scope variables it uses are `knowledge_graph` and `num_nodes`.

stellargraph/layer/knowledge_graph.py
```diff
@@ -175,27 +172,66 @@ def rank_edges_against_all_nodes(model, test_data, known_edges_graph):
...
+            pred,
+            *,
```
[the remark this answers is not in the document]
huonw: https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html#keyword-only-arguments

stellargraph/layer/knowledge_graph.py
huonw: I'm generally resistant about adding too much documentation for scoped and private functions, because maintaining documentation is error-prone, and it just drifts. However, I will move it out of `ComplEx` to `_ranks_from_score_columns` (or something) and give it a doc string.

huonw:
> The FB15K numbers still seem a little low but I don't think this is a blocking concern seeing as the WN18 results are matching - any idea whats going on?

Unclear, I tried with a few different parameters (e.g. increasing the embedding dimension from 150 to 200), and tweaking the learning rate, and they made a bit of difference (e.g. I kept the dimension at 200), but I got bored of waiting to do this hyperparameter optimisation work. If it's deemed important, I can put more effort into it/write a script, or even just switch to the slower Adagrad optimiser (one potential problem is our regularisation doesn't _quite_ match theirs, because Keras's `BinaryCrossentropy` takes the mean, but the paper and https://github.com/ttrouill/complex doesn't, so one needs to scale the L2 normalisation lambda appropriately).

> Also not a blocking concern, just an idea for future work. I think users with powerful a CPU and/or GPU might notice a slowdown when using rank_edges_against_all_nodes because the numpy ops are single threaded. We might be able to speed this up using tensorflow ops/layer/model for the bulk ops. What do you think?

That's not a bad idea, but I'd prefer to do it as future work, to keep this PR focused on the filtered-ranks-computation.

stellargraph/layer/knowledge_graph.py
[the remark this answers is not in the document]
kieranricardo: cool 👍

kieranricardo:
@huonw oh one more thing, you should update the changelog because `ComplEx` is a new non-experimental algorithm 🎉

huonw:
I haven't been bothering to update the CHANGELOG patch-by-patch because I find it just results in merge conflicts. Since I'm "slacking" off in this respect, I've instead made sure to go over the git log before each release like https://github.com/stellargraph/stellargraph/pull/930.

PantelisElinas:
@huonw just letting you know that I won't be able to review this until next week. So, if you are satisfied with the one review, then go ahead and merge.

P.

huonw:
Thanks for letting me know 👍 

I'll merge this with just @kieranricardo's review.

Issue #725, opened 2020-01-24
Tidy up github landing page


PantelisElinas:
### Description

Our github landing page has grown over the last few months as we have added more algorithms and features. As it stands, it is difficult for new users to find important information, e.g., the link to our forum or introductory material to get started with graph ML and StellarGraph.

We need to re-structure our landing page for usability. I'm not sure what all these changes should be but it might be a good idea to consult with our comms team.

### User Story

**As** a StellarGraph user

**I want to** easily locate the most important information about the StellarGraph library

**So that** I can make full use of the available resources, demos, tutorials, and developers.

### Done Checklist

- [ ] Landing page content updated
- [ ] CHANGELOG.md updated

Issue #992, opened 2020-03-03
Add "get started" basic tutorial notebook(s) to teach users how to create a StellarGraph and do basic ML workflow.


timpitman:
### Description

Split from #910 - Our current set of demo notebooks are good to show particular algorithms, but don't teach the basics of how to create a StellarGraph from a dataset on disk, then use it for a basic ML workflow.

We can create two (or possibly one?) tutorial notebooks that show:
1. how to load/preprocess data to build a StellarGraph
    * can be based on existing notebooks, but with a focus on documenting the c ore workflow, rather than specific algorithm or results.
    * need to cover homogeneous and heterogeneous data sets. (possibly CORA for homog; could include our own CSV for hetero. ) 

~~2. the very basics of the machine learning workflow (something very simple like GCN node classification)
    * important now that are examples are using the `datasets` module to return a `StelalrGraph` object rather than explicit code in the notebook.~~

### User Story

**As a new user to StellarGraph, I want a basic tutorial showing me how to load data and construct a StellarGraph, ~~then construct the basic machine learning workflow~~ so that I can understand the library and use it on my own datasets**

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
I've split https://github.com/stellargraph/stellargraph/issues/1087 off to cover how we surface these (and other) notebooks, leaving this one to just create the notebooks.

PR #1081, opened 2020-03-15
Rewrite the top-level README to be more focused

huonw:
This tweaks our top-level README to try to be a bit more focused and helpful:

- rewrite some of the text to be a bit more relevant to data scientists, e.g. reading a bit more naturally by being _slightly_ less precise/exact, and trying to draw the connection to their real data and problems
- providing a concrete workflow to get started in two forms:
  - a Getting Started section: this includes providing basic Installation instructions, in addition to the detailed ones (users familiar with their environment and Python tooling can start with one of the basic ones, while the expanded section provides more guidance for those who need it)
  - a code example (GCN), that tries to emphasise how much of the workflow just builds on the normal ML/TF workflow, and finishes with links to the full GCN demo as well as the other examples
- providing a "quick links" for getting help, as well as making sure that section can be found easily with a search (it mentions "help", "support", "contact us")
- removing/moving some of the content that's more internal/developer focused

This doesn't touch the algorithms or references sections, since these are somewhat standalone, are good concepts, and can be tweaked as independent work. This also doesn't touch any other READMEs, such as those for the demos.

There's potentially many more tweaks we could do. Let me know if you think any are good.

Rendered form: https://github.com/stellargraph/stellargraph/blob/feature/725-readme/README.md

See: #725


README.md
```diff
@@ -38,90 +38,133 @@
...
+from sklearn import model_selection
+
+#### Data preparation ####
+
+# Use Pandas to load and preprocess the data
+nodes = pd.read_csv("nodes.csv") # columns of node data
+edges = pd.read_csv("edges.csv") # "source" & "target" columns for each link between nodes
+
+node_classes = pd.read_csv("labels.csv") # a column of each node's class/label
+targets = pd.get_dummies(node_classes) # one-hot encoding
```
kjun9: Thoughts on making this smaller?
```suggestion
# load data into pandas dataframes
nodes, edges, targets = load_my_data()
```

README.md
```diff
@@ -38,90 +38,133 @@
...
-  using the GraphSAGE [3] algorithm given node features and training labels.
-  - See the demo in folder `demos/node-classification-node2vec` for an example of how to predict attributes of nodes
-  using the Node2Vec [1] algorithm for nodes without features, unsupervised node representation learning, and
-  supervised classifier training for the downstream task.
-  - See the demo in folder `demos/node-classification-hinsage` for examples of how to predict attributes of nodes
-  using the HinSAGE algorithm for given node features and training labels.
+# two layers of GCN, each with hidden dimension 16
+gcn = sg.layer.GCN(
+    layer_sizes=[16, 16], activations=["relu", "relu"], generator=generator, dropout=0.5
+)
```
kjun9: Thoughts on removing some of the optional parameters here? I think `relu` should be default already
```suggestion
gcn = sg.layer.GCN(layer_sizes=[16, 16], generator=generator)
```

CONTRIBUTING.md
```diff
@@ -66,6 +66,10 @@ StellarGraph considers courtesy and respect for others an essential part of the
 
 9. Create a pull request on GitHub from your branch to `develop` branch. If you have already discussed the new features on GitHub with the developers and they are aware of what the pull request contains, then the developers will endeavour to approve the pull request promptly.
 
+#### Continuous Integration (CI)
+
+A pull request can only be merged if it passes tests. StellarGraph uses Buildkite CI to let a computero do the testing, and results are reported directly in each pull request. The full Buildkite pipeline can be viewed at <https://buildkite.com/stellar/stellargraph-public/>.
```
kjun9:
```suggestion
A pull request can only be merged if it passes tests. StellarGraph uses Buildkite CI to let a computer do the testing, and results are reported directly in each pull request. The full Buildkite pipeline can be viewed at <https://buildkite.com/stellar/stellargraph-public/>.
```

README.md
kjun9: Or even
```suggestion
x_inp, x_out = sg.layer.GCN(layer_sizes=[16, 16], generator=generator).build()
```

README.md
```diff
@@ -38,90 +38,133 @@
...
-  - See the demo in folder `demos/link-prediction-hinsage` for an example of how to predict
-  movie ratings between users and movies using a Heterogeneous generalisation of GraphSAGE model, which we call HinSAGE.
+#### Training and prediction ####
 
--->
+# Use TensorFlow Keras to create a model and work with it  (e.g. training and evaluation)
+model = tf.keras.Model(inputs=x_inp, outputs=predictions)
+model.compile("adam", loss="categorical_crossentropy", metrics=["accuracy"])
 
+model.fit(train_gen, epochs=5)
```
kjun9: Thoughts on calling `generator.flow` inline here, similar to `evaluate`?
```suggestion
model.fit(generator.flow(train_targets.index, train_targets), epochs=5)
```
Seems like would be nice if we supported `generator.flow(train_targets)` using pandas dataframe index by default actually

README.md
PantelisElinas: Why not simplify by using one of the existing datasets in `sg.datasets`?

README.md
```diff
@@ -38,90 +38,133 @@
...
    * [References](#references)
 
 ## Introduction
-**StellarGraph** is a Python library for machine learning on graph-structured (or equivalently, network-structured) data.
 
-Graph-structured data represent entities, e.g., people, as nodes (or equivalently, vertices),
-and relationships between entities, e.g., friendship, as links (or
-equivalently, edges). Nodes and links may have associated attributes such as age, income, and time when
-a friendship was established, etc. StellarGraph supports analysis of both homogeneous networks (with nodes and links of one type) and heterogeneous networks (with more than one type of nodes and/or links).
+Graph-structured data represent entities as nodes (aka vertices) and relationships between them as edges (aka links), along with associated data as attributes. For example, a graph can contain people as nodes and friendships between them as links, with data like a person's age and the date a friendship was established. StellarGraph supports analysis of both homogeneous networks (with nodes and links of one type) and heterogeneous networks (with more than one type of nodes and/or links).
```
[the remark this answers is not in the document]
huonw: Good idea.

README.md
huonw: I was trying to have this be more of a StellarGraph workflow taster, than showing of GCN in particular. In particular, I was trying to make it easy for a data scientist to see how their data fits into this, by showing that the "load the data" step is just normal Pandas. The built-in datasets package up all of that data prep to keep our algorithm demos focused on the algorithms, but my motivation with this was a bit broader.

I think I'll go with Kevin's suggestion, although tweaking the comment slightly.

README.md
huonw: I like it, although I think I'd prefer to have `build` be separate.

huonw:
I've updated this with a few changes:

- added @kjun9's suggestions
- tweaked the comments/exact code from those suggestions very slightly
- expanded the "StellarGraph supports analysis of ..." sentence into a list, covering more than just the homogeneous vs. heterogeneous split
- moved the "The StellarGraph library offers state-of-the-art algorithms ..." section to be first in the Introduction, since it's the important bit, and the discussion of what graph data is is more clarification (I could easily be convinced this isn't an improvement, and some other approach is better :smile:)
- added more links, including to some of our blog posts

README.md
```diff
@@ -38,90 +38,132 @@
 </p>
 
 
-# Table of Contents
+**StellarGraph** is a Python library for machine learning on [graphs and networks](https://en.wikipedia.org/wiki/Graph_\(discrete_mathematics\)).
```
timpitman: This is a great concise summary, however the line of text gets a bit lost between the badges and contents.

README.md
```diff
@@ -38,90 +38,132 @@
...
-The StellarGraph library implements several state-of-the-art algorithms for applying machine learning methods to discover patterns and answer questions using graph-structured data.
-
-The StellarGraph library can be used to solve tasks using graph-structured data, such as:
 - Representation learning for nodes and edges, to be used for visualisation and various downstream machine learning tasks;
 - Classification and attribute inference of nodes or edges;
 - Link prediction;
-- Interpretation of node classification through calculated importances of edges and neighbour nodes for selected target nodes [8].
-
-We provide [examples](https://github.com/stellargraph/stellargraph/tree/master/demos/) of using `StellarGraph` to solve such tasks using several real-world datasets.
+- Interpretation of node classification through [calculated importances of edges and neighbour nodes for selected target nodes](https://medium.com/stellargraph/https-medium-com-stellargraph-saliency-maps-for-graph-machine-learning-5cca536974da) [8].
```
timpitman: This feels like too much jargon for an introduction?

README.md
```diff
@@ -38,90 +38,132 @@
...
-and relationships between entities, e.g., friendship, as links (or
-equivalently, edges). Nodes and links may have associated attributes such as age, income, and time when
-a friendship was established, etc. StellarGraph supports analysis of both homogeneous networks (with nodes and links of one type) and heterogeneous networks (with more than one type of nodes and/or links).
+The StellarGraph library offers state-of-the-art algorithms for [graph machine learning](https://medium.com/stellargraph/knowing-your-neighbours-machine-learning-on-graphs-9b7c3d0d5896), making it easy to discover patterns and answer questions about graph-structured data. It can solve many machine learning tasks:
 
-The StellarGraph library implements several state-of-the-art algorithms for applying machine learning methods to discover patterns and answer questions using graph-structured data.
-
-The StellarGraph library can be used to solve tasks using graph-structured data, such as:
 - Representation learning for nodes and edges, to be used for visualisation and various downstream machine learning tasks;
 - Classification and attribute inference of nodes or edges;
```
timpitman: should we pre-empt the addition of graph classification here?

README.md
```diff
@@ -38,90 +38,132 @@
...
+| Relational Graph Convolutional Network [11] | The RGCN algorithm performs semi-supervised learning for node representation and node classification on knowledge graphs. RGCN extends GCN to directed graphs with multiple edge types and works with both sparse and dense adjacency matrices.|
+| ComplEx[12] | The ComplEx algorithm computes embeddings for nodes (entities) and edge types (relations) in knowledge graphs, and can use these for link prediction |
+| GraphWave [13] | GraphWave calculates unsupervised structural embeddings via wavelet diffusion through the graph. |
+| Watch Your Step [14] | The Watch Your Step algorithm computes node embeddings by using adjacency powers to simulate expected random walks. |
+| Deep Graph Infomax [15] | Deep Graph Infomax trains unsupervised GNNs to maximize the shared information between node level and graph level features. |
 
 ## Installation
+
 StellarGraph is a Python 3 library and we recommend using Python version `3.6.*`. The required Python version
```
timpitman: we could just say 3.6?

README.md
```diff
@@ -38,90 +38,132 @@
...
+StellarGraph is built on [TensorFlow 2](https://tensorflow.org/) and its [Keras high-level API](https://www.tensorflow.org/guide/keras), as well as [Pandas](https://pandas.pydata.org) and [NumPy](https://www.numpy.org). It is thus user-friendly, modular and extensible. It interoperates smoothly with code that builds on these, such as the standard Keras layers and [scikit-learn](http://scikit-learn.github.io/stable), so it is easy to augment the core graph machine learning algorithms provided by StellarGraph.
 
 ## Getting Started
 
-To get started with StellarGraph you'll need data structured as a homogeneous or heterogeneous graph, including
-attributes for the entities represented as graph nodes.
-[NetworkX](https://networkx.github.io/) is used to represent the graph and [Pandas](https://pandas.pydata.org/)
-or [Numpy](https://www.numpy.org/) are used to store node attributes.
+[The numerous detailed and narrated examples](https://github.com/stellargraph/stellargraph/tree/master/demos/) are a good way to get started with StellarGraph. There is likely to be one that is similar to your data or your problem (if not, [let us know](#getting-help)).
```
timpitman: At some point in future, we should expand this to link to some 'tutorial' content - the readme page for the demos doesn't tell people where to start; it's just an index.  We could even link to our machine learning on graphs course?

timpitman: > we should expand this to link to some 'tutorial' content

covered by #992 actually

README.md
huonw: I'll try moving the badges above the `# StellarGraph Machine Learning Library` title, and see if that works better.

README.md
huonw: I'm in two minds, one is that the graph classification works should land soon and so we can just short cut to including, and the other is that people arriving after this lands (if it lands first) to our front page will see this but no evidence of how to do so. Similarly for the dynamic/temporal graph work.

README.md
huonw: I'll remove everything from "through ..." onwards (but keep the link).

README.md
huonw: I think we should either make the `demos` landing page have guidance, to avoid making this document too huge, or switch to doing it all on read-the-docs, which case this link would switch to point there.

README.md
timpitman: Hopefully people will be curious and raise an issue or discourse thread if they can't find it?!

README.md
```diff
@@ -38,91 +36,135 @@
...
+If you get stuck or have a problem, there's many ways to make progress and get help or support:
+
+- [Read the documentation](https://stellargraph.readthedocs.io)
+- [Consult the examples](https://github.com/stellargraph/stellargraph/tree/master/demos/)
+- Contact us:
+  - [Ask questions and discuss problems on the StellarGraph Discourse forum](https://community.stellargraph.io)
+  - [File an issue](https://github.com/stellargraph/stellargraph/issues/new/choose)
+
+
+## Example: GCN
```
habiba-h: I think the GCN example is unnecessary here.

README.md
```diff
@@ -38,91 +36,135 @@
...
-The StellarGraph library implements several state-of-the-art algorithms for applying machine learning methods to discover patterns and answer questions using graph-structured data.
-
-The StellarGraph library can be used to solve tasks using graph-structured data, such as:
 - Representation learning for nodes and edges, to be used for visualisation and various downstream machine learning tasks;
 - Classification and attribute inference of nodes or edges;
 - Link prediction;
-- Interpretation of node classification through calculated importances of edges and neighbour nodes for selected target nodes [8].
-
-We provide [examples](https://github.com/stellargraph/stellargraph/tree/master/demos/) of using `StellarGraph` to solve such tasks using several real-world datasets.
+- [Interpretation of node classification](https://medium.com/stellargraph/https-medium-com-stellargraph-saliency-maps-for-graph-machine-learning-5cca536974da) [8].
```
habiba-h: Having only 1 bullet point with a link looks awkward to me.

README.md
huonw: Could you expand on why you think so?

I added the example because it's widely recommended (e.g. https://medium.com/@meakaakka/a-beginners-guide-to-writing-a-kickass-readme-7ac01da88ab3 https://www.makeareadme.com) and it lets people get a sense for the library and how it uses Pandas and Keras and so on at a glance, before they're committed and digging through the demos.

README.md
huonw: That's a fair concern. I've added a link to https://medium.com/stellargraph/can-graph-machine-learning-identify-hate-speech-in-online-social-networks-58e3b80c9f7e, and once we have graph classification and #1090, we can add a link to that one too.

README.md
habiba-h: It looks abrupt to me. The readme is meant for a general introduction. An example, that also, a specific one looks out of place. Pointing to the detailed GCN demo, as you did later, in my opinion would have been enough.  Also, from the link you refer above, it says to give a concise example. The GCN is this form would confuse and turn me away more than making it simpler for me, or I would just skip this section, but then again it could be just me, go with the widely recommended choice.

huonw: I don't think pointing to the GCN demo is quite enough: at least at the moment, https://github.com/stellargraph/stellargraph/blob/master/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb is too specific to serve as a good high-level example, focusing on the "big picture" view, since it has to explain every step of the process. My goal with this example is to give that big picture view of how StellarGraph fits into the eco-system, in particular emphasising that connects smoothly to Pandas and TensorFlow/Keras, rather than being a fundamentally new ML paradigm to learn.

It's definitely true that we should be concerned about it being confusing and intimidating, so I've restructured it to interleave real headings and text with the code. Do you think this is better?

README.md
huonw: #1053 has merged, so I've added it here!

README.md
```diff
@@ -38,91 +36,147 @@
...
+graph = sg.StellarGraph(nodes, edges)
+
+generator = sg.mapper.FullBatchNodeGenerator(graph, method="gcn")
+
+# two layers of GCN, each with hidden dimension 16
+gcn = sg.layer.GCN(layer_sizes=[16, 16], generator=generator)
+x_inp, x_out = gcn.build() # create the input and output TensorFlow tensors
+
+# use TensorFlow Keras to add a layer to compute the (one-hot) predictions
+predictions = tf.keras.layers.Dense(units=len(ground_truth_targets.columns), activation="softmax")(x_out)
```
habiba-h: The `import tensorflow as tf` comes later. This line should come after that.

A block that states all the steps about building the model should be together. 
Also,  I think model building part with the prediction layer should come together. 
Since `model = tf.keras.Model(inputs=x_inp, outputs=predictions)` needs `x_input` and `predictions` so it would be less confusing if they are in one place together.

README.md
```diff
@@ -38,91 +36,147 @@
...
+
+#### Training and evaluation
+
+These input and output tensors can be used to create a TensorFlow Keras model and work with it such as training and evaluation with the functions offered by Keras. StellarGraph's data generators make it simple to construct the required Keras Sequences for input data.
+
+```python
+import tensorflow as tf
+
+model = tf.keras.Model(inputs=x_inp, outputs=predictions)
+model.compile("adam", loss="categorical_crossentropy", metrics=["accuracy"])
```
habiba-h: Compile step with its parameters could maybe seperate.

README.md
```diff
@@ -38,91 +36,147 @@
...
+
+model.fit(generator.flow(train_targets.index, train_targets), epochs=5)
```
habiba-h: Training and Evaluation section.

README.md
```diff
@@ -38,91 +36,147 @@
...
+
+generator = sg.mapper.FullBatchNodeGenerator(graph, method="gcn")
+
+# two layers of GCN, each with hidden dimension 16
+gcn = sg.layer.GCN(layer_sizes=[16, 16], generator=generator)
+x_inp, x_out = gcn.build() # create the input and output TensorFlow tensors
+
+# use TensorFlow Keras to add a layer to compute the (one-hot) predictions
+predictions = tf.keras.layers.Dense(units=len(ground_truth_targets.columns), activation="softmax")(x_out)
+```
```
habiba-h: The way you  have short comments for all the steps above, it would be nice to have the same for these following steps.

README.md
habiba-h: The example looks more readable! Nice! 
Added some further comments incase you find them helpful.

README.md
```diff
@@ -38,91 +36,147 @@
 </p>
 
 
-# Table of Contents
+# StellarGraph Machine Learning Library
+
+**StellarGraph** is a Python library for machine learning on [graphs and networks](https://en.wikipedia.org/wiki/Graph_\(discrete_mathematics\)).
+
+## Table of Contents
```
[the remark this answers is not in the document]
huonw: > Algorithms short teaser of all the methods implemented in the library

I put this towards the end because I felt it's mainly helpful for someone familiar with graphs and graph ML. For instance, someone who is looking for a specific algorithm. They might either come here through a google search for that algorithm name or searching for graph ML libraries, and then either ctrl-F for the specific algorithm or jump to the section.

For someone unfamiliar with graph ML (or, at least, unfamiliar with the specific algorithms), I don't think the current table of algorithms is very easily digested, and so I'm concerned readers will get distracted before wading through it all.

> Installation the structure right now tells a user how to install with demo before installing the library. I think the basic library installation should be upfront. The demo are the optional bit that if a user is interested in testing the functionality of the library before they can install with the demos.

I put this later because it's a bit of a "reference" section. If a user has decided to install StellarGraph, they can easily jump to the relevant section; if they haven't decided to install it, it's mostly distraction. For the latter, one thing it is useful for might be to emphasise how easy it is to install, so I've added:

```markdown
It is thus also [easy to install with `pip` or Anaconda](#Installation).
```

to the end of the last paragraph of the introduction.

> Sample demos here you point out to the demos directory.
> 4.1 Example The detailed example should follow right after the general introduction to the sample demos. (the example looks much more user friendly to me now :-)). I have some other comments that I will add to the example.

To be specific, you think it's better to have a "Sample demos" section than a "Getting started" one? I feel that the "getting started" phrasing is more obvious as a place for users to look: it is talking about what they want to do (start using StellarGraph), rather than the route we're suggesting they use (look at demos).

> Getting Help _I think the help is for all the things above, installation, understanding of approaches, demos etc. so should come at the end of everything. _

As touched upon my some of my earlier comments, I don't see this as a linear document because there's some sections that many uses will not look at (because they don't need to, for their particular use-case/background, e.g. if `pip install stellargraph` is fine for a user, they don't need to look at the Anaconda or docker installation instructions). In addition, I think we should be emphasising the getting help section: we are not getting too many issues at the moment, so we should be happy to help anyone with any problem related to StellarGraph. It's better for someone to ask a question that has already been answered (but they didn't find) than to not ask and give up using StellarGraph.

README.md
huonw: @timpitman thoughts?

README.md
```diff
@@ -38,91 +36,147 @@
...
 
-* Node classification and regression
-  - See the demo in folder `demos/node-classification-graphsage` for an example of how to predict attributes of nodes
-  using the GraphSAGE [3] algorithm given node features and training labels.
-  - See the demo in folder `demos/node-classification-node2vec` for an example of how to predict attributes of nodes
-  using the Node2Vec [1] algorithm for nodes without features, unsupervised node representation learning, and
-  supervised classifier training for the downstream task.
-  - See the demo in folder `demos/node-classification-hinsage` for examples of how to predict attributes of nodes
-  using the HinSAGE algorithm for given node features and training labels.
+## Getting Help
```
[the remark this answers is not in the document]
huonw: (I replied to this on your comment on the table of contents)

README.md
huonw: I've  given it a comment and so separated it slightly from the `model = ...` line; is that what you're looking for?

README.md
huonw: I don't know what this means?

README.md
timpitman: I think it looks good, thanks!

README.md
timpitman: I like the idea of improving the demos landing page (ie README in the demos folder).  That could even optionally be embedded in our readthedocs. documentation.

README.md
```diff
@@ -38,91 +36,147 @@
...
-- Interpretation of node classification through calculated importances of edges and neighbour nodes for selected target nodes [8].
-
-We provide [examples](https://github.com/stellargraph/stellargraph/tree/master/demos/) of using `StellarGraph` to solve such tasks using several real-world datasets.
+- [Interpretation of node classification](https://medium.com/stellargraph/https-medium-com-stellargraph-saliency-maps-for-graph-machine-learning-5cca536974da) [8].
 
+Graph-structured data represent entities as nodes (or vertices) and relationships between them as edges (or links), along with associated data as attributes. For example, a graph can contain people as nodes and friendships between them as links, with data like a person's age and the date a friendship was established. StellarGraph supports analysis of many kinds of graphs:
 
-## Guiding Principles
+- homogeneous (with nodes and links of one type),
+- heterogeneous (with more than one type of nodes and/or links)
```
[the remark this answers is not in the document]
habiba-h: I don't have a confident opinion about the style but having only bullet point with a comma at the end bothered me :-). 
I usually do commas and the comma and when having bullets that are not starting as standalone sentences. I don't know when I started doing that but held on to that convention. I am down with any popular style guide :-).

README.md
habiba-h: In the attached screenshot, the model definition comes as a first line of the 2nd block. To me it would look more appropriate it this was not cut-off from the 3 last lines of the previous block. 
Like, the model is taking as input from the layers defined above. So they all seem to be a complete thing together. 

Screen Shot 2020-03-23 at 5 54 17 pm

README.md
```diff
@@ -38,91 +36,147 @@
...
-  movie ratings between users and movies using a Heterogeneous generalisation of GraphSAGE model, which we call HinSAGE.
+This is the only part that is specific to StellarGraph. The machine learning model consists of some graph convolution layers followed by a layer to compute the actual predictions as a TensorFlow tensor. StellarGraph makes it easy to construct all of these layers via the `GCN` model class. It also makes it easy to get input data in the right format via the `StellarGraph` graph data type and a data generator.
 
--->
+```python
+import stellargraph as sg
 
+# convert the raw data into StellarGraph's graph format for faster operations
+graph = sg.StellarGraph(nodes, edges)
+
```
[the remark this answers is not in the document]
habiba-h: > * three has a reputation as being a good number for writing https://en.wikipedia.org/wiki/Rule_of_three_(writing)

I didn't know about the rule of three. Learnt something new :-D.

README.md
habiba-h: Sorry the comments weren't linked at the right place. I meant was the the 

> Training and Evaluation

can start with the `compile` part. And the `model` line could be part of previous block. See the previous comment about that.

README.md
huonw: Sounds good, I'll move the `model = ` line and tweak the text a bit.

README.md
huonw: Ah, cool. It's now like this.

huonw:
Thanks for the reviews everyone, I think it's now much better than my first version! 😄

Issue #819, opened 2020-02-09
Some notebooks out of memory when running on CI


timpitman:
### Describe the bug

Some notebooks can't be run on CI (with a t2large) as they run out of memory:

```
stellargraph-metapath2vec.ipynb
node-link-importance-demo-gat.ipynb
node-link-importance-demo-gcn-sparse.ipynb
```

huonw:
It seems the RGCN notebook fails too, e.g. https://buildkite.com/stellar/stellargraph-public/builds/1880#1b11f48a-f2b1-4faf-aba7-e8d9de6769cd/372-588


```
ResourceExhaustedError:  OOM when allocating tensor with shape[8285,8285] and type float on /job:localhost/replica:0/task:0/device:CPU:0 by allocator cpu
```

huonw:
#1744 enables the interpretability notebooks on CI, so all that remains is the RGCN notebook `demos/node-classification/rgcn-node-classification.ipynb`:

https://github.com/stellargraph/stellargraph/blob/70d99525f1d1635215adc4a46c9bed3fa2dba759/.github/workflows/ci.yml#L159

PR #1085, opened 2020-03-17
Feature/gcn lstm

habiba-h:
**GCN-LSTM** is stack of graph convolution and lstm layers  used for time series prediction on spatio-temporal data. 
 We use a 2-layer graph convolution network to leverage the graph structure. The output of the graph convolution network is fed into an LSTM based sequence to sequence model, that is jointly trained on the output of gcn and the historical speeds of a segment.

This work is inspired by the paper: [T-GCN: A Temporal Graph Convolutional Network for Traffic Prediction](https://ieeexplore.ieee.org/document/8809901).

The authors have made available the implementation of their model in their [github repo](https://github.com/lehaifeng/T-GCN). 

There has been a few differences in the architecture proposed in the paper and the implementation with regards to the graph convolution component, such as: 

-[https://github.com/lehaifeng/T-GCN/issues/18]( https://github.com/lehaifeng/T-GCN/issues/18)

-[https://github.com/lehaifeng/T-GCN/issues/14](https://github.com/lehaifeng/T-GCN/issues/14)

**Architecture:** 

- We use the 2-layer graph convolution proposed by  [Kipf and Welling (ICLR2017)](https://arxiv.org/abs/1609.02907). 

- The output of gcn feeds in an LSTM layer that combines gcn output with the time-series history to learn a combined spatio-temporal end to end model.

- A dense and dropout layer is further added  as it helps improve the model performance.  Note, these final two layers are not part of the architecture proposed in the TGCN paper. 

#### Reviewer Checklist
- [ ] The code for gcn-lstm class is available.
- [ ] The  gcn-lstm class implements the architecture explained above.
- [ ]  Demo notebook runs and clearly demonstrates how to use GCN-LSTM for spatio-temporal data.


demos/spatio-temporal/input_data.py
```diff
@@ -0,0 +1,58 @@
...
+
+@author: 
+"""
+
+import numpy as np
+import pandas as pd
+import pickle as pkl
+
+def load_sz_data(dataset):
+    sz_adj = pd.read_csv(r'data/sz_adj.csv',header=None)
```
huonw: What's the `sz` data?

demos/spatio-temporal/input_data.py
```diff
@@ -0,0 +1,58 @@
...
+    adj = np.mat(sz_adj)
+    sz_tf = pd.read_csv(r'data/sz_speed.csv')
+    return sz_tf, adj
+
+def load_los_data(dataset):
+    los_adj = pd.read_csv(r'data/los_adj.csv',header=None)
```
huonw: What's the `los` data? It would be best if we could add it to `stellargraph/datasets/datasets.py` instead of having an extra script for the demo. If it ends up being too awkward to add it there, feel free to open an issue for follow up work.

stellargraph/layer/gcn_lstm.py
```diff
@@ -0,0 +1,347 @@
...
+    Model,
+)
+from tensorflow.keras.layers import Input, Layer, Lambda, Dropout, Reshape, LSTM, Dense
+
+from ..mapper import FullBatchGenerator
+from .misc import SqueezedSparseConversion
+from .preprocessing_layer import GraphPreProcessingLayer
+
+
+class GraphConvolution(Layer):
```
huonw: As far as I can tell, this differs to the existing `gcn.GraphConvolution` layer in four ways:

1. the adjacency matrix is a constant
2. it doesn't squeeze out the batch dimension
3. it does some transposes in the matrix multiplications
4. it doesn't have special handling for `final_layer=True`

Is this accurate? Could you expand on why 1 and 3 in particular are needed?

stellargraph/layer/__init__.py
```diff
@@ -36,3 +36,4 @@
 from .rgcn import *
 from .watch_your_step import *
 from .knowledge_graph import *
+from .gcn_lstm import *
```
huonw: Both `gcn` and `gcn_lstm` now include a class called `GraphConvolution`, which means these `*` imports will shadow each other as the `stellargraph.layer.GraphConvolution` name. I think in particular the `gcn_lstm` one will "win" so existing code importing `stellargraph.layer.GraphConvolution` will switch from using GCN one to using GCN-LSTM one. They're not interchangeable so that code will break.

One way to fix this would be to rename the `gcn_lstm` `GraphConvolution` to something like `FixedAdjacencyGraphConvolution` or even just `LSTMGraphConvolution` so that it no longer overlaps with the `gcn` one.

stellargraph/layer/gcn_lstm.py
```diff
@@ -0,0 +1,347 @@
...
+        # Add optional bias & apply activation
+        if self.bias is not None:
+            output += self.bias
+
+        output = self.activation(output)
+
+        return output
+
+
+class Graph_Convolution_LSTM(Model):
```
huonw: Our other "models" aren't actually formal models, meaning they don't inherit from `tf.keras.Model`. I think it would be good to stick to that.

Also, Python style is to not have underscores in class names, e.g.

```suggestion
class GraphConvolutionLSTM(Model):
```

but potentially it's best to break the "style" slightly and call it:

```suggestion
class GCN_LSTM(Model):
```

stellargraph/layer/gcn_lstm.py
```diff
@@ -0,0 +1,347 @@
...
+        self.bias_regularizer = regularizers.get(bias_regularizer)
+        self.bias_constraint = constraints.get(bias_constraint)
+
+        self.gcn_1 = GraphConvolution(units = layer_sizes[0], A=adj, activation=self.activations[0])
+        self.gcn_2 = GraphConvolution(units = layer_sizes[0], A=adj, activation=self.activations[1])
+        self.lstm = LSTM(self.layer_sizes[1], return_sequences=False)
+        self.dense = Dense(self.outputs, activation=self.activations[2])
+        self.dropout = Dropout(self.dropout)
+
+    def call(self, inputs):
```
huonw: Echoing the comment about the name, our "model" classes usually have two methods:

```python
def __call__(self, inputs): # like this `call`, but note the double-underscore, to make it work like `some_model(inputs)`
    ...

def build(self): # builds appropriate input tensors and then applies the model
    inputs = ...
    outputs = self(inputs)
    return inputs, outputs
```

demos/spatio-temporal/input_data.py
habiba-h: This is the alternative dataset of trajectories of taxis. I had it as alternative but not using it. I will remove it.

demos/spatio-temporal/input_data.py
habiba-h: Yes, it can be easily added to the `stellargraph/datasets/datasets.py`. 
 There are a number of other methods that I think of as data related such as formatting the data in a time series to be fed into a time series forecasting model. So having it all in one utility file where anyone interested in just doing spatio-temporal stuff can look them, instead of looking them up in a centralized  utility file.  This was just for convenience. I can merge it in the `stellargraph/datasets/datasets.py`.

stellargraph/layer/gcn_lstm.py
habiba-h: > As far as I can tell, this differs to the existing `gcn.GraphConvolution` layer in four ways:
> 
> 1. the adjacency matrix is a constant

Yes,  it is. There are two pragmatic reasons for that but neither are binding. 
1. The temporal data  means structure is fixed but features are changing. So its from performance perspective better to just load it at initialization.  However, we can make it more general for dynamic graphs and also to make it more aligned with the already existing `graph_convolution` layer. 
2. The graph convolution happens for each timestep of each observation. For example, for a  sequence of 10 timesteps, gcn is performed for each one of them with the same static graph but different time (feature), so seems like a big overhead of passing the adjacency matrix in each build. 
I am sure there is a more optimal way of doing this. This was just meant to produce a working version.

> 2. it doesn't squeeze out the batch dimension

Yes and this is by design. `Keras` require the batch dimension whereas `GCN` doesn't so even though a unit  batch dimension is passed it gets squeezed inside the layer. However, in the `gcn_lstm`, the data is passed as a 3D tensor `[batch, sequence, nodes]`, don't need to squeeze the batch dimension. 
Maybe we can add the not squeezing as an option to the existing `graph_convolution` layer but didn't want to update it as a first pass until I have the actual dimensions I need to `gcn_lstm`  figured out.  This can be merged in the more polished version. 

> 3. it does some transposes in the matrix multiplications

Yes,  due to the dimension. Typically, you have nodes times features dimensions. Here we have batch times features times nodes dimension. Since graph convolution happens on the feature dimension, so i transpose it before convolving on the feature dimension and then flip it back. 

> 4. it doesn't have special handling for `final_layer=True`

Since here `graph_convolution` is stacked over the RNN layer, i.e. it feeds into the RNN so its never a `final_layer`, so its never true. I can pass it as false when I define the architecture, so this can easily be handled with the `graph_convolution` layer we already have. I am just keeping it false without any special handling. I'll add comment to explain this in the code. 

> Is this accurate? Could you expand on why 1 and 3 in particular are needed?

See above. Any thoughts on these design choices are really appreciated!

stellargraph/layer/__init__.py
habiba-h: I didn't think about it causing conflicts. I'll fix the name. 
However, why would the `gcn_lstm`'s `GraphConvolution` would win  in the import?

stellargraph/layer/gcn_lstm.py
huonw: > Yes, it is. There are two pragmatic reasons for that but neither are binding.

Makes sense; seems fine to start with it as written for now 👍 

> However, in the gcn_lstm, the data is passed as a 3D tensor [batch, sequence, nodes]

And `batch > 1` here?

> Yes, due to the dimension. Typically, you have nodes times features dimensions. Here we have batch times features times nodes dimension. Since graph convolution happens on the feature dimension, so i transpose it before convolving on the feature dimension and then flip it back.

Ah, right, our other GCN layers have input `[batch, nodes, features]`, makes sense!

> Since here graph_convolution is stacked over the RNN layer, i.e. it feeds into the RNN so its never a final_layer, so its never true. I can pass it as false when I define the architecture, so this can easily be handled with the graph_convolution layer we already have. I am just keeping it false without any special handling. I'll add comment to explain this in the code.

The other GCN layers are usually stacked into another dense layer to compute predictions (or similar), so they're not the "final layer" of the (whole) model either. The `final_layer` handling is so that the full-batch layer only yields predictions for the nodes that were passed to the `.flow` method, rather than all nodes.

Given that, this code currently seems to just yield predictions for all nodes. I think this means it's not possible to train on only a subset of labelled nodes? (For instance, do a train/test split across the nodes, rather than the times?) I guess this isn't too important for these sort of datasets, because it's time that is interesting and so it's assumed we'll always have information for every node?

stellargraph/layer/gcn_lstm.py
```diff
@@ -332,16 +281,69 @@ def __init__(
...
+
+        Returns:
+            tuple: `(x_inp, x_out)`, where `x_inp` is a list of Keras/TensorFlow
+            input tensors for the GCN model and `x_out` is a tensor of the GCN model output.
+        """
+        # Inputs for features
+        x_t = Input(batch_shape=(None, self.n_features, self.n_nodes))
+
+        # Indices to gather for model output
+        out_indices_t = Input(batch_shape=(None, self.n_nodes), dtype="int32")
```
kieranricardo, line 341: I think `out_indices_t` is unused here, so the model could be simplified to just have the node features`x_t` as input.

stellargraph/layer/gcn_lstm.py
```diff
@@ -332,16 +281,69 @@ def __init__(
...
+
+        x_inp = [x_t, out_indices_t]  # + A_placeholders
+        x_out = self(x_inp)
+
+        return x_inp[0], x_out
+
+    def _node_model(self):
```
kieranricardo, line 348: looks like `_node_model` is an unused hidden function that wraps `in_out_tensors` - I think this can be safely deleted

```suggestion
```

stellargraph/layer/gcn_lstm.py
```diff
@@ -332,16 +281,69 @@ def __init__(
...
+        # Activation function for each gcn layer
+        if lstm_activations is None:
+            lstm_activations = ["tanh"] * n_lstm_layers
+        elif len(lstm_activations) != n_lstm_layers:
+            raise ValueError(
+                "Invalid number of activations; require one function per lstm layer"
+            )
+        self.lstm_activations = lstm_activations
+
+        self.gcn_1 = FixedAdjacencyGraphConvolution(
```
kieranricardo, line 302: The `layer_sizes` doc string implies that the user can specify any number of gcn layers to use but it looks like `GraphConvolutionLSTM` is hard coded to use 2 layers. I think we should change this to handle any number of gcn layers.

```suggestion
        self.gcn_layers = [FixedAdjacencyGraphConvolution(units, self.adj, act) for units, act in zip(gc_layer_sizes, gc_activations)]
```

stellargraph/layer/gcn_lstm.py
```diff
@@ -332,16 +281,69 @@ def __init__(
...
+            units=self.gc_layer_sizes[0], A=self.adj, activation=self.gc_activations[0]
+        )
+        self.gcn_2 = FixedAdjacencyGraphConvolution(
+            units=self.gc_layer_sizes[1], A=self.adj, activation=self.gc_activations[1]
+        )
+        self.lstm = LSTM(
```
kieranricardo, line 308: Same as my comment above, the docstring implies the user can specify any number of `LSTM` layers but this is hard coded to only use 1. It would be cool if we can specify any number of `LSTM` units!

```suggestion
        self.lstm_layers = [
            LSTM(units, act, return_sequences=False) for units, act in zip(self.lstm_layer_sizes, self.lstm_activations)
        ]
```

stellargraph/layer/gcn_lstm.py
```diff
@@ -0,0 +1,349 @@
...
+            "kernel_constraint": constraints.serialize(self.kernel_constraint),
+            "bias_initializer": initializers.serialize(self.bias_initializer),
+            "bias_regularizer": regularizers.serialize(self.bias_regularizer),
+            "bias_constraint": constraints.serialize(self.bias_constraint),
+        }
+
+        base_config = super().get_config()
+        return {**base_config, **config}
+
+    def calculate_laplacian(self, adj):
```
kieranricardo, line 119: can we use something from our utils to calculate this? and if we can't it'd be great to have this in utils :)

stellargraph/layer/gcn_lstm.py
```diff
@@ -0,0 +1,349 @@
...
+           1. 2-layer graph convolutional network (gcn) proposed by Kipf & Welling in  https://arxiv.org/abs/1609.02907 (ICLR 2017).
+           2. 1 LSTM layer
+           3. 1 Dense layer
+           4. 1 Dropout layer
+           The last two layers consistently showed better performance and regularization experimentally.
+        Notes:
+          - The normalized Lapalacian matrix is provided as input to
+            Keras methods. 
+            
+        Args:
```
kieranricardo, line 229: It looks like this docstring args is missing a few things

stellargraph/layer/gcn_lstm.py
```diff
@@ -0,0 +1,349 @@
...
+            kernel_regularizer (str or func, optional): The regulariser to use for the weights of each layer.
+            kernel_constraint (str or func, optional): The constraint to use for the weights of each layer.
+            bias_initializer (str or func, optional): The initialiser to use for the bias of each layer.
+            bias_regularizer (str or func, optional): The regulariser to use for the bias of each layer.
+            bias_constraint (str or func, optional): The constraint to use for the bias of each layer.
+        """
+
+    def __init__(
+        self,
+        outputs,
```
kieranricardo, line 245: I think this argument is number of nodes? we can just derive this from `adj` so I think we can leave this out
```suggestion
```

stellargraph/datasets/datasets.py
```diff
@@ -803,3 +803,54 @@ def load(self):
...
+    name="METR-LA",
+    directory_name="METR-LA",
+    url="https://github.com/lehaifeng/T-GCN/tree/master/data",
+    url_archive_format="n/a",
+    expected_files=["los_speed.csv", "los_adj.csv"],
+    description="This traffic dataset contains traffic information collected from loop detectors in the highway of Los Angeles County (Jagadish et al., 2014).",
+    source="https://github.com/lehaifeng/T-GCN/tree/master/data",
+):
+    def load(self):
+        los_adj = pd.read_csv(r"data/los_adj.csv", header=None)
```
kieranricardo, line 819: To get this in a similar format to the rest of our dataset this should call `self.download()` and then load the dataset from `~/stellargraph-datasets`. Something like:

```
def load(self):
    self.download()
    paths = [self._resolve_path(f) for f in expected_files]
    los_adj = pd.read_csv(paths[1], header=None)
    adj = np.mat(los_adj)
    los_tf = pd.read_csv(paths[0])
```

huonw: I think `download()` won't do the right thing here, because `url` needs to point to a single file that contains the data (e.g. a zip file, or similar). Getting an appropriate zip file out of a GitHub repo is a bit fiddly: we've taken to just uploading it to figshare.com separately.

Given all that, and in the interests of getting this landed sooner rather than later, I think it would be appropriate to leave this class as is, but call it experimental:

```python
@experimental(reason="the data isn't downloaded automatically", issues=[9999])
class METR_LA(
```

where a follow-up issue is filed about this, and `9999` is replaced with its number.

(It's great to this parsing code all packaged up!)

stellargraph/datasets/datasets.py
```diff
@@ -849,3 +850,55 @@ def load(self):
         )
 
         return StellarGraph(nodes=nodes, edges=edges, edge_weight_column="time"), edges
+
+
+@experimental(reason="the data isn't downloaded automatically", issues=[9999])
```
huonw, line 855: The 9999 in my suggestion here and in `test-demo-notebooks.sh` was a placeholder that needs to be replaced by a real issue number. At the moment, this is creating a link like https://github.com/stellargraph/stellargraph/issues/9999, which doesn't currently exist. Could you open an issue about this and then replace the 9999 here and in `test-demo-notebooks.sh`?

habiba-h: Ah. Ok! I'll fix this.

stellargraph/layer/gcn_lstm.py
```diff
@@ -212,41 +195,41 @@ def call(self, features):
 class GraphConvolutionLSTM:
 
     """
-        A stack of 2 Graph Convolutional layers followed by an LSTM, Dropout and,  Dense layer.
+        A stack of N1 Graph Convolutional layers followed by N2 LSTM, Dropout and,  Dense layer.
```
kieranricardo, line 198: I think this might read a bit clearer:

```suggestion
        A stack of N1 Graph Convolutional layers followed by N2 LSTM layers, a Dropout layer, and a Dense layer.
```

stellargraph/layer/gcn_lstm.py
```diff
@@ -212,41 +195,41 @@ def call(self, features):
...
         The implementation of the above paper is based on one graph convolution layer stacked with a GRU layer.
-        The StellarGraph implementation is built as a stack of the following layers:
-           1. 2-layer graph convolutional network (gcn) proposed by Kipf & Welling in  https://arxiv.org/abs/1609.02907 (ICLR 2017).
-           2. 1 LSTM layer
+        The StellarGraph implementation is built as a stack of the following set of layers:
+           1. User specified no. of Graph Convolutional layers
+           2. User specified no. of LSTM layers
            3. 1 Dense layer
            4. 1 Dropout layer
            The last two layers consistently showed better performance and regularization experimentally.
```
kieranricardo, line 207: I think the docstring has changed a fair bit now, do you better performance compared to the implementation in the paper?

stellargraph/layer/gcn_lstm.py
```diff
@@ -212,41 +195,41 @@ def call(self, features):
...
-        Args:
-            layer_sizes (list of int): Output sizes of GCN layers in the stack.
-            bias (bool): If True, a bias vector is learnt for each layer in the GCN model.
-            dropout (float): Dropout rate applied to input features of each GCN layer.
-            activations (list of str or func): Activations applied to each layer's output;
+       Args:
+           seq_len: No. of LSTM cells
+           adj: unweighted/weighted adjacency matrix of [no.of nodes by no. of nodes dimension
+           gc_layers: No. of Graph Convolution  layers in the stack. The output of each layer is equal to sequence length.
+           lstm_layer_size (list of int): Output sizes of LSTM layers in the stack.
```
kieranricardo, line 212: small typoe
```suggestion
           lstm_layer_sizes (list of int): Output sizes of LSTM layers in the stack.
```

stellargraph/layer/gcn_lstm.py
```diff
@@ -47,6 +38,7 @@ class FixedAdjacencyGraphConvolution(Layer):
 
      Args:
         units (int): dimensionality of output feature vectors
+        A (N x N): weighted/unweighted adjacency matrix
         activation (str or func): nonlinear activation applied to layer's output to obtain output features
         use_bias (bool): toggles an optional bias
         final_layer (bool): If False the layer returns output for all nodes,
```
kieranricardo, line 44: I don't think there's a `final_layer` arg here

```suggestion
```

stellargraph/layer/gcn_lstm.py
habiba-h: Actually its very hard to compare with the paper. The paper reports the results different from what I get when I run their code. And their implementation is quite different from what they propose in the paper :-). 
But this implementation does better than the simple LSTM or Naive baseline.

.buildkite/steps/test-demo-notebooks.sh
```diff
@@ -33,10 +33,17 @@ f=${NOTEBOOKS[$INDEX]}
...
+    'node-link-importance-demo-gat.ipynb' | 'node-link-importance-demo-gcn.ipynb' | 'node-link-importance-demo-gcn-sparse.ipynb' | 'rgcn-aifb-node-classification-example.ipynb' | \
+    'stellargraph-metapath2vec.ipynb' | 'gcn-lstm-LA.ipynb')
     # These notebooks do not yet work on CI:
     # FIXME #818: datasets can't be downloaded
     # FIXME #819: out-of-memory
+    # FIXME #849: CI does not have neo4j
+    # FIXME #907: socialcomputing.asu.edu is down
+    # FIXME #1303: METR_LA dataset can't be downloaded automatically
+    # FIXME #818: datasets can't be downloaded
+    # FIXME #819: out-of-memory
```
huonw, line 45: This has added a whole pile more lines then it needs to: the #849 and #907 fixme's have been resurrected and the #818 and #819 ones have been duplicated
```suggestion
    # FIXME #1303: METR_LA dataset can't be downloaded automatically
```

stellargraph/datasets/datasets.py
huonw: Great! Reminder:

```suggestion
@experimental(reason="the data isn't downloaded automatically", issues=[1303])
```

Issue #1003, opened 2020-03-03
Write Unit Tests for deep graph Infomax


kieranricardo:
This issue is to write unit tests for deep graph Infomax to get it out of experimental.

PR #1086, opened 2020-03-17
Deep Graph Infomax unit tests

kieranricardo:
This PR implements more unit tests for Deep Graph Infomax to remove the experimental tag.

See #1003

Reviewer checklist:

- [ ] Deep Graph Infomax sufficiently tested
- [ ] experimental tags removed
- [ ] changelog updates


stellargraph/layer/deep_graph_infomax.py
```diff
@@ -114,7 +114,8 @@ def build(self):
 
         x_inp, node_feats = self.base_model.build(multiplicity=1)
         # identity layer so we can attach a name to the tensor
-        node_feats = Lambda(lambda x: x, name=self._unique_id)(node_feats)
+        rand_string = "".join(random.choice(ascii_lowercase) for _ in range(5))
```
huonw: What's this for?

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -148,8 +149,10 @@ def embedding_model(self, model):
         """
 
         try:
-            x_emb_out = model.get_layer(self._unique_id).output
-        except ValueError:
+            x_emb_out = [
+                layer for layer in model.layers if self._unique_id in layer.name
```
huonw:
```suggestion
                layer for layer in model.layers if layer.name.startswith(self._unique_id)
```

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -148,8 +149,10 @@ def embedding_model(self, model):
         """
 
         try:
-            x_emb_out = model.get_layer(self._unique_id).output
-        except ValueError:
+            x_emb_out = [
```
huonw: This could be written as `next(layer for layer ...).output` catching `StopIteration` (https://stackoverflow.com/a/8534381/1256624), rather than building the list and indexing it; if you were so inclined.

tests/layer/test_deep_graph_infomax.py
```diff
@@ -43,14 +47,37 @@ def test_dgi():
     assert embeddings.shape == (len(G.nodes()), emb_dim)
 
 
-def test_dgi_embedding_model_wrong_model():
+@pytest.mark.parametrize("model_type", [GCN, APPNP, GAT])
```
huonw: No PPNP here because it doesn't support sparsity?

kieranricardo: yep that's right.

kieranricardo:
> Is it worth also having a test for behaviour of calling `build` twice, to both ensure we've got (and keep) the behaviour we want and catch the mutation problem we noticed on #1001?
> 

I added this test in initially, unfortunately the mutation issue we noticed in #1001 is still here due to the way `GCN` stores the layers it builds in `self_layers`.

stellargraph/layer/deep_graph_infomax.py
kieranricardo: nice!

stellargraph/layer/deep_graph_infomax.py
kieranricardo: I noticed that `keras` gives every layer a unique name:

```
import tensorflow as tf

print(tf.keras.layers.Input(5).name, tf.keras.layers.Input(5).name)
```

 I think this is so that if you combine layers from different models no two layers will have the same name - which would throw an error. 

This is to try and mimic that behaviour, but at the moment its kinda pointless because of the mutation behaviour in GCN that I talk about below.

huonw:
> I added this test in initially, unfortunately the mutation issue we noticed in #1001 is still here due to the way GCN stores the layers it builds in self_layers.

Based on the discussion we had on slack, it turns out `DeepGraphInfomax` is incorrectly stateless #1088. If you wanted to make it (but only DGI) stateful here and make sure that aspect is tested, that'd be fine by me. But it'd also be fine if you do it separately.

kieranricardo:
> Based on the discussion we had on slack, it turns out `DeepGraphInfomax` is incorrectly stateless #1088. If you wanted to make it (but only DGI) stateful here and make sure that aspect is tested, that'd be fine by me. But it'd also be fine if you do it separately.
 
I not entirely sure but I think in this case `DeepGraphInfomax` inherits (wraps?) the statefulness of  the `base_model`. Either way I'll add in a statefulness test and fix `DeepGraphInfomax` if the test fails.

tests/reproducibility/test_deep_graph_infomax.py
```diff
@@ -0,0 +1,61 @@
...
+    corrupted_generator = CorruptedGenerator(generator)
+    gen = corrupted_generator.flow(G.nodes())
+
+    assert_reproducible(lambda: dgi(generator, gen, model_type), num_iter=3)
+
+
+@pytest.mark.parametrize("model_type", [GCN, APPNP, GAT, PPNP])
+def test_dgi_dense(model_type):
+    G = example_graph_random()
+    generator = FullBatchNodeGenerator(G, sparse=True)
```
huonw:
```suggestion
    generator = FullBatchNodeGenerator(G, sparse=False)
```

I think?

Alternatively, you could parameterise (using [skipping](http://doc.pytest.org/en/latest/skipping.html#skipping-test-functions)):

```python
@pytest.mark.parametrize("model_type", [GCN, APPNP, GAT, PPNP])
@pytest.mark.parametrize("sparse", [False, True])
def test_dgi(model_type, sparse):
    if sparse and model is PPNP:
        pytest.skip("PPNP doesn't support sparse=True")

    G = example_graph_random()
    generator = FullBatchNodeGenerator(G, sparse=sparse)
    ...
```

(We could do a similar thing to reduce the duplication in the main tests, too.)

Also, this test seems to be currently running with `sparse=True` and passing for `PPNP`; are we sure that it doesn't support sparseness?

tests/reproducibility/test_deep_graph_infomax.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.mark.parametrize("model_type", [GCN, APPNP, GAT])
+def test_dgi_sparse(model_type):
+    G = example_graph_random()
+    generator = FullBatchNodeGenerator(G, sparse=True)
+    corrupted_generator = CorruptedGenerator(generator)
+    gen = corrupted_generator.flow(G.nodes())
+
+    assert_reproducible(lambda: dgi(generator, gen, model_type), num_iter=3)
```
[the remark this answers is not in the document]
kieranricardo: just to make things a bit quicker

tests/reproducibility/test_deep_graph_infomax.py
kieranricardo: `PPNP()` won't throw and error if passed a generator with `Sparse=True` but the results would be meaningless (maybe it should crash though?). So I don't think its worth testing as this may fail in the future. fyi creating `FullBatchNodeGenerator(G, sparse=True, method="ppnp")` will throw an error

kieranricardo:
@huonw I forgot about the weights in the `Discriminator` layer! I'll make this layer a private variable.

tests/reproducibility/test_deep_graph_infomax.py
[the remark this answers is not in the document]
huonw: I think we should handle that by reducing the global `num_iter`, rather than tweaking it per-model.

tests/reproducibility/test_deep_graph_infomax.py
huonw: > PPNP() won't throw and error if passed a generator with Sparse=True but the results would be meaningless (maybe it should crash though?).

If it gives nonsense results, and we can catch it easily, yeah, we should. Should be a separate issue/PR though.

> (We could do a similar thing to reduce the duplication in the main tests, too.)

Did you try this?

kieranricardo: I'll make a seperate issue/PR
> Did you try this?
and yep, I put in there - much nicer! didn't realize you got stack pytest.param like that

huonw: > I'll make a seperate issue/PR

👍 

> and yep, I put in there - much nicer! didn't realize you got stack pytest.param like that

I saw that you put it into `tests/reproducibility/test_deep_graph_infomax.py`, that parenthetical and my quote was about the main tests in `tests/layer/test_deep_graph_infomax.py`.

kieranricardo: oh forgot to do it for `layer/test_deep_graph_infomax.py`  my bad, I've added it in now

PR #1092, opened 2020-03-17
Switch the `input_dim` argument to a real argument

huonw:
If we're going to support this argument, we should include it as a real argument. This PR switches it to a real argument (and documents it) in the classes for which it is the only `kwargs` that StellarGraph uses; that is, after this patch the `kwargs` parameter is not used for anything in these classes except for passing through to `tf.keras.Layer`.

See: #801


huonw:
Ah, I was a bit unclear: I've updated the PR description. I'm only updating the ones for which `input_dim` is the only thing in `kwargs` that StellarGraph uses; it can be changed and documented in those ones as part of converting all of those other `kwargs`.

Issue #1094, opened 2020-03-17
Cluster GCN hits SparseEfficiencyWarnings


huonw:
### Describe the bug

Cluster GCN does some operations on Scipy sparse matrices that are slow because they involve changing the sparsity structure. This is somewhat similar to #931 #932, where a 15&times; speedup was observed from resolving this warning.

### To Reproduce

In the first cell of the cluster GCN notebook https://github.com/stellargraph/stellargraph/blob/develop/demos/node-classification/cluster-gcn/cluster-gcn-node-classification.ipynb change `ignore` to `error`

```diff
- warnings.simplefilter("ignore", SparseEfficiencyWarning)
+ warnings.simplefilter("error", SparseEfficiencyWarning)
```

Or, run the Cluster-GCN tests, e.g. on CI: https://buildkite.com/stellar/stellargraph-public/builds/2369#e230600c-2c0d-4024-a837-213da5587d65/278-406

### Observed behavior

In the notebook, the `model.fit` cell fails:

```
SparseEfficiencyWarning                   Traceback (most recent call last)

...

~/projects/stellargraph/stellargraph-3/stellargraph/mapper/mini_batch_node_generators.py in __getitem__(self, index)
    287         if self.normalize_adj:
    288             # add self loops
--> 289             adj_cluster.setdiag(1)  # add self loops
    290             degree_matrix_diag = 1.0 / (adj_cluster.sum(axis=1) + 1)
    291             degree_matrix_diag = np.squeeze(np.asarray(degree_matrix_diag))

...

~/.pyenv/versions/sg3/lib/python3.6/site-packages/scipy/sparse/compressed.py in _set_many(self, i, j, x)
    899             warn("Changing the sparsity structure of a {}_matrix is expensive."
    900                  " lil_matrix is more efficient.".format(self.format),
--> 901                  SparseEfficiencyWarning, stacklevel=3)
    902             # replace where possible
    903             mask = offsets > -1

SparseEfficiencyWarning: Changing the sparsity structure of a csr_matrix is expensive. lil_matrix is more efficient.
```

In the tests

```
...
tests/mapper/test_cluster_gcn_node_mapper.py::test_ClusterNodeSquence
/usr/local/lib/python3.6/site-packages/scipy/sparse/_index.py:126: SparseEfficiencyWarning: Changing the sparsity structure of a csr_matrix is expensive. lil_matrix is more efficient.
 self._set_arrayXarray(i, j, x)
```

### Expected behavior

[...]

PR #1095, opened 2020-03-18
Optimise Cluster-GCN adjacency normalization

huonw:
This cuts the time for training in the demo notebook from ~60ms/step to ~30ms/step, and reduces the time for the included benchmark of the time to iterate over a whole epoch (times in milliseconds):

|  q |    code |  min | median |  max |
|---:|--------:|-----:|-------:|-----:|
|  1 | this PR | 18.3 |   19.2 | 23.5 |
|    | develop | 38.7 |   40.6 | 50.8 |
|  2 | this PR |  9.7 |   10.4 | 14.9 |
|    | develop | 26.9 |   28.2 | 35.3 |
| 10 | this PR |  3.4 |    3.8 |  6.5 |
|    | develop | 19.7 |   21.2 | 30.0 |

Before this patch, normalizing the adjacency matrix was a huge chunk of the time required. In particular the `setdiag` call (which resulted in the `SparseEfficiencyWarning`s) was about 60% of the time of each `__getitem__` call, based on profiling, and the time in `StellarGraph.to_adjacency_matrix` was only 26% of the `__getitem__` time.

After this patch, the total time for normalisation decreases to 13% of the new overall time (that is,  the time in `to_adjacency_matrix` has increased to 72%, and the total time for normalisation has decreased to 13%. All in all, since nothing else chanced, this means the time to normalise has gotten about 10&times; faster. (This matches independent benchmarks I did while experimenting with the best way to rewrite this.)

I investigated other places to put the densification (the `.toarray()` call), such as after computing `diag` (not noticeably faster) or before doing the row multiplication (much slower as the size of each cluster increases, with constant density).

This doesn't try to share code with `core.utils.normalize_adj` (yet) because that is designed to output sparse matrices, doesn't handle self loops and doesn't handle the lambda diagonal enhancement, and thus there may not actually be much code to share.

See: #1094


stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -276,6 +276,28 @@ def __len__(self):
...
+        #     NA + NI + λN(diag(A) + I) =
+        #     NA + N(I + λ(diag(A) + I)) =
+        #     NA + λN(diag(A) + (1 + 1/λ) I))
+        degrees = np.asarray(adj_cluster.sum(axis=1)).ravel()
+        normalization = 1 / (degrees + 1)
+
+        # N * A: multiply rows manually
+        norm_adj = adj_cluster.multiply(normalization[:, None]).toarray()
+
+        # N * (1 + lambda) * (diag(A) + I): work with the diagonals directly
```
kjun9: I'm guessing it expands to be the same thing..? but the equation further above makes more sense to me `NA + λN(diag(A) + (1 + 1/λ) I))`, and looks more like the code below:
```
diag = NI
diag_addition = Nλ(diag(A) + (1 + 1/λ)I)
```

huonw: Oh, yeah, this isn't the same, it was an incorrect old version.

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -285,16 +307,9 @@ def __getitem__(self, index):
         # The operations to normalize the adjacency matrix are too slow.
         # Either optimize this or implement as a layer(?)
```
PantelisElinas: Do we still need this comment, or can we re-write to only say that we might be able to improve further as a layer?

huonw: Ah, yeah, I forgot to update it. 👍

Issue #900, opened 2020-02-20
Examples section in readthedocs is outdated


timpitman:
### Describe the bug

Examples section at https://stellargraph.readthedocs.io/en/latest/quickstart.html#getting-started refers to scripts that no longer exist, when it should refer to demo notebooks.  It also discusses downloading the CORA dataset, which is no longer required.

Ideally, we should have a table linking to our demos which clearly indicates the type of problem, and the algorithm used.

kjun9:
@timpitman I'm wondering if this whole page should just be identical to our top-level README - it seems like the contents are pretty much the same (introduction, getting started, installation, examples). I think it's less likely to get outdated that way, and it seems like we should just be able to point to the markdown file https://stackoverflow.com/questions/46278683/include-my-markdown-readme-into-sphinx

kjun9:
> it seems like we should just be able to point to the markdown file

Hmm seems it doesn't work well enough actually, the table of contents aren't handled nicely and internal links get broken. I'll try some alternatives

kjun9:
Managed to get something working using `recommonmark` and `sphinx-markdown-tables` #1096

PR #1096, opened 2020-03-18
Replace quickstart.txt with readme

kjun9:
This takes the approach of just replacing the existing `quickstart.txt` of our readthedocs documentation with our top-level README using:
* `recommonmark` to include markdown files directly into sphinx docs
* `sphinx-markdown-tables` to render markdown tables correctly
* creating a symlink `docs/README.md` that points to our top-level README

This ends up having to introduce two other readthedocs specific files in the repo:
* docs/requirements.txt - to specify the new dependencies to build the docs
* ~.readthedocs.yml - to specify the location of the requirements file~ oops there was already an existing one

I've preferred to go with this approach instead of rewriting specific sections in the existing `quickstart.txt` document since the content/intention looked mostly the same to me and it'll be easier to maintain going forward.. but let me know if there's a reason why the sphinx doc should be different from our top level readme that I'm not thinking of.

See documentation here: https://stellargraph--1096.org.readthedocs.build/en/1096/index.html

See #900


kjun9:
> there's two tables of contents

Yeah this one is difficult to get rid of, but I could try to at least have the second TOC be at the level below `StellarGraph Machine Learning Library` instead of the same level which it currently is. It probably means I have to update the readme file.

> API docs now seem to be broken

hmm good catch! Not sure what happened there I'll take a look

> there's some broken links or missing content in "indices and tables" section at the bottom

I'm guessing that's related to the API docs being broken

kjun9:
@timpitman 

Turns out the issue was just caused by the TOC section heading being at the same level as the top-level heading (i.e. in the README we had `# Stellargraph Machine Learning Library` and `# Table of Contents`) - I've updated the README to have the TOC at the next level down instead here, but it should actually be resolved with #1081 too. With this, the API docs and the links at the bottom are also not broken anymore.

It's still maybe slightly strange seeing an inner table of contents:
image

but I'm not sure there's a way around that with this approach. What do you think?

kjun9:
> 3. a few other projects have a README-like page as their actual landing page, like https://requests.readthedocs.io/en/master/ and https://papermill.readthedocs.io/en/latest/ I wonder if we could do something like this instead?

Yeah that seems like a good idea 👍  and doing this maybe resolves points 1 and 2 too.

> 4. can we somehow rewrite any links in the README to stellargraph.readthedocs.io/en/stable/FOO to just FOO, so that they're always pointing to the pages of the same version (rather than jumping to the stable versions)?

I think once we start customising the markdown, I'm not sure if we'd wanna use `recommonmark`, I didn't have much luck finding documentation around customisations like that.

kjun9:
https://github.com/stellargraph/stellargraph/issues/1156

PR #1100, opened 2020-03-18
Store embeddings statefully in ComplEx and DistMult models

huonw:
Our models are meant to be stateful collections of layers, meaning each `build` call should give input/output tensors using the same underlying trainable weights. For the knowledge graph models `ComplEx` and `DistMult`, this means storing the node and edge-type embeddings in the model itself.

See: #1088


huonw:
Switched these to be private, per https://github.com/stellargraph/stellargraph/pull/1102#discussion_r394670416 .

Issue #1089, opened 2020-03-17
The `build` method on models is potentially misleading


huonw:
### Description

Per https://github.com/stellargraph/stellargraph/issues/1088, models are designed to be stateful. They all have a `build` method that is meant to return input and output tensors that are views into this state for use with `tf.keras.Model`.

This is a little misleading (at least to @huonw), because a `build` method (or things with names like that) typically comes in two forms:

- `build` on a stateful object that returns nothing, and is thus purely a stateful initialisation/mutation (such as Keras's [`Layer.build`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#build), that "is typically used to create the weights of Layer subclasses")
- `build` on a stateless (or, at least, just "configuration" state) builder/factory object, that returns a stateful object, where multiple calls are either disallowed entirely (e.g. throw an exception) or return different state (this is the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern), see also https://stackoverflow.com/a/14429493/1256624 and https://stackoverflow.com/a/53540830/1256624 for some discussion about multiple calls)

The model `build` method is somewhere in between: it should do no mutation, and instead just reshape the (already built) internal state into the appropriate input and output tensors. In particular, multiple calls of `build` should yield tensors that connect to the same underlying state and so mutations (e.g. training) through one set of input/output tensors will "change" any other sets of input/output tensors.

This at least tripped up @huonw and @kieranricardo and potentially will trip up others too. We could consider a different name that conveys the flow of state more accurately.

PR #1101, opened 2020-03-18
Store weights statefully in Attri2vec

huonw:
Our models are meant to be stateful collections of layers, meaning each `build` call should give input/output tensors using the same underlying trainable weights. For `Attri2vec`, this means storing the dense layers and the link-prediction embeddings in the model itself.

See: #1088

### DIFF
--- a/stellargraph/layer/attri2vec.py
+++ b/stellargraph/layer/attri2vec.py
@@ -93,6 +93,22 @@ class Attri2Vec:
         # Feature dimensions for each layer
         self.dims = [self.input_feature_size] + layer_sizes
 
+        # store the trainable layers
+        self._layers = [
+            Dense(layer_size, activation=self.activation, use_bias=self.bias)
+            for layer_size in layer_sizes
+        ]
+
+        if self.multiplicity == 1:
+            self._output_embedding = None
+        else:
+            self._output_embedding = Embedding(
+                self.input_node_num,
+                layer_sizes[-1],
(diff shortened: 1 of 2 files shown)


stellargraph/layer/attri2vec.py
```diff
@@ -182,13 +195,9 @@ def link_model(self):
...
         x_inp_dst = Input(shape=(1,))
-        output_embedding = Embedding(
-            self.input_node_num,
-            self.dims[self.n_layers],
-            input_length=1,
-            name="output_embedding",
-        )
-        x_out_dst = output_embedding(x_inp_dst)
+
+        assert isinstance(self._output_embedding, Embedding)
```
PantelisElinas: Do we ever or is there ever a need for calling `link_model` or `node_model` directly rather than indirectly via a call to `build`? I understand this ticket is not exactly about this issue but it seems to me like the `*_model` methods should be "private". Anyway, just mentioning this, if you feel we need a ticket for this, otherwise, it all looks good as far as I can tell.

huonw: Yeah good point, but I've filed https://github.com/stellargraph/stellargraph/issues/1116, because it seems more than just `Attri2Vec` have `node_model` and `link_model` methods in addition to `build` (also cc #1089): `APPNP`, `GAT`, `GCN`, `GraphSAGE`, `PPNP`.

PR #1102, opened 2020-03-18
Store embeddings statefully in Watch Your Step

huonw:
Our models are meant to be stateful collections of layers, meaning each `build` call should give input/output tensors using the same underlying trainable weights. For `WatchYourStep`, this means storing the left and right embeddings in the model itself.

See: #1088

### DIFF
--- a/stellargraph/layer/watch_your_step.py
+++ b/stellargraph/layer/watch_your_step.py
@@ -181,13 +181,29 @@ class WatchYourStep:
 
         self.embedding_dimension = embedding_dimension
 
-        self.attention_regularizer = attention_regularizer
-        self.attention_initializer = attention_initializer
-        self.attention_constraint = attention_constraint
-
-        self.embeddings_initializer = embeddings_initializer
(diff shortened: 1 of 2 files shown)


stellargraph/layer/watch_your_step.py
```diff
@@ -181,13 +181,29 @@ def __init__(
...
         self.embedding_dimension = embedding_dimension
 
-        self.attention_regularizer = attention_regularizer
-        self.attention_initializer = attention_initializer
-        self.attention_constraint = attention_constraint
-
-        self.embeddings_initializer = embeddings_initializer
-        self.embeddings_regularizer = embeddings_regularizer
-        self.embeddings_constraint = embeddings_constraint
+        self._left_embedding = Embedding(
```
kieranricardo: Just reviewed you ComplEx and Distmult stateful PR - in that PR the embeddings are not hidden but here there are. Can you say why you went for different approaches?

huonw: No particular reason. I will update #1100.

PR #1104, opened 2020-03-18
Move watch_your_step.get_embeddings to WatchYourStep.embeddings method

huonw:
This moves the `get_embeddings` function to be a method on `WatchYourStep`, similar to `ComplEx`, `DistMult` and `DeepGraphInfomax`. This method refers to the layers stored within the `WatchYourStep` model instance.

See: #1109


huonw:
I'm going to wait for #1102 for this, since it will allow us to simplify the embeddings function dramatically.

PR #1106, opened 2020-03-18
Use stored embedding layers for ComplEx and DistMult .embeddings

huonw:
This saves us from needing to pull layers out of a Keras model using carefully constructed layer names.

See: #1109


huonw:
I'm going to wait for #1100 since it will allow us to simplify the embeddings function dramatically.

Issue #714, opened 2020-01-22
Top-level __all__ doesn't match attempted reexports


huonw:
### Describe the bug

Our top level `__init__` file imports and reexports types like `Ensemble` and `TemperatureCalibration`, but these are not listed in `__all__`. This is somewhat unconventional (as I understand it), and, if it's on purpose, should probably be explicitly noted:

https://github.com/stellargraph/stellargraph/blob/10e62006907dd5968286f33648d1054e9c961c1b/stellargraph/__init__.py#L22-L30

https://github.com/stellargraph/stellargraph/blob/10e62006907dd5968286f33648d1054e9c961c1b/stellargraph/__init__.py#L40-L49

### To Reproduce

1. View `stellargraph/__init__.py`

### Observed behavior

`__all__` does not list many of the names in the rest of `__init__.py`

### Expected behavior

`__all__` should match the names in the rest of the file, or there should be some explanation otherwise.

### Environment

N/A

### Additional context

N/A

PantelisElinas:
Hi @huonw 

I don't recall why it is as it is. We should probably update as necessary.

P.

PR #1107, opened 2020-03-18
Update and deprecate top-level exports

kjun9:
See #714


stellargraph/__init__.py
```diff
@@ -21,11 +21,23 @@
...
     "data",
+    "calibration",
+    "ensemble",
+    "interpretability",
+    "losses",
     "layer",
     "mapper",
     "utils",
+    "custom_keras_layers",
+    "expected_calibration_error",
```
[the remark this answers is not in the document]
kjun9: Yeah, only `StellarGraph` and `StellarDiGraph` really makes sense to me to export at the top-level, it's not clear to me why the rest of the classes/functions need special treatment. One reason for their previous inclusion might've been because they were under `utils` and we didn't want users to have to write `from stellargraph.utils.calibration import IsotonicCalibration` etc.. but now with `utils` gone maybe it makes sense to just remove them from the top-level export

kjun9:
@huonw 

I've tried removing the other top-level exports and added the deprecation warnings like you suggested (but this also suffers from #1117 I'm guessing). I kept the classes from `core` as top level exports since `core` is not part of the top-level `__all__` at all.

stellargraph/__init__.py
```diff
@@ -69,3 +71,41 @@
     "PPNPPropagationLayer": layer.ppnp.PPNPPropagationLayer,
     "APPNPPropagationLayer": layer.appnp.APPNPPropagationLayer,
 }
+
+
+def _top_level_deprecation_warning(name, path):
+    msg = (
```
huonw: FWIW, from my perspective, I think it'd be fine to not bother wrapping the string, and just write:

```python
warnings.warn(
    f"'{name}' is no longer available at the top-level. Please use 'stellargraph.{path}.{name}' instead.",
    DeprecationWarning,
)
```

This seems slightly easier to understand/work with, but maybe PyCharm or whatever interfere?

stellargraph/__init__.py
```diff
@@ -69,3 +71,41 @@
...
+        f"'{name}' is no longer available at the top-level. "
+        f"Please use 'stellargraph.{path}.{name}' instead."
+    )
+    warnings.warn(msg, DeprecationWarning)
```
huonw: The problem here isn't just caused by the default behaviour of `DeprecationWarning`, but also by point 1 of #1107: setting the `stacklevel` argument to point to the problematic call from the user. If the user calls `stellargraph.expected_calibration_error(...)`, the stack looks like:

1. user code
2. `stellargraph.expected_calibration_error`
3. `_top_level_deprecation_warning`

and so we need 

```suggestion
    warnings.warn(msg, DeprecationWarning, stacklevel=3)
```
to point at the user code.

If I make this change locally, I see the warning locally, and in a Python 3.7 shell.

kjun9: makes sense, okay I'll add stacklevel=3 here

kjun9:
Hi @PantelisElinas I've added you as reviewer - for now this can wait in case you're available soon to check my assumption that the ensemble and calibration functions/classes don't need to be re-exported at the top-level, but I'll also eventually just go ahead with the merge if that's not the case.

kjun9:
> if this is the way it should be.

Classes and functions in `ensemble` and `calibration` didn't particularly occur to me as something to be treated differently from things in other modules, so that's the motivation for removing them from the top-level so that we can be consistent now and in future, but not that there's a way it *should* be if that makes sense.

Thanks for taking a look!

Issue #1009, opened 2020-03-04
Some StellarGraph methods have outdated/unused functionality


huonw:
### Describe the bug

A few methods in the new StellarGraph class have arguments and functionality that isn't used or isn't useful:

- `info`: `show_attributes` and `sample` are no longer used https://github.com/stellargraph/stellargraph/blob/278a252cda91fc2c8e9c6285e9cea1c3a6f2fbdb/stellargraph/core/graph.py#L782

- `nodes_of_type`: `node_type = None` is weird, asking for nodes of a specific type without specifying a type: https://github.com/stellargraph/stellargraph/blob/278a252cda91fc2c8e9c6285e9cea1c3a6f2fbdb/stellargraph/core/graph.py#L596

### To Reproduce

1. Check `graph.py` for the things above.

### Observed behavior

See above

### Expected behavior

We should clean up some functionality

### Environment

Operating system: N/A

Python version: N/A

Package versions: StellarGraph 0.11.0b 278a252cda91fc2c8e9c6285e9cea1c3a6f2fbdb

### Additional context

N/A

PR #1110, opened 2020-03-18
Deprecate show_attributes and sample parameters to StellarGraph.info

huonw:
This parameters no longer make sense now that we've moved away from NetworkX:

- `show_attributes`: we no longer have a concept of attributes outside of node features and edge weights, unlike the NetworkX-backed `StellarGraph` which could have arbitrary data associated with each entity
- `sample`: using NumPy and Pandas is fast enough without any sampling, so there's not as good a reason to support that extra code path
 
See: #1009


stellargraph/core/graph.py
```diff
@@ -821,7 +821,7 @@ def _unique_type_triples(self, *, return_counts, selector=slice(None)):
 
         return zip(*unique_ets)
 
-    def info(self, show_attributes=True, sample=None, truncate=20):
+    def info(self, show_attributes=None, sample=None, truncate=20):
```
habiba-h: if `show_attributes` and `sample` are not used, why are we still keeping it in the `info` definition?

huonw: Good question: it's not truly needed, but it's helpful for our users. If a user has existing code that has `some_graph.info(sample=20)` or `some_graph.info(False)`, removing the parameters to make this `def info(self, truncate=20)`  will cause their code to break when they upgrade their version of StellarGraph. For example, the first one will have an exception
```python
>>> some_graph.info(sample=20)
TypeError: info() got an unexpected keyword argument 'sample'
```
and the second will end up passing `False` to `truncate`, which doesn't give the behaviour they were expecting.

By leaving them here with a warning, the code will continue to work, and the warning will hopefully help them migrate to the better form `some_graph.info()`. After we do a release or two with the deprecations, we can then properly remove them This is similar to how we had a variety of deprecated things in the 0.10 release, and then removed them soon after we did the release in https://github.com/stellargraph/stellargraph/pull/969.

huonw:
Good catch! The warning is being emitted, but it's seems it's being suppressed by default, from https://docs.python.org/3/library/warnings.html#warning-categories

> `DeprecationWarning`: Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in __main__).

For instance, forcing all warnings to be shown with:

```
import warnings
warnings.simplefilter("always")
```

shows

```
>>> print(SG.info(show_attributes = "yes"))
/Users/wil9dw/projects/stellargraph/stellargraph-3/stellargraph/core/graph.py:843: DeprecationWarning: 'show_attributes' is no longer used, remove it from the 'info()' call
  DeprecationWarning,
StellarGraph: Undirected multigraph
 Nodes: 5, Edges: 4

 Node types:
  default: [5]
    Features: none
    Edge types: default-friend->default, default-temporal->default

 Edge types:
    default-temporal->default: [2]
    default-friend->default: [2]
```

I think the problem here is I haven't configured the `stacklevel` parameter correctly, so it doesn't look the code is running in `__main__` (as in, the main script or a REPL session or jupyter notebook). Unfortunately I don't know how to test this, and it probably affects all our other DeprecationWarnings too 😦  Your example demonstrates this too: `SG = sg.StellarGraph(G)` should've hit https://github.com/stellargraph/stellargraph/blob/cdcde873f7677a25d8c18124ef88a599f31ca7d9/stellargraph/core/graph.py#L253-L256

I've opened https://github.com/stellargraph/stellargraph/issues/1117.


> Is it because I am starting with a networkx graph and then converting it to StellarGraph?

No, that won't affect it, all methods of construction result in a `StellarGraph` instance and thus all use the same `info` method.

habiba-h:
Thanks for explaining this! Helpful to know whats going on under the hood. 
And yes when I tried just `SG = sg.StellarGraph(G)` I didn't get the `networkx` warning.

huonw:
I've fixed it here by adding [`stacklevel=2` to the `warn` call](https://docs.python.org/3/library/warnings.html#warnings.warn). In a python 3.7 shell, this now shows:

```python
>>> print(SG.info(show_attributes = "yes"))
__main__:1: DeprecationWarning: 'show_attributes' is no longer used, remove it from the 'info()' call
...
```

and in a Jupyter notebook (Python 3.6 or Python 3.7) it shows:

```python
print(SG.info(show_attributes=True))
```
> ```
> StellarGraph: Undirected multigraph
>  Nodes: 5, Edges: 4
> 
>  Node types:
>   default: [5]
>     Features: none
>     Edge types: default-friend->default, default-temporal->default
> 
>  Edge types:
>     default-temporal->default: [2]
>     default-friend->default: [2]
> .../lib/python3.6/site-packages/ipykernel/__main__.py:1: DeprecationWarning: 'show_attributes' is no longer used, remove it from the 'info()' call
>   if __name__ == '__main__':
> ```

Handling the other cases (and optimising the Python 3.6 behaviour) can happen as part of #1117.

huonw:
Yeah, that's part of what #1117 is tracking.

PR #1111, opened 2020-03-18
Fold StellarGraph nodes_of_type functionality into nodes(node_type=)

huonw:
The `StellarGraph` `nodes_of_type` method has a lot of overlap with the `nodes`s method, especially given `nodes_of_type(None)` is exactly equivalent to `nodes()`. This patch moves the `nodes_of_type` functionality into the `nodes` method by adding a `node_type` parameter to the latter, which can be used like `some_graph.nodes(node_type="foo")`. The old `nodes_of_type` method is thus deprecated and users should migrate to the new `nodes` form.

See: #1009


kjun9:
Hmm it's a bit peculiar to me that `nodes()` and `nodes_of_type()` are separate in the first place? I think having one method to do both jobs seems likely to be more convenient for sharing code between homogeneous and heterogeneous algorithms. Thoughts on / would it be possible to just merge the two methods together and deprecate `nodes_of_type`?

huonw:
That's a very good idea!

huonw:
I've implemented that. What do you think?

kjun9:
> What do you think?

I like it 👍 it seems like there's some failures due to the fact that `nodes_of_type` used to return a list, but `nodes` returns a dataframe index though

huonw:
I'm assuming that not many people will be relying on the details of `generator.node_list`, so I've fixed that in the minimal way in the downstream code (i.e. in `IntegratedGradientsGAT`), rather than pessimise everyone by converting to a list eagerly.

Issue #1120, opened 2020-03-19
Remove experimental tag from temporal random walk


kjun9:
### Description

The `TemporalRandomWalk` class now has some unit tests and a demo notebook. This issue captures any remaining work to remove the experimental tag from the class, which might just be removing the tag and updating the changelog, or the inclusion of more unit tests that were missed.

### User Story

**As** a data scientist

**I want to** use the `TemporalRandomWalk` class as a non-experimental feature

**So that** I know it's going to work

PR #1121, opened 2020-03-19
Remove experimental tag from TemporalRandomWalk

kjun9:
This marks the final step to complete StellarGraph's implementation of the algorithm from the reference paper: Continuous-Time Dynamic Network Embeddings.

See #1120


kjun9:
> Will you be updating README.md separately?

Ah no, I'd forgotten. I'll update it here, thanks for the reminder

Issue #1123, opened 2020-03-19
GCN node classification notebook doesn't have enough explanation to serve as first-machine-learning notebook


huonw:
### Describe the bug

We're currently thinking that we'll use the GCN node classification notebook as an "intro to graph ML with StellarGraph", however, it doesn't currently have enough explanation to justify this. For instance, there's no introduction at the start.

### To Reproduce

1. Look at `demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb` https://github.com/stellargraph/stellargraph/blob/develop/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb

### Observed behavior

Currently on develop (5c26f2da9bcba2c62bd4225e9177ff4c01e7f748), https://github.com/stellargraph/stellargraph/blob/5c26f2da9bcba2c62bd4225e9177ff4c01e7f748/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb , the notebook starts like:

![image](https://user-images.githubusercontent.com/1203825/77118333-15ad5900-6a88-11ea-8233-264ccc47008f.png)

### Expected behavior

We should hand-hold the reader to get them up to speed on graph ML and in particular the parts that are unique to StellarGraph.

### Environment

Operating system: N/A

Python version: N/A

Package versions: StellarGraph 5c26f2da9bcba2c62bd4225e9177ff4c01e7f748

### Additional context

N/A

PR #1074, opened 2020-03-12
Add demos for loading data from Pandas and NetworkX

huonw:
This adds two notebooks into a new `demos/basics` folder

- "Loading data into StellarGraph from Pandas"
- "Loading data into StellarGraph from NetworkX"

These notebooks walk through building a `StellarGraph` object in many cases:

- "homogeneous graph without features" 
- "homogeneous graph with features"
- "homogeneous graph with edge weights"
- "directed graphs"
- "heterogenous graphs"

They're designed to be both tutorials and a reminder. Someone with a particular type of graph can jump to a given section.

See: #992

PR #1125, opened 2020-03-20
Add more explanation to the GCN node classification demo

huonw:
This adds more text to the GCN node classification demo to be more suitable as an introduction to graph machine learning with StellarGraph. We're planning to use this as the example of how to create models and generators (and so on), and so this PR makes sure that is well explained.

It also makes a couple of minor adjustments/simplifications to the code, such as:

- switching to `pandas_series.value_counts()` instead of `set(...)` or `Counter(...)` for understanding the subject labels (in addition to having slightly nicer formatting, this has the benefit that it should be stable: rerunning the notebook won't change the ordering)
- reducing the `patience` value in the early stopping to something that may actually take effect: previous it was larger than the number of epochs (50 vs. 40)
- simplifying how the TSNE plot is rendered

See: #1123


demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb
```diff
@@ -4,14 +4,36 @@
    "cell_type": "markdown",
```
PantelisElinas: <p>There is no "deep machine learning" specialization. You might want to say, "The StellarGraph library supports many state-of-the-art graph machine learning algorithms..." You might also want to put a link to our introductory Medium article that already goes into detail about GCN (https://medium.com/stellargraph/knowing-your-neighbours-machine-learning-on-graphs-9b7c3d0d5896).</p><p> </p><p>I think "sometimes called node classification or even node attribute inference." can be safely removed from this sentence without loss or simplify to say " commonly known as node classification." As long as you consistently use the latter term in the rest of the document and not use the "node attribute inference" term then you don't have to mention it here.</p><p> </p><p>You might want to mention that Cora is a directed graph but we are going to treat it as undirected.</p><p> </p><p> </p><blockquote>StellarGraph supports <a href="https://github.com/stellargraph/stellargraph/raw/d09fefd5325ee9f895640defac4006ecac8f1c0b/demos/node-classification/README.md" target="_blank">other algorithms for doing node classification</a>, as well as many <a href="https://github.com/stellargraph/stellargraph/raw/d09fefd5325ee9f895640defac4006ecac8f1c0b/demos/README.md" target="_blank">other tasks</a>. </blockquote><p> </p><p>Should probably drop the word "doing" from above. Also, can we be more specific about the other tasks, and maybe re-write to say "as well as, link prediction, interpretabililty, community detection, graph classification, etc." and link each to the individual folder?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

PantelisElinas: <p>Shall we mention here importing StellarGraph as sg to try and steer people towards associating this short name with our library like how everyone imports Pandas as pd and Numpy as np?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

PantelisElinas: <blockquote>We can create retrieve a <code>StellarGraph</code> graph object representing this Cora dataset... </blockquote><p> </p><p>Needs a re-write.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

PantelisElinas: <blockquote>...we will use one-hot vectors that will be fed into a soft-max Keras layer during training.</blockquote><p> </p><p>These vectors are not input to the soft-max layer. They are used in calculating the loss by comparing with the output of the soft-max layer.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

PantelisElinas: <p>We have a slide with a picture that describes the dataset including the TF encoding (https://en.wikipedia.org/wiki/Tf%E2%80%93idf). It might be useful to add here if you want. I will send it via slack.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

PantelisElinas: <blockquote>...is a list of hidden feature sizes of each layer in the model...</blockquote><p> </p><p>The layer is commonly referred to as hidden not the individual units. Saying something like, "<code>layer_sizes</code>  denotes the number of hidden layers and the sizes of each layer. In this example, we use two GCN layers with 16 units each."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

PantelisElinas: <p>I think calling <code>build</code>  is important for how StellarGraph models are used. I suggest a brief explanation of what this call achieves might be useful.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

PantelisElinas: <p>You might want to emphasize here that StellarGraph models are built using custom Keras layers and as such, we can use them together with standard Keras layers. For example, here is how to use a Keras Dense layer for classification using the node embeddings from the GCN layers.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

PantelisElinas: <p>Using the ModelCheckpoint  callback to retrieve the best model is not longer necessary as EarlyStopping  has introduced the parameter <code>restore_best_weights</code>.  Note that the default values is <code>False</code>  so that it has to be explicitly set to <code>True</code>  to get the best model. Using <code>ModelCheckpoint</code>  was how we had to do it. Many of our notebooks have been updated accordingly, seems like a good opportunity to update this one as well.</p><p> </p><p>On the patience vs number of epochs decisions, e.g., reduce the former but leave the latter as is, I would suggest that we set patience to what it was and just increase the number of epochs. The point is that you want to train as much as possible to find the best model given that early stopping will help avoid overfitting. If you set patience too small, then the training might end before the optimizer had a chance to find the best model.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

PantelisElinas: <blockquote>...to view how the loss function decreases and accuracy increases (hopefully!) over the training epochs:</blockquote><p> </p><p>Maybe be a bit more neutral here (no need to be hopeful of an outcome) and say something like, "we can examine the behavior of the loss function and training metrics by visualizing their values during training using <code>StellarGraph</code>'s  <code>plot_history</code>  utility funtion."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='39'/>

PantelisElinas: <p>We can remove this extra step as I suggested above.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='42'/>

PantelisElinas: <blockquote>..meaning the each embedding.</blockquote><p> </p><p>Re-write</p><p> </p><blockquote>There's many tools...</blockquote><p> </p><p>There are...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='56'/>

PantelisElinas: <p>Yes, that should be fine.</p>

PantelisElinas: <p>If that was there before, I must not have seen it. It is ok then.</p>

PantelisElinas: <p>I agree that the old values for patience and epochs made little sense so updating them is the right thing to do. I was only suggesting that the update should have been the other way around. The value for patience should be something reasonable so there is nothing special about it being 50. Generally, it is preferable to train as long as possible but also give the algorithm a fair chance to overcome a local minimum, hence generally you want a largish patience value but not too large with respect to the number of epochs as we had it setup before. I think we had set the number of epochs to a small number so that it runs fast; not sure why patience was so high though, probably an oversight. </p><p> </p><p> </p>

timpitman: <p>The StellarGraph links here refer to a specific PR here when I test; I think they should point to master? They don't seem to work correctly inside the github notebook renderer either, eg: <a href="https://github.com/stellargraph/stellargraph/blob/cf18446ab0c9f5ba553c1c38c2cf6e62537fc655/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb" target="_blank">https://github.com/stellargraph/stellargraph/blob/cf18446ab0c9f5ba553c1c38c2cf6e62537fc655/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb</a></p><p> </p><p>Also the display of "This demo..." and "StellarGraph supports..." paragraphs looks as though they're quoted?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

timpitman: <p>broken link to the notebook here (may work once other work is merged into master?)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

timpitman: <p>I think this has lost the explanation of 'patience', which is useful?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

timpitman: <p>Why the increase from 40 to 200 epochs here - did you aim to be testing the 'patience' setting which wasn't relevant before?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='38'/>

huonw: It was there before, so I'll leave it as.

huonw: Not so much testing, as better practice to train for a while and have a relatively high patience. https://github.com/stellargraph/stellargraph/pull/1125#discussion_r396158701

huonw: I tried a few different ways to phrase the explanation, and decided it was better to just link to the Keras docs and let them explain the parameters. That is, not try to teach too much about Keras here too. What do you think?

huonw: You're right, it's dependent on #1074.

huonw: > The StellarGraph links here refer to a specific PR here when I test; I think they should point to master? They don't seem to work correctly inside the github notebook renderer either, eg: https://github.com/stellargraph/stellargraph/blob/cf18446ab0c9f5ba553c1c38c2cf6e62537fc655/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb

Could be more specific about which links and their destination that you find unexpected?

I clicked on every link in this section in the GitHub rendering you linked, and all but one went where I expected:

| name | status |
|---|---|
| See all other demos | ✅ |
| The StellarGraph library | ✅  |
| graphs | ❌[1] |
| https://linqs.soe.ucsc.edu/data | ✅ |
| "Knowing Your Neighbours: Machine Learning on Graphs" | ✅ |
| Pandas | ✅ |
| scikit-learn | ✅ |
| StellarGraph | ✅ |
| TensorFlow Keras | ✅ |
| other algorithms for node classification |  ✅ |
| other tasks | ✅ |
| Graph Convolutional Networks (GCN): Semi-Supervised Classification with Graph Convolutional Networks | ✅ |

1. pointed to <https://en.wikipedia.org/wiki/Graph_%5C(discrete_mathematics%5C>, fixed by updating to <https://en.wikipedia.org/wiki/Graph_%28discrete_mathematics%29>)


> Also the display of "This demo..." and "StellarGraph supports..." paragraphs looks as though they're quoted?

The quoting is on purpose: they're meant to be asides and/or providing quick context/helpful links, but there's no markdown to generate the real semantic markup (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside) and writing the HTML directly doesn't for what I want with this (a "side note" providing more info):

- there's no special rendering for `<aside>` in jupyter lab: image
- GitHub strips out the `<aside>` tag: https://gist.github.com/huonw/87106c1524cd99e73f8db3c53afc1d26

timpitman: I see, yes that works fine.

timpitman: Thanks for explaining about the quotes. The bad links were in the cell I think NBreview referenced, specifically the paragraph:

```StellarGraph supports other algorithms for doing node classification, as well as many other tasks such as link prediction, community detection, and graph classification.```

huonw: Ah, right. That's https://github.com/stellargraph/stellargraph/issues/1139. I'll remove it.

Issue #1109, opened 2020-03-18
WatchYourStep ComplEx and DistMult embeddings can be extracted by any model


huonw:
### Describe the bug

In #1001, there's an `embedding_model` method, that extracts some layers from a keras model to create a new model, for computing node embeddings.

https://github.com/stellargraph/stellargraph/blob/02c5f086ecdd605c7a9409264b34cedbb1a6cfbc/stellargraph/layer/deep_graph_infomax.py#L139-L163

It is careful to check that the layers match the `DeepGraphInfomax` instance that it is working with by adding a unique identifier to the layer names. This also ensures that multiple DGI models can be used within a single overall graph.

The `WatchYourStep`, `ComplEx` and `DistMult` offer similar functionality, but they are not as strict, so it can be possible to mix up models, which may restrict how we can change the API in future.

https://github.com/stellargraph/stellargraph/blob/02c5f086ecdd605c7a9409264b34cedbb1a6cfbc/stellargraph/layer/watch_your_step.py#L108-L127

https://github.com/stellargraph/stellargraph/blob/02c5f086ecdd605c7a9409264b34cedbb1a6cfbc/stellargraph/layer/knowledge_graph.py#L119-L138

https://github.com/stellargraph/stellargraph/blob/02c5f086ecdd605c7a9409264b34cedbb1a6cfbc/stellargraph/layer/knowledge_graph.py#L364-L380

### To Reproduce

- Construct two `DistMult` (or other) models, and then call `distmult2.embeddings(Model(*distmult1.build()))`

### Observed behavior

The above works.

### Expected behavior

It should fail.

### Environment

Operating system: N/A

Python version: N/A

Package versions: StellarGraph 02c5f086ecdd605c7a9409264b34cedbb1a6cfbc

### Additional context

N/A

PR #1126, opened 2020-03-20
Omit model argument in DeepGraphInfomax.embedding_model via statefulness

huonw:
We can use the statefulness of models (#1088) to simplify how to construct the `DeepGraphInfomax`'s embedding model. In particular, the base model should be stateful, so multiple calls to `build` will give input/output tensors that all tie into the same trainable weights, so the embedding model can be built by creating a new set of base-model input/output tensors instead of picking a layer out of a pre-build Keras model.

See: #1109 (at least, this is along the same lines as #1104 and #1106)

### DIFF
--- a/stellargraph/layer/deep_graph_infomax.py
+++ b/stellargraph/layer/deep_graph_infomax.py
@@ -93,8 +93,6 @@ class DeepGraphInfomax:
         self._corruptible_inputs_idxs = [0]
 
         self._discriminator = DGIDiscriminator()
-        self._base_input = None
-        self._base_output = None
 
     def build(self):
         """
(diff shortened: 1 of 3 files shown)


stellargraph/layer/deep_graph_infomax.py
```diff
@@ -89,11 +89,12 @@ def __init__(self, base_model):
         self.base_model = base_model
 
         self._node_feats = None
-        self._unique_id = f"DEEP_GRAPH_INFOMAX_{id(self)}"
         # specific to full batch models
         self._corruptible_inputs_idxs = [0]
 
         self._discriminator = DGIDiscriminator()
+        self._base_input = None
```
kieranricardo: are these variable's used?

huonw: Oh, from an old version. Removed now.

Issue #713, opened 2020-01-22
'from .foo import *' reexports are invisible in documentation


huonw:
### Describe the bug

We've got a lot of `__init__` files that reexport some or all of their submodule contents, e.g.:

https://github.com/stellargraph/stellargraph/blob/10e62006907dd5968286f33648d1054e9c961c1b/stellargraph/__init__.py#L32-L46

https://github.com/stellargraph/stellargraph/blob/10e62006907dd5968286f33648d1054e9c961c1b/stellargraph/utils/__init__.py#L22-L24

However this doesn't seem to be visible in the documentation, e.g.:

- `StellarGraph` is only listed as `stellargraph.core.StellarGraph` https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.core.StellarGraph, but is available as `stellargraph.StellarGraph`
- `Ensemble` is only listed as `stellargraph.utils.ensemble.Ensemble` https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.utils.ensemble.Ensemble, but is available as both `stellargraph.Ensemble` and `stellargraph.utils.Ensemble`

(develop at the time of writing: 10e62006907dd5968286f33648d1054e9c961c1b).

### To Reproduce

1. Go to the API documentation https://stellargraph.readthedocs.io/en/latest/api.html
2. Compare the import paths against `__init__` files

### Observed behavior

There's only a single listing of each of the types, without any mention of reexports.

### Expected behavior

Some sort of indication of more convenient import paths, e.g. only listing them.

### Environment

N/A

### Additional context

N/A

kjun9:
Tensorflow's docs seems to have a dropdown to view aliases - e.g. https://www.tensorflow.org/api_docs/python/tf/keras/losses/BinaryCrossentropy#view-aliases

Maybe we can do something like that?

PR #1127, opened 2020-03-20
Document top-level reexports instead of core module

kjun9:
This replaces `stellargraph.core.*` with `stellargraph.*` in our documentation. I thought this was a reasonable way to go about tackling #713 since we now only reexport classes in `core` (see #1107 ) and we don't actually have `core` included in `__all__`, so users should be encouraged to use these top-level exports.

This approach does mean that documenting anything else we might reexport at top-level that's outside of `core` is still awkward in the future, but doesn't seem too likely that we'd want to do that?

See: https://stellargraph--1127.org.readthedocs.build/en/1127/api.html#module-stellargraph

See #713

### DIFF
--- a/docs/api.txt
+++ b/docs/api.txt
@@ -4,7 +4,7 @@ StellarGraph API
 Core
 ----------------
 
-.. automodule:: stellargraph.core
+.. automodule:: stellargraph
   :members: StellarGraph, StellarDiGraph, GraphSchema
 
 
--- a/stellargraph/__init__.py
+++ b/stellargraph/__init__.py
@@ -14,10 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""
-Stellar Machine Learning Library
-
-"""
 
 __all__ = [
     "data",


timpitman:
This PR resolves the issues with "core", however #713 does list other issues, eg:
> Ensemble is only listed as stellargraph.utils.ensemble.Ensemble https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.utils.ensemble.Ensemble, but is available as both stellargraph.Ensemble and stellargraph.utils.Ensemble

kjun9:
@timpitman 

https://github.com/stellargraph/stellargraph/pull/1107 ended up deprecating every other top-level reexport except those from `core` (so e.g. `Ensemble` is no longer re-exported at top-level), so that's why I went with this approach, what do you think?

huonw:
FWIW, there still reexports in some places, such as `stellargraph.layer.GCN` is a reexport of ``stellargraph.layer.gcn.GCN``, but it's documented under the latter. (https://stellargraph.readthedocs.io/en/latest/api.html#stellargraph.layer.gcn.GCN).

kjun9:
> FWIW, there still reexports in some places, such as stellargraph.layer.GCN is a reexport of stellargraph.layer.gcn.GCN, but it's documented under the latter.

Seems like this is the case for everything in `stellargraph.layer` - I might try something in a separate PR

PR #1128, opened 2020-03-20
Replace kwargs with real args for attri2vec

kjun9:
See #801

### DIFF
--- a/stellargraph/layer/attri2vec.py
+++ b/stellargraph/layer/attri2vec.py
@@ -34,8 +34,8 @@ def _require_without_generator(value, name):
         return value
     else:
         raise ValueError(
-            f"'{name}' is None. 'input_dim', 'node_num' and 'multiplicity' must be specified when a "
-            f"Generator is not provided."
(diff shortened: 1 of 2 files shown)


stellargraph/layer/attri2vec.py
```diff
@@ -29,6 +29,16 @@
 from ..mapper import Attri2VecLinkGenerator, Attri2VecNodeGenerator
 
 
+def _require_without_generator(value, name):
+    if value is not None:
+        return value
+    else:
+        raise ValueError(
+            f"'{name}' is None. 'input_dim', 'node_num' and 'multiplicity' must be specified when a "
```
huonw: For the phrasing of this error message, I wonder if we can go for something more formulaic (consistency makes things easier to parse, IME), e.g.

```python
f"{name}: expected a value for 'input_dim', 'node_num' and 'multiplicity' when 'generator' is not provided, found {name}=None."
```

kjun9: Thanks, I'll do the same for https://github.com/stellargraph/stellargraph/pull/1130

Issue #1135, opened 2020-03-23
Unit tests for graph_log_likelihood


kieranricardo:
This PR is to write unit tests for `graph_log_likelihood`. See equation 4 of https://arxiv.org/pdf/1710.09599.pdf. This is the last experimental blocker for `Watch Your Step`

PR #1138, opened 2020-03-24
graph_log_likelihood unit test

kieranricardo:
This PR adds a unit test for `graph_log_likelihood`. I also found and fixed a wrong sign bug in `graph_log_likelihood`, this had the effect of:

- decreasing demo classification accuracy from 78% to 67%: it now matches the papers reported accuracy of 67%
- demo embeddings TSNE are now less well separated and appear less visually similar to the paper

So in terms of accuracy it matches the paper better, but in terms of TSNE plot it matches the paper worse. TSNE is very parameter dependant and I couldn't find the paper's parameters so this is a less reliable metric than classification accuracy.

See #1135 and equation 4 of https://arxiv.org/pdf/1710.09599.pdf


kieranricardo:
> Why does using this "incorrect" loss function give a better result?

I think we may have accidently stumbled across a better loss function.

Running this code:

```
from matplotlib import pyplot as plt
import numpy as np

# expected walk value for one node pair
e = 5

# potential values of sigma in range (0, 1)
# lets see which sigma gives the lowest loss
sigmas = np.linspace(0.0001, 0.9999, 1000)

# graph log llikelihood function when there's no edge
loss_1 = np.abs(-e * np.log(sigmas) - np.log(1 - sigmas))
loss_2 = np.abs(e * np.log(sigmas) - np.log(1 - sigmas))

plt.title("Graph Log Likelihood Loss")
plt.plot(sigmas, loss_1)
plt.ylabel("loss")
plt.xlabel("sigma")
plt.show()

plt.title("Custom Loss")
plt.plot(sigmas, loss_2)
plt.ylabel("loss")
plt.xlabel("sigma")
plt.show()
```

Gives:

image

The local minima of the custom loss (my old buggy graph log likelihood loss) is roughly the same as the graph log likelihood but it looks to me that the custom loss local minima will be easier to find with gradient descent bc it's gradient doesn't flatten out so much.

Issue #1116, opened 2020-03-18
Some models have public `node_model` and `link_model` methods in addition to `build`


huonw:
### Describe the bug

We've mostly switched to using the `build` for creating tensors for our models (e.g. #339 & PR #385, #410 & PR #543), but there's many classes that still have `node_model` and `link_model` methods. We've moved away from using them in demos (in #543), so we should potentially make these methods entirely private or otherwise remove them if they're not useful. (Although, #440 explicitly added `link_model`, so maybe it is useful?)

Model classes with these methods:

- [ ] `APPNP`
- [ ] `Attri2Vec`
- [ ] `GAT`
- [ ] `GCN`
- [ ] `GraphSAGE`
- [ ] `PPNP`

(#1089 is also tangentially related.)

This was noticed in https://github.com/stellargraph/stellargraph/pull/1101#discussion_r394673564.

### To Reproduce

1. `git grep 'def \(node\|link\)_model'`

### Observed behavior

```
stellargraph/layer/appnp.py:    def link_model(self):
stellargraph/layer/appnp.py:    def node_model(self):
stellargraph/layer/attri2vec.py:    def node_model(self):
stellargraph/layer/attri2vec.py:    def link_model(self):
stellargraph/layer/gcn.py:    def link_model(self):
stellargraph/layer/gcn.py:    def node_model(self):
stellargraph/layer/graph_attention.py:    def link_model(self):
stellargraph/layer/graph_attention.py:    def node_model(self):
stellargraph/layer/graphsage.py:    def node_model(self):
stellargraph/layer/graphsage.py:    def link_model(self):
stellargraph/layer/ppnp.py:    def link_model(self):
stellargraph/layer/ppnp.py:    def node_model(self):
```

### Expected behavior

If a model has a node or link model method, it should be private, like `def _node_model`.

### Environment

Operating system: N/A

Python version: N/A

Package versions: StellarGraph https://github.com/stellargraph/stellargraph/commit/2641535bdf42ffa53600e91a88d9af93d96cd640

### Additional context

N/A

Issue #1117, opened 2020-03-19
Some DeprecationWarnings don't show up in jupyter notebooks


huonw:
### Describe the bug

We have a few places that use `DeprecationWarning`s to flag things that are changing and help users migrate... however, these aren't being shown by default in scripts or Jupyter notebooks. The `DeprecationWarning` type only shows up for code in `__main__` by default (i.e. a script or REPL session or jupyter notebook ),  from https://docs.python.org/3/library/warnings.html#warning-categories

> `DeprecationWarning`: Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in __main__).

We should be configuring our deprecation warnings to appear in those circumstances. Some options:
1. using the `stacklevel` parameter to ensure they point to the call of the deprecated function, not the call of `warnings.warn` 
   - this works for Jupyter for all Python versions, but not for the Python REPL or a script in Python 3.6, the default was changed in 3.7 https://www.python.org/dev/peps/pep-0565/, but we could add a `warnings.filterwarnings("once", category=DeprecationWarning, module="__main__")` call to our `__init__.py` file to override Python's defaults for Python 3.6 (probably not a good idea...)
   - (I think we should be doing this anyway, since problem is the caller, not the `warnings.warn` call itself)
2. switching to using `FutureWarning`, but, per that table, it's not quite right (https://github.com/ipython/ipython/issues/8478 is related, and mentions Pandas using `FutureWarning` so that it is shown by default)
3. similarly, doing what NumPy does, which is creating a custom warning class `VisibleDeprecationWarning` that isn't suppressed by default https://github.com/numpy/numpy/blob/965b41d418e6100c1afae0b6f818a7ef152bc25d/numpy/_globals.py#L44-L51 (we could potentially just use that, too)

I think we should do only 1, without changing the warnings filtering, because we're mostly focused on Jupyter notebooks, but I could easily be convinced of other options. For instance, the thinking with Python's defaults seems to be these deprecation (and "pending deprecation") warnings get caught by a test suite outside of `__main__` (and hence, not showing them by default is ok); for instance, pytest will capture and show all warnings by default. However, I imagine many users of StellarGraph won't have a test suite exercising their graph ML, since that seems to be relatively hard/rare in ML circle, and thus if they have separate files containing all the ML parts that are imported/run by a script or notebook they won't see these warnings. Therefore, 2 or 3 are tempting.

[...]

PR #1140, opened 2020-03-24
Enhancement/1089 build method clarity

kieranricardo:
This PR cleans up the API for creating models. It makes several core changes:

- On `APPNP`, `Attri2Vec`, `GCN`, `GAT`, `GraphSAGE`, `HinSAGE`, `PPNP`, the `node_model` and `link_model` methods are now private. Migration:
  - `model.node_model()`: use `model.in_out_tensors()` (if the model was created with a node generator) or `model.in_out_tensors(multiplicity=1)` (otherwise)
  - `model.link_model()`: use `model.in_out_tensors()` (if the model was created with a link generator) or `model.in_out_tensors(multiplicity=1)` (otherwise)
- `build` is renamed to `in_out_tensors` for all model classes. Migration: replace `model.build()` with `model.in_out_tensors()`

In support of this, it makes some other changes:

- tests and demos now call `in_out_tensors` instead of `node_model`/`link_model`/`build`/`default_model`
- adds a `multiplicity` keyword arg into `in_out_tensors` for models with a `multiplicity` attribute

See #1089 and #1116


huonw:
This is going to be a bit annoying, but I feel like we really should have deprecations here. This is such a massive change that I imagine almost all users of StellarGraph will hit it (only those doing pure Node2vec or similar won't, I think), and as it stands, I think they'll hit an error like `AttributeError: 'GCN' object has no attribute 'build'` which is... rather unhelpful. It's such a fundamental change that we should potentially retain these deprecations for more than one release cycle. I can see a few approaches to implementing this (whichever you choose, you'll need to verify that it works as expected, though):

#### manual build methods

```python
def build(self, *args, **kwargs):
    """Deprecated: use :meth:`in_out_tensors`."""
    warnings.warn("The 'build' method is deprecated, use 'in_out_tensors' instead.", DeprecationWarning, stacklevel=2)
    return self.in_out_tensors(*args, **kwargs)
```
#### defining a single function and setting it in each class

We could do the previous one somewhat magically, by defining a single instance of it (e.g. in `stellargraph/layer/misc.py`):

```python
def deprecated_model_build(object, *args, **kwargs):
    """Deprecated: use :meth:`in_out_tensors`."""
    warnings.warn("The 'build' method is deprecated, use 'in_out_tensors' instead.", DeprecationWarning, stacklevel=2)
    return object.in_out_tensors(*args, **kwargs)
```

and then each model includes something like:

```python
from .misc import deprecated_model_build
class GCN:
    ...
    build = deprecated_model_build
```

#### defining a base class for models with a `build` method

```python
class Model:
    def build(self, *args, **kwargs):
        """Deprecated: use :meth:`in_out_tensors`."""
        warnings.warn("The 'build' method is deprecated, use 'in_out_tensors' instead.", DeprecationWarning, stacklevel=2)
        return self.in_out_tensors(*args, **kwargs)

...

class GCN(Model):
    ...

class GraphSAGE(Model):
    ...
```


--- 

We should consider doing the same for the `node_model` and `link_model` methods, although that's less major of a change. That is, retaining the public `node_model` and `link_model` methods with their current behaviour, but have them emit a `DeprecationWarning`.

Also, I feel like this really didn't need to have quite this many changes in it? The overall change isn't that large, but conceptually it's quite a lot, e.g. the last two changes seem pretty orthogonal to this change? (As in, things that don't depend on the name of the `build` method)

> - deprecates default_model
> - adds a multiplicity keyword arg into in_out_tensors for models with a multiplicity attribute

kieranricardo:
> Who are they intended to be used by?

I think having a seperate `_node_model` and `_link_model` now mostly just reduces code duplication because most `_link_model` functions end up calling `_node_model` twice

kieranricardo:
@huonw 

> > adds a multiplicity keyword arg into in_out_tensors for models with a multiplicity attribute

Some models like `attri2vec` could be created with a `generator` of any `multiplicity`, and then used to create either a link or a node model with `link_model` and `node_model` regardless of the `generator.multiplicity`. When I made `node_model` and `link_model` private, the `attri2vec` `build` function had no `multiplicity` kwarg so this got rid of the functionality of using any `attri2vec` instance to create a link and node model. I added the `multiplicity` kwarg for `attri2vec` and a couple of other models to keep this functionality.

> > deprecates default_model

I felt this fit into unifying the API to only use `in_out_tensors`, but I think you have a good point that this may be better suited to another PR.

kieranricardo:
@huonw good point about the deprecations, I think I'll go for the `deprecated_model_build` approach because it seems easier to extend to `node_model` and `link_model` than your 3rd suggestion and less effort than your first suggestion 👍

huonw:
> I felt this fit into unifying the API to only use in_out_tensors, but I think you have a good point that this may be better suited to another PR.

Yeah, it's easy to do as a separate PR; in fact, I'd already opened https://github.com/stellargraph/stellargraph/pull/1145 doing it before I'd seen this PR.

stellargraph/layer/graph_attention.py
```diff
@@ -843,6 +843,10 @@ def __init__(
                 )
             )
 
+        self.node_model = deprecated_model_function(self._node_model, "node_model")
```
huonw: Sorry to re-review before being suggested, but my original suggestion was careful to create these statically in each class. To be more specific, this would be like

```python
class GAT:
    ...
    def _node_model(self):
        ...

    node_model = deprecated_model_function(_node_model, "node_model")
    link_model = deprecated_model_function(_link_model, "link_model")
    build = deprecated_model_function(in_out_tensors, "build")
```

This approach should mean they're still shown in the documentation, unlike dynamically assigning them when building each instance. After doing this, and moving the doc string from `deprecated_model_function` to `_function_wrapper`, the docs look like:


image

Compared to the current approach of this PR of https://stellargraph--1140.org.readthedocs.build/en/1140/api.html#stellargraph.layer.graph_attention.GAT.in_out_tensors which just shows `in_out_tensors`.

kieranricardo: nice good point!

kieranricardo: @huonw my current approach isn't triggering deprecation warnings, can you see anything wrong?

huonw: > my current approach isn't triggering deprecation warnings, can you see anything wrong?

What do you mean by this specifically? How are you testing?

Per #1117, the `stacklevel` needs to be set correctly (although it looks like it is), and, even with this, they'll only show up in Jupyter, or in script run with Python 3.7 not Python 3.6.

kieranricardo: I'm testing some models in `test_deprecated_model_function` in `test_misc.py`, and I've also tried using the depracted functions in jupyter notebooks. The functions work but I'm not getting a warning

kieranricardo: i'm getting jupyter warnings now nevermind, must just be my test

tests/layer/test_misc.py
```diff
@@ -105,3 +108,91 @@ def test_squeezedsparseconversion_axis():
...
+
+    try:
+        x_in, x_out = sg_model._link_model()
+        with pytest.warns(DeprecationWarning):
+            x_in, x_out = sg_model.link_model()
+    except AttributeError:
+        pass
+
+
+def test_deprecated_model_functions():
```
huonw: Epic test!

stellargraph/layer/rgcn.py
```diff
@@ -575,11 +575,16 @@ def build(self):
...
-        if self.multiplicity == 1:
+        if multiplicity == 1:
             return self._node_model()
 
         else:
             raise NotImplementedError(
                 "Currently only node prediction if supported for RGCN."
             )
+
+    node_model = deprecated_model_function(_node_model, "node_model")
```
huonw: The `_node_model` method already wasn't public, so I'm not sure this is necessary?

demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb
```diff
@@ -509,7 +509,7 @@
     }
    ],
    "source": [
-    "x_inp, x_out = gcn.build()\n",
+    "x_inp, x_out = gcn.in_out_tensors()\n",
```
huonw: There's text above that should be updated ``via the `GCN.build` method:`` -> ``via the `GCN.in_out_tensors` method:``.

tests/layer/test_hinsage.py
```diff
@@ -270,41 +270,6 @@ def test_hinsage_apply():
     assert actual == pytest.approx(expected)
 
 
-def test_hinsage_default_model():
```
kjun9: Should we be trying to keep/update this test?
```python
def test_hinsage_in_out_tensors():
    ...
    xin, xout = hs.in_out_tensors()
    ...
```

huonw: Why delete this test? Despite the name, it was using `build`?

demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb
kieranricardo: nice catch! I found another one in the cluster gcn demo

tests/layer/test_hinsage.py
kieranricardo: ah I was too hasty to delete this one, I'll ad it back and rename it

kieranricardo: I was too hasty to delete this one, I'll ad it back and rename it

kieranricardo: @kjun9 (easier to reply to one thread)
actually I think this test is a duplicate of `test_hinsage_apply` now that we've removed `default_model`

stellargraph/layer/rgcn.py
kieranricardo: nice catch!

kieranricardo:
thanks for your speedy reviews @kjun9 and @huonw!

PR #1142, opened 2020-03-24
Replace kwargs with real args in graphsage

kjun9:
See #801


stellargraph/layer/graphsage.py
```diff
@@ -815,7 +830,16 @@ def __init__(
...
             self._get_sizes_from_generator(generator)
         else:
-            self._get_sizes_from_keywords(kwargs)
+            self.n_samples = _require_without_generator(n_samples, "n_samples")
+            self.input_feature_size = _require_without_generator(input_dim, "input_dim")
+            self.multiplicity = _require_without_generator(multiplicity, "multiplicity")
+            # Check the number of samples and the layer sizes are consistent
+            if len(self.n_samples) != self.max_hops:
+                raise ValueError(
+                    "Mismatched lengths: neighbourhood sample sizes {} versus layer sizes {}".format(
```
huonw: I think this check could be moved out of the `else` and out of `_get_sizes_from_generator`? Or, instead maybe reworded for each case, e.g. this one `"n_samples: expected one sample size for each of the {self.max_hops} layers, found {len(self._n_samples)} sample sizes"`

However, this is probably fine/better as separate clean-up work, given there's lots of error messages that can be tweaked here.

kjun9: Yeah I might just quickly reword this one with your suggestion, that seems slightly more helpful

PR #1146, opened 2020-03-24
Add VisibleDeprecationWarning for more obvious deprecations

huonw:
See: #1117


huonw:
Per https://github.com/stellargraph/stellargraph/issues/1117#issuecomment-616311317, we've decided not to do this because #1144 fixes the major problems.

PR #1150, opened 2020-03-24
Update API docs to include layer re-exports instead

kjun9:
`stellargraph.layer` re-exports various things in inner modules, so this updates our docs to document the re-exports instead.

This replaces the use of `automodule :members:` for each of the models, with `autoclass` instead, so that we can keep individual sections for each type of model while documenting all of them under the `layer` module.

I moved `KGTripleGenerator` to the generators section to avoid documenting it as `stellargraph.layer.KGTripleGenerator`, but maybe you intended it to be documented with the models? @huonw 

link for this PR's docs: https://stellargraph--1150.org.readthedocs.build/en/1150/api.html

Part of #713


huonw:
Nice! The docs look like they're closer to optimal, it seems the only remaining reexports that aren't handled are the `stellargraph.datasets.datasets` ones.

> This replaces the use of automodule :members: for each of the models, with autoclass instead, so that we can keep individual sections for each type of model while documenting all of them under the layer module.

Sphinx doesn't support multiple `automodule` sections for the same module, with different member lists?

> I moved KGTripleGenerator to the generators section to avoid documenting it as stellargraph.layer.KGTripleGenerator, but maybe you intended it to be documented with the models? @huonw

Yeah, I though it makes more sense to document functionality that will be used together, together. As it is, the list of generators is just a giant list of random bits of functionality that tie into later models, whereas it'd make more sense in my mind to have a "full batch methods" section (with `FullBatchNodeGenerator`, `FullBatchLinkGenerator` `GCN`, `GAT`, ..., possibly in sub-sections), a "sampling methods" section (with `GraphSAGENodeGenerator`, `HinSAGELinkGenerator`, `GraphSAGE`, ...), a "knowledge graph" section, etc. (Each of these might have subsections too, of course).

However, that's a bigger refactoring, that should can be done elsewhere, so I'm happy for this to switch it back to be more uniform.

kjun9:
> Sphinx doesn't support multiple automodule sections for the same module, with different member lists?

Doing something like:

```
section A
----------
.. automodule stellargraph.layer
  :members: A

section B
----------
.. automodule stellargraph.layer
  :members: B
```
complains about having `layer` module appear multiple times when building, and the module index doesn't get built correctly I think.

kjun9:
> it seems the only remaining reexports that aren't handled are the stellargraph.datasets.datasets ones.

I can add these here too?

huonw:
I guess so, but it seems like those are likely to be a different/simpler style and so better to do elsewhere?

kjun9:
> I guess so, but it seems like those are likely to be a different/simpler style and so better to do elsewhere?

I've opened https://github.com/stellargraph/stellargraph/pull/1157

docs/api.txt
```diff
@@ -19,106 +19,159 @@ Generators
 -----------
 
 .. automodule:: stellargraph.mapper
-  :members: FullBatchNodeGenerator, FullBatchLinkGenerator, GraphSAGENodeGenerator, DirectedGraphSAGENodeGenerator, DirectedGraphSAGELinkGenerator, ClusterNodeGenerator, GraphSAGELinkGenerator, HinSAGENodeGenerator, HinSAGELinkGenerator, Attri2VecNodeGenerator, Attri2VecLinkGenerator, RelationalFullBatchNodeGenerator, AdjacencyPowerGenerator, GraphWaveGenerator, CorruptedGenerator, GraphGenerator
+  :members: FullBatchNodeGenerator, FullBatchLinkGenerator, GraphSAGENodeGenerator, DirectedGraphSAGENodeGenerator, DirectedGraphSAGELinkGenerator, ClusterNodeGenerator, GraphSAGELinkGenerator, HinSAGENodeGenerator, HinSAGELinkGenerator, Attri2VecNodeGenerator, Attri2VecLinkGenerator, RelationalFullBatchNodeGenerator, AdjacencyPowerGenerator, GraphWaveGenerator, CorruptedGenerator, GraphGenerator, KGTripleGenerator
 
 
-GraphSAGE model
-----------------
+Layer
```
huonw: Should we call this:

```suggestion
Layers
```

or even

```suggestion
Layers and models
```

(Our other sections don't match the module names directly, like `Generators` for `stellargraph.mapper`).

kjun9:
Yeah I initially wanted have a go at achieving something like:
```
Layer
------

.. automodule stellargraph.layer
    # members with their docstrings updated via some custom code in `conf.py` 
    # to just say something like:
    # stellargraph.layer.GraphSAGE - alias of stellargraph.layer.graphsage.GraphSAGE
    :members:

GraphSAGE
^^^^^^^^^^^^

.. automodule stellargraph.layer.graphsage
    :members:

```
but decided it wasn't worth the effort in the end to figure out properly if that was feasible. Our demos refer to the re-exported classes instead of the inner modules so I still thought it was an improvement to have these documented instead.

PR #1157, opened 2020-03-24
Document datasets module reexports

kjun9:
This replaces `stellargraph.datasets.datasets.*` with `stellargraph.datasets.*` in our API docs

Part of #713 

See docs: https://stellargraph--1157.org.readthedocs.build/en/1157/api.html#module-stellargraph.datasets

### DIFF
--- a/stellargraph/datasets/__init__.py
+++ b/stellargraph/datasets/__init__.py
@@ -19,4 +19,19 @@ The datasets package contains classes to download sample datasets
 
 """
 
+__all__ = [
+    "Cora",
+    "CiteSeer",
+    "PubMedDiabetes",
+    "BlogCatalog3",
+    "MovieLens",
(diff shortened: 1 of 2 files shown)


stellargraph/datasets/__init__.py
```diff
@@ -19,4 +19,19 @@
 
 """
 
+__all__ = [
```
huonw: It's a little bit of a pest that we have to duplicate the list of definitions here. Could you say why it was needed just so I understand (and potentially add a comment, if it's subtle)?

kjun9: It seemed like `:members:` in `api.txt` looks for `__all__` - the docs for this section came out empty without this

kjun9: I've added a comment

Issue #1105, opened 2020-03-18
Deep Graph Infomax GraphSAGE


kieranricardo:
Add Deep Graph Infomax support for GraphSAGE.

Issue #1022, opened 2020-03-05
Hinsage link prediction demo「AttributeError: 'MovieLens' object has no attribute 'load'」


fangxi:
I use the demo ipynb file from demo/linkprediction/hinsage
`dataset = datasets.MovieLens()`
`display(HTML(dataset.description))`
`G, edges_with_ratings = dataset.load()`
show then error as below:

> AttributeError: 'MovieLens' object has no attribute 'load'

huonw:
Hi @fangxi, thanks for using StellarGraph and especially for filing an issue!

I suspect the problem here is the version of `stellargraph` doesn't match the version of the notebook you are running. Since you haven't listed it, I cannot tell what version of `stellargraph` you are using, but I guess it is the latest release `v0.10.0`? If that is the case, please try the notebook that was part of that release, because we've made changes since then: https://github.com/stellargraph/stellargraph/blob/v0.10.0/demos/link-prediction/hinsage/movielens-recommender.ipynb (note the `v0.10.0` in the path, instead of `develop`, which is the default).

For future reference, you can get to specific versions using the "Branch"/"Tag" dropdown menu on GitHub:

![image](https://user-images.githubusercontent.com/1203825/75944983-2341db00-5eed-11ea-8037-436dd296966e.png)

Or, with git on the command line with `git checkout v0.10.0` to jump to the tag for that release.

fangxi:
> Or, with git on the command line with `git checkout v0.10.0` to jump to the tag for that release.

Thanks for the help! git checkout works, I'm able to run the demo successfully.


However, when I try to use hinsage link prediction on my own dataset, in fit_generator stage, Index error occurs


To Reproduce, I use following code to build a simple heterogeneous graph and make hinsage prediction
```
# for test
user_nodes = pd.DataFrame(
    {"y": [1.5, 0.2, 3, 6]}, index=[3,4,5,6]
    
)

wifi_nodes = pd.DataFrame(
    {"x": [0.4, 0.1, 0.9]}, index=[0,1,2]
    
)

loan_edges = pd.DataFrame({"source":[3,4,5,6],"target":[0, 1, 2, 0]}, index=[38776,38777,38778,38779])
apply_edges = pd.DataFrame({"source":[5,3],"target":[0, 1]}, index=[38780,38781])


Gs = sg.StellarGraph(
    nodes = {'user':user_nodes,'wifikey':wifi_nodes},
    edges = {'loan':loan_edges,'apply':apply_edges})

[...]

Issue #987, opened 2020-02-28
Representation of a graph in an unsupervised manner using GCN


sbhttchryy:
Given a networkx graph G, and a subgraph in it called SG, I want to find an embedding for the subgraph SG, using a GCN. My aim is to preserve the structural information  and get an embedding for SG.  Is there any way that could be done with this library?

Thank you very much.

huonw:
Hi @sbhttchryy, similar to https://github.com/stellargraph/stellargraph/issues/693, we don't explicitly have support for graph/subgraph embeddings, but a technique like computing unsupervised embeddings for nodes and combining them (e.g. mean) would be a way to get started.

We have an example of computing unsupervised embeddings for nodes with GraphSAGE in https://github.com/stellargraph/stellargraph/blob/v0.10.0/demos/embeddings/embeddings-unsupervised-graphsage-cora.ipynb . I suspect a similar technique should work for GCN if you replace `GraphSAGELinkGenerator` &rarr; `FullBatchLinkGenerator`, and `GraphSAGE` &rarr; `GCN`.

We don't currently have support for RGCN link prediction (link prediction is how the unsupervised tasks work), but heterogeneous link prediction can be performed with HinSAGE, via `HinSAGELinkGenerator` and the `HinSAGE` model, and so the notebook above could be adapted to use those too.

If you're interested in RGCN, another algorithm that may work is `ComplEx`, designed for "knowledge graphs"/triple-stores (https://stellargraph.readthedocs.io/en/latest/api.html#module-stellargraph.mapper.knowledge_graph). It's currently experimental because we don't have an example for it, but the tests do show a bit about how it can be used https://github.com/stellargraph/stellargraph/blob/v0.10.0/tests/layer/test_knowledge_graph.py . This algorithm is unsupervised, in a sense, because it only works with the triples that exist in the graph.

huonw:
> heterogeneous link prediction can be performed with HinSAGE, via HinSAGELinkGenerator and the HinSAGE model, and so the notebook above could be adapted to use those too

Ah, just as a warning, there's potentially some care required when doing unsupervised HinSAGE, if there's multiple node types (https://github.com/stellargraph/stellargraph/issues/1022). However, if you're interested in RGCN, I am inferring that you have only one node type and so may not be affect.

PR #1162, opened 2020-03-25
Feature/1105 dgi graphsage

kieranricardo:
This PR implements `DeepGraphInfomax` for `GraphSAGE` and `DirectedGraphSAGE`. It makes the following additions:

- `CorruptedGenerator` and `CorruptedNodeSequence` support for `GraphSAGENodeGenerator` and `DirectedGraphSAGENodeGenerator`
- `DeepGraphInfomax` support for `GraphSAGENode` and `DirectedGraphSAGENode`
-  `GraphSAGE`, `APPNP`, and `GAT` options for the demo notebook
-  a `ReadOut` layer
- `multiplicity` arg for `GraphSAGE.in_out_tensors` for compatibility with `DeepGraphInfomax`

A note about the demo. The DGI paper doesn't test GraphSAGE on Cora and initial tests yielded poor results. To mimic a fullbatch GCN training on Cora I set the `num_samples` to `5` and the `batch_size` to `2700` (Cora has an average node degree of 4 and 2700 nodes) and this gives much better results (64% -> 74%).

See #1105


demos/embeddings/deep-graph-infomax-cora.ipynb
```diff
@@ -6,7 +6,7 @@
    "source": [
```
huonw: <p>Having this be conditional means that we're not testing the notebook with each of the algorithms, only whatever is hard coded, and requires extra conditionals to get the right generator (etc.). In this particular case, the latter isn't too bad, but it is very slightly confusing (i.e. a user reading this would have to "decode" it to be sure they're doing the right thing).</p><p> </p><p>An alternative would be to have this notebook do all/most of the algorithms directly, as a comparison of their performance. It could potentially be framed something like</p><p> </p><ul><li>teach how to use DGI with one algorithm (GCN for simplicity, or whichever gives the best results) with all of the explanations, and demonstrate the downstream task(s) like node classification and TSNE</li><li>run just the DGI -&gt; node classification training task for all of the other algorithms, only explaining whatever differences (since there's less need for explanation, the cells could be larger, with less intermediate output, so the notebook doesn't explode in size).</li><li>Give a comparison summary table/dataframe of the results for each of them at the end as a conclusion</li></ul><p> </p><p>What do you think?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

huonw: <p>It looks like the model is created and compiled twice? Is this just an artefact from ReviewNB?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

stellargraph/mapper/sequences.py
```diff
@@ -653,25 +653,48 @@ class CorruptedNodeSequence(Sequence):
...
+            shuffled_feats = stacked_feats.reshape(-1, features[0].shape[-1])
+            shuffled_idxs = np.random.permutation(shuffled_feats.shape[0])
+            shuffled_feats = shuffled_feats[shuffled_idxs, :]
+            shuffled_feats = shuffled_feats.reshape(stacked_feats.shape)
+            shuffled_feats = np.split(
+                shuffled_feats, np.cumsum([y.shape[1] for y in features])[:-1], axis=1
+            )
 
-        return [shuffled_feats] + inputs, self.targets
+            return shuffled_feats + features, self.targets[: stacked_feats.shape[0], :]
```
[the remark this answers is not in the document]
kieranricardo: If the batch size doesn't exactly divide the number of nodes then the last batch of an epoch will be smaller - `num_nodes % batch_size`. So that slicing it just there for the last epoch. It would be alot nicer though if this wasn't the case!

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -145,8 +181,10 @@ def embedding_model(self):
         # these tensors should link into the weights that get trained by `build`
         x_emb_in, x_emb_out = self.base_model.in_out_tensors(multiplicity=1)
 
-        squeeze_layer = Lambda(lambda x: K.squeeze(x, axis=0), name="squeeze")
-        x_emb_out = squeeze_layer(x_emb_out)
+        # squeeze out batch dim of full batch models
+        if len(x_emb_out.shape) == 3:
```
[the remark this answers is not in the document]
kieranricardo: The non-full batch methods output `batch_size x embedding_dim` tensors, whereas the fullbatch methods output `1 x num_output_nodes x embeddings`.

good idea! I'll deprecate in this PR

stellargraph/mapper/sequences.py
```diff
@@ -653,25 +653,48 @@ class CorruptedNodeSequence(Sequence):
...
+
+            features = inputs[0]
+            shuffled_idxs = np.random.permutation(features.shape[1])
+            shuffled_feats = features[:, shuffled_idxs, :]
+
+            return [shuffled_feats] + inputs, self.targets
+
+        else:
+
+            features = inputs
```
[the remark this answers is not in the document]
kieranricardo: yep that's right, its the approach the paper uses (page 7, **Inductive earning on large graphs**)

stellargraph/mapper/sequences.py
```diff
@@ -653,25 +653,48 @@ class CorruptedNodeSequence(Sequence):
...
+            stacked_feats = np.concatenate(features, axis=1)
```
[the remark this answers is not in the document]
kieranricardo: yep that's right.

Wow that's quite a speed up, nice!

stellargraph/mapper/sequences.py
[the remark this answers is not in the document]
huonw: I would've thought that it would already be sliced by the `GraphSAGENodeGenerator` to have a matching length, though, because the targets array should have exactly one row for each node in the batch?

huonw:
> It seems like we can/should generalise this to HinSAGE to give us unsupervised HinSAGE (#987, #1022) as future work, but we should probably consider how we separate the specialized-to-each-class corruption logic from DeepGraphInfomax and CorruptedSequence before we do that. Could you file an issue for both of these?

I filed #1173 for HinSAGE + DGI.

demos/embeddings/deep-graph-infomax-cora.ipynb
kieranricardo: yeah I think you're right that this might be confusing - especially for new users. I'll make the first section just about one algo, and then do a comparison summary to demonstrate that DGI can be used for a lot of different algos 👍

stellargraph/layer/deep_graph_infomax.py
```diff
@@ -76,9 +108,16 @@ class DeepGraphInfomax:
 
     def __init__(self, base_model):
 
-        if not isinstance(base_model, (GCN, GAT, APPNP, PPNP)):
+        if isinstance(base_model, (GCN, GAT, APPNP, PPNP)):
```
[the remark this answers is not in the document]
kieranricardo: There's definitely room for better abstraction/generalization! I quite like the `GraphModel` approach with `corruptible_input_indices` but agree that its better to keep this PR focused on GraphSAGE.

stellargraph/mapper/sequences.py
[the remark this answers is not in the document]
kieranricardo: The `targets` array in this case is created in `CorruptedNodeSequence`, and is more like a batch targets array (because each node has the exact same targets).

demos/embeddings/deep-graph-infomax-cora.ipynb
kieranricardo: <p>nah that's a real bug :(</p>

kieranricardo: I've added this and the results for APPNP and GAT are 30% and 40% respectively which seem a bit low - i'll investigate and try and figure out whats going on 🤔

stellargraph/mapper/sequences.py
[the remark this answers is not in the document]
huonw: Oh, right!

demos/embeddings/deep-graph-infomax-cora.ipynb
kieranricardo: APPNP just applies a MLP and then does nieghbood smoothing, so I think it makes sense that thie doesnt work for APPNP. Not sure about GAT though

huonw: Could you explain a bit more about your reasoning why that process means APPNP "should" fail? I feel like one-layer GCN like used here is fairly similar conceptually (squashing the MLP and neighbourhood smoothing into a single step) and that seems to go ok?

huonw: <p>Maybe we could be more specific about those code changes, e.g.</p><p> </p><blockquote>only requires changing the generator to be appropriate for each algorithm, and adjusting the data dimensions for full batch methods</blockquote>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

huonw: <ul><li>Instead of "squeeze batch dim out" maybe the comment could be adjusted to something like "for full batch models, squeeze out the batch dim (which is 1)". (I guess this also applies to the comment in the GCN version above.)</li><li>Why a batch size of 2700 for GraphSAGE? I think that results in two batches, one of size 2700 and one of size 8? Maybe we could go with either a less unusual number (like 1000), or definitely the full graph (<code>G.number_of_nodes()</code> )?</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

huonw: <p>Maybe GAT is struggling with a small number of attention heads (<code>attn_heads</code> defaults to 1, I think)?</p><p> </p><p>My original suggestion was to write each section out as its own cells to make these sort of per-model tweaks easier. One approach would be to have a function that takes a model and a generator, e.g.</p><p> </p><blockquote>The process for training with Deep Graph Infomax is almost identical for every model, requiring only the model and generator to be wrapped in <code>DeepGraphInfomax</code>  and <code>CorruptedGenerator</code>  respectively. We can define a function to do this for us, to make it easy to train different models for a comparison. The StellarGraph library ensures that the process for predicting with each different model is similar (using the <code>flow</code>  methods), so the functions can include the evaluation code too. The only per-model customisation required in this function is removing the batch dimension for full batch methods like GCN, GAT and APPNP.</blockquote><pre class="ql-syntax" spellcheck="false">def run_deep_graph_infomax(base_model, generator):
&nbsp; &nbsp; corrupted_generator = CorruptedGenerator(generator)
&nbsp; &nbsp; gen = corrupted_generator.flow(G.nodes())

&nbsp; &nbsp; infomax = DeepGraphInfomax(base_model)
&nbsp; &nbsp; x_in, x_out = infomax.in_out_tensors()

&nbsp; &nbsp; model = Model(inputs=x_in, outputs=x_out)
&nbsp; &nbsp; model.compile(loss=tf.nn.sigmoid_cross_entropy_with_logits, optimizer=Adam(lr=1e-3))

&nbsp; &nbsp; history = model.fit(gen, epochs=epochs, verbose=0, callbacks=[es])

&nbsp; &nbsp; x_emb_in, x_emb_out = base_model.in_out_tensors()

&nbsp; &nbsp; # for full batch models, squeeze out the batch dim (which is 1)
&nbsp; &nbsp; if isinstance(base_model, (GAT, GCN, APPNP)):
&nbsp; &nbsp; &nbsp; &nbsp; x_emb_out = tf.squeeze(x_emb_out, axis=0)

&nbsp; &nbsp; emb_model = Model(inputs=x_emb_in, outputs=x_emb_out)

&nbsp; &nbsp; test_gen = generator.flow(test_subjects.index)
&nbsp; &nbsp; train_gen = generator.flow(train_subjects.index)

&nbsp; &nbsp; test_embeddings = emb_model.predict(test_gen)
&nbsp; &nbsp; train_embeddings = emb_model.predict(train_gen)

&nbsp; &nbsp; lr = LogisticRegression(multi_class="auto", solver="lbfgs")
&nbsp; &nbsp; lr.fit(train_embeddings, train_subjects)

&nbsp; &nbsp; y_pred = lr.predict(test_embeddings)
&nbsp; &nbsp; acc = (y_pred == test_subjects).mean()

&nbsp; &nbsp; return acc
</pre><p> </p><p>and then it can be used for each of them like follows (with this we wouldn't have to re-run GCN and could reuse the full-batch generator):</p><p> </p><pre class="ql-syntax" spellcheck="false">gat_model = GAT(layer_sizes=[512], activations=["relu"], attn_heads=8, generator=generator)
gat_acc = run_deep_graph_infomax(gat_model, generator)
</pre><p> </p><pre class="ql-syntax" spellcheck="false">appnp_model = APPNP(layer_sizes=[512], activations=["relu"], generator=generator)
appnp_acc = run_deep_graph_infomax(appnp_model, generator)
</pre><p> </p><pre class="ql-syntax" spellcheck="false">graphsage_generator = GraphSAGENodeGenerator(G, batch_size=2700, num_samples=[5])
graphsage_model = GraphSAGE(layer_sizes=[512], activations=["relu"], generator=generator)
graphsage_acc = run_deep_graph_infomax(graphsage_model, generator)
</pre><p> </p><p>And then the final results would use the various <code>*_acc</code> variables.</p><p> </p><p>(It might make sense to rename the GCN <code>generator</code> to <code>full_batch_generator</code>, if we go with this.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

huonw:
> It seems like we can/should generalise this to HinSAGE to give us unsupervised HinSAGE (#987, #1022) as future work, but we should probably consider how we separate the specialized-to-each-class corruption logic from DeepGraphInfomax and CorruptedSequence before we do that. Could you file an issue for both of these?

I filed https://github.com/stellargraph/stellargraph/issues/1176 about the corruption logic.

stellargraph/layer/deep_graph_infomax.py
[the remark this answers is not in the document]
huonw: https://github.com/stellargraph/stellargraph/issues/1176 (linked below too, but here to keep the thread consistent)

demos/embeddings/deep-graph-infomax-cora.ipynb
kieranricardo: I reafctored the notebook like you suggested, looks much nicer 👍 I decreased layer size and epochs here to make the whole thing run abit faster and because I didn't notice much of a drop in performance. 

I'm still getting the low accuracy for GAT w/ the larger number of heads tho :(

kieranricardo: aside from this results issue I think its ready for a re-review

huonw, line 6: <p>I know you mentioned choosing a lower number of epochs to make this run faster, but this kinda looks like the loss is still decreasing?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

huonw, line 6: <p>This isn't part of this PR directly (and so doesn't need to be fixed here), but I heard there were "standard" train/test/val splits for Cora? Do you know anything about this?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

huonw, line 6: <p>This isn't part of this PR directly (and so doesn't need to be fixed here), but it looks like this number is pretty close to the supervised task: https://github.com/stellargraph/stellargraph/blob/develop/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb</p><p> </p><p>Maybe we could mention this; something like (please edit!)</p><pre class="ql-syntax" spellcheck="false">This accuracy is close to that for [training a supervised GCN model end-to-end](../node-classification/gcn/gcn-cora-node-classification-example.ipynb), suggesting that Deep Graph Infomax is an effective method for unsupervised training.
</pre><p>(Note the relative link is important, so that it works in jupyterlab/colab etc.)</p><p> </p><p>Of course the setup is a bit different here; this has a only a single GCN layer with a much large embedding size, vs the two 16-dim GCN layers in the supervised one.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

huonw, line 6: <p>I don't think we need to create a new <code>FullBatchNodeGenerator</code> for each of these methods and instead reuse the one defined above (potentially after renaming it); or do you think it's clearer for users this way?</p><p> </p><p>Also, per the suggestion earlier, I think it would better to not overwrite <code>base_model</code>, because that means a user can't jump around and easily rerun cells; if they run the whole notebook and then want to go back to an earlier cell (e.g. adjust the logistic regression parameters for the first GCN), they won't be able to do it properly without recreating the right <code>base_model</code> variable. Specifically, I think we can call this <code>gat_model</code>, with <code>gcn_model</code> earlier and <code>graphsage_model</code> later. The same applies for the <code>generator</code> variables.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

huonw, line 6: <p>I think we can just save the <code>gcn_acc</code> from the first section and not have to rerun with GCN? I.e. adjust the second "Extracting Embeddings and Logistic Regression" cell to end with:</p><p> </p><pre class="ql-syntax" spellcheck="false">gcn_acc = (y_pred == test_subjects).mean()

print(f"Test classification accuracy: {gcn_acc}")
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

kieranricardo:
update on eager execution fixes:

For some reason calling `sess.run(...)` within `Sequence.__getitem__()` causes an error. To get around this I changed any `Sequence.__getitem__()` that called `node_features` to call `node_features_tensors` and natively process the tensors. This was made harder by the fact that almost all models don't test the `fit` function! i'll make an issue to add this in.

demos/embeddings/deep-graph-infomax-cora.ipynb
kieranricardo: great idea for the relevant link! I didn't edit in the end bc I think your comment states that nicely :)

kieranricardo: bupmed epochs upto 100 and its near convergence now.

kieranricardo: yeah I think that's what I used when I replicated the paper's results and I think its what most papers use. but I don't know (well haven't looked into) how/where to get these splits

huonw:
Nice!

PR #1165, opened 2020-03-25
Fix Conda package dependencies and test failures

timpitman:
`conda build` fails at the test stage due to moved saliency_maps code - the test imports section needs to be updated due to API changes this release.

In addition, the requirements are very old and haven't been correct for the last few Conda releases of StellarGraph - TensorFlow needs to be updated to version 2.1.   Unfortunately, this is only possible currently for Windows and Linux - Anaconda have not yet released a Mac version of TensorFlow 2.1: https://github.com/tensorflow/tensorflow/issues/35754

A fresh attempt at `meta.yaml` was generated using `conda skeleton pypi stellargraph` (with the v0.11 already on PyPI) - suggested package changes were then made to the original `meta.yaml`.

See: #1166

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## [0.11.1](https://github.com/stellargraph/stellargraph/tree/v0.11.1)
+
+### Bug fixes and other changes
+
+- The [Conda package for StellarGraph](https://anaconda.org/stellargraph/stellargraph) has been updated to require TensorFlow 2.1, as TensorFlow 2.0 is no longer supported.  This means that StellarGraph will currently install via Conda on Linux and Windows - Mac support is waiting on the [Tensorflow 2.1 release to Conda](https://github.com/tensorflow/tensorflow/issues/35754). [\#1165](https://github.com/stellargraph/stellargraph/pull/1165)
+
 ## [0.11.0](https://github.com/stellargraph/stellargraph/tree/v0.11.0)
 
 [Full Changelog](https://github.com/stellargraph/stellargraph/compare/v0.10.0...v0.11.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bug fixes and other changes
 
-- The [Conda package for StellarGraph](https://anaconda.org/stellargraph/stellargraph) has been updated to require TensorFlow 2.1, as TensorFlow 2.0 is no longer supported.  This means that StellarGraph will currently install via Conda on Linux and Windows - Mac support is waiting on the [Tensorflow 2.1 release to Conda](https://github.com/tensorflow/tensorflow/issues/35754). [\#1165](https://github.com/stellargraph/stellargraph/pull/1165)
+- The [Conda package for StellarGraph](https://anaconda.org/stellargraph/stellargraph) has been updated to require TensorFlow 2.1, as TensorFlow 2.0 is no longer supported.  As a result, StellarGraph will currently install via Conda on Linux and Windows - Mac support is waiting on the [Tensorflow 2.1 release to Conda](https://github.com/tensorflow/tensorflow/issues/35754). [\#1165](https://github.com/stellargraph/stellargraph/pull/1165)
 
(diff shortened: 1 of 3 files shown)


meta.yaml
```diff
@@ -16,38 +16,38 @@ requirements:
...
     - gensim >=3.4.0
     - matplotlib >=2.2
-    - networkx >=2.2,<2.4
+    - networkx >=2.2
     - numpy >=1.14
     - pandas >=0.24
     - pip
     - python
-    - scikit-learn >=0.20
+    - scikit_learn >=0.20
```
kjun9: Shouldn't this still be kebab case?
```suggestion
    - scikit-learn >=0.20
```
and for line 34 too

timpitman: this form actually came out of the automated `conda skeleton pypi stellargraph` but have fixed now.

timpitman:
Have discovered that Conda doesn't support tensorflow 2.1 on Mac yet https://github.com/tensorflow/tensorflow/issues/35754 therefore it's not really possible for us to release this - it can't actually build!  the last few releases on Condaof stellargraph don't even work.

timpitman:
I've now tested this branch on Linux, and it successfully builds via Conda and installing the resulting build passes tests.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## [0.11.1](https://github.com/stellargraph/stellargraph/tree/v0.11.1)
+
```
huonw: Maybe we could provide a short blurb here that makes it clear that this is just fixing the anaconda issues?

```suggestion

This bugfix release contains the same code as 0.11.0, and just fixes the metadata in the Anaconda package so that it can be installed successfully.

```

Or fold it into the other text. My concern is that a user wouldn't directly understand the consequence (and how it affects them) of the fix currently described.

meta.yaml
```diff
@@ -23,7 +23,7 @@ requirements:
     - python
     - scikit-learn >=0.20
     - scipy >=1.1.0
-    - tensorflow 1.14.*
+    - tensorflow >=2.1.0
     - ipython
     - ipykernel
   run:
```
huonw: As a question not directly related to this PR, do we really need to duplicate so much between the `host` and `run` sections?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+### Bug fixes and other changes
+
+- The [Conda package for StellarGraph](https://anaconda.org/stellargraph/stellargraph) has been updated to require TensorFlow 2.1, as TensorFlow 2.0 is no longer supported.  As a result, StellarGraph will currently install via Conda on Linux and Windows - Mac support is waiting on the [Tensorflow 2.1 release to Conda](https://github.com/tensorflow/tensorflow/issues/35754). [\#1165](https://github.com/stellargraph/stellargraph/pull/1165)
```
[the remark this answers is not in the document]
timpitman: This is a good idea - cleaner than a closed issue which tensorFlow themselves can't fix - I've filed https://github.com/ContinuumIO/anaconda-issues/issues/11697

meta.yaml
timpitman: Conda's documentation and examples aren't as easy to follow as I'd like - I was thinking the `run_exports` option might help, but I think it could cause other problems.

huonw:
Oh, wait, merging with `develop` isn't correct for this PR because it will pull in everything that has landed since 0.11. I think the minimal fix for that problem (#1170) is to add a `USER neo4j` line to the `stellargraph-neo4j` Dockerfiles, i.e. make

https://github.com/stellargraph/stellargraph/blob/94d28efc23719c0bb95077ce37f35b557463e518/docker/stellargraph-neo4j/Dockerfile#L3

look like

```dockerfile
SHELL ["/bin/bash", "-euo", "pipefail", "-c"]

USER neo4j
```

somewhat similar to the develop version.

https://github.com/stellargraph/stellargraph/blob/c80ee17e75a892ef072ececd006d112dadf8bacb/docker/stellargraph-neo4j/Dockerfile#L1-L8

PR #1167, opened 2020-03-27
Store features as TensorFlow tensors, not NumPy arrays

kieranricardo:
This PR stores features as a dictionary of `Tensor`s instead of a dictionary of `np.array`s, and uses `tf.nn.embedding_lookup` to access the features. 

Testing with GraphSAGE on CORA using my machine (MacOS + 9th gen quad core i7) and the following hyperparameters the training time decreases from 10s to 3s per epoch.

```
batch_size = 200
num_samples = [20, 20]
generator = GraphSAGENodeGenerator(G, batch_size, num_samples)
train_gen = generator.flow(node_subjects.index, targets, shuffle=True)
graphsage = GraphSAGE(
generator=generator, layer_sizes=[32, 32], activations=["relu", "relu"]
```

~See #1168~
See: #1228


stellargraph/core/graph.py
```diff
@@ -734,7 +735,7 @@ def node_features(self, nodes, node_type=None):
             node_type (hashable): the type of the nodes.
 
         Returns:
-            Numpy array containing the node features for the requested nodes.
+            tensorflow Tensor containing the node features for the requested nodes.
```
huonw: I think we should consider carefully how we execute this transition; because this may break downstream users catastrophically. For instance, we could leave the default behaviour as returning numpy arrays but have a `tensor=True` flag. Do you have thoughts?

kieranricardo: @kjun9 I think this is the same point you raised. Yeah I think you're both right that we need to leave to a default option to return numpy arrays, I'm slightly more inclined to go with Kevin's idea of two seperate functions - so we know what type each function returns.

kieranricardo:
@huonw 

- Actually I think this break anything that uses `.numpy()` in eager mode. I'll add a fix for this
- We should see a speedup for `HinSAGE` but I haven't tested this. `GCN` shouldn't see any change because it castes the features to a `numpy.array` but we can/should change the types in `FullBatchSequence` to `tf.Tensor`

stellargraph/core/graph.py
```diff
@@ -743,19 +744,27 @@ def node_features(self, nodes, node_type=None):
         all_valid = valid.all()
         valid_ilocs = node_ilocs if all_valid else node_ilocs[valid]
 
+        # check the types of the input nodes
+        types = np.unique(self._nodes.type_of_iloc(valid_ilocs))
```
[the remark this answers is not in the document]
kieranricardo: good point, I'll revert this function back and update the error catching in `NodeData`

stellargraph/core/graph.py
```diff
@@ -773,6 +782,7 @@ def node_features(self, nodes, node_type=None):
         sampled = self._nodes.features(node_type, valid_ilocs)
         features = np.zeros((len(nodes), sampled.shape[1]))
         features[valid] = sampled
+        features = tf.convert_to_tensor(features)
```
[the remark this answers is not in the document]
kieranricardo: nice!

kieranricardo:
bit of an update re non eager execution. So now all the `test_graph.py` and `test/layer` tests pass (except those tests that rely on `.numpy()` for comparison) in non-eager mode. 

I ran into a problem where running `model.fit(...)` with a `Sequence` that calls `.node_features()` (which is now distinct from `.node_features_tensors`)  in `Sequence.__getitem__()` in non-eager mode  throws this error:

`RuntimeError: The Session graph is empty. Add operations to the graph before calling run().`

To make things worse most models dont actually test `.fit`! To get around this error I swapped out `.node_features()` for `.node_features_tensors()` and replaced any numpy feaature processing with tensorflow, so now almost all of our feature processing is done in tensorflow with the exception of full batch methods.

I also removed a redundant `node_features` call in `ClusterNodeGenerator`.

stellargraph/core/element_data.py
```diff
@@ -279,8 +281,10 @@ def features(self, type_name, id_ilocs) -> np.ndarray:
             raise ValueError("unknown IDs")
 
         try:
-            return self._features[type_name][feature_ilocs, :]
-        except IndexError:
+            return tf.nn.embedding_lookup(
+                self._features[type_name], feature_ilocs.astype(int),
+            )
+        except InvalidArgumentError:
```
huonw, line 287: I think `tf.nn.embedding_lookup` may not throw this error when executing on a GPU, e.g. https://www.tensorflow.org/api_docs/python/tf/compat/v1/nn/embedding_lookup

> validate_indices: DEPRECATED. If this operation is assigned to CPU, values in indices are always validated to be within range. If assigned to GPU, out-of-bound indices result in safe but unspecified behavior, which may include raising an error.

Although the non-compat version https://www.tensorflow.org/api_docs/python/tf/nn/embedding_lookup doesn't discuss this at all.

We might have to switch to manual verification, like:

```python
iloc_range = self._type_element_ilocs[type_name
feature_ilocs = id_ilocs - iloc_range.start

if (feature_ilocs < 0).any() or (feature_ilocs >= len(iloc_range)).any():
    raise ValueError("unknown IDs")

return tf.nn.embedding_lookup(...)
```

Hopefully this doesn't destroy the wins we get from using `embedding_lookup`.

I guess we could potentially switch the subtraction, validation and type casting logic to use tensorflow too, maybe as a `tf.function`? (maybe this whole function should be a `tf.function`?)

It would be good to report the results of `test_benchmark_get_features` (in `test_graph.py`) before and after each refactoring. In the best case, we'd report the progress on GPU as well as on CPU.

stellargraph/core/element_data.py
```diff
@@ -279,8 +281,10 @@ def features(self, type_name, id_ilocs) -> np.ndarray:
             raise ValueError("unknown IDs")
 
         try:
-            return self._features[type_name][feature_ilocs, :]
-        except IndexError:
+            return tf.nn.embedding_lookup(
+                self._features[type_name], feature_ilocs.astype(int),
```
huonw, line 285: Why does this now need casting?

stellargraph/core/graph.py
```diff
@@ -756,8 +758,19 @@ def node_features(self, nodes, node_type=None):
 
             node_type = types[0]
 
+        # handles eager execution and non-eager execution
+        if tf.executing_eagerly():
```
huonw, line 762: Hm, this is pretty complicated... and I suspect the `else` case is catastrophically slow? (Something that can be verified with the `test_benchmark_get_features`.)

Do you think this is particularly valuable to support?

stellargraph/core/graph.py
```diff
@@ -722,7 +724,7 @@ def check_graph_for_ml(self, features=True):
 
         # TODO: check the feature node_ids against the graph node ids?
 
-    def node_features(self, nodes, node_type=None):
+    def node_features(self, nodes, node_type=None) -> np.ndarray:
```
huonw, line 727: This is a large amount of duplication of subtle code!

Maybe this one could be entirely replaced with something like:

```python
return self.node_features_tensor(nodes, node_type).numpy()
```

tests/layer/test_hinsage.py
```diff
@@ -573,15 +573,21 @@ def test_hinsage_from_generator():
...
 
     def transform_neighbours(neighs, dim):
-        return np.expand_dims(
-            neighs.reshape(1, dim, int(neighs.shape[1] / dim), neighs.shape[2]).sum(
-                axis=-1
+        return tf.reduce_mean(
+            tf.expand_dims(
+                tf.reduce_sum(
+                    tf.reshape(
+                        neighs, (1, dim, int(neighs.shape[1] / dim), neighs.shape[2])
```
huonw, line 580: This is some serious [Hadouken indentation](http://i.imgur.com/BtjZedW.jpg)!

Maybe:

```python
new_shape = (1, dim ,...)
summed = tf.reduce_sum(tf.reshape(neighs, new_shape)))
return tf.reduce_mean(tf.expand_dims(summed, -1), axis=2)
```

stellargraph/core/graph.py
kieranricardo: yeah the `else` is catastrophically slow here :( 

I we dont support it then `subgraph` and `to_networkx` won't work in non-eager mode. fullbatch methods also wouldn't work as is but refactoring them to handle tensors natively would fix this

stellargraph/core/graph.py
kieranricardo: unfortunately that won't work for non eager mode - but if we decide to not support it then yeah definitely!

stellargraph/core/graph.py
kieranricardo: i was planning on doing the fullbatch refactoring in a seperate PR but maybe its worth just doing it here, thoughts?

tests/layer/test_hinsage.py
kieranricardo: haahaha i hadn't heard of that before! that some impressive nesting

stellargraph/core/graph.py
huonw: We could wrap it in the `Session()` magic that you're doing inside this function, I think?

stellargraph/core/graph.py
huonw: Unless it's very complicated, I think it's fine to do here; if this PR was (at least, in our heads) reframed to be "switch StellarGraph to storing tensorflow tensors instead of numpy arrays".

stellargraph/core/graph.py
kieranricardo: good idea!

stellargraph/core/element_data.py
kieranricardo: `node_ilocs` are of type `uint8`, unfortunately `embedding_lookup` only support `int32` and `int64`

kieranricardo:
@huonw it looks like tensorflow will eagerly store the constants on GPU, I'll do manual device placement to ensure everything is stored on CPU by default. This should also make it pretty easy if we decide to support an option for storing features and adj matrices on GPU

and I'll be sure to report the benchmarks 👍

kieranricardo:
```
------------------------------------------------------------------------------------------ benchmark 'StellarGraph node features': 8 tests -------------------------------------------------------------------------------------------
Name (time in us)                                              Min                   Max                Mean             StdDev              Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_get_features[specify-4] (0001_develop)      74.0510 (1.0)        315.5710 (1.0)       79.9056 (1.0)      16.3182 (1.0)       76.1150 (1.0)       1.9910 (1.0)       278;711       12.5148 (1.0)        6118           1
test_benchmark_get_features[specify-1] (0001_develop)      74.4400 (1.01)       354.4430 (1.12)      81.6790 (1.02)     18.1429 (1.11)      77.3600 (1.02)      2.2555 (1.13)     554;1230       12.2431 (0.98)      10357           1
test_benchmark_get_features[infer-1] (0001_develop)       141.4190 (1.91)       555.2240 (1.76)     153.1598 (1.92)     27.6718 (1.70)     146.6435 (1.93)      3.7310 (1.87)      135;230        6.5291 (0.52)       2362           1
test_benchmark_get_features[infer-4] (0001_develop)       141.8080 (1.92)       445.1760 (1.41)     157.0700 (1.97)     31.6779 (1.94)     146.9510 (1.93)      7.0322 (3.53)      240;419        6.3666 (0.51)       2893           1
test_benchmark_get_features[specify-1] (0002_sampler)     159.8100 (2.16)       731.9850 (2.32)     197.1657 (2.47)     52.2096 (3.20)     180.2670 (2.37)     14.7425 (7.40)      465;717        5.0719 (0.41)       4123           1
test_benchmark_get_features[specify-4] (0002_sampler)     163.4500 (2.21)       646.4190 (2.05)     196.6817 (2.46)     53.6475 (3.29)     176.0180 (2.31)     13.0950 (6.58)      254;408        5.0844 (0.41)       2384           1
test_benchmark_get_features[infer-4] (0002_sampler)       236.5120 (3.19)     1,274.0850 (4.04)     283.4384 (3.55)     73.6335 (4.51)     255.3900 (3.36)     21.4935 (10.80)     340;507        3.5281 (0.28)       2947           1
test_benchmark_get_features[infer-1] (0002_sampler)       251.5710 (3.40)       717.6690 (2.27)     285.6156 (3.57)     71.6478 (4.39)     256.8270 (3.37)     13.6642 (6.86)        55;89        3.5012 (0.28)        513           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
(tf2-stellar) Kierans-MacBook-Pro:stellargraph kieranricardo$ pytest-benchmark compare 0001_develop 0002_sampler-speedup

------------------------------------------------------------------------------------------ benchmark 'StellarGraph node features': 8 tests -------------------------------------------------------------------------------------------
Name (time in us)                                              Min                   Max                Mean             StdDev              Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_get_features[specify-4] (0001_develop)      74.0510 (1.0)        315.5710 (1.0)       79.9056 (1.0)      16.3182 (1.0)       76.1150 (1.0)       1.9910 (1.0)       278;711       12.5148 (1.0)        6118           1
test_benchmark_get_features[specify-1] (0001_develop)      74.4400 (1.01)       354.4430 (1.12)      81.6790 (1.02)     18.1429 (1.11)      77.3600 (1.02)      2.2555 (1.13)     554;1230       12.2431 (0.98)      10357           1
test_benchmark_get_features[infer-1] (0001_develop)       141.4190 (1.91)       555.2240 (1.76)     153.1598 (1.92)     27.6718 (1.70)     146.6435 (1.93)      3.7310 (1.87)      135;230        6.5291 (0.52)       2362           1
test_benchmark_get_features[infer-4] (0001_develop)       141.8080 (1.92)       445.1760 (1.41)     157.0700 (1.97)     31.6779 (1.94)     146.9510 (1.93)      7.0322 (3.53)      240;419        6.3666 (0.51)       2893           1
test_benchmark_get_features[specify-1] (0002_sampler)     159.8100 (2.16)       731.9850 (2.32)     197.1657 (2.47)     52.2096 (3.20)     180.2670 (2.37)     14.7425 (7.40)      465;717        5.0719 (0.41)       4123           1
test_benchmark_get_features[specify-4] (0002_sampler)     163.4500 (2.21)       646.4190 (2.05)     196.6817 (2.46)     53.6475 (3.29)     176.0180 (2.31)     13.0950 (6.58)      254;408        5.0844 (0.41)       2384           1
test_benchmark_get_features[infer-4] (0002_sampler)       236.5120 (3.19)     1,274.0850 (4.04)     283.4384 (3.55)     73.6335 (4.51)     255.3900 (3.36)     21.4935 (10.80)     340;507        3.5281 (0.28)       2947           1
test_benchmark_get_features[infer-1] (0002_sampler)       251.5710 (3.40)       717.6690 (2.27)     285.6156 (3.57)     71.6478 (4.39)     256.8270 (3.37)     13.6642 (6.86)        55;89        3.5012 (0.28)        513           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


```

stellargraph/core/element_data.py
kieranricardo: good news for us, if run on a gpu it will return zero tensors with the right shape for invalid and negative indices. so we don't need to change the verification here.

I'm gonna look at using a `tf.function` for this

kieranricardo:
I tried using a `tf.function` and I got a slight slowdown - this is even after refactoring the benchmark to call the function once before benchmarking to build the tf graph.
```
---------------------------------------------------------------------------------- benchmark 'StellarGraph node features': 4 tests ----------------------------------------------------------------------------------
Name (time in us)                               Min                 Max                Mean             StdDev              Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_get_features[specify-4]     205.8430 (1.0)      571.4800 (1.0)      217.4520 (1.0)      24.1863 (1.0)      210.0280 (1.0)       3.9815 (1.13)      196;443        4.5987 (1.0)        2604           1
test_benchmark_get_features[specify-1]     206.7150 (1.00)     823.0610 (1.44)     219.3296 (1.01)     29.7938 (1.23)     210.4320 (1.00)      3.5337 (1.0)       193;480        4.5593 (0.99)       2565           1
test_benchmark_get_features[infer-1]       288.6790 (1.40)     633.8360 (1.11)     310.7310 (1.43)     35.9383 (1.49)     294.5550 (1.40)     21.5960 (6.11)      177;178        3.2182 (0.70)       1413           1
test_benchmark_get_features[infer-4]       289.6960 (1.41)     759.3240 (1.33)     306.4269 (1.41)     34.6038 (1.43)     294.5040 (1.40)      6.8915 (1.95)      120;287        3.2634 (0.71)       1520           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

kieranricardo:
```
------------------------------------------------------------------------------------------- benchmark 'StellarGraph node features': 8 tests -------------------------------------------------------------------------------------------
Name (time in us)                                              Min                   Max                Mean              StdDev              Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_get_features[specify-1] (0002_develop)     197.5870 (1.0)        459.5120 (1.0)      209.4526 (1.0)       26.3674 (1.0)      203.2815 (1.0)       5.2440 (1.0)       246;492        4.7743 (1.0)        3990           1
test_benchmark_get_features[specify-4] (0002_develop)     198.4220 (1.00)       641.0320 (1.40)     212.1275 (1.01)      30.1935 (1.15)     204.7575 (1.01)      5.3080 (1.01)      239;551        4.7141 (0.99)       3876           1
test_benchmark_get_features[infer-4] (0002_develop)       274.1250 (1.39)       656.1530 (1.43)     290.3003 (1.39)      35.0277 (1.33)     279.5615 (1.38)      6.7170 (1.28)      220;333        3.4447 (0.72)       2772           1
test_benchmark_get_features[infer-1] (0002_develop)       275.2780 (1.39)       619.7210 (1.35)     295.6064 (1.41)      38.6036 (1.46)     285.1620 (1.40)      6.5800 (1.25)      136;275        3.3829 (0.71)       1717           1
test_benchmark_get_features[specify-1] (0001_sampler)     346.6450 (1.75)       636.0980 (1.38)     381.2819 (1.82)      59.4843 (2.26)     353.8430 (1.74)     33.3838 (6.37)        17;19        2.6227 (0.55)        173           1
test_benchmark_get_features[specify-4] (0001_sampler)     377.9050 (1.91)     5,683.8260 (12.37)    536.5454 (2.56)     684.0001 (25.94)    442.0600 (2.17)     47.1105 (8.98)         3;15        1.8638 (0.39)        167           1
test_benchmark_get_features[infer-4] (0001_sampler)       459.1140 (2.32)     5,793.4770 (12.61)    636.4216 (3.04)     696.6652 (26.42)    531.1770 (2.61)     49.3570 (9.41)         3;16        1.5713 (0.33)        163           1
test_benchmark_get_features[infer-1] (0001_sampler)       546.4290 (2.77)       915.3580 (1.99)     597.7073 (2.85)     112.8566 (4.28)     556.1980 (2.74)     37.0220 (7.06)          1;1        1.6731 (0.35)         10           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

huonw:
> It seems like we only get the tensorflow speedup if its used in a Sequence object

Do you have any thoughts about why this might be? That seems a bit peculiar to me?

stellargraph/mapper/full_batch_generators.py
```diff
@@ -117,7 +117,7 @@ def __init__(
             self.use_sparse = sparse
 
         # Get the features for the nodes
-        self.features = G.node_features(self.node_list)
+        self.features = G.node_features_tensors(self.node_list)
```
huonw, line 120: Does this get computed eagerly and realised into memory here? (This is just a general question, and doesn't/shouldn't need changes here.)

kieranricardo:
> Do you have any thoughts about what the mechanism might be here? That seems a bit peculiar to me?

Huh, the mystery runs a bit deeper. Check out this timing code:

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```
from stellargraph.datasets import Cora
import tensorflow as tf
import numpy as np

G, targets = Cora().load()

%time _ = [real_array[lookup, :].shape for _ in range(10)]
%time _ = [tf.nn.embedding_lookup(real_tensor, lookup).shape for _ in range(10)]
```

**Output:**

```
CPU times: user 34.1 ms, sys: 154 µs, total: 34.3 ms
Wall time: 34.4 ms
CPU times: user 7.1 ms, sys: 719 µs, total: 7.82 ms
Wall time: 3.1 ms
```

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


So tf is faster, cool, buuut when I run this:

```
fake_array = np.random.random(real_array.shape).astype(real_array.dtype)
fake_tensor = tf.constant(fake_array)

%time _ = [fake_array[lookup, :].shape for _ in range(10)]
%time _ = [tf.nn.embedding_lookup(fake_tensor, lookup).shape for _ in range(10)]
```

**Output:**

```
CPU times: user 3.06 ms, sys: 50 µs, total: 3.11 ms
Wall time: 3.15 ms
CPU times: user 8.48 ms, sys: 739 µs, total: 9.21 ms
Wall time: 3.55 ms
```

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Somehow numpy is now 10x faster than it was before on an array with the same type and size?!?!

huonw:
Good catch! 

In any case, in terms of actually substantial things, the type and shape aren't the only properties of a numpy array; there's [flags](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.flags.html) and, probably more relevantly, strides that describes the data order:

```python
real_array.strides # (4, 10832)
fake_array.strides # (5732, 4)
```

That is, the elements of the first axis of `real_array` are placed contiguously (4 bytes apart, which is the size of a float32) (Fortran or F order), while the elements of `fake_array`'s *second* axis are placed contiguously in memory (C order). That is the arrays are arranged like

```
real: [N0F0, N1F0, N2F0, ..., N0F1, N1F1, ..., N2707F1432]
fake: [N0F0, N0F1, N0F2, ... N1F0, N1F1, ..., N2707F1432]
```

(This can be additionally verified by `1433 * 4 == 5732` and `2708 * 4 == 10832` and by noting the `C_CONTIGUOUS` and `F_CONTIGUOUS` flags via `real_array.flags` and similarly for `fake_array`.)

This means that selecting the features for a node from `real` has to copy floats individually, stepping by 10kb each time, while from `fake` can copy a whole row with an efficient loop (which can use SIMD instructions to copy up to 64 bytes at a time). The memory/cache behaviour of these are very different and the latter is much more likely to get close to maxing out the CPU's memory bandwidth (specifically, a read of any memory will load a whole 64-byte cache line, so the `real_array` form uses only 4/64 = 1/16th of the data that is transferred).

All in all, I wonder if the performance improvement we see here on CPU is just because TensorFlow is switching the layout to C-order, rather than F-order. I've opened #1225 that does the switch only, which doesn't show any change on our microbenchmarks but does improve something like your GraphSAGE code from 10.3s to 4.33s (see #1225 for the full benchmark to copy-paste so we can compare this PR exactly).

I wonder if we should land #1225, but keep working on this because getting TensorFlow end-to-end, especially for full-batch-on-GPU methods, would be awesome.

---

That said, this layout difference doesn't explain the difference between the sequence form and the isolated form. I wonder if our benchmark is just not doing the same thing as the GraphSAGE-on-Cora example (e.g. smaller feature data?) and so if there's some increased constant with TF, it'll look slower.

---

Minor thing, but using `%timeit` (with the `-n` argument) is better than `%time`, e.g. it avoids needing to do the manual loop and gives a sense of the variance (https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-timeit). In this case, it tells me basically the same thing:

```python
%timeit -n 20 real_array[lookup, :].shape
# 3.56 ms ± 210 µs per loop (mean ± std. dev. of 7 runs, 20 loops each)
%timeit -n 20 tf.nn.embedding_lookup(real_tensor, lookup).shape
# 677 µs ± 51.5 µs per loop (mean ± std. dev. of 7 runs, 20 loops each)
%timeit -n 20 fake_array[lookup, :].shape
# 693 µs ± 92.9 µs per loop (mean ± std. dev. of 7 runs, 20 loops each)
%timeit -n 20 tf.nn.embedding_lookup(fake_tensor, lookup).shape
# 670 µs ± 31.6 µs per loop (mean ± std. dev. of 7 runs, 20 loops each)
```

kieranricardo:
awesome explanation @huonw, thanks for figuring this one out!

Yeah merging #1225 sounds good, its only one line of code for a huge speedup!

stellargraph/mapper/sequences.py
```diff
@@ -656,16 +662,19 @@ def __init__(self, base_generator):
...
-            self.targets = np.tile([1.0, 0.0], reps=(base_generator.batch_size, 1))
+            batch_reps = (base_generator.batch_size, 1)
+            target = tf.constant([[1.0, 0.0]], dtype=tf.float32)
         else:
             raise TypeError(
                 f"base_generator: expected FullBatchSequence, SparseFullBatchSequence, "
                 f"or NodeSequence, found {type(base_generator).__name__}"
             )
+        with tf.device("/CPU:0"):
+
```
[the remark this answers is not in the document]
kieranricardo: yep bad formatting on my part

stellargraph/mapper/full_batch_generators.py
kieranricardo: AFAIK in eager mode yeah that get realised into memory here, but in graph mode I think these stay as symbolic tensors and every time a keras model accesses them it runs the feature lookup code. 

I think in the case of `model.fit(..., epochs=something > 1)` the tensorflow graph optimizer recognizes that this op's output is constant and realizes the symbolic tensor as a constant tensor for the entirety of the `.fit` call. So tensorflow only does the feature lookup at the start of each `fit` or `predict` call.

stellargraph/layer/cluster_gcn.py
```diff
@@ -372,7 +372,7 @@ def in_out_tensors(self):
             and `x_out` is a tensor for the Cluster-GCN model output.
         """
         # Placeholder for node features
-        N_feat = self.generator.features.shape[1]
+        N_feat = self.generator.feature_size
```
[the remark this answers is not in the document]
kieranricardo: yep I think that will 👍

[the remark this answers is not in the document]
kieranricardo: not in graph mode unfortunately. When we initialize the  layers `tf.shape(self.generator.features)[1]` is still a symbolic tensor so it gives this error:

```
TypeError: Error converting shape to a TensorShape: Dimension value must be integer or None or have an __index__ method, got <tf.Tensor 'strided_slice_12:0' shape=() dtype=int32>.
```

when creating the input layers

stellargraph/mapper/sequences.py
```diff
@@ -674,31 +683,46 @@ def __getitem__(self, index):
...
+                shuffled_feats = list(tf.split(shuffled_feats, size_of_splits, axis=1))
+
+                targets = tf.squeeze(
+                    tf.gather(self.targets, indices=[tf.range(head_nodes)]), axis=0
+                )
 
         return shuffled_feats + inputs, targets
+
+
+def _get_tf_dim(tensor, axis):
```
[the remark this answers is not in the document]
kieranricardo: `tf.shape(tensor)[axis]` works even in graph mode 👍

stellargraph/core/graph.py
```diff
@@ -736,6 +738,33 @@ def node_features(self, nodes, node_type=None):
         Returns:
             Numpy array containing the node features for the requested nodes.
         """
+        feature_tensor = self.node_features_tensors(nodes, node_type)
+
+        if tf.executing_eagerly():
+            feature_arr = feature_tensor.numpy()
```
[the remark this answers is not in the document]
kieranricardo: yeah I think that's clearer 👍

stellargraph/mapper/sequences.py
```diff
@@ -674,31 +683,46 @@ def __getitem__(self, index):
...
-            shuffled_feats = [features[:, shuffled_idxs, :]]
-            targets = self.targets
+            if isinstance(
+                self.base_generator, (FullBatchSequence, SparseFullBatchSequence)
+            ):
+                features = inputs[0]
 
-        else:
+                shuffled_feats = [
+                    tf.expand_dims(
```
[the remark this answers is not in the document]
kieranricardo: yeah I like that, might make it easier to generalize to graph classification layer on 👍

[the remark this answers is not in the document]
huonw: Another (better?) option would be a pair of `tf.reshape` calls similar to the GraphSAGE code, maybe?

stellargraph/core/element_data.py
kieranricardo: At least as it is, this function is always running on CPU and so will always return an exception.

stellargraph/mapper/full_batch_generators.py
huonw: Cool, the latter sounds like exactly what we want 🏎

kieranricardo:
> Also, I'm a bit concerned about all the `with tf.device(...):`s that this requires....

Yeah I thought about allowing the users to just specify what device they want but the majority of the time I think people will go with the defaults which could lead to unexpected function behaviour when something is run on GPU. e.g. with the `embedding_lookup` function this function has different behaviour on GPU which we don't test at all! Having different default behaviour of our functions on different systems which we don't test seems risky to me. What do you think?

Also these `with device(..)..` calls won't be necessary when everything is wrapped in a `tf.data.Dataset` which AFAIK automatically defaults to CPU with everything. So this PR puts us into a bit of an awkward spot but it will get better in the future! 

One possibility could be to keep this PR open, merge the other tensorflow PRs into this PR, and then merge to develop once everything is native. That seems like a horrendous and difficult merge in two weeks time but maybe its worth it. Maybe this is really bad practice? Thoughts?

stellargraph/core/element_data.py
huonw: Right, but that's only because we're forcing it to run on CPU. I'll move this discussion to the main PR.

stellargraph/connector/neo4j/mapper.py
```diff
@@ -144,17 +145,17 @@ def sample_features(self, head_nodes, batch_num):
...
-            self.graph.node_features(layer_nodes, node_type)
-            for layer_nodes in nodes_per_hop
-        ]
-
-        # Resize features for sampled nodes
-        batch_feats = [
-            np.reshape(a, (len(head_nodes), -1 if np.size(a) > 0 else 0, a.shape[1]))
-            for a in batch_feats
-        ]
+        with tf.device("/CPU:0"):
```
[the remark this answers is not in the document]
kieranricardo: missed it, nice catch!

huonw:
>  with the embedding_lookup function this function has different behaviour on GPU which we don't test at all

Following on from https://github.com/stellargraph/stellargraph/pull/1167#discussion_r402047664, can we just switch the behaviour to always be like the GPU one? That is, always just fill in zeros? It's not obvious to me if that's possible via the `embedding_lookup` API directly, but maybe there's some variation we can use that's close enough?

>  function has different behaviour on GPU which we don't test at all! Having different default behaviour of our functions on different systems which we don't test seems risky to me. What do you think?

I think it's a very valid concern, and one that probably applies even now. There may be GPU vs. CPU behaviours already that we're not aware of, because we're not testing it. I've opened #1227 to cover this.

> One possibility could be to keep this PR open, merge the other tensorflow PRs into this PR, and then merge to develop once everything is native. That seems like a horrendous and difficult merge in two weeks time but maybe its worth it. Thoughts?

That sounds like too much work to be worth it, unfortunately. I wonder if we can do most of #1206 while leaving the features as NumPy arrays? I think you mentioned that tf datasets only use CPU anyway?

---

Also, since #1225 has now merged and done the major part of #1168, I've opened https://github.com/stellargraph/stellargraph/issues/1228 and switched this PR to addressing that specifically.

kieranricardo:
> I wonder if we can do most of #1206 while leaving the features as NumPy arrays

I think that since #1225 is merged this PR adds a lot of extra complexity for very little (if any) performance gain. I'm going to sideline this for now and look at changing things over to `tf` datasets while keeping our backend in numpy/pandas until the last moment.

huonw:
Sounds good to me. We can maybe roll this out progressively, where it makes sense, e.g., start using `tf.<operation>` instead of `np.<operation>`.

PR #1174, opened 2020-03-30
1042-unit tests for graph classification

PantelisElinas:
Hi @kjun9 

I have added some unit tests for the graph classification model and related classes. Can you have a look, let me know if I missed anything?

This relates to #1045 

Regards,

P.


tests/layer/test_gcnsupervisedgraphclassification.py
```diff
@@ -0,0 +1,118 @@
...
+        model = GCNSupervisedGraphClassification(
+            layer_sizes=[16], activations=["relu"], generator=self.generator
+        )
+
+        assert len(model.layer_sizes) == 1
+        assert len(model.activations) == 1
+        assert model.layer_sizes[0] == 16
+        assert model.activations[0] == "relu"
+
+        with pytest.raises(TypeError):
```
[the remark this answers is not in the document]
huonw: It's not just good for reading and understanding, but it also helps ensure the test continues to check the right thing in future. If there's changes that modify the order in which things are checked or other changes to what's expected of parameters (e.g. `layer_sizes` switches to needing a tuple instead of a list).

tests/layer/test_gcnsupervisedgraphclassification.py
```diff
@@ -0,0 +1,118 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from stellargraph.layer.graph_classification import *
+from stellargraph.mapper import GraphGenerator, FullBatchNodeGenerator
+import pytest
+from ..test_utils.graphs import example_graph_random
+
+
+class Test_GCNSupervisedGraphConvolution:
```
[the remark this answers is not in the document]
huonw: Given the file already incorporates the type name, and, classes in pytest are purely an organisational thing (https://docs.pytest.org/en/latest/goodpractices.html#test-discovery "methods inside Test prefixed test classes **(without an __init__ method)**", i.e. the class can't do any special initialisation when being built), I think we could also just not have the class at all here or in the other file.

Since the `graphs` and `generator` values are small/cheap they can be promoted to the top-level and use `generator` instead of `self.generator`.  (If they were expensive to build, they should become fixtures so that they're only executed for tests that actually need them, not every time pytest searches for tests (e.g. #791); this applies with or without the class.)

(This is a small nit.)

stellargraph/mapper/graph_generator.py
```diff
@@ -96,6 +96,16 @@ def flow(self, graph_ilocs, targets=None, batch_size=1, name=None):
                     f"expected targets to be the same length as node_ids, found {len(targets)} vs {len(graph_ilocs)}"
                 )
 
+        if not isinstance(batch_size, int):
+            raise TypeError(
+                f"expected batch_size to be integer type, found {type(batch_size)}"
```
huonw: I often forget this too, but `type(...)` returns a class value that doesn't format very nicely, and one instead has to pull out the (string) name manually:

```suggestion
                f"expected batch_size to be integer type, found {type(batch_size).__name__}"
```

```python
>>> batch_size = 1.2
>>> print(f"expected batch_size to be integer type, found {type(batch_size)}")
expected batch_size to be integer type, found <class 'float'>
>>> print(f"expected batch_size to be integer type, found {type(batch_size).__name__}")
expected batch_size to be integer type, found float
```

tests/layer/test_gcnsupervisedgraphclassification.py
```diff
@@ -0,0 +1,118 @@
...
+
+        x_in, x_out = model.in_out_tensors()
+
+        assert len(x_in) == 3
+        assert len(x_in[0].shape) == 3
+        assert x_in[0].shape[-1] == 4  # the node feature dimensionality
+        assert len(x_out.shape) == 2
+        assert x_out.shape[-1] == layer_sizes[-1]
+
+    def test_stateful(self):
```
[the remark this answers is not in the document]
huonw: (#1088 is probably related.)

tests/mapper/test_graph_generator.py
```diff
@@ -0,0 +1,138 @@
...
+
+class Test_GraphGenerator:
+
+    graphs = [
+        example_graph_random(feature_size=4, n_nodes=6),
+        example_graph_random(feature_size=4, n_nodes=5),
+        example_graph_random(feature_size=4, n_nodes=3),
+    ]
+
+    graphs_nx = [
```
huonw: Small nit: this and the next two are only used in a single test each; maybe they can be just defined within the relevant test rather than scoped globally?

tests/mapper/test_graph_generator.py
```diff
@@ -0,0 +1,138 @@
...
+        generator = GraphGenerator(graphs=self.graphs)
+
+        seq = generator.flow(graph_ilocs=[0, 1, 2], batch_size=2)
+        assert isinstance(seq, GraphSequence)
+
+        assert len(seq) == 2  # two batches
+
+        # The first batch should be size 2 and the second batch size 1
+        batch_0 = seq[0]
+        assert batch_0[0][0].shape[0] == 2
```
huonw: If you wanted, you could reduce how much code there is to maintain here by incorporating `test_generator_flow_check_padding` into this one. Specifically, instead of just checking one element of each shape, check the whole tuple, like `.shape == (2, 6, ...)` here and `.shape == (2, 3, ...)` for `batch_1`. 

In addition to reducing the code, I think this would also be good as check that the tensors are in the expected order, because we'll be checking the whole shapes and so capture the differences between the feature and adjacency matrices too.

tests/mapper/test_graph_generator.py
```diff
@@ -0,0 +1,138 @@
...
+        # The largest graph has 6 nodes vs 3 for the smallest one.
+        # Check that the data matrices have the correct size 6
+        batch = seq[0]
+
+        assert batch[0][0].shape[1] == 6
+        assert batch[0][1].shape[1] == 6
+        assert batch[0][2].shape[1] == 6
+
+        for mask in batch[0][1]:
+            assert np.sum(mask) == 6 or np.sum(mask) == 3
```
huonw: Given we're passing in the graph ilocs in a fixed order, I think we know which mask is which and so can check them directly?

```suggestion
        assert np.sum(batch[0][1][0]) == 6
        assert np.sum(batch[0][1][1]) == 3
```

PantelisElinas: Currently, this order cannot be guaranteed because when calling `flow` a `GraphSequence` object is created and the last statement in `GraphSequence.__init__` is to call `on_epoch_end` that shuffles the graph indexes. So, when the single batch is requested in the test, the order of the graphs within the batch is unknown. Maybe when #1043 is solved we can revisit this test.

huonw: Ah, right 👍

tests/mapper/test_graph_generator.py
PantelisElinas: I can't compare the entire shape because of the random shuffling of the graphs when `on_epoch_end` is called; there is no guarantee of the size for the second item in the shape of the arrays. This is because the latter is set based on the largest graph in the batch and since the first batch has 2 graphs we don't know which two of sizes 6, 5, or 3 are in the batch. Essentially, the second value in batch_0[0][0].shape could be either 6 or 5. Similarly, for batch_1 we don't know which of the 3 graphs was left over for the second batch so it could be any of 6, 5, or 3 depending on which two are in the first batch.

PantelisElinas: Also, I prefer to have a separate test for padding because then it is clear what the test is for and also I use a different sequence generator with 2 graphs with guaranteed maximum size. Although, in this case I can check the full shapes so I will update accordingly.

PantelisElinas:
@kjun9 

> ...but I think it would be good to have one to cover that we're actually transforming the data correctly - maybe just a test for `GraphSequence` object to check that `__getitem__` actually returns data that we expect given a particular set of graphs?

Actually, there are tests for this behaviour in `test_graph_generator.py` where the sizes of the batches are checked. See for example `test_generator_flow_check_padding` and `test_generator_flow_no_targets`.

PantelisElinas:
@kevin @huonw 

can I remove the experimental tags if/when you approve this pull request?

huonw:
> I did look through the code to see if any aspect of this "single style" seemed particularly relevant for naming.. but it seems rather generic to me, almost like a Default/Uniform/Standard/Basic, maybe we'd have a better idea once we have a second example of a graph generator?

FYI @kjun9, https://github.com/stellargraph/stellargraph/pull/1053#discussion_r391962238 is relevant (as is https://github.com/stellargraph/stellargraph/pull/1053#discussion_r390102345 more tangentially).

huonw:
https://github.com/stellargraph/stellargraph/issues/1198

PR #1175, opened 2020-03-31
Fix broken tests for upcoming tensorflow 2.2 upgrade

kjun9:
This addresses two issues from upgrading:
* `Model.compile` no longer throws ValueError for bad values for loss and metrics. It seems like metrics and loss are now built lazily when you call `Model.fit`
* shuffling was implemented incorrectly for `PaddedGraphSequence` but this wasn't caught prior to the tensorflow upgrade for some reason

See #1078


kjun9:
`test_stateful in tests.layer.test_gcn_supervised_graph_classification` is now failing with Tensorflow 2.2 RC3 https://buildkite.com/stellar/stellargraph-public/builds/3087

huonw:
It seems like both of the fixes here should work with tf 2.1, could we land them without the `tensorflow==2.2.0rc3` version pinning? Now that we've got #1122/#1248, we'll be testing against the RC every morning  and so don't need a separate PR for it.

kjun9:
Sure I can do that

Issue #1169, opened 2020-03-27
Integrate and run on OGB datasets


huonw:
### Description

... [Insert a brief description of the actions involved to complete this task.]

### User Story

**As** a ...

**I want to** [accomplish some things]

**So that** [achieve a larger goal]

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
Hm, it seems the OGB datasets:

- can be quite large (N = https://ogb.stanford.edu/docs/nodeprop, L = https://ogb.stanford.edu/docs/linkprop/, G = https://ogb.stanford.edu/docs/graphprop/, ... = <https://snap.stanford.edu/ogb/data>)

   | task | dataset  | size (MB) | data file url                                                                                                                       |
   |------|----------|----------:|---------------------------------------------------------------------------------------------------------------------------|
   | N    | proteins |       212 | [.../nodeproppred/proteinfunc_v3.zip](https://snap.stanford.edu/ogb/data/nodeproppred/proteinfunc_v3.zip)                 |
   | N    | products |      1461 | [.../nodeproppred/products.zip](https://snap.stanford.edu/ogb/data/nodeproppred/products.zip)                             |
   | L    | ppa      |       391 | [.../linkproppred/ppassoc_v4.zip](https://snap.stanford.edu/ogb/data/linkproppred/ppassoc_v4.zip)                         |
   | G    | hiv      |         2 | [.../graphproppred/csv_mol_download/hiv.zip](https://snap.stanford.edu/ogb/data/graphproppred/csv_mol_download/hiv.zip)   |
   | G    | pcba     |        38 | [.../graphproppred/csv_mol_download/pcba.zip](https://snap.stanford.edu/ogb/data/graphproppred/csv_mol_download/pcba.zip) |
   | G    | ppi      |      2991 | [.../graphproppred/ogbg_ppi_medium.zip](https://snap.stanford.edu/ogb/data/graphproppred/ogbg_ppi_medium.zip)             |

- are very slow to download; I tried downloading the a few of files above with wget: 
[...]

PR #1177, opened 2020-03-31
Add loaders for some OGB datasets

huonw:
See: #1169


huonw:
The 1.0.1 versions that we were testing with here failed on CI in the OGB code itself, because that used too much memory for the testing instances; before even getting to the StellarGraph code.

There's a been a few recent releases of OGB (see https://github.com/snap-stanford/ogb/releases), I don't know if this will help.

In any case, this is stalled and lower priority for now, so closing.

Issue #1178, opened 2020-03-31
Consider updating random walkers to optionally take hyperparameters upon construction rather than in `run`.


kjun9:
### Description

Related: https://github.com/stellargraph/stellargraph/issues/919 and https://github.com/stellargraph/stellargraph/pull/536#discussion_r400583830

It seems like it's useful to be able to instantiate a random walker object that already has its hyperparameters defined - for example, being able to pass a walker object with defined hyper-parameters, to be used by `UnsupervisedSampler`, seems like a more compelling use-case, compared to being able to instantiate the walker and use `run` with different hyperparameters. 

The way we use `SampledBreadthFirstWalk` objects in our generators also doesn't make use of being able to use `run` with different parameters each time (only the set of `nodes` changes).

Issue #919, opened 2020-02-21
Allow UnsupervisedSampler to take "walker" as a parameter


ericbroda:
### Description

UnsupervisedSampler is limited to using "UniformRandomWalk" as the node walker/sampler.  It would be ideal to allow the walker to be selected/provided by the developer.

### User Story

As a developer I would like flexibility in choosing which walker I can use, and also be able to write my own walkers and have them used by UnsupervisedSampler

So that I can techniques to improve model performance

### Suggested Solution:

1. Update signature of UnsupervisedSampler __init__ method to include an optional "walker" parameter:
~~~~
def __init__(self, G, nodes=None, length=2, number_of_walks=1, seed=None, walker=None):
~~~~

2.  Modify code (line 59) in __init__  to use the provided walker as follows:
~~~~
# self.walker = UniformRandomWalk(G, seed=seed)
if walker:
    self.walker = walker
else:
    self.walker = UniformRandomWalk(G, seed=seed)
~~~~

Developers that wish to use this functionality would now call UnsupervisedSampler as follows:
~~~~
: 
: 
from stellargraph.data.explorer import BiasedRandomWalk
walker = BiasedRandomWalk(Gs, seed=42)
unsupervisedSamples = UnsupervisedSampler(Gs, nodes=labels, length=walk_length, number_of_walks=num_walks, seed=42, walker=walker)
: 
: 
~~~~

kjun9:
Thanks for your suggestion @ericbroda ! I agree with you that it makes sense to allow the walker being used to run the unsupervised sampler be flexible/implemented by a user.

I can think of a few more pieces of work on top of your suggested solution that would make this feature more useable long term:
- a custom walker from the user should probably be implementing one or two methods in an existing parent class in the library, to guide the user how a "walker" should behave. Currently this parent class would be `GraphWalk` but it's being used too broadly at the moment and is not really designed as an interface that can readily be implemented by users. I'm thinking it might make sense split it into two different classes `GraphWalk` and `GraphSampler` and go from there (happy to explain this more if you're interested)
[...]

PR #1179, opened 2020-03-31
Allow instantiating random walk classes with pre-defined hyper-parameters

kjun9:
This updates the `__init__` and `run` methods for the random walker classes:
* `UniformRandomWalk`
* `BiasedRandomWalk`
* `UniformRandomMetaPathWalk`
* `TemporalRandomWalk`

which are all supposed to be suitable for obtaining context pairs for unsupervised learning. This refactoring would allow users to instantiate a walker object with its hyperparameters pre-defined.

If this seems reasonable, I was hoping to do a bit more follow-up to make #919 and #536 possible:
* define a separate class for random walkers instead of `GraphWalk` (which currently also contains BFS-type walkers for the SAGE algorithms) - maybe just `RandomWalk`?
* update `UnsupervisedSampler` to optionally take a `RandomWalk` object as a parameter.

Thoughts?

See #1178

### DIFF
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -42,7 +42,7 @@ def _default_if_none(value, default, name, ensure_not_none=True):
     value = value if value is not None else default
     if ensure_not_none and value is None:
         raise ValueError(
-            f"{name}: expected a value to be specified in either `__init__` or `run`, found: None."
+            f"{name}: expected a value to be specified in either `__init__` or `run`, found None in both"
         )
     return value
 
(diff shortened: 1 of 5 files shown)


stellargraph/data/explorer.py
```diff
@@ -38,6 +38,15 @@
 from ..random import random_state
 
 
+def _default_if_none(value, default, name, ensure_not_none=True):
+    value = value if value is not None else default
+    if ensure_not_none and value is None:
+        raise ValueError(
+            f"{name}: expected a value to be specified in either `__init__` or `run`, found: None."
```
huonw: Minor tweak, potentially:

```suggestion
            f"{name}: expected a value to be specified in either `__init__` or `run`, found None in both"
```

stellargraph/data/explorer.py
```diff
@@ -179,22 +188,38 @@ def _check_sizes(self, n_size):
 class UniformRandomWalk(GraphWalk):
     """
     Performs uniform random walks on the given graph
+
+    Args:
+        graph (StellarGraph): Graph to traverse
+        graph_schema (GraphSchema, optional): Graph schema
```
huonw: I think none of these classes use `graph_schema` at all, so we could potentially not accept the argument in each of them (theoretically we would have `GraphWalk` not call `create_graph_schema` too, which I guess could happen as part of splitting out `RandomWalk`). That is

```suggestion
```

and
```python
def __init__(self, graph, n=None, length=None, seed=None):
    super().__init__(graph, graph_schema=None, seed=seed) # or super().__init__(graph, seed=seed)
```

and similarly for each of the other classes.

Thoughts?

kjun9:
> because those want to pass around an object that can just have .run(start_nodes) called without also having to pass through all the possible arguments for each run

Yep! As it stands, it won't work for `TemporalRandomWalk.run` since its parameters are a bit different, so would have to think of something

stellargraph/data/explorer.py
kjun9: Makes sense, I'll update them

Issue #1037, opened 2020-03-06
Investigate storing to Neo4j


huonw:
### Description

With #1034, we'll understand somewhat what it takes to load data from Neo4j. We should do the same for storing data to Neo4j. We can think about things like:

- putting data in from scratch (that is, starting with data in a `StellarGraph` and maybe some ML results, and filling out an empty Neo4j graph). For example, one might want it for visualisation purposes.
- adding ML results on to an existing graph (potentially having read that graph out of Neo4j, via #1034), potentially in different forms:
  - node attribute inference
  - link prediction, adding new links
  - embeddings (likely similar to node attribute inference)

A good goal might be to demonstrate all of this functionality in a notebook, like #1034.

### User Story

**As** a data scientist

**I want to** write data into Neo4j

**So that** it can be used in neo4j-based workflows and tooling

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
https://dzone.com/articles/tips-for-fast-batch-updates-of-graph-structures-wi is potentially a good quick/high-level intro to ways to make this fast.

huonw:
We now have a brief discussion of this at the end of the demo notebook https://github.com/stellargraph/stellargraph/blob/develop/demos/basics/loading-saving-neo4j.ipynb introduced in #1034 and #1187 in the "Saving predictions into Neo4j" section. It just uses `UNWIND` + `SET` to start with.

Issue #1034, opened 2020-03-06
Have a demo notebook of loading data from neo4j


huonw:
### Description

It would be good to investigate how hard it is to load data from Neo4j into a StellarGraph object for loading. This might include things like:

- filtering only a subgraph
- loading features
- doing feature transformations

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
The existing graphsage notebooks demonstrate at least loading features from Neo4j, but not edges.

PR #1184, opened 2020-03-31
Add a demo notebook that demonstrates loading from and saving to neo4j

huonw:
This adds a notebook to `basics/` that demonstrates a few ways one might load graphs and subgraphs from Neo4j into memory (as Pandas DataFrames) and then convert those into `StellarGraph` objects. It follows basically the same structure as the Pandas and NetworkX ones, except has an additional "save results" section at the end too.

See: #1034 and #1037


demos/basics/loading-saving-neo4j.ipynb
```diff
@@ -0,0 +1,2038 @@
+{
```
kjun9, line 1: <blockquote>We can use single Cypher query to retrieve an identifier for the source and target of each node is sufficient.</blockquote><p>"We can use a single Cypher query to retrieve the identifiers of the source and target nodes of each edge."?</p><p> </p><blockquote>ensure the DataFrame columns have the easiest name.</blockquote><p>"ensure the DataFrame columns match the defaults"?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1184/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

kjun9, line 1: <blockquote>and one type of edge that touches it.</blockquote><p>maybe "and one type of edge between them"? or "one type of edge connecting them"?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1184/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

kjun9, line 1: <blockquote>One way sort of a subgraph someone might be interested is one where the nodes and/or edges satisfy certain criteria.</blockquote><p>"One type of subgraph someone might be interested..."?</p><p> </p><p>Also in general, this <code>Subgraphs</code>  sections feels more like a cypher tutorial to me, rather than anything <code>stellargraph</code>  specific - I understand that it comes from us believing it would be a common use-case since you probably wouldn't load ALL your neo4j data into memory, but the exact way the user should reduce their graph is probably going to be very specific to their data. Thoughts?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1184/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='49'/>

huonw: Wow, I really mangled that first sentence!

huonw: This ~whole notebook is a cypher tutorial, since it's actually not doing anything StellarGraph-y that isn't already explained in `loading-pandas.ipynb`. I agree that the exact reduction is specific to their data and their use case, but I feel that "filter the elements (in isolation)" and "k-hop subgraphs" are likely to be the most common (or at least, serve as close-enough examples) and so we can spell them out to get people started, and reduce the barriers to working with StellarGraph.

kjun9: > This ~whole notebook is a cypher tutorial, since it's actually not doing anything StellarGraph-y that isn't already explained in loading-pandas.ipynb.

I guess so, but there's some relationship between the query being run and the way the stellargraph constructor is used that the notebook is demonstrating (e.g. run a query to obtain heterogeneous graph then use it to create a heterogeneous stellargraph), whereas for the subgraph one feels more like it's purely in cypher (i.e. stellargraph doesn't know/care that it's a subgraph). That said, I'm fine with keeping it since I agree it's likely to be a common use-case.

huonw: Ah, I understand the distinction you were seeing, and it makes sense to me now. I'll leave it as is, but we can always rethink.

PR #1187, opened 2020-04-01
Allow flexible walker for unsupervised sampler using new base class RandomWalk

kjun9:
This introduces a new base class `RandomWalk` which is a more specific type compared to the existing `GraphWalk` class. A `RandomWalk` class is defined as having a `run` method which takes a list of node IDs and returns a list of random walks (list of list of node IDs). Currently these classes belong to this type:

* `UniformRandomWalk`
* `BiasedRandomWalk`
* `UniformRandomMetaPathWalk`

It's useful to define this type, since these are all suitable for use with `UnsupervisedSampler`, and so we can introduce a new `walker` parameter to `UnsupervisedSampler` to allow for some flexibility instead of being hardcoded to use `UniformRandomWalk`. 

Note: `TemporalRandomWalk` could theoretically belong in this family too with some adjustments - currently its `run` method does not take a list of nodes as input, so it might make sense for it to have two different `run` methods - one that follows the reference paper (i.e. takes "number of context windows", etc as parameters), and one that fits into our definition of a `RandomWalk` by making some assumptions (e.g. calculate the parameters based on set of nodes).

also Note: I did try to keep random refactoring to a minimum.. but I thought it was reasonable enough to update the error handling/messages for the classes that were brought into the new `RandomWalk` base class, instead of using the existing checks in `GraphWalk`.

See #919 

Related https://github.com/stellargraph/stellargraph/pull/536#discussion_r400583830

### DIFF
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -56,6 +56,12 @@ def _ensure_int(value, name, min_value):
 
 
 class RandomWalk(ABC):
+    """
+    Abstract base class for Random Walk classes. A Random Walk class must implement a ``run`` method
+    which takes an iterable of node IDs and returns a list of walks. Each walk is a list of node IDs
+    that contains the starting node as its first element.
+    """
+
     def __init__(self, graph, seed=None):
         if not isinstance(graph, StellarGraph):
             raise TypeError("Graph must be a StellarGraph or StellarDiGraph.")
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -34,7 +34,7 @@ from scipy.special import softmax
 from ..core.schema import GraphSchema
 from ..core.graph import StellarGraph
 from ..core.utils import is_real_iterable
-from ..core.experimental import experimental
+from ..core.validation import require_integer_in_range
 from ..random import random_state
 from abc import ABC, abstractmethod
 
@@ -46,13 +46,6 @@ def _default_if_none(value, default, name, ensure_not_none=True):
             f"{name}: expected a value to be specified in either `__init__` or `run`, found None in both"
         )
     return value
-
-
-def _ensure_int(value, name, min_value):
-    if not isinstance(value, int) or value < min_value:
-        raise ValueError(
-            f"{name}: expected an integer greater than or equal to {min_value}, found: {value}"
-        )
 
 
 class RandomWalk(ABC):
@@ -81,7 +74,7 @@ class RandomWalk(ABC):
             # Restore the random state
             return self._random_state
         # seed the random number generator
-        _ensure_int(seed, "seed", min_value=0)
+        require_integer_in_range(seed, "seed", min_val=0)
         rs, _ = random_state(seed)
         return rs
 
@@ -96,8 +89,8 @@ class RandomWalk(ABC):
                 stacklevel=3,
             )
 
-        _ensure_int(n, "n", min_value=1)
-        _ensure_int(length, "length", min_value=1)
+        require_integer_in_range(n, "n", min_val=1)
+        require_integer_in_range(length, "length", min_val=1)
 
     @abstractmethod
     def run(self, nodes, **kwargs):
@@ -159,7 +152,7 @@ class GraphWalk(object):
             The random state as determined by the seed.
         """
         if seed is None:
-            # Restore the random state
+            # Use the class's random state
             return self._random_state
         # seed the random number generator
         rs, _ = random_state(seed)
@@ -259,7 +252,7 @@ class UniformRandomWalk(RandomWalk):
         self.n = n
         self.length = length
 
-    def run(self, nodes, n=None, length=None, seed=None):
+    def run(self, nodes, *, n=None, length=None, seed=None):
         """
         Perform a random walk starting from the root nodes. Optional parameters default to using the
         values passed in during construction.
@@ -359,7 +352,9 @@ class BiasedRandomWalk(RandomWalk):
         self.q = q
         self.weighted = weighted
 
-    def run(self, nodes, n=None, length=None, p=None, q=None, seed=None, weighted=None):
+    def run(
+        self, nodes, *, n=None, length=None, p=None, q=None, seed=None, weighted=None
+    ):
 
         """
         Perform a random walk starting from the root nodes. Optional parameters default to using the
@@ -517,7 +512,7 @@ class UniformRandomMetaPathWalk(RandomWalk):
         self.length = length
         self.metapaths = metapaths
 
-    def run(self, nodes, n=None, length=None, metapaths=None, seed=None):
+    def run(self, nodes, *, n=None, length=None, metapaths=None, seed=None):
         """
         Performs metapath-driven uniform random walks on heterogeneous graphs.
 
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -388,13 +388,13 @@ class BiasedRandomWalk(RandomWalk):
         rs = self._get_random_state(seed)
 
         if weighted:
-
             # Check that all edge weights are greater than or equal to 0.
             # Also, if the given graph is a MultiGraph, then check that there are no two edges between
             # the same two nodes with different weights.
             for node in self.graph.nodes():
                 # TODO Encapsulate edge weights
                 for neighbor in self.graph.neighbors(node):
+
                     wts = set()
                     name = f"Edge weight between ({node}) and ({neighbor})"
                     for weight in self.graph._edge_weights(node, neighbor):
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -36,6 +36,7 @@ from ..core.graph import StellarGraph
 from ..core.utils import is_real_iterable
 from ..core.experimental import experimental
 from ..random import random_state
+from abc import ABC, abstractmethod
 
 
 def _default_if_none(value, default, name, ensure_not_none=True):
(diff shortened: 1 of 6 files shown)


stellargraph/data/explorer.py
```diff
@@ -47,6 +48,62 @@ def _default_if_none(value, default, name, ensure_not_none=True):
     return value
 
 
+def _ensure_int(value, name, min_value):
```
huonw, line 51: There's `require_integer_in_range` in `core.validation` that does this too.

stellargraph/data/explorer.py
```diff
@@ -47,6 +48,62 @@ def _default_if_none(value, default, name, ensure_not_none=True):
...
+    def _get_random_state(self, seed):
+        """
+        Args:
+            seed: The optional seed value for a given run.
+
+        Returns:
+            The random state as determined by the seed.
+        """
+        if seed is None:
+            # Restore the random state
```
huonw, line 81: Is "restore" the right word here? Maybe

```suggestion
            # Use the class's random state
```

stellargraph/data/unsupervised_sampler.py
```diff
@@ -56,23 +63,12 @@ def __init__(self, G, nodes=None, length=2, number_of_walks=1, seed=None):
...
-        #             "({}) Only Uniform Random Walks are possible".format(
-        #                 type(self).__name__
-        #             )
-        #         )
-        #     else:
-        #         self.walker = walker(G, seed=seed)
-        # else:
-        #         self.walker = UniformRandomWalk(G, seed=seed)
+        if walker is not None:
+            self.walker = walker
```
huonw, line 67: Should we emit a warning or something if the other parameters are non-default (i.e. that they're being ignored)? I feel the `seed` one in particular is 'dangerous', since the user presumably intended to pass it to the walker when they constructed it (maybe without realising that's what they intended). Thoughts?

stellargraph/data/explorer.py
```diff
@@ -47,6 +48,62 @@ def _default_if_none(value, default, name, ensure_not_none=True):
...
+                "No root node IDs given. An empty list will be returned as a result.",
+                RuntimeWarning,
+                stacklevel=3,
+            )
+
+        _ensure_int(n, "n", min_value=1)
+        _ensure_int(length, "length", min_value=1)
+
+    @abstractmethod
+    def run(self, nodes, **kwargs):
```
huonw, line 103: If we're (slightly) refactoring here, would

```suggestion
    def run(self, start_nodes, **kwargs):
```
be slightly better as an API? (not that it matters much at all)

huonw, line 103: Also, theoretically the subclasses don't currently satisfy this signature because there's no `*args` here, or `*` on the subclasses. For instance, if `rw = UniformRandomWalk(...)`, then `rw.run(start_nodes, 10)` will work and pass 10, but that isn't allowed by the abstract signature here. We could either update this to be:

```suggestion
    def run(self, nodes, *args, **kwargs):
```

or update each of the run methods to make the non-`nodes` params keyword only, like 

```python
def run(self, nodes, *, n=None, length=None, seed=None):
```

Thoughts?

stellargraph/data/unsupervised_sampler.py
kjun9: That's a good point, I'll add warnings for when the other parameters are specified but ignored

stellargraph/data/explorer.py
kjun9: This would be a breaking change, since there are a number of notebooks that use `BiasedRandomWalk.run` directly for instance. Do you think it's worth doing?

kjun9: > update each of the run methods to make the non-nodes params keyword only

I like this approach, I think we want to force anything other than `nodes` to be keyword only

huonw: I think it's only breaking if they do `.run(nodes=...)`? If so, then let's leave it.

huonw: Sounds reasonable.

kjun9: Hmm tests still pass, but pycharm doesn't seem to like it...
image
whereas it wasn't complaining before.

huonw: Huh; I think that might be a PyCharm bug? Because it seems to be that the keyword-only one is the one that matches the behaviour of the abstract one:

```python
class X:
  def run_abstract(self, nodes, **kwargs):
    print(nodes, kwargs)
  def run_develop(self, nodes, n=None, length=None, seed=None):
    print(nodes, n, length, seed)
  def run_keyword_only(self, nodes, *, n=None, length=None, seed=None):
    print(nodes, n, length, seed)

x = X()

# positional
x.run_abstract([1], 2) # TypeError: run_abstract() takes 2 positional arguments but 3 were given
x.run_keyword_only([1], 2) # TypeError: run_keyword_only() takes 2 positional arguments but 3 were given
x.run_develop([1], 2) # [1] 2 None None

# keyword
x.run_abstract([1], n=2) # [1] {'n': 2}
x.run_keyword_only([1], n=2) # [1] 2 None None
x.run_develop([1], n=2) # [1] 2 None None
```

I wonder if there's a public bug tracker we can search/file against?

kjun9: There's a number of issues that seem somewhat related: https://youtrack.jetbrains.com/issues/PY?q=signature%20does%20not%20match%20base

I don't see one that specifically talks about using `*` though - I can adapt your example a bit and file an issue

kjun9: Hmm I upgraded pycharm and now the error seems to be gone actually... which I probably should've done before filing the issue 😅 Maybe the positional one should show an error though (which it still doesn't)

kjun9: I went through our `node2vec` and `metapath2vec` demos and they all seem to pass in `nodes` as a keyword argument when they run the random walk, so I'll leave it.

tests/data/test_unsupervised_sampler.py
```diff
@@ -117,3 +117,15 @@ def test_walker_custom(line_graph):
     for context_pairs, labels in batches:
         for node, neighbour in context_pairs[labels == 1]:
             assert node == neighbour
+
+
+def test_ignored_param_warning(line_graph):
+    walker = UniformRandomWalk(line_graph, n=2, length=3)
+    with pytest.warns(UserWarning):
```
huonw, line 124:
```suggestion
    with pytest.warns(UserWarning, match="walker, length"):
```
or going all out
```suggestion
    with pytest.warns(UserWarning, match="walker, length: 'walker' was provided so 'length' parameter will be ignored"):
```

and so on for the other ones.

I wonder if makes sense to just make these `ValueError`s too, instead of warnings; it means slightly fewer code paths for us to think about and less risk of mixing something up (e.g. we might refactor `UnsupervisedSampler` to use `seed` in the `__init__` unconditionally and a warning wouldn't necessarily be as obvious to us), but doesn't really matter that much. Thoughts?

tests/data/test_unsupervised_sampler.py
```diff
@@ -19,65 +19,113 @@
...
+            if label == 1:
+                assert context in set(line_graph.neighbors(target))
+
+
+def test_walker_uniform_random(line_graph):
+    length = 3
+    number_of_walks = 2
+    batch_size = 4
+
+    walker = UniformRandomWalk(line_graph, n=number_of_walks, length=length)
```
huonw, line 93: What's this testing? It looks like it's almost testing that `UnsupervisedSampler(..., walker=None)` uses `UniformRandomWalk` internally, but it's not quite doing the validation for that. It also has a lot of overlap with `CustomWalker`?

kjun9: It's just testing that the batches are consistent with the walker object provided instead of the default values for `number_of_walks` and `length` for `UnsupervisedSampler`. I guess the custom walker test also does a similar job, but I thought I'd include one that uses one of our actual `RandomWalk` classes.

huonw: Ah, I see. Makes sense. Maybe a brief comment to provide context?

tests/data/test_unsupervised_sampler.py
kjun9: Hmm that's true, yeah I can update it to be `ValueError` instead. It doesn't quite catch everything though since:
```python
sampler = UnsupervisedSampler(g, seed=None, walker=some_walker)
```
won't raise an error if the user intended to use no seed, since `None` is the default value. 

Maybe we should consider just deprecating the other parameters and use the `walker` parameter as the only way to specify additional parameters - for instance:
```python
# use defaults
sampler = UnsupervisedSampler(g)

# use a seed
sampler = UnsupervisedSampler(g, walker=UniformRandomWalk(g, seed=0))
```

kjun9:
> We could consider looking at the repositories that use this (because I think there's quite a few) and see if they customise the parameters much to get a little bit of data about this.

Do you mean repositories in here...? https://github.com/stellargraph/stellargraph/network/dependents?dependent_type=REPOSITORY&package_id=UGFja2FnZS0xNjI2NTY5NDU%3D

> It's fine by me for this to be follow up work too.

I've filed https://github.com/stellargraph/stellargraph/issues/1199

huonw:
> Do you mean repositories in here...? https://github.com/stellargraph/stellargraph/network/dependents?dependent_type=REPOSITORY&package_id=UGFja2FnZS0xNjI2NTY5NDU%3D

Yeah, but that's only 13 out of ~40. The spreadsheet that I'll link on slack is more complete.

Issue #1166, opened 2020-03-25
Update Conda `meta.yaml` and document how it should be maintained


kjun9:
### Describe the bug

`conda build` failed during release 0.11.0 from having outdated information in `meta.yaml`:
```
Traceback (most recent call last):
  File "/Users/jun021/miniconda3/envs/stellargraph-dev/conda-bld/stellargraph_1585130950934/test_tmp/run_test.py", line 20, in <module>
import: 'stellargraph'
    import stellargraph.utils.saliency_maps
import: 'stellargraph.core'
ModuleNotFoundError: No module named 'stellargraph.utils.saliency_maps'
import: 'stellargraph.data'
import: 'stellargraph.layer'
import: 'stellargraph.mapper'
import: 'stellargraph.utils'
import: 'stellargraph.utils.saliency_maps'
Tests failed for stellargraph-0.11.0-py_0.tar.bz2 - moving package to /Users/jun021/miniconda3/envs/stellargraph-dev/conda-bld/broken
WARNING:conda_build.build:Tests failed for stellargraph-0.11.0-py_0.tar.bz2 - moving package to /Users/jun021/miniconda3/envs/stellargraph-dev/conda-bld/broken
WARNING conda_build.build:tests_failed(2231): Tests failed for stellargraph-0.11.0-py_0.tar.bz2 - moving package to /Users/jun021/miniconda3/envs/stellargraph-dev/conda-bld/broken
TESTS FAILED: stellargraph-0.11.0-py_0.tar.bz2
```

The requirements defined in `meta.yaml` were also not updated for a few releases so despite building successfully, those releases uploaded in conda cloud are broken. We should update `meta.yaml` to be more optimised and easier to maintain, and document when to make updates (e.g. during release).

timpitman:
See https://github.com/stellargraph/stellargraph/pull/1165

timpitman:
#1165 improves the Conda package, and integrates it into CI so that it's continually tested and we'll know as soon as it breaks - that way it will be updated immediately inside PR's, rather than waiting for the release.

Issue #627, opened 2020-01-08
Simplify or automate the release procedure


timpitman:
https://github.com/stellargraph/stellargraph/pull/624 brought the release procedure inside the git repository.  Now we can streamline the procedure, and look at automating some parts with Buildkite.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1189, opened 2020-04-01
Build Conda package on CI (#1166)

timpitman:
This PR creates a Buildkite stage to run `conda build` on CI (using the Anaconda docker image), which will validate our Conda package.

There are also improvements to the `meta.yaml` file which defines the stellargraph package:
* brings in the changes from v0.11.1 which are in master, but didn't merge correctly into develop
* tweaks dependencies (eg python 3.6 min, no cap on version of networkx)
* expands the `test` section to bundle and run our unit tests on the built package which can validate issues like broken tensorflow dependencies
  * note that it may be more practical long-term to only run a subset of our tests - suggestions welcome

note that I've temporarily removed other stages from the pipeline to speed up development, but will restore for the final PR.

See: #1166


.buildkite/pipeline.yml
```diff
@@ -15,227 +15,14 @@ plugins: &plugins
     mode: checkout
 
 steps:
-  - label: ":docker::python: build 3.6 runner"
```
huonw, line 18: (Reminder: restore all this deleted code 😄)

.buildkite/pipeline.yml
```diff
@@ -15,227 +15,14 @@ plugins: &plugins
...
+  - label: ":python: test Conda build"
     <<: *timeout
-    key: "runner-3_6"
+    command: ".buildkite/steps/conda-build.sh"
```
huonw, line 20: This script is one command, so, from my perspective, it would be fine to do this as 

```suggestion
    command: "conda build . --no-anaconda-upload"
```

meta.yaml
```diff
@@ -16,39 +16,50 @@ requirements:
...
     - stellargraph.utils
-    - stellargraph.utils.saliency_maps
-
+  requires:
+    - pytest
+    - pytest-benchmark >= 3.1
+  source_files:
+    - tests
+  commands:
+    - pip install rdflib
```
huonw, line 61: Why this here? Could it be part of the `requires`? If not, maybe a comment.

meta.yaml
```diff
@@ -16,39 +16,50 @@ requirements:
...
+    - py.test -ra tests/ -p no:cacheprovider --benchmark-disable --color=yes
```
huonw, line 62: This build step takes a long time (12 minutes, according to https://buildkite.com/stellar/stellargraph-public/builds/2857#783ced6d-b185-4363-a480-fe567df8a27b) and most of it is running these tests (8 minutes, https://buildkite.com/stellar/stellargraph-public/builds/2857#783ced6d-b185-4363-a480-fe567df8a27b/167-1184). Do you think it's worth running all of these tests here?

One option would be only doing a subset, as a smoke test, like:

```suggestion
    - py.test -ra tests/core tests/layer -p no:cacheprovider --benchmark-disable --color=yes
```

What do you think?

huonw:
(I edited a reference to #1166 into the PR description because references in the title don't become links, and I think they also don't create the back-reference on the other end.)

meta.yaml
timpitman: rdflib is here as there's no Conda package, however it's only used in one test - running a subset of tests allows us to remove it.

meta.yaml
timpitman: good idea (floated in the original description of this PR that a subset may be more appropriate).

huonw: Sorry about missing the PR description bit!

meta.yaml
```diff
@@ -16,39 +16,53 @@ requirements:
...
     - stellargraph.layer
     - stellargraph.mapper
     - stellargraph.utils
-    - stellargraph.utils.saliency_maps
-
+  requires:
+    - pytest
+    - pytest-benchmark >= 3.1
+  source_files:
+    # we only bundle and run a subset of tests as a 'smoke test'
```
huonw, line 59: Much faster! 👍 

```
============ 299 passed, 1 skipped, 15 warnings in 77.59s (0:01:17) ============
```

timpitman:
@kjun9 I added you as a reviewer to see if you're happy with this PR as a resolution to #1166?

kjun9:
> I added you as a reviewer to see if you're happy with this PR as a resolution to #1166?

Ah I didn't mention explicitly in #1166 but I also wasn't sure if we were using the `requirements.host` section correctly?

From reading [the docs](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#host), it seems like we might just need `python`, not all of our dependencies - should we try this?

timpitman:
> we might just need `python`, not all of our dependencies - should we try this?

you're right - Conda is quite intelligent here, and even selects a version of python to match the set of dependencies used later for 'run' - this works well.

timpitman:
I also added build kite upload of the artefact - useful currently as it can't build on Mac due to tensorflow 2.1 being unsupported.

.buildkite/steps/conda-build.sh
```diff
@@ -0,0 +1,14 @@
...
+
+echo "+++ :snake: :construction_worker: conda build"
+conda build . --no-anaconda-upload || exitCode=$?
+# if successful, then upload to buildkite
+if [ "${exitCode}" == "0" ]; then
+  conda_package="$(conda build . --output)"
+  buildkite-agent artifact upload "${conda_package}"
+fi
+
+exit $exitCode
```
huonw, line 14: I think in this case, because there's no code to run if `conda build .` fails (unlike notebooks and tests where we want to upload the notebooks and JUnit xml files, respectively), we can just have this be straight line and rely on the `set -e` to have the script stop running if the build fails:

```suggestion
conda build . --no-anaconda-upload

echo "+++ :snake::buildkite: upload package"
conda_package="$(conda build . --output)"
buildkite-agent artifact upload "${conda_package}"
```

huonw:
Uploading the package is a great idea 👍 It's part of the way to #627. Should we update the release procedure to downloading from buildkite instead of a manual `conda build .` (or at least indicate it exists there)?

PR #1190, opened 2020-04-01
Remove the use of utils in cora lp demo

kjun9:
This doesn't remove the actual utils directory yet, since it contains the metapath2vec demo code which needs to be notebook-ified, but this does remove the use of utils in the node2vec demo. Some of the markdown in the notebook was also updated to explain more bits of code that have ended up in the notebook.

See #896


demos/link-prediction/random-walks/cora-lp-demo.ipynb
```diff
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
```
huonw, line 4: <p>Maybe we don't need to have the same comments twice just with "graph" -&gt; "graph_test"; this cell could just have a comment like</p><p> </p><pre class="ql-syntax" spellcheck="false"># Do the same process to compute a training subset from within the test graph
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

huonw, line 4: <p>Maybe the functions here could be marked to correspond to the lists steps to give the user a little more guidance for this large cell? E.g.</p><p> </p><pre class="ql-syntax" spellcheck="false"># 1. link embeddings
...

# 2. training classifier
...

# 3. evaluate classifier
...
</pre><p>Or maybe that doesn't make sense?</p><p> </p><p>Also in <code>link_examples_to_features</code>, I think the <code>isinstance(binary_operator, str)</code> support code isn't used and the <code>operator_func</code> dict isn't defined.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

huonw, line 4: <p>How long does <code>evaluate_link_prediction_model</code> take to run? If it's relatively short, potentially we don't need to have the <code>Finished training link prediction model</code> line, which makes the output below slightly harder to parse?</p><p> </p><p>In addition. In fact, we could potentially remove all of the <code>print</code> s here and instead display the results via a Pandas DataFrame, since that has a nice Jupyter rendering and makes things aligned, e.g.:</p><p> </p><pre class="ql-syntax" spellcheck="false">def run_link_prediction(...):
    clf = ...
    score = ...
    return {"classifier": ..., ...}
</pre><p> </p><pre class="ql-syntax" spellcheck="false">results = [run_link_prediction(op) for op in ...]
best_results = ...

print(f"Best result ...")

pd.DataFrame(
    [(result["binary_operator"].__name__, result["score"] for result in results],
    columns=("name", "ROC AUC score")
)
</pre><p>(This does flip the "Best result" printing to be before the dataframe, but I think that's ok.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

huonw, line 4: <p>This is qualitatively quite different to how it was before. Maybe it's just because of the outlier at (2.4, 2) that's causing the main region to be much more compressed than it was? Thoughts?</p><p> </p><p>Also, it's not directly relevant to the changes in this PR, but this suffers from colourblindness problems like other plots we've talked about. Assuming the <code>alpha</code>  differences aren't critical, it could be switched to something like:</p><p> </p><pre class="ql-syntax" spellcheck="false">plt.scatter(X_transformed[:, 0], X_transformed[:, 1], c=link_labels_test, cmap="jet", alpha=0.8)
</pre><p>or for more direct control over the colours</p><p> </p><pre class="ql-syntax" spellcheck="false">plt.scatter(X_transformed[:, 0], X_transformed[:, 1], c=np.where(link_labels_test == 1, "g", "r"), alpha=0.8)
</pre><p>(this is reproducing the current numbers, but could be switched to <code>"b", "r"</code>  or something like that.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

kjun9: <blockquote>Maybe the functions here could be marked to correspond to the lists steps to give the user a little more guidance for this large cell?</blockquote><p>Yeah that makes sense to me</p><p> </p><blockquote>Also in&nbsp;<code>link_examples_to_features</code>, I think the&nbsp;<code>isinstance(binary_operator, str)</code>&nbsp;support code isn't used and the&nbsp;<code>operator_func</code>&nbsp;dict isn't defined.</blockquote><p>good catch!</p>

kjun9: <p>Nice :+1:</p>

kjun9: <p>Seems like the scores are actually quite different from before too... I'll investigate a bit, but might also just be related to updating the dataset loading, and not necessarily the link prediction code</p><p> </p><p>There's a comment above the cell about which colour represents what, so I'll go with the direct control version with "b" and "r"</p><p> </p>

kjun9: Hmm actually I think I've done the evaluation incorrectly - there needs to be yet another `train_test_split` if I were to replicate existing code, I'll make an update

kjun9:
I've made a few more adjustments to make it more closely follow what was there before... but the results seem to have gotten worse actually - I'll look into it a bit further, but what I've gathered so far:
* dataset loading
  * Previously used networkx to obtain largest connected component which gave `Largest subgraph statistics: 2485 nodes, 5069 edges`
  * With `dataset.load(largest_connected_component_only=True, str_node_ids=True)` we get more edges: `Nodes: 2485, Edges: 5209`
* edge splitter
  * previously used edge splitter with networkx graphs
  * now used with StellarGraph objects
* Training Word2Vec model
  * Previously used some kind of custom node2vec code to run random walks (in `utils/node2vec`)
  * Now using `BiasedRandomWalk` class in the library

huonw:
> we get more edges

I think there's multi-edges in the Cora dataset, which get eliminated with a `nx.Graph` but not with the multigraph `StellarGraph`.

> previously used edge splitter with networkx graphs, now used with StellarGraph objects

The StellarGraph "implementation" is just doing conversions, so hopefully this shouldn't affect it too much:

https://github.com/stellargraph/stellargraph/blob/d9e49e1810a7f999e04ad29283aae4222b2ce28c/stellargraph/data/edge_splitter.py#L64-L71

kjun9:
I tried a switch to compare results using `BiasedRandomWalk` vs. the biased random walk in utils, and still getting the same (different from original) results, so will try reducing the multi-edges now to see if that helps. Seems a bit odd to me that a citation network contains multi-edges?

kjun9:
Also tried: 
* removing the multi-edges, which does make the number of edges match the original, but doesn't improve the scores. 
* tried using the entire set of data/graph loading cells from old notebook, but still no luck

The problem has to be somewhere before the link classifier training, since the embedding visualisation still doesn't look correct, and that only uses the node2vec features + binary operator

So now going to open both notebooks side by side and go through each cell...

kjun9:
Turned out to be a silly typo.. `walk_length` was set incorrectly. Now the plot looks good:
image

kjun9:
@PantelisElinas It'd be good to get your review on this one, since I'm not entirely sure if there's enough explanation for the train-test splitting in the notebook... the splits are:

* we create `graph_test` and test examples/labels using edge splitter on `graph`
* we create `graph_train` and training examples/labels using edge splitter on `graph_test`, with `graph` as the master so that the negative edge sampling doesn't accidentally sample a positive edge from `graph`.
* Using a node2vec model trained on `graph_train` we:
  * split the training examples/labels further into train/test sets within this set
  * train the classifier on the train-train examples/labels
  * evaluate the classifier on the train-test examples/labels
* Using a node2vec model trained on `graph_test` we:
  * evaluate the classifier on these test examples/labels

I'm not sure if I got my head around why we need to evaluate do another split on the training set to evaluate the classifier, instead of just evaluating on the original test set? What roughly made sense to me was that we want scores using embeddings from both `graph_train` and `graph_test`, so that's the explanation I've included as a comment in the notebook, but not sure if that's the full/correct story.

PantelisElinas:
@kjun9 

> I'm not sure if I got my head around why we need to evaluate do another split on the training set to evaluate the classifier, instead of just evaluating on the original test set? 

The reason you need the extract split is for selecting the binary operator. You shouldn't use your test data to compare operators and decide on the best one to use.

So, given the train graph and the training edge data, you do a 75/25 split, you use 75% of the data to train you binary classifier and 25% of the data to evaluate its performance for a given `binary_operator`. This value you store in `score_graph_train`. Using the same split of data you perform the same operation on the next available `binary_operator` and so on. Now, you can compare performance between the operators and select the best performing one. Then you evaluate your model's performance on the test set. The idea here is that your test data should not be used for model selection, i.e., in this case selecting the `binary_operator`. You can further select a better model tuning any of the Node2Vec parameters but we don't show that in this notebook; again, for this you would need that additional 75/25 split of your training data.

The "error" in your `run_link_prediction(binary_operator)` is that it should just return `score_graph_train` and not proceed with calculating `score_graph_test`. Once you have the `score_graph_train` values for all `binary_operators` then you can select the best performing one and calculate and report `score_graph_test`. The reason I said "error" and not error is that there is nothing wrong with evaluating all operators on the test data, **unless** you want to do model selection which I think is what this notebook should demonstrate and hence should be updated accordingly.

One important detail is when you do the 75/25 split, make sure that it is done once and all operators are trained and evaluated on the same data. Otherwise, it is not exactly correct to use the calculated `score_graph_train` values to select the best operator; just in case one operator receives a split of the data that is easier to classify.

Let me know if this doesn't clarify things.

BTW, the rest of the notebook looks really good.

kjun9:
> The reason you need the extract split is for selecting the binary operator ... The idea here is that your test data should not be used for model selection, i.e., in this case selecting the binary_operator. You can further select a better model tuning any of the Node2Vec parameters but we don't show that in this notebook

Ah I see! Okay I'll update the notebook to mention more explicitly that the 75/25 split is for model selection and that in this notebook we are demonstrating model selection with a single variable parameter being the binary operator. Then only use the best operator for calculating `score_graph_test`.

> One important detail is when you do the 75/25 split, make sure that it is done once and all operators are trained and evaluated on the same data.

Makes sense 👍  I'll update the notebook

kjun9:
@huonw @PantelisElinas this is ready for review again - I've tried to incorporate more explanation of the splits in the notebook

demos/link-prediction/random-walks/cora-lp-demo.ipynb
huonw, line 4: <p>We've tried to have these buttons end up under the title e.g. https://github.com/stellargraph/stellargraph/blob/develop/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb</p><p> </p><p>This is done by creating a markdown cell that only contains the title, and having any other introductory text be a separate cell after that.</p><p> </p><p>Also, notebook could do with a better title, e.g. "Link prediction with node2vec on Cora" or something.</p><p> </p><p>(I know it's not directly related to the changes in this PR, but I think this PR is enough of a "rewrite the demo" that this can be done here.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

huonw, line 4: <p>I think the use of a list here is a good improvement, but I had to re-read it several times to understand all the moving parts, and particularly how the "splits"/"sets" connect to the graphs. Maybe it could be expanded (sorry for the poor formatting, reviewnb's markdown editor kinda sucks):</p><p> </p><p>---</p><p> </p><p>We're going to tackle link prediction as a supervised learning problem on top of node representations/embeddings. The embeddings are computed with the unsupervised node2vec algorithm. After obtaining embeddings, a binary classifier can be used to predict a link, or not, ... [rest of first paragraph]</p><p> </p><p>There's four steps:</p><p> </p><ol><li>Obtain embeddings for each node</li><li>For each set of hyperparameters, train a classifier</li><li>Select the classifier that performs the best</li><li>Evaluate the selected classifier on unseen data to validate its ability to generalise</li></ol><p> </p><p>We have to carefully split the data to avoid data leakage and evaluate the algorithms correctly data:</p><p> </p><ul><li>For computing node embeddings, a <strong>train graph</strong></li><li>For training classifiers, a <strong>classifier training set</strong> of positive and negative edges that weren't used for computing node embeddings</li><li>For choosing the best classifier, an <strong>model selection test set</strong> of positive and negative edges that weren't used for computing node embeddings or training the classifier</li><li>For the final evaluation, a <strong>test graph</strong> with more edges than the train graph to compute test node embeddings, and a <strong>test set</strong> of positive and negative edges not used for computing the test node embeddings or for classifier training or model selection</li></ul><p> </p><p>---</p><p> </p><p>The bolded things should then match the names of the variables (but I think you've already done that?)</p><p> </p><p>Potentially the first paragraph and the "there's four steps" description can move to the introduction too and this section just focuses on the splits.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

huonw, line 4: <p>Maybe the <code>embedding_test</code>  computation could move to be part of an explicit "test the best model section" (not sure of a good heading name), so that it's in context?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

huonw, line 4: <p>Maybe this cell could have a heading and/or description before it, since at the moment this is 3 code cells in a row. (Especially useful if the <code>embeddings_test</code>  computation moves to be close to here too)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

kjun9: I've updated the title and made it its own cell 👍

kjun9: Thanks, yeah I think it makes sense to have the first paragraph + "four steps..." in the introduction, so I've moved it there now - also helps to break up this big markdown section. 

The variable name vs how I refer to the splits in english were slightly different (**Train Graph** vs `graph_train`) so I've put the variable names in parens too.

kjun9: I've moved `embeddings_test` here and gave it a brief title/description 👍

demos/link-prediction/random-walks/cora-lp-demo.ipynb
```diff
@@ -4,18 +4,7 @@
...
```
kjun9, line 4: <p><code>l1</code>  happened to win this time, so that's why the plot is flipped, but I think it still looks as good</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

Issue #1134, opened 2020-03-23
Deep Graph Infomax for Graph Classification


kieranricardo:
This issue is to implement DGI for graph classification. The original DGI paper (https://arxiv.org/pdf/1809.10341.pdf) proposes a method of doing this by randomly sampling a negative graph instead of scrambling the feature matrix.

huonw:
Per slack, another approach is https://arxiv.org/abs/1908.01000 ("InfoGraph: Unsupervised and Semi-supervised Graph-Level Representation Learning via Mutual Information Maximization").

PR #1193, opened 2020-04-02
Make DGI easy to extend by moving corruption logic to Generator classes

huonw:
This adds 2 non-abstract (defaulted) methods to the `Generator` base class from #1194:

- `corrupt_inputs`: takes in a list of model input values (tensors/numpy arrays etc.) created by a `flow`'d sequence and returns some corruption of them for DGI
- `corruptible_input_indices`: the input tensor locations that get corrupted by `corrupt_inputs` and so need duplicated input tensors

Notably, the model classes don't have to know anything about corruption or DGI, it's purely a generator/sequence-level concern. Any model that uses that generator with corruption support can thus be trained in an unsupervised fashion. For instance, new models using `FullBatchNodeGenerator` will automatically be able to be trained with DGI.

Things to think about before merging this PR:

- reproducibility: maybe `corrupt_inputs` should take a random state?
- passing `base_generator` vs. the corrupted generator into `DeepGraphInfomax`
- how does this work with `tf.data`-based generators?
- maybe instead of defaulted methods on `Generator` we should add a subclass:
  ```python
  class CorruptibleGenerator(Generator):
      @abc.abstractmethod
      def corrupt_inputs(self, ...): ...
      @abc.abstractmethod
      def corruptible_input_indices(self): ...
   ```
   and then checks for "corruption support" are `isinstance(generator, CorruptibleGenerator)`.
- adding docs

This builds off #1194.

### DIFF
--- a/stellargraph/layer/deep_graph_infomax.py
+++ b/stellargraph/layer/deep_graph_infomax.py
@@ -16,6 +16,7 @@
 
 from . import GCN, GAT, APPNP, PPNP
 from .misc import deprecated_model_function
+from ..mapper import Generator
 
 from tensorflow.keras.layers import Input, Lambda, Layer, GlobalAveragePooling1D
 import tensorflow as tf
@@ -72,26 +73,44 @@ class DeepGraphInfomax:
 
     Args:
         base_model: the base stellargraph model class
+        base_generator (Generator):
     """
 
-    def __init__(self, base_model):
+    def __init__(self, base_model, base_generator=None):
 
-        if not isinstance(base_model, (GCN, GAT, APPNP, PPNP)):
-            raise TypeError(
-                f"base_model: expected GCN, GAT, APPNP or PPNP found {type(base_model).__name__}"
-            )
-
-        if base_model.multiplicity != 1:
+        if base_generator is None:
+            # legacy code path for the original implementation
             warnings.warn(
-                f"base_model: expected a node model (multiplicity = 1), found a link model (multiplicity = {base_model.multiplicity}). Base model tensors will be constructed as for a node model.",
+                f"the 'base_generator' parameter will be required in future; pass in the base model's generator, like `DeepGraphInfomax(base_model, base_generator)`",
+                DeprecationWarning,
                 stacklevel=2,
+            )
+            if not isinstance(base_model, (GCN, GAT, APPNP, PPNP)):
+                raise TypeError(
+                    f"base_model: expected GCN, GAT, APPNP or PPNP found {type(base_model).__name__}"
+                )
+
+            if base_model.multiplicity != 1:
+                warnings.warn(
+                    f"base_model: expected a node model (multiplicity = 1), found a link model (multiplicity = {base_model.multiplicity}). Base model tensors will be constructed as for a node model.",
+                    stacklevel=2,
+                )
+
+            # specific to full batch models
+            self._corruptible_inputs_idxs = [0]
+        elif isinstance(base_generator, Generator):
+            self._corruptible_inputs_idxs = base_generator.corruptible_input_indices()
+
+            if self._corruptible_inputs_idxs is None:
+                raise TypeError(
+                    "base_generator: expected a Generator that supports corruption, found {type(base_generator).__name__}"
+                )
+        else:
+            raise TypeError(
+                f"base_generator: expected a Generator, found {type(base_generator).__name__}"
             )
 
         self.base_model = base_model
-
-        self._node_feats = None
-        # specific to full batch models
-        self._corruptible_inputs_idxs = [0]
 
         self._discriminator = DGIDiscriminator()
 
--- a/stellargraph/mapper/sequences.py
+++ b/stellargraph/mapper/sequences.py
@@ -648,17 +648,12 @@ class CorruptedNodeSequence(Sequence):
     data for training Deep Graph Infomax.
 
     Args:
+        corrupt_inputs: a callable that returns a list of corrupted input values
         base_generator: the uncorrupted Sequence object.
     """
 
-    def __init__(self, base_generator):
-
-        if not isinstance(base_generator, (FullBatchSequence, SparseFullBatchSequence)):
-            raise TypeError(
-                f"base_generator: expected FullBatchSequence or SparseFullBatchSequence, "
-                f"found {type(base_generator).__name__}"
-            )
-
+    def __init__(self, corrupt_inputs, base_generator):
+        self.corrupt_inputs = corrupt_inputs
         self.base_generator = base_generator
         self.targets = np.zeros((1, len(base_generator.target_indices), 2))
         self.targets[0, :, 0] = 1.0
@@ -669,9 +664,5 @@ class CorruptedNodeSequence(Sequence):
     def __getitem__(self, index):
 
         inputs, _ = self.base_generator[index]
-        features = inputs[0]
 
-        shuffled_idxs = np.random.permutation(features.shape[1])
-        shuffled_feats = features[:, shuffled_idxs, :]
-
-        return [shuffled_feats] + inputs, self.targets
+        return self.corrupt_inputs(inputs) + inputs, self.targets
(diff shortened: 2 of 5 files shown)


stellargraph/layer/deep_graph_infomax.py
```diff
@@ -72,27 +73,45 @@ class DeepGraphInfomax:
 
     Args:
         base_model: the base stellargraph model class
+        base_generator (Generator):
     """
 
-    def __init__(self, base_model):
+    def __init__(self, base_model, base_generator=None):
```
huonw, line 79: Potentially we want to pass the `CorruptedGenerator` here, but I feel passing the base model's generator has a nice symmetry here.

stellargraph/mapper/sequences.py
```diff
@@ -648,17 +648,12 @@ class CorruptedNodeSequence(Sequence):
...
-        if not isinstance(base_generator, (FullBatchSequence, SparseFullBatchSequence)):
-            raise TypeError(
-                f"base_generator: expected FullBatchSequence or SparseFullBatchSequence, "
-                f"found {type(base_generator).__name__}"
-            )
-
+    def __init__(self, corrupt_inputs, base_generator):
+        self.corrupt_inputs = corrupt_inputs
         self.base_generator = base_generator
         self.targets = np.zeros((1, len(base_generator.target_indices), 2))
```
huonw, line 658: I'm not quite sure how to handle generalising this; potentially a `Generator` class property like:

```python
@classmethod
@abc.abstractmethod
def batch_dimensions(cls):
    ...

class FullBatchGenerator(...):
   @classmethod
   def batch_dimensions(cls):
       return 2

class BatchedNodeGenerator(...):
   @classmethod
   def batch_dimensions(cls):
       return 1
```

which can then be passed through to `__getitem__`:

```python
batch_shape = inputs[0].shape[:batch_dimensions]
targets = np.tile([0, 1], batch_shape + (1,))
```

It's a bit unfortunate that this results in recomputing the target array for each batch, but I guess there's some caching that's possible.

kieranricardo: I reckon this `Sequence` logic is the trickiest bit. I think it may be worth still having per model code in here because the shuffling/corrupting could be drastically different for different models/training algorithms e.g. graph classification, hinsage, shuffling the adjacency matrix instead of features.

kieranricardo: or atleast it feels to me that we haven't got enough examples of what this corruption will look like for different algorithms to generalize it yet

stellargraph/layer/deep_graph_infomax.py
kieranricardo: I'm inclined to go with `CorruptedGenerator ` here because most (all?) models are constructed with their training generator, also `CorruptedGenerator` will have the advantage that the check for `CorruptibleGenerator` has already been done.

kieranricardo:
this corruptable generator approach is awesome!

kieranricardo:
I reckon for `tf.data.Datasets` we can do something like:

```
CorruptedGenerator(...):
    def __init__(self, base_gen):
        ...
    def flow(....):
        base_flow = self.base_gen.flow(...)
        if isinstance(base_flow, Dataset):
            return base_flow.map(shuffle_func)
```

with `shuffle_func` maybe stored in `base_gen`?

stellargraph/mapper/sequences.py
huonw: I think pretty much all options I can think of for shuffling are handled by:

- `corrupt_inputs` being a method on the generator (so it has access to anything in `self`)
- `corrupt_inputs` taking all of the input tensors as input, because it can then pick and choose anything

To be clear, my comment above here is attached to the `self.targets` line and is talking about that, and I don't quite follow how your code connects?

huonw:
> I reckon for tf.data.Datasets we can do something like:

Ah good idea. Since the generator knows that it's returning a `Dataset` from its own `flow`, I think `shuffle_func` could just be `corrupt_inputs` as per Sequences: `return base_flow.map(self.base_gen.corrupt_inputs)`?

---

Slightly prompted by https://github.com/stellargraph/stellargraph/pull/1193#discussion_r402048330 and by pondering this for the last few hours, I'm not quite happy with the code here. I feel like this corruption logic isn't a fixed behaviour of each generator, but actually a separate set of possibilities associated with each one. For instance, one could have a full-batch node generator corrupter that manipulates the adjacency in addition to the node features.

So maybe a better API is something like:

```python
class DGICorrupter(abc.ABC):
    @abc.abstractmethod
    def corruptible_input_indices(self): ...
    @abc.abstractmethod
    def corrupt_inputs(self, inputs): ...

class CorruptedGenerator(Generator):
    def __init__(self, base_generator, corrupter: DGICorrupter):
        ...
```

so that a user can slot in any `corrupter`.

The library might provide corrupters like

```python
class FullBatchNodeGeneratorFeaturesShuffler(DGICorrupter):
    def corruptible_input_indices(self):
        return [0]
    def corrupt_inputs(self, inputs): 
        # ... shuffling code ...

class FullBatchNodeGeneratorFeaturesAdjacencyShuffler(DGICorrupter):
    def corruptible_input_indices(self):
        return [0, 2]
    def corrupt_inputs(self, inputs): 
        # ... shuffling code ...
```

that can be used like:

```python
CorruptedGenerator(base_generator, FullBatchNodeGeneratorFeaturesShuffler())
```

That's a bit verbose, but I think there's a few ways we could improve it:

- consider avoiding duplicating the input tensors (so that we don't need `corruptible_input_indices` and thus can just pass in the `corrupt_inputs` functions directly, no need to have `corruptible_input_indices`) (discussion on Slack)
- have a `default_corrupter()` method on `Generator`, and `def __init__(self, base_generator, corrupter = None):`, so the `default_corrupter` method is called is used if `corrupter` is `None`
- have the builtin corrupters available on the generator type, like `CorruptedGenerator(generator, generator.shuffle_features)` and `CorruptedGenerator(generator, generator.shuffle_features_and_adjancency)` for full batch methods (or whatever, for which ever styles we support)

stellargraph/mapper/sequences.py
kieranricardo: ah sorry I didn't realise this was just refferring to the targets line, scratch that comment! 

Just talking about the targets, I think this approach might not work for graph classification. DGI graph classification takes its corrupted nodes from a seperate trainig graph. and so we will likely end up with a different number of true and corrupted nodes. 

Maybe `corrupt_inputs` could be `corrupt_inputs_outputs`?

huonw:
Notes to self: I think we can cover full batch corruption, graphsage and HinSAGE with a type like

```python
class ShuffleTensors:
    def __init__(self, which_tensors):
        ...
```

where `which_tensors` is an sequence of sequences of integers, where the integers are indices into the input tensor array, the inner sequence represents sets tensors that are all shuffled together (e.g. `(0, 1)` means that `corrupted[0]` and `corrupted[1]` are sampled from the union of `input[0]` and `input[1]`), and the outer sequence represents groups of those sets.

huonw:
> I should summarise this discussion

Ideas for possible alternates ways to write this were:

1. stacking the inputs like `normal, normal, ... normal, corrupted, corrupted, ... corrupted` with targets like `[1, 1, ... 1, 0, 0, ... 0]` (i.e. double the batch size of the underlying generator)
   - this would allow having a non-equal 1 and 0 samples (e.g. for #1134, the `InfoGraph` paper suggests using more than 1 negative samples per positive sample)
   - we'd need to pass down the mask to the DGI layer, so that it could work out which values were positive to use for the summary input to the discriminator

2. having alternating batches of `normal, normal, ..., normal`, targets `1, 1, ..., 1`, and `corrupted, corrupted, ... corrupted`, targets `0, 0, ..., 0` (i.e. double the number of batches of the underlying generator)
   - doesn't work because the discriminator for each batch needs to have the normal ones to compare against

We also brainstormed a bit about generalising DGI to work with more complicated models (e.g. a dense layer on top of a GCN model, or a completely custom set of a layers)

- pass input layers/tensors to a `__call__` method, so that it can duplicate them as appropriate
- do something similar to `APPNP.propagate_model`, where DGI takes the whole Keras model and manipulates that (although that's maybe somewhat fragile, e.g. #1213?)

huonw:
The approach above is different enough to this PR that it was easier to start from scratch; so I've moved it to #1243 (and #1247).

Issue #1124, opened 2020-03-20
Deep Graph Infomax demo could potentially be in demos/embeddings rather than node-classification


huonw:
### Describe the bug

The Deep Graph Infomax algorithm is an unsupervised method, i.e. it is typically used to compute embeddings which can be used for downstream tasks. It has a demo at https://github.com/stellargraph/stellargraph/blob/3348ad45845bef09c13a39d3432f35a2b0a1a450/demos/node-classification/deep-graph-infomax/gcn-deep-graph-infomax-cora-example.ipynb

 For other algorithms/demos like this, we typically put them in `demos/embeddings` (with the downstream task at the end still), rather than in node-classification, since they can be used for much more than that.

For example, https://github.com/stellargraph/stellargraph/blob/3348ad45845bef09c13a39d3432f35a2b0a1a450/demos/embeddings/embeddings-unsupervised-graphsage-cora.ipynb

### To Reproduce

1. See if https://github.com/stellargraph/stellargraph/tree/develop/demos/node-classification includes Deep Graph Infomax
2. See if https://github.com/stellargraph/stellargraph/tree/develop/demos/embeddings includes it

### Observed behavior

1 has DGI, 2 doesn't.

### Expected behavior

1 shouldn't have DGI, 2 should.

### Environment

Operating system: N/A

Python version: N/A

Package versions: StellarGraph 3348ad45845bef09c13a39d3432f35a2b0a1a450

### Additional context

N/A

Issue #1087, opened 2020-03-17
Provide clearer guidance on how to find an appropriate demo


huonw:
### Description

Currently, someone looking at our demos lands at https://github.com/stellargraph/stellargraph/tree/develop/demos (or https://github.com/stellargraph/stellargraph/tree/master/demos) and then has to orient themselves and find an appropriate place to visit for their problem. We can try to provide more task-oriented guidance, in addition than just listing the algorithms and high-level task descriptions (which works well for someone familiar with graph ML):

1. someone new to the library and just wants to understand how it works (relates to #992)
2. someone with a specific dataset and problem who may not know all the graph-jargon or the algorithm jargon

For 1, we can provide links to specific notebooks.

For 2, we might provide more generic descriptions of the problems so people can orient themselves. The same applies within each subfolder, e.g. if someone wants to do inductive NAI on a homogeneous graph with node features, how do they work that out from https://github.com/stellargraph/stellargraph/tree/develop/demos/node-classification especially if they're only passingly familiar with what a "homogeneous" graph is in this context.

This relates to #992 and #910.

huonw:
I'm working on this now. My two ideas are:

- a table that has columns like "node classification", "link prediction", "representation learning", "model interpretability", "graph classification" with a row for each algorithm, and then links within the cells as appropriate (this might be better as a rewrite of the top-level README algorithm table)
- a flow-chart/question-answer style list

I'll experiment and see what works nicely.

timpitman:
> * a table that has columns like "node classification", "link prediction", "representation learning", "model interpretability", "graph classification" with a row for each algorithm, and then links within the cells as appropriate (this might be better as a rewrite of the top-level README algorithm table)

I like the idea of a better table if we can make it fit ok in the markdown

huonw:
This is now mostly done with #1200, but there's a few left-over pieces of work in the notes there:

- this only rewrites the basics, node-classification, link-prediction, graph-classification, embeddings, interpretability sub-readmes, the others are left for now to avoid this PR spiralling completely out of control (it would be good to land this before tomorrow morning) and will be done in follow up work
- we should probably add a glossary somewhere, so we don't have to explain/remind what terms like heterogeneous, inductive, ... means
- node-classification/node2vec is a bit weird: it's "really" a unsupervised algorithm and node classification is an downstream task on top of the embeddings computed with it, we could potentially fold it into embeddings (similar to https://github.com/stellargraph/stellargraph/issues/1124)
[...]

PR #1200, opened 2020-04-03
Rewrite demo READMEs to have more guidance

huonw:
This rewrites our demos READMEs to provide more guidance to users, focusing on two users:

1. people who have a graph and some vague idea of the task they're doing, but no idea of what algorithms to use and potentially not clear on the graph ML terminology (like "node classification" or "link prediction")
2. people who are familiar with algorithms (e.g. because they've read literature, or because they've started using an algorithm and need to consult a demo to remind themselves how to use it)

For 1, the top-level README now has a compact description of each task, so that people can jump to the right place. For 2, it also has a large table of all algorithms and what tasks they solve (that will become links).

For the individual task folders, the READMEs are rewritten to list each demo, rather than each algorithm. The table does include listing the algorithm(s) used, and other notable things about the demos, such as the types of graphs it uses (node features, heterogeneous, etc.) and any additional tasks demonstrated beyond just the main one (node classification on top of representation learning, etc.). This is designed to serve both 1 and 2: 1 can look at the columns related to their data/task, and 2 can search for the algorithms in which they are interested.

Rendered demos/README: https://github.com/stellargraph/stellargraph/tree/feature/1087-demo-guidance/demos

Notes:

- this adds more mentions of TensorFlow and Keras
- this makes the "all demos" binder link point to the README because that opens as a rendered document in Jupyter Lab, and the links all work
- we're having to use a HTML table to get hover text for explanations to keep the table narrow (without an affordance to indicate that there is hover text, unfortunately, due to Github filtering things out like a `style: 'text-decoration: underline;'` attribute, and not natively styling `<abbr>` or `<u>` elements)
- the table in the top-level demos readme is generated by `scripts/demo_table.py`, because maintaining it by hand will be impossible (e.g. keeping the hover text correct, having appropriate links, having links that are valid)
- the table is checked to be valid (including the links being valid) on CI using the `--action=compare` flag, and failures add an informative annotation (demonstration of a failure: https://buildkite.com/stellar/stellargraph-public/builds/2990)
- this only rewrites the `basics`, `node-classification`, `link-prediction`, `graph-classification`, `embeddings`, `interpretability` sub-readmes, the others are left for now to avoid this PR spiralling completely out of control (it would be good to land this before tomorrow morning) and will be done in follow up work
- we should probably add a glossary somewhere, so we don't have to explain/remind what terms like heterogeneous, inductive, ... means
- `node-classification/node2vec` is a bit weird: it's "really" a unsupervised algorithm and node classification is an downstream task on top of the embeddings computed with it, we could potentially fold it into `embeddings` (similar to https://github.com/stellargraph/stellargraph/issues/1124)
- `community_detection` could be folded into `embeddings` too, because it's also a downstream task on top of unsupervised GraphSAGE (maybe we could have a more explicit "downstream tasks" demo in `embeddings`, and then all of the others can just refer to that?)
- (moving demos around will break any links to them, but we could leave a notebook that just contained `Demo moved to [...](...)`.)

See: #1087


demos/node-classification/README.md
```diff
@@ -1,90 +1,42 @@
...
-8. On Calibration of Modern Neural Networks. C. Guo, G. Pleiss, Y. Sun, and K. Q. Weinberger.
-ICML 2017. ([link](https://geoffpleiss.com/nn_calibration))
-
-9. Simplifying Graph Convolutional Networks. F. Wu, T. Zhang, A. H. de Souza, C. Fifty, T. Yu, and K. Q. Weinberger.
-arXiv:1902.07153. ([link](https://arxiv.org/abs/1902.07153))
-
-10.	Predict then propagate: Graph neural networks meet personalized PageRank. J. Klicpera, A. Bojchevski, A., and S. Günnemann, ICLR, 2019, arXiv:1810.05997.([link](https://arxiv.org/abs/1810.05997))
-
-11. Cluster-GCN: An Efficient Algorithm for Training Deep and Large Graph Convolutional Networks. W. Chiang, X. Liu, S. Si, Y. Li, S. Bengio, and C. Hsiej, KDD, 2019, arXiv:1905.07953.([link](https://arxiv.org/abs/1905.07953))
+# Node classification using StellarGraph
```
timpitman, line 1: Overall, this new section works really well and the table provides an easy way to see which demos to use

demos/node-classification/README.md
```diff
@@ -1,90 +1,42 @@
...
+
+[StellarGraph](https://github.com/stellargraph/stellargraph) provides numerous algorithms for doing node classification on graphs. This folder contains demos of all of them to explain how they work and how to use them.
+
+A node classification task predicts an attribute of about each node in a graph. For instance, labelling each node with a categorical class (binary classification or multiclass classification), or predicting a continuous number (regression). It is supervised or semi-supervised, where the model is trained using a subset of nodes that have ground-truth labels.
```
timpitman, line 5: Possible typo here?
```suggestion
A node classification task predicts an attribute for each node in a graph. For instance, labelling each node with a categorical class (binary classification or multiclass classification), or predicting a continuous number (regression). It is supervised or semi-supervised, where the model is trained using a subset of nodes that have ground-truth labels.
```

demos/node-classification/README.md
```diff
@@ -1,90 +1,42 @@
...
+[ppnp]: ppnp/ppnp-cora-node-classification-example.ipynb
+[attri2vec]: attri2vec/attri2vec-citeseer-node-classification-example.ipynb
+[graphsage]: graphsage/graphsage-cora-node-classification-example.ipynb
+[graphsage-inductive]: graphsage/graphsage-pubmed-inductive-node-classification-example.ipynb
+[graphsage-directed]: graphsage/directed-graphsage-on-cora-example.ipynb
+[hinsage]: hinsage/README.md
+[node2vec]: node2vec/stellargraph-node2vec-node-classification.ipynb
+[node2vec-weighted]: node2vec/stellargraph-node2vec-weighted-random-walks.ipynb
+
+See [the root README](../../README.md) or each algorithm's documentation for the relevant citation(s).
```
timpitman, line 42: Referring back to the main README here is good - we only need the collection of citations in one place.

demos/README.md
```diff
@@ -1,95 +1,68 @@
...
+  - For graphs/networks: [**graph classification**](./graph-classification)
+  - Adjusting predictions scores to be probabilities (for any model): [**calibration**](./calibration)
+- Representation learning or computing embedding vectors (including unsupervised tasks):
+  - For nodes/vertices/entities and edges/links/connections: [**embeddings**](./embeddings) (and [**community detection**](./community_detection) for using these to cluster nodes)
+  - For graphs/networks: [**graph classification**](./graph-classification) (only supervised)
+- Ensembling models to reduce prediction variance: [**ensembles**](./ensembles)
+- Loading data into a `StellarGraph` object, with Pandas or NetworkX: [**basics**](./basics)
+- Experimental: running GraphSAGE on data stored in Neo4j: [**neo4j connector**](./connector/neo4j)
+
+## Find a demo for an algorithm
```
timpitman, line 22: This table does take a while to read, but I think it's a reasonable compromise in order to be able to make it fit vertically in markdown; I can't think of a better way.
Some possible suggestions:
* do we need to use all 'tick' glyphs here - some simpler combination of shapes and colours may be clearer?
* there's a mix in this and the other table of ✔️ and ✔ The smaller tick looks better but is too small; can we change the size enough with markdown to make it work?
* there's some small tick-box glyphs which don't look quite right, and aren't in the key below, eg:"☑ US"

demos/README.md
```diff
@@ -1,95 +1,68 @@
...
+
+| algorithm                                             | *heter.*    | *D* | *EW* | *NF* | NC | I(NC) | LP | RL                     | ind.            | GC |
```
timpitman, line 24: Is there anything possible with alt-text on images that could give mouseover hints to the columns?

demos/README.md
huonw: We've chatted a bit on slack, but in summary; I've switched to text, and it seems easier.

demos/README.md
huonw: We've chatted a bit on slack, in summary: I've switched to raw HTML, so I can use `<span title="...">` to get mouseover hints. It's generated by the `demo_table.py` script so that it's easier to update and maintain, and it is even checked on CI so that people don't accidentally edit this table when they should be editing the script.

huonw:
I've updated this (and updated the PR description), so I think it's ready for review.

demos/basics/README.md
```diff
@@ -1,6 +1,10 @@
-## Table of Contents
+# StellarGraph basics
 
-This folder contains examples of loading data into a `StellarGraph` object, which is the format used by the machine learning algorithms in this library.
+[StellarGraph](https://github.com/stellargraph/stellargraph) has support for loading data via Pandas and NetworkX. This folder contains examples of the loading data into a `StellarGraph` object, which is the format used by the machine learning algorithms in this library.
 
-- [`loading-pandas`](loading-pandas.ipynb) shows the recommend way to load data, using Pandas
+## Find demos for a format
+
+- [`loading-pandas`](loading-pandas.ipynb) shows the recommend way to load data, using Pandas (supporting any input format that Pandas supports, including CSV files and SQL databases)
```
kjun9, line 7:
```suggestion
- [`loading-pandas`](loading-pandas.ipynb) shows the recommended way to load data, using Pandas (supporting any input format that Pandas supports, including CSV files and SQL databases)
```

demos/embeddings/README.md
```diff
@@ -1,39 +1,35 @@
...
-**1.** Node2Vec: Scalable Feature Learning for Networks. A. Grover, J. Leskovec. ACM SIGKDD International Conference
-on Knowledge Discovery and Data Mining (KDD), 2016. ([link](https://snap.stanford.edu/node2vec/))
+| demo | algorithm(s) | training method | node features | downstream tasks shown |
+|---|---|---|---|---|
+| [Deep Graph Infomax][dgi] | GCN, GAT, PPNP, APPNP, GraphSAGE | `DeepGraphInfomax` (mutual information) | yes | visualisation, node classification |
+| [Unsupervised GraphSAGE][graphsage] | GraphSAGE | `UnsupervisedSampler` (link prediction) | yes | visualisation, node classification |
+| [Attri2Vec][attri2vec] | Attri2Vec | `UnsupervisedSampler` (link prediction) | yes | visualisation |
+| [Metapath2Vec][metapath2vec] | Metapath2Vec | natively unsupervised | | visualisation |
+| [Node2Vec][node2vec] | Node2Vec | natively unsupervised | | visualisation |
+| [Watch Your Step][wys] | Watch Your Step | natively unsupervised | | visualisation, node classification |
```
kieranricardo, line 24: Should the node features column contain "no" instead of being empty?

```suggestion
| [Watch Your Step][wys] | Watch Your Step | natively unsupervised | no | visualisation, node classification |
```

demos/README.md
```diff
@@ -1,95 +1,282 @@
...
+    <td>
+      <span title='UnsupervisedSampler, using link prediction'><a href='embeddings/embeddings-unsupervised-graphsage-cora.ipynb'>US</a></span>
+      <span title='DeepGraphInfomax, using mutual information'><a href='embeddings/deep-graph-infomax-cora.ipynb'>DGI</a></span>
+    </td>
+    <td><a href='node-classification/graphsage/graphsage-pubmed-inductive-node-classification-example.ipynb'>demo</a></td>
+    <td></td>
+  </tr>
+  <tr>
+    <td>HinSAGE</td>
+    <td>yes</td>
```
kieranricardo, line 190: Should this be "yes, nodes" (i hope i got the right column! this is meant to be the "heterogeneous" column of the hinsage row)

```suggestion
    <td>yes, nodes</td>
```

demos/link-prediction/README.md
```diff
@@ -1,62 +1,33 @@
...
+A link prediction task predicts an attribute of links/edges in a graph. For instance, predicting whether a link/edge that isn't already in the graph should exist (binary classification, or recommendation, or knowledge base completion, in a knowledge graph), or even labelling existing links with a categorical class (binary classification or multiclass classification), or predicting a continuous number (regression). It is supervised or semi-supervised, where the model is trained using a subset of links/edges that have ground-truth labels. For predicting edge existence, the ground-truth may just be whether the edge exists in the original data, rather than a separate label.
 
-The examples folder structure is shown below.
+Link prediction can also be done as a downstream task from node representation learning/embeddings, by combining node embedding vectors for the source and target nodes of the edge and training a supervised or semi-supervised classifier against the result. Unsupervised algorithms that can be used in this manner include random walk-based methods like Metapath2Vec. StellarGraph provides [demos of unsupervised algorithms](../embeddings).
```
kieranricardo, line 7: I'm not sure on this one, but I think some of these are unsupervised because the edges aren't explicitly labelled - they just already exist in the data.

demos/README.md
timpitman: Thanks, I find the new version much quicker to interpret; and as a bonus it doesn't need a key.

demos/graph-classification/README.md
```diff
@@ -0,0 +1,17 @@
+# Graph classification using StellarGraph
+
+[StellarGraph](https://github.com/stellargraph/stellargraph) allows for doing graph classification.
```
timpitman, line 3: possibly some alternate language to "allows for doing"? eg:

```suggestion
[StellarGraph](https://github.com/stellargraph/stellargraph) provides algorithms for graph classification.
```

demos/interpretability/README.md
```diff
@@ -0,0 +1,23 @@
+# Interpretability of node classification results using StellarGraph
+
+[StellarGraph](https://github.com/stellargraph/stellargraph) has support for inspecting several different algorithms for [node classification][nc] to understand or interpret how they came to a decision. This folder contains demos of all of them to explain how they work and how to use them as part of a TensorFlow Keras data science workflow.
```
timpitman, line 3: good SEO :)

demos/interpretability/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Interpreting a model involves training and making predictions for a model, and then analysing the predictions and the model to find which neighbours and which features had the largest influence on the prediction.
+
+## Find algorithms and demos
+
+This table lists interpretability demos, including the algorithms used.
+
+| demo | algorithm(s) |
+|---|---|
+| [GCN (dense)][gcn] | GCN, Integrated Gradients |
```
timpitman, line 13: This line doesn't render correctly for me?
Screen Shot 2020-04-07 at 3 30 13 pm

demos/embeddings/README.md
huonw: I thought that it was easier to read/scan the tables with whitespace for negatives, instead of `no`. For instance, for a table with more columns like link prediction:

| Demo | Algorithm | Node features | Heterogeneous | Temporal |
|---|---|---|---|---|
| [GCN][gcn] | GCN | yes | no | no |
| [Attri2Vec][attri2vec] | Attri2Vec | yes |no | no |
| [GraphSAGE][graphsage] | GraphSAGE | yes |no | no |
| [HinSAGE][hinsage] | HinSAGE | yes | yes | no |
| [Node2Vec][node2vec] | Node2Vec | no | no | no|
| [CTDNE][ctdne] | CTDNE | no | no | yes |
| [ComplEx][complex] | ComplEx | no | yes, multiple edge types | no |
| [DistMult][distmult] | DistMult | no | yes, multiple edge types | no |

[gcn]: gcn/cora-gcn-links-example.ipynb
[attri2vec]: attri2vec/stellargraph-attri2vec-DBLP.ipynb
[graphsage]: graphsage/cora-links-example.ipynb
[hinsage]: hinsage/movielens-recommender.ipynb
[node2vec]: random-walks/cora-lp-demo.ipynb
[ctdne]: random-walks/ctdne-link-prediction.ipynb
[complex]: knowledge-graphs/complex.ipynb
[distmult]: knowledge-graphs/distmult.ipynb

I find that much noisier, and, e.g. the `yes` for heterogeneous is harder to see at a glance.

The same is true for this one, but it is less bad, since there's only one columns with a few blanks. However, I think consistency is good and so leave no/negatives out.

demos/link-prediction/README.md
huonw: I think that's covered by the last sentence of the previous paragraph?

> For predicting edge existence, the ground-truth may just be whether the edge exists in the original data, rather than a separate label.

demos/README.md
huonw: I think HinSAGE supports different edge types. That is, it reasons about `source type --[edge type]-- target type` triples at each level of a sample, rather than just the source and target types.

https://github.com/stellargraph/stellargraph/blob/e10ab7b91a8d30a9fb21b6f475b1e4554ea6643d/stellargraph/data/explorer.py#L734-L754

.buildkite/pipeline.yml
```diff
@@ -197,6 +197,15 @@ steps:
     <<: *timeout
     command: "scripts/whitespace.sh --ci"
 
+  - label: ":book: check demo table"
+    command: "python scripts/demo_table.py --action=compare"
```
timpitman, line 201: I wonder if we should add (possibly in the script itself) output that makes it clearer what someone should do when CI breaks (ie instructing them how run the script locally, eg like we do for black)

huonw: We do: https://buildkite.com/stellar/stellargraph-public/builds/2990

(The link in the PR description was pointing to the wrong place before)

demos/graph-classification/README.md
huonw: Yeah, that new phrasing is way better...

demos/interpretability/README.md
huonw: Oops, I updated the link names without fixing this.

demos/interpretability/README.md
huonw: :)

demos/embeddings/README.md
kieranricardo: oh yeah I see what you mean - the blanks are way easier to processs 👍

demos/README.md
kieranricardo: yep youre right actually :+1:

.buildkite/pipeline.yml
timpitman: that looks perfect 👍 👍

demos/graph-classification/README.md
```diff
@@ -0,0 +1,17 @@
+# Graph classification using StellarGraph
+
+[StellarGraph](https://github.com/stellargraph/stellargraph) provides an algorithm for graph classification. This folder contains a demo to explain how it works and how to use it as part of a TensorFlow Keras data science workflow.
```
timpitman, line 3: we just need to be careful here as we'll very soon have more than one

huonw: Yup!

Issue #1201, opened 2020-04-03
~3 GCN layers that are nearly identical


huonw:
### Description

Currently there's three layers that implement the GCN layer:

- `GraphConvolution`
- `ClusterGraphConvolution`
- `GraphClassificationConvolution`

They all do basically the same thing, with just minor tweaks/differences to their `call` method.

https://github.com/stellargraph/stellargraph/blob/7012767a937865a4952bbf8c755efaad64353b86/stellargraph/layer/cluster_gcn.py#L182-L221

https://github.com/stellargraph/stellargraph/blob/7012767a937865a4952bbf8c755efaad64353b86/stellargraph/layer/gcn.py#L178-L223

https://github.com/stellargraph/stellargraph/blob/7012767a937865a4952bbf8c755efaad64353b86/stellargraph/layer/graph_classification.py#L84-L94

We should be able to merge them

PR #1204, opened 2020-04-06
Gather output nodes in full batch models, not the layers

huonw:
The core layer in a full batch method (like `GraphConvolution` inside `GCN`) natively computes information about a whole graph, but this doesn't directly work with machine learning (and, in particular, how it works with `Keras`). In particular, there's usually only a subset of labelled nodes, so training needs to only compute candidate embeddings for those nodes to compare against the ground-truth target array.

We currently handle this in full batch methods by having each layer know whether it is the `final_layer` in the model, and do `tf.gather` using a tensor of indices of relevant nodes. This means that the individual full-batch layers have to:
- always accept the tensor of indices, even if they won't use it (and, potentially, even if they are being used as part of a model that doesn't ever use it)
- do the appropriate `tf.gather` invocation to select the relevant nodes

This patch adjusts this approach to make the output filtering a model level concern: layers always compute all the information, and models that need to can do a `tf.gather` call (via the new `layer.misc.GatherIndices`) to select out the relevant output elements. This has a few benefits:

- it seems conceptually more "correct" to me, because it's the training/using of a overall model that needs to filter out elements, _not_ the operation of an individual layer
- it reduces the amount of code required noticeably, and will reduce if further when we remove the left-over-`final_layer` detection (that is designed to help us migrate).
- it makes #1201 easier (in #1205): graph classification doesn't have any output indices to feed into the graph convolution layers, and so this patch makes `GraphConvolution` closer to the behaviour required for that
- it may also ease implementing models that incorporate information from multiple convolution layers, because it's easy to make sure all the convolution layers compute embeddings for all nodes

See: #1201

### DIFF
--- a/stellargraph/layer/appnp.py
+++ b/stellargraph/layer/appnp.py
@@ -20,7 +20,7 @@ import tensorflow.keras.backend as K
 
 from ..mapper import FullBatchGenerator
 from .preprocessing_layer import GraphPreProcessingLayer
-from .misc import SqueezedSparseConversion, deprecated_model_function
+from .misc import SqueezedSparseConversion, deprecated_model_function, GatherIndices
 
 
 class APPNPPropagationLayer(Layer):
@@ -34,23 +34,15 @@ class APPNPPropagationLayer(Layer):
         Keras requires this batch dimension, and for full-batch methods
         we only have a single "batch".
 
-      - There are three inputs required, the node features, the output
-        indices (the nodes that are to be selected in the final layer)
+      - There are two inputs required, the node features,
         and the normalized graph Laplacian matrix
 
       - This class assumes that the normalized Laplacian matrix is passed as
         input to the Keras methods.
 
-      - The output indices are used when ``final_layer=True`` and the returned outputs
-        are the final-layer features for the nodes indexed by output indices.
-
-      - If ``final_layer=False`` all the node features are output in the same ordering as
-        given by the adjacency matrix.
-
     Args:
         units (int): dimensionality of output feature vectors
-        final_layer (bool): If False the layer returns output for all nodes,
-                            if True it returns the subset specified by the indices passed to it.
+        final_layer (bool): Deprecated, use ``tf.gather`` or :class:`GatherIndices`
         teleport_probability: "probability" of returning to the starting node in the propogation step as desribed  in
         the paper (alpha in the paper)
         input_dim (int, optional): the size of the input shape, if known.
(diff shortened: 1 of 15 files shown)


stellargraph/layer/appnp.py
```diff
@@ -149,18 +140,13 @@ def call(self, inputs):
...
         output = (1 - self.teleport_probability) * K.dot(
             A, propagated_features
         ) + self.teleport_probability * features
 
-        # On the final layer we gather the nodes referenced by the indices
-        if self.final_layer:
-            output = K.gather(output, out_indices)
-
         # Add batch dimension back if we removed it
         if batch_dim == 1:
```
PantelisElinas, line 151: I wonder since we check `batch_dim != 1` and raise `ValueError` in line 135, do we need to check `batch_dim == 1` here again? We would never get to this line if it wasn't true in the first place because of the earlier check.

PantelisElinas, line 151: I wonder since we check `batch_dim != 1` and raise `ValueError` in line 135, do we need to check `batch_dim == 1` here again? We would never get to this line if it wasn't true in the first place because of the earlier check.

stellargraph/layer/gcn.py
```diff
@@ -211,10 +200,6 @@ def call(self, inputs):
             output += self.bias
         output = self.activation(output)
 
-        # On the final layer we gather the nodes referenced by the indices
-        if self.final_layer:
-            output = K.gather(output, out_indices)
-
         # Add batch dimension back if we removed it
         # print("BATCH DIM:", batch_dim)
         if batch_dim == 1:
```
PantelisElinas, line 205: Same as before, I don't think we need to check that `batch_dim==1` here.

huonw: For this one, the `batch_dim == 1` special case is removed entirely in #1205.

stellargraph/layer/appnp.py
huonw: You're right that this condition is always true, but changing this here (and in the other locations that this has been copy-pasted around: GCN, RGCN, PPNP and GAT) I feel is better handled separately. And, in particular, I feel that it would be better to do something more like #1205 which removes the `batch_dim == 1` special case entirely, so that, for instance, these layers can be used for graph classification.

huonw: I've filed https://github.com/stellargraph/stellargraph/issues/1214 for this work.

PR #1205, opened 2020-04-06
Merge Cluster-GCN and GCN graph classification into GCN layer

huonw:
`ClusterGraphConvolution` and `GraphConvolution` were practically identical, with just some minor differences in the details of the code, and `GraphClassificationConvolution` only differed by:

1. not using any output indices, which is moved out of `GraphConvolution` in #1204
2. supporting a batch of multiple graphs (i.e. multiple sets of node features and adjacency matrices); the core of this patch is thus expanding `GraphConvolution` to handle a batch size > 1 for dense matrices, not sparse ones (yet, since it's harder: Keras's `.batch_dot` doesn't seem to support sparse matrices).

That is, by building on #1204, this patch can merge all three of our implementations of the GCN layer.

This deprecates `ClusterGraphConvolution` because it has been released for a while, but removes `GraphClassificationConvolution` entirely, because it has only been released `@experimental`-y.

See: #1201


stellargraph/layer/gcn.py
```diff
@@ -143,8 +149,8 @@ def build(self, input_shapes):
             input_shapes (list of int): shapes of the layer's inputs (node features and adjacency matrix)
 
         """
-        feat_shape = input_shapes[0]
-        input_dim = int(feat_shape[-1])
+        feature_shape, *As_shapes = input_shapes
```
kjun9, line 152: Doesn't seem like `As_shapes` is used in this method, should we just leave it as `feature_shape = input_shapes[0]`?

stellargraph/layer/gcn.py
```diff
@@ -181,30 +187,33 @@ def call(self, inputs):
...
-
-        # Remove singleton batch dimension
-        features = K.squeeze(features, 0)
 
         # Calculate the layer operation of GCN
         A = As[0]
-        h_graph = K.dot(A, features)
+        if K.is_sparse(A):
+            # batch_dot doesn't support sparse tensors, so we special case them to only work with a
+            # single batch element (and the adjacency matrix without a batch dimension)
```
kjun9, line 195: So is this something we don't need to support? Or is it something to fix?

kjun9: Hmm I see 🤔 

Ah by "is this something we don't need to support" I actually meant do we know if users will run into these errors (in which case, maybe suggesting to use dense matrices would be better than suggesting that the batch dimension in wrong?), but sounds like they shouldn't if they are using the generators we provide.

huonw: Yeah, they shouldn't hit it with our generators. If the generators are behaving incorrectly, hopefully that's something we'd catch first.

(Also, it turns out a sparse batch dot may not be _that_ hard: https://github.com/tensorflow/tensorflow/issues/9210#issuecomment-497889961)

I'll file an issue.

PR #1208, opened 2020-04-06
Remove duplication in APPNP.{in_out_tensors,propagate_model}

huonw:
The `APPNP` model `propagate_model` method ended up reimplementing both the `Input` tensor creation from `in_out_tensors` and the layer application from `__call__`. This PR merges `APPNP.propagate_model` and `APPNP.in_out_tensors` by factoring out the common code into a `_tensors` private method.

It also adds a test that `propagate_model` is implemented correctly, since it's now slightly less obvious (also, #1213).


stellargraph/layer/appnp.py
```diff
@@ -362,14 +362,7 @@ def __call__(self, x):
...
 
-    def in_out_tensors(self, multiplicity=None):
-        """
-        Builds an APPNP model for node or link prediction
-
-        Returns:
-            tuple: `(x_inp, x_out)`, where `x_inp` is a list of Keras/TensorFlow
-            input tensors for the model and `x_out` is a tensor of the model output.
-        """
+    def _tensors(self, multiplicity, extra_layers):
```
kieranricardo, line 365: Im not sure if I'm missing someting, but `extra_layers` doesn't seem to be used here so I think this is always just creating a regular APPNP model

huonw: Oh, oops. Missed the loop.

kieranricardo:
@huonw why the ppnp demo start failing?

huonw:
Yeah, I noticed but haven't investigated yet. I'm putting this PR on the back burner for a little while to get my blog post out, but will look into it later this afternoon or tomorrow.

huonw:
The problem was the `propagate_model` version _replaces_ the `APPNP` model's dense layer (i.e. base_model -> propagate), but I originally had it done in addition (i.e. base_model -> dense -> propagate). I've refactored slightly to make it easy to do a swap.

stellargraph/layer/appnp.py
```diff
@@ -349,7 +333,26 @@ def __call__(self, x):
 
         return h_layer
 
-    def _tensors(self, multiplicity, extra_layers):
+    def __call__(self, x):
```
kieranricardo, line 336: It looks like `__call__` now just wraps the `_run` function. Could you say why you went for this approach instead of something like adding a `feature_layers` kwarg in `__call__? 
 e.g.

```
def __call__(self, x, feature_layers=None):
    ...
    if feature_layers is None:
        feature_layers = self.feature_layers
    ...
```

huonw: I didn't think it makes sense for a user to see or use the `feature_layers` argument, so it's better to not provide it at all, rather than having it hanging around and potentially confusing things (and also, if it exists, we have to support it forever, or risk breaking users if we change things). What do you think?

kieranricardo: Cool, I think that's a good point. The the user shouldnt even need to use the `feature_layers` arg

Issue #1207, opened 2020-04-06
Allow D^(-1)A propagation for PaddedGraphGenerator objects 


PantelisElinas:
### Description

The `PaddedGraphGenerator` class only allows for adjacency matrix normalisation based on the GCN model. However, some algorithms such as DGCNN use a different normalisation such as D^-1. This ticket is about updating the `PaddedGraphGenerator` and `PaddedGraphSequence` classes to allow the user to specify how to normalise the graph adjacency matrices.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed

PR #1212, opened 2020-04-06
DGCNN implementation

PantelisElinas:
This implements the DGCNN supervised graph classification algorithm from [1].

One difference between this implementation and [1] is the adjacency normalisation method used. See #1207. 

This pull request depends on #1210.

1. An End-to-End Deep Learning Architecture for Graph Classification, M. Zhang et al, AAAI, 2018, https://www.cse.wustl.edu/~muhan/papers/AAAI_2018_DGCNN.pdf


stellargraph/layer/sort_pooling.py
```diff
@@ -0,0 +1,99 @@
...
+
+                where B is the batch size, N is the number of nodes in the largest graph in the batch, and
+                F_i is the dimensionality of node features output from the i-th convolutional layer.
+
+        Returns:
+            Keras Tensor that represents the output of the layer.
+        """
+        outputs = tf.map_fn(
+            lambda x: tf.gather(
+                x, tf.argsort(x[..., -1], axis=0, direction="DESCENDING")
```
huonw, line 84: Also, I think this currently doesn't handle padding perfectly; any padding values may be sorted higher than some elements, and this may lead to peculiar behaviour if as graphs are grouped together differently.

For a small example of what I'm thinking, suppose `k=2` and `batch_size=2`, and we've got three graphs G1, G2 and G3, with 2, 2, 3 nodes respectively. If the network gives G1's nodes sorting score 1 and -1, then for a batch `G1, G2`, the final output will include both of those nodes, but for a batch `G1, G3`, I think the sorting tensor may look like:

```
[
 [..., 1],
 [..., -1],
 [..., 0] # padding
]
```

(For ease of identifying the elements I'm assuming that there's no biases/they cancel out for the all-zero paddings.)

And thus the chosen elements will the higher node and the padding `[[..., 1], [..., 0]]`.

Should this take a `mask` argument and obey it (somehow)?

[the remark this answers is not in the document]
PantelisElinas: This is addressed in #1210.

PantelisElinas: This is a valid concern but belongs in #1210 and not here.

stellargraph/layer/graph_classification.py
```diff
@@ -250,3 +243,123 @@ def in_out_tensors(self):
         return x_inp, x_out
 
     build = deprecated_model_function(in_out_tensors, "build")
+
+
+@experimental(reason="Missing unit tests and generally untested.")
+class DeepGraphConvolutionalNeuralNetwork(GCNSupervisedGraphClassification):
```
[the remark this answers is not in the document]
PantelisElinas: I did consider this option and I decided against it for several reasons. 

First of all, I don't like monstrous model classes that try to implement many different models in one by considering different combinations of parameter values. For example, if in the future we want a model that replaces the GCN layers with GAT or another layer, then we have to add more parameters. I find that having to keep track that the correct combination of parameters is given for each model implemented by a class is going to be difficult to maintain in the future and difficult to explain to the end-user. 

In addition, model classes should be named after the algorithm implemented; so just like we have model classes for GAT, GCN, GraphSAGE, APPNP, etc. the DGCNN model has a clearly identifiable name in the literature (although I did opt out for using the full name instead of the acronym) and our model class should be named after it instead of hiding it behind the generic `GCNSupervisedGraphClassification` name. 

Finally, it was no accident that I could implement DGCNN by subclassing the `GCNSupervisedGraphClassification` class; it was by design and I prefer to keep it this way.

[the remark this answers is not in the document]
huonw: It's a good point that it's good to have names that match the papers for ease of reference/reproduction.
 
However, I'm concerned that we're going to end up with an explosion of models for each paper as special cases each with their own implementation, rather than distilling out the core pieces that slot together in reusable ways (especially when the pieces are so orthogonal, in this case). For instance, other models use outputs from multiple convolution layers (e.g. UGraphEmb) suggesting it's a useful/reuseable technique.

I think we can work together find a middle ground that offers the benefits of both:

- the flexible model class, to reduce how much work we do, and offer expert users full control
- named/easy-to-access models for published papers

For instance, we could have the `GCNSupervisedGraphClassification` class support all of the arguments (with sensible defaults), and then subclasses that slot in the appropriate arguments in the `__init__`, and then don't need to override any other functions:

```python
class DeepGraphConvolutionalNeuralNetwork(GCNSupervisedGraphClassification):
    """
    Args:
        ...
        k (int): size (number of rows) of output tensor.
        ...
        kwargs: any other arguments for :class:`GCNSupervisedGraphClassification`.
    """
    def __init__(self, layer_sizes, activations, k, generator, **kwargs):
        super().__init__(
            layer_sizes=layer_sizes,
            activations=activations,
            generator=generator,
            pooling=SortPooling(k=k),
            pool_all_gcn_layers=True,
            **kwargs,
       )
```

Instead of a subclass, this could even be a function:

```python
def deep_graph_convolutional_neural_network(layer_sizes, activations, k, generator, **kwargs):
    return GCNSupervisedGraphClassification
        layer_sizes=layer_sizes,
        activations=activations,
        generator=generator,
        pooling=SortPooling(k=k),
        pool_all_gcn_layers=True,
        **kwargs,
    )
```

What do you think?

huonw, line 249: On the topic of naming (separate comment, because it's orthogonal to the code-sharing discussion above), we have `GCN` not `GraphConvolutionalNetwork` and `GAT` not `GraphAttentionNetwork` etc. Thus, maybe this should be:

```suggestion
class DGCNN(GCNSupervisedGraphClassification):
```

PantelisElinas: Main reason I went with the full name was that I felt we are oversaturated with acronyms and thought we might want to start naming classes using the full name of the algorithms. 

I leave it up to you to decide which is best.

PantelisElinas:
@huonw 

(Looks like the GraphSAGE reproducibility test has failed for the latest build and I cannot re-run it without pushing new code. Maybe you can re-run the tests?) 

I have updated the current implementation to use the updated `SortPooling` layer. I had to modify the `SortPooling` layer to allow to flatten the output tensor as described in the paper and required as input to the Keras CNN layers. The model now also uses the correct adjacency matrix normalisation (multiply D^-1xA instead of the GCN-like D^-1/2xAxD^-1/2) and also adds self loops (uses A+I instead of A). 

I think the reference implementation is now very close to what is proposed in the paper. 

While I am working on a demo using the larger PROTEINS datasets, you can use our existing graph classification demo with MUTAG to verify that we can reproduce the reported results. See below for the code you need to change in that demo notebook to run the experiments. 

The accuracy on MUTAG using 10-fold cross validation is reported in the paper (Table 1) as ~85.83% plus/minus 1.66. Using the below code, I get 84.3% and std: 4.9%. So the mean is about the same but the standard deviation is a bit higher. I had a look at the 10 accuracies (one for each fold) and they are [0.93859655, 0.84210527, 0.84210527, 0.8070175, 0.8684211, 0.81578946, 0.8684211, 0.8333333, 0.8703703, 0.7407407]. If we exclude the first and last that look like outliers, then performance is 84.3% and std: 2.3%, much close to the reported values in the paper. I don't know why we get these extreme results. I hope working with the larger PROTEINS dataset, we are going to get a result with lower variance and still similar to that reported in the paper.

You can verify the above result using our graph classification demo notebook. You need to replace the method `create_graph_classification_model` with the following (everything else should be the same),

```
def create_graph_classification_model(generator, k=20):

    dgcnn_model = DeepGraphConvolutionalNeuralNetwork(
        layer_sizes=[32, 32, 32, 1],
        activations=["tanh", "tanh", "tanh", "tanh"],
        k=k,
        generator=generator,
        kernel_regularizer="l2",
        bias_regularizer="l2"
    )
    #
    x_inp, x_out = dgcnn_model.in_out_tensors()

    # Add 1D CNN layer
    predictions = Conv1D(filters=16, kernel_size=97, strides=97)(x_out)
    predictions = MaxPool1D(pool_size=2)(predictions)

    predictions = Conv1D(filters=32, kernel_size=5, strides=1)(predictions)

    predictions = Dense(units=128, activation="relu")(predictions)
    predictions = Dropout(rate=0.5)(predictions)
    predictions = Dense(units=1, activation="sigmoid")(predictions)
    
    k_model = Model(inputs=x_inp, outputs=predictions)

    k_model.compile(
        optimizer=Adam(lr=0.001),
        loss=binary_crossentropy,
        metrics=["acc"],
    )

    return k_model

```

huonw:
> (Looks like the GraphSAGE reproducibility test has failed for the latest build and I cannot re-run it without pushing new code. Maybe you can re-run the tests?)

Done.

README.md
```diff
@@ -184,7 +184,7 @@ The StellarGraph library currently includes the following algorithms for graph m
 | Deep Graph Infomax [15] | Deep Graph Infomax trains unsupervised GNNs to maximize the shared information between node level and graph level features. |
 | Continuous-Time Dynamic Network Embeddings (CTDNE) [16] | Supports time-respecting random walks which can be used in a similar way as in Node2Vec for unsupervised representation learning. |
 | DistMult [17] | The DistMult algorithm computes embeddings for nodes (entities) and edge types (relations) in knowledge graphs, and can use these for link prediction |
-
+| DGCNN [18] | The Deep Graph Convolutional Neural Network (DGCNN) algorithm for supervised graph classification. |
```
huonw, line 187:
```suggestion
| DGCNN [18] | The Deep Graph Convolutional Neural Network (DGCNN) algorithm for supervised graph classification. |

```

stellargraph/core/utils.py
```diff
@@ -32,18 +32,22 @@ def is_real_iterable(x):
...
         symmetric: True if symmetric normalization or False if left-only normalization
-
+        add_self_loops: True if self loops are to be added before normalization, i.e., use A+I where A is the adjacency
+            matrix and I is a square identity matrix of the same size as A.
     Returns:
         Return a sparse normalized adjacency matrix.
     """
 
+    if add_self_loops:
+        adj = adj + sp.diags(np.ones(adj.shape[0]) - adj.diagonal())
```
huonw, line 49: What's the reason for subtracting off existing self-loops? The papers I can remember all just use `A + I`, not `A + I - diag(A)` or similar, and e.g. the GCN code uses just `A + I` too (https://github.com/tkipf/gcn/blob/dab6e6b0d936c6590fb1a1b3f8fd7cef9fc4c858/gcn/utils.py#L134). The doc-string is currently slightly misleading too.

```suggestion
        adj = adj + sp.diags(np.ones(adj.shape[0]))
```

(This is a question so I can understand the approach, as I see that it's duplicated in several other places in this file too.)

stellargraph/layer/graph_classification.py
```diff
@@ -189,3 +180,145 @@ def in_out_tensors(self):
...
+            dropout=dropout,
+            kernel_initializer=kernel_initializer,
+            kernel_regularizer=kernel_regularizer,
+            kernel_constraint=kernel_constraint,
+            bias_initializer=bias_initializer,
+            bias_regularizer=bias_regularizer,
+            bias_constraint=bias_constraint,
+        )
+
+        if not isinstance(k, int):
```
huonw, line 276: (Just an FYI, doesn't have to change here)

This validation seems better suited to `SortPooling`, and could be done more simply with `core.validation.require_integer_in_range(k, "k", min_val=1)`:

https://github.com/stellargraph/stellargraph/blob/c5af1bf37bb73cd9b2d5b6d30ce971f0a9331e9d/stellargraph/core/validation.py#L59-L67

stellargraph/layer/sort_pooling.py
```diff
@@ -64,7 +64,7 @@ def compute_output_shape(self, input_shapes):
             An input shape tuple.
         """
         if self.flatten_output:
-            return input_shapes[0], self.k * input_shapes[2]
+            return input_shapes[0], self.k * input_shapes[2], 1
```
huonw, line 67: Just so I understand, what's the reason for the extra `1` here?

stellargraph/layer/graph_classification.py
```diff
@@ -189,3 +180,145 @@ def in_out_tensors(self):
         return x_inp, x_out
 
     build = deprecated_model_function(in_out_tensors, "build")
+
+
+@experimental(reason="Missing unit tests and generally untested.")
```
huonw, line 185:
```suggestion
@experimental(reason="Missing unit tests and generally untested.", issues=[1297])
```

stellargraph/layer/sort_pooling.py
PantelisElinas: The `Conv1D` layer requires input to be 3D tensor. See https://www.tensorflow.org/api_docs/python/tf/keras/layers/Conv1D

stellargraph/core/utils.py
PantelisElinas: In all the papers, it is assumed (as far as I understand) that the input graph has no self loops. So, the operation A+I correctly adds a self loop for every node in the graph. The reason for this operation is to include the node's features in the calculation. For most datasets we use, e.g., Cora, MUTAG, PROTEINS, it is true that all nodes do not have self loops and hence, A+I works correctly. The problem arises if someone has a dataset with some nodes already having self loops or even all nodes already having self loops; this could be the case for custom datasets. Then A+I increases the weights for the nodes' self features unnecessarily. I don't know if having a 2 instead of 1 will really make a big difference but being careful about it seems reasonable. I suspect papers use A+I in all the equations for convenience; the practical implementation of this operation needs to be done a bit more carefully.

Our own GCN normalisation routine `core.utils.GCN_Aadj_feats_op` (which I have to admit I don't know if we still use) uses the same calculation and this is where the above line comes from in the first place.

huonw: Fair enough.

> The reason for this operation is to include the node's features in the calculation

A different framing is that `A + I` is emphasising the self-nodes features rather than just including, in which case doing `A + I` directly seems like a reasonable interpretation. (I think [GIN](https://arxiv.org/abs/1810.00826) effectively does this, via `A + (1 + eps) * I` for some trainable (or not) `eps`.)

What do you think?

> the same calculation and this is where the above line comes from in the first place

Indeed: a better way to do this would've been to move (rather than copy) that calculation, and have `GCN_Aadj_feats_op` call `normalize_adj(..., add_self_loops=True)`, but this can be done elsewhere/elsewhen.

> which I have to admit I don't know if we still use

For future reference, this sort of question can be easily answered (in the affirmative, in this case) with `git grep`:

```
$ git grep GCN_Aadj_feats_op
...
stellargraph/mapper/full_batch_generators.py:from ..core.utils import GCN_Aadj_feats_op, PPNP_Aadj_feats_op
stellargraph/mapper/full_batch_generators.py:            self.features, self.Aadj = GCN_Aadj_feats_op(
...
```

PantelisElinas: > Fair enough.
> 
> > The reason for this operation is to include the node's features in the calculation
> 
> A different framing is that `A + I` is emphasising the self-nodes features rather than just including, in which case doing `A + I` directly seems like a reasonable interpretation. (I think [GIN](https://arxiv.org/abs/1810.00826) effectively does this, via `A + (1 + eps) * I` for some trainable (or not) `eps`.)
> 
> What do you think?
> 

This might be another reasonable way to explain the A+I operation. My above quoted statement is based on how it is described in prior-art such as for example the DGCNN paper states, "Our proposed graph convolution adopts a propagation D^-1A (with A=A+I, my comment) which is very similar to that of PK, except for preserving nodes' self-information over iterations."

Also Kipf's GCN paper states, "Here, A = A + I is the adjacency matrix of the undirected graph G with added self-connections." which does not imply this operation is meant to emphasise nodes' self features but just to include them. 

That said, GIN does indeed extend the concept of simply adding self-connections so...basically, I have no idea what the correct justification for A+I  should be!

huonw: Yeah, that makes sense. Maybe we can (eventually) offer all the options and it's up to a user to decide which one applies to their graph.

Issue #850, opened 2020-02-12
Add brute force importance in stellargraph library


kjun9:
### Description

Currently the brute-force approach of calculating node/link importance is implemented in the demo notebooks, however this functionality could be moved into the library:

* it will remove duplication of this logic across the demos
* it makes it easier for us to compare the brute force vs integrated gradient approaches in tests
* the brute force approach can be an option for users when using any of the other algorithms that do not currently have the integrated gradient approach implemented

### User Story

**As** a data scientist

**I want to** use the brute force importance calculations in stellargraph

**So that** compare with IG importances or use it for algorithms that don't have IG implementations

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

kjun9:
After having thought about and experimented a bit, I'm now thinking of doing a few pieces leading up to this work that roughly seems to me like it'd work for methods that use full-batch or sampled generators. 

Initially wanted to just create a corrupting-wrapper sort of thing around the `StellarGraph` class (https://github.com/stellargraph/stellargraph/pull/1217) but it ended up looking a bit too fragile to me given that there's no guarantee that things don't start breaking if you change the underlying graph in any given stellargraph pipeline.

So now, the eventual resulting API I'm thinking for brute force importance is something like:

```python
model = ... # some node model
generator = ... # some node generator

target_node_id = ... # node to make predictions for
target_class = ... # class for which you want to compare prediction scores for (for multi-class classification)
node_to_exclude = ... # you're wondering how important this node is

# baseline prediction without any "corruption"
pred_baseline = model.predict(generator.flow([target_node_id]))[0]

# sequence with missing node
seq = generator.flow_exclude_node([target_node_id], node_to_exclude)
pred_exclude_node = model.predict(seq)[0]

# take the difference to get the importance of excluded node on the target node
importance_score = pred_baseline[target_class] - pred_exclude_node[target_class]
```

## 1. Implement `flow_exclude_node` and `flow_exclude_edge` for full batch generators only

This looks like it'd be relatively straightforward for full batch generators, since their `flow` methods simply create a `FullBatchSequence` using the feature and adjacency matrices
```python
class FullBatchGenerator:
    def flow_exclude_node(ids, node_to_exclude):
[...]

PR #1217, opened 2020-04-07
Add brute force node importance

kjun9:
This adds brute force node importance with a class that overrides the `node_features` method in `StellarGraph`

See #850

### DIFF
--- a/stellargraph/interpretability/saliency_maps/brute_force.py
+++ b/stellargraph/interpretability/saliency_maps/brute_force.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 Data61, CSIRO
+#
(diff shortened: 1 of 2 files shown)


kjun9:
I think this ends up looking a bit clunky still, and I'm a bit unconvinced that it will be reliable since it assumes that given a trained model, we can always define a function that looks like:
```python
def predict(graph, node) -> score
```
Currently an entire stellargraph ML pipeline has the freedom to be defined in a way that's quite tightly coupled with the original graph that it was created with, so I'm not sure if this will always be possible. 

For GCN, the dimensions in a model/layer comes from the dimensions in a generator, and a generator is defined using a graph, so we can be careful that the dimensions are kept consistent even when we feed in new graphs.. but you could theoretically have a generator and model that does something a lot more tightly coupled with the graph. Not sure how likely all this is though.

Issue #1158, opened 2020-03-24
Implement the "Random" strategy for knowledge-graph completion evaluation


huonw:
### Description

In [1], the authors discuss different ways of evaluating the performance of a knowledge graph algorithm. In particular, for deciding the rank of a true edge amongst the negative edges when there are ties. In section 3.2, they show that some algorithms have an "unusual score distribution", with many ties for the valid edge. In section 4, they give three approaches for handling the ties:

- Top: the correct triplet is ranked first amongst its ties (for instance, given scores [123, 1, 1, 1], where the correct triplet is one of the 1s, the rank would be 2)
- Bottom: the correct triplet is ranked last amongst its ties (for instance, in the example above, rank would be 4)
- Random: the correct triplet has its rank randomly chosen between the Top and Bottom ranks (for instance, in the example above, the rank would randomly be one of 2, 3, or 4)

They recommend using Random.

Sun, Zhiqing, Shikhar Vashishth, Soumya Sanyal, Partha Talukdar, and Yiming Yang. “A Re-Evaluation of Knowledge Graph Completion Methods.” ArXiv:1911.03903 [Cs], November 10, 2019. http://arxiv.org/abs/1911.03903.

PR #1223, opened 2020-04-08
Support 'random' and 'bottom' knowledge graph rank eval strategies

huonw:
This follows [1] to add two new evaluation strategy for knowledge graphs, particularly focused on how ties are broken when ranking edges.

Recall that our knowledge graph ranking code compares the model's computed score for an edge like `E = (s, r, t)` (`s` = source, `r` = relationship type, `t` = target) against the scores of corruptions `E' = (s, r, n)` for all `n` in the graph and similarly for `(n, r, t)`. The rank of `E` is the number of corrupted edge `E'` scores that are greater than `E`'s (ignoring edges in the train, test or validation set for "filtered" mode).

If there's a tie where score of `E` = score of `E'` (for one or more `E'`), this is implicitly ranking `E` higher than all of those tied `E'`. [1] shows that this misleading, because an algorithm that gives the same score to every edge (true or corrupted) gets `rank = 1` for every edge under this approach.

The paper describes 3 ranking strategies for breaking ties:

- Top: the true edge is ranked before all tied corruptions (as described above)
- Bottom: the true edge is ranked after all tied corruptions, dramatically penalising ties
- Random: the true edge is given a rank somewhere between top and bottom randomly (which averages out across many ties)

The paper describes the "Random" one as fairest/best, and so we default to it.

Changing the strategy makes little difference for "simple" algorithms like ComplEx and DistMult, because ties are rare (in a probabilistic framing, there's a very small chance that two bilinear products of 3 "random" vectors ends up being equal). This is more useful for non-linear/neural-network-based models that use activations functions like RelU, since these make it easy to end up with many (or all) zeros in hidden layers, and thus more quantisation (and more ties) in scores. We may implement them in future.

[1]: Sun, Zhiqing, Shikhar Vashishth, Soumya Sanyal, Partha Talukdar, and Yiming Yang. “A Re-Evaluation of Knowledge Graph Completion Methods.” ArXiv:1911.03903 [Cs], November 10, 2019. http://arxiv.org/abs/1911.03903.

See: #1158


huonw:
Oh, oops, I haven't pushed the unit test yet, so not quite ready for review.

huonw:
Ok this is ready for review now; sorry for assigning you too early @kieranricardo.

PR #1232, opened 2020-04-09
Ehancement/adj list by type

kieranricardo:


huonw:
What's the status of this PR? Does it relate to #1399?

kieranricardo:
woah i don't even remember this PR! I think i very briefly explored something similar #1399 but i didn't get very far. I'll close this

Issue #1069, opened 2020-03-12
StellarGraph.info: include information about edge weights


huonw:
### Description

The `info` method shows somewhat detailed information about node types like the count, the size of the features, and any edges that touch on that node type. We could basic descriptive statistics about the edge weights too, e.g. if a graph had weights that range from a min of 0.3 to a max of 2.1, it might show something like

```
StellarGraph: Undirected multigraph

...

 Edge types:
    paper-cites->paper: [5429]
        Weights: range=[0.3, 2.1], mean=1.5, std=0.1
```

An edge type with the default weight (all weights 1) could show the conventional thing `Weights: range=[1.0, 1.0], mean=1.0, std=0.0` or some special case `Weights: default (all 1.0)`.

### User Story

**As** a data scientist

**I want to** get a high-level view of the graph's edge weights

**So that** can quickly validate construction seemed correct and move onto real ML

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1220, opened 2020-04-08
Create an explicit API for handling node ilocs

kieranricardo:
This PR makes a number of changes to support iloc usage:

- the nodes are stored as ilocs in `StellarGraph._edges` instead of node ids
- where possible `StellarGraph` functions now have a `use_ilocs=False`  argument which specifies that any nodes input to the function are ilocs and node ilocs should be returned.
- `StellarGraph_get_index_for_nodes` will be made public
- tests for `use_ilocs=True` 

To keep the scope of this PR small, this PR does not change any samplers/generators. Many of these could be updated to natively use ilocs by passing `use_ilocs=True` to `StellarGraph` functions.

See #870

PR #1240, opened 2020-04-15
Add edge weight info in graph info

kjun9:
See #1069 

I noticed that current behaviour is to show the edge type triples `A-R->B` and `B-R->A` separately in the info string even when the graph is undirected - should we be collapsing them?


stellargraph/core/graph.py
```diff
@@ -832,6 +833,13 @@ def _unique_type_triples(self, *, return_counts, selector=slice(None)):
 
         return zip(*unique_ets)
 
+    def _edge_metrics_by_type_triple(self, *, selector=slice(None), metrics):
+        src_ty, rel_ty, tgt_ty = self._edge_type_triples(selector)
+        df = pd.DataFrame(
+            self._edges.weights, index=[src_ty, rel_ty, tgt_ty], columns=["weight"]
```
huonw, line 839: This might be slightly clearer just using columns:

```suggestion
            {"src_ty": src_ty, "rel_ty": rel_ty, "tgt_ty": rel_ty, "weight": self._edges.weights}
```

and then `df.groupby(["src_ty", "rel_ty", "tgt_ty"])` on the next line.

It also turns out to be slightly faster because constructing the MultiIndex seems to be quite expensive. However, the difference is small enough that we should be focusing on the clarity, so I'm happy to go with whatever you prefer.

The following tables show using 4 columns (`columns`) vs the weight column + multiindex (`index`), for various numbers of edges (`n`), for creating the DataFrame (`creation`), running the groupby on an existing data frame (`groupby`) and the sum of the two (`total`). Times in milliseconds averaged across 100 runs.

1 node type, 1 edge type:
<table>
  <thead>
    <tr>
      <th></th>
      <th colspan="2" halign="left">creation</th>
      <th colspan="2" halign="left">groupby</th>
      <th colspan="2" halign="left">total</th>
    </tr>
    <tr>
      <th></th>
      <th>columns</th>
      <th>index</th>
      <th>columns</th>
      <th>index</th>
      <th>columns</th>
      <th>index</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>10</th>
      <td>0.68</td>
      <td>1.56</td>
      <td>1.58</td>
      <td>1.24</td>
      <td>2.26</td>
      <td>2.79</td>
    </tr>
    <tr>
      <th>100</th>
      <td>0.53</td>
      <td>1.27</td>
      <td>1.69</td>
      <td>1.27</td>
      <td>2.22</td>
      <td>2.54</td>
    </tr>
    <tr>
      <th>1000</th>
      <td>0.54</td>
      <td>1.32</td>
      <td>1.43</td>
      <td>1.18</td>
      <td>1.97</td>
      <td>2.49</td>
    </tr>
    <tr>
      <th>10000</th>
      <td>0.53</td>
      <td>1.32</td>
      <td>1.73</td>
      <td>1.39</td>
      <td>2.27</td>
      <td>2.70</td>
    </tr>
    <tr>
      <th>100000</th>
      <td>2.62</td>
      <td>2.82</td>
      <td>4.25</td>
      <td>6.83</td>
      <td>6.87</td>
      <td>9.65</td>
    </tr>
    <tr>
      <th>1000000</th>
      <td>5.27</td>
      <td>31.72</td>
      <td>48.55</td>
      <td>29.56</td>
      <td>53.83</td>
      <td>61.28</td>
    </tr>
  </tbody>
</table>

1 node type, 10 edge types:

<table>
  <thead>
    <tr>
      <th></th>
      <th colspan="2" halign="left">creation</th>
      <th colspan="2" halign="left">groupby</th>
      <th colspan="2" halign="left">total</th>
    </tr>
    <tr>
      <th></th>
      <th>columns</th>
      <th>index</th>
      <th>columns</th>
      <th>index</th>
      <th>columns</th>
      <th>index</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>10</th>
      <td>0.64</td>
      <td>1.32</td>
      <td>1.76</td>
      <td>1.21</td>
      <td>2.40</td>
      <td>2.53</td>
    </tr>
    <tr>
      <th>100</th>
      <td>0.57</td>
      <td>1.22</td>
      <td>1.49</td>
      <td>1.23</td>
      <td>2.05</td>
      <td>2.46</td>
    </tr>
    <tr>
      <th>1000</th>
      <td>0.53</td>
      <td>1.22</td>
      <td>1.46</td>
      <td>1.27</td>
      <td>1.99</td>
      <td>2.49</td>
    </tr>
    <tr>
      <th>10000</th>
      <td>0.68</td>
      <td>1.41</td>
      <td>1.74</td>
      <td>1.30</td>
      <td>2.42</td>
      <td>2.71</td>
    </tr>
    <tr>
      <th>100000</th>
      <td>1.24</td>
      <td>4.01</td>
      <td>4.74</td>
      <td>6.25</td>
      <td>5.97</td>
      <td>10.26</td>
    </tr>
    <tr>
      <th>1000000</th>
      <td>5.37</td>
      <td>32.43</td>
      <td>51.92</td>
      <td>31.78</td>
      <td>57.29</td>
      <td>64.21</td>
    </tr>
  </tbody>
</table>


10 node types, 10 edge types:

<table>
  <thead>
    <tr>
      <th></th>
      <th colspan="2" halign="left">creation</th>
      <th colspan="2" halign="left">groupby</th>
      <th colspan="2" halign="left">total</th>
    </tr>
    <tr>
      <th></th>
      <th>columns</th>
      <th>index</th>
      <th>columns</th>
      <th>index</th>
      <th>columns</th>
      <th>index</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>10</th>
      <td>0.62</td>
      <td>1.40</td>
      <td>1.46</td>
      <td>1.27</td>
      <td>2.08</td>
      <td>2.67</td>
    </tr>
    <tr>
      <th>100</th>
      <td>0.58</td>
      <td>1.21</td>
      <td>1.57</td>
      <td>1.21</td>
      <td>2.15</td>
      <td>2.42</td>
    </tr>
    <tr>
      <th>1000</th>
      <td>0.53</td>
      <td>1.21</td>
      <td>1.59</td>
      <td>1.20</td>
      <td>2.11</td>
      <td>2.41</td>
    </tr>
    <tr>
      <th>10000</th>
      <td>0.63</td>
      <td>1.46</td>
      <td>2.06</td>
      <td>1.43</td>
      <td>2.68</td>
      <td>2.89</td>
    </tr>
    <tr>
      <th>100000</th>
      <td>3.13</td>
      <td>6.28</td>
      <td>4.82</td>
      <td>3.96</td>
      <td>7.95</td>
      <td>10.24</td>
    </tr>
    <tr>
      <th>1000000</th>
      <td>4.50</td>
      <td>30.51</td>
      <td>52.23</td>
      <td>27.22</td>
      <td>56.73</td>
      <td>57.72</td>
    </tr>
  </tbody>
</table>

<details><summary>Code</summary>

```python
import pandas as pd

import numpy as np
import timeit

def time(g, rep=100):
    return timeit.timeit(g, number=rep) / rep

def f(N, etypes, ntypes, rep=100):
    s = np.random.randint(ntypes, size=N)
    e = np.random.randint(etypes, size=N)
    t = np.random.randint(ntypes, size=N)
    w = np.random.rand(N)

    col = lambda: pd.DataFrame({"stype": s, "etype": e, "ttype": t, "weight": w})

    col_create = time(col)
    df = col()
    col_group = time(lambda: df.groupby(["stype", "etype", "ttype"])["weight"].mean())
    
    ind = lambda: pd.DataFrame(w, index=[s, e, t], columns=["weight"])
    ind_create = time(ind)
    df = ind()
    ind_group = time(lambda: pd.DataFrame(w, index=[s, e, t], columns=["weight"]))
    return (col_create, ind_create, col_group, ind_group, col_create + col_group, ind_create + ind_group)

ns = [10, 100, 1000, 10000, 100000, 1000000]
columns = pd.MultiIndex.from_product([["creation", "groupby", "total"], ["columns", "index"]])

times_1_1 = pd.DataFrame([f(n, 1, 1) for n in ns], index=ns, columns=columns)
times_1_10 = pd.DataFrame([f(n, 1, 10) for n in ns], index=ns, columns=columns)
times_10_10 = pd.DataFrame([f(n, 10, 10) for n in ns], index=ns, columns=columns)

for x in [times_1_1, times_1_10, times_10_10]:
    print((x * 1000).round(2).to_html().replace(' border="1" class="dataframe"', ""))
```

</details>

huonw, line 839: If this supports a `selector` argument, I think it needs to be applied here too:

```suggestion
            self._edges.weights[selector], index=[src_ty, rel_ty, tgt_ty], columns=["weight"]
```

or else the length of this will be different to the length of the index.

Alternatively, we could not support `selector` at all and just call `src_ty, rel_ty, tgt_ty = self._edge_type_triples()`.

stellargraph/core/graph.py
```diff
@@ -906,18 +926,30 @@ def str_node_type(count, nt):
             sep="\n  ",
         )
 
-        # FIXME: it would be better for the schema to just include the counts directly
-        unique_ets = self._unique_type_triples(return_counts=True)
```
huonw, line 910: I think we can remove the `return_counts` parameter from `_unique_type_triples` now, because it's now unused after this?

stellargraph/core/graph.py
```diff
@@ -832,6 +833,17 @@ def _unique_type_triples(self, *, return_counts, selector=slice(None)):
 
         return zip(*unique_ets)
 
+    def _edge_metrics_by_type_triple(self, *, selector=slice(None), metrics):
+        src_ty, rel_ty, tgt_ty = self._edge_type_triples(selector)
+        df = pd.DataFrame(
+            self._edges.weights, index=[src_ty, rel_ty, tgt_ty], columns=["weight"]
+        )
+
+        if len(df):
```
huonw, line 842: group-by doesn't work with an empty dataframe? It's unfortunate that it doesn't just give an empty dataframe in return?

stellargraph/core/graph.py
```diff
@@ -906,18 +930,28 @@ def str_node_type(count, nt):
...
+        et_metrics = self._edge_metrics_by_type_triple(metrics=metric_names)
+
         edge_types = sorted(
             (
-                (count, EdgeType(src_ty, rel_ty, tgt_ty))
-                for src_ty, rel_ty, tgt_ty, count in unique_ets
+                (
+                    metrics["count"],
+                    EdgeType(src_ty, rel_ty, tgt_ty),
+                    _WeightInfo(
```
huonw, line 941: This named tuple is defined a long way away, so it's not obvious what it stores. I think it'd be ok to either define it within this function so that the connection is stronger, or, since all of this processing is tightly scoped, not define it at all by just passing around the `metrics` value.

```python
(metrics["count"], EdgeType(src_ty, rel_ty, tgt_ty), metrics)
```

What do you think?

stellargraph/core/graph.py
kjun9: hmm I think it's the `agg` that fails due to not having numeric columns to aggregate, and actually it seems to work if I explicitly cast the `weight` column to be numeric
```python
df["weights"] = pd.to_numeric(df["weights"])
```
so I'll switch to that and just return a dataframe

stellargraph/core/graph.py
kjun9: Nice! I'm happy to go with columns, I think it looks clearer too avoiding the `level=[0,1,2]`

stellargraph/core/graph.py
kjun9: Yeah I think just passing around `metrics` is probably fine here, it's only being used in this function 👍 I'll update

stellargraph/core/graph.py
kjun9: I might just remove `selector`, rather than make `_edge_metrics_by_type_triple` more flexible than it needs to be when it's only being used in the `info` method for this specific purpose.

stellargraph/core/graph.py
huonw: Ah, hm. With the new form using `dict`, it seems like we may not need this work around as aggregation works directly:

```python
pd.DataFrame({"a": [], "b": []}).groupby("a").agg("mean")
# Empty DataFrame
# Columns: [b]
# Index: []
```

It seems like the exception you mention only occurs when using `columns=[...]` to set the names like the old form:

```python
pd.DataFrame([], columns=["a", "b"]).groupby("a").agg("mean")
# pandas.core.base.DataError: No numeric types to aggregate
```

stellargraph/core/graph.py
```diff
@@ -895,6 +900,18 @@ def str_node_type(count, nt):
...
             return f"{nt}: [{count}]\n    Features: {feature_text}\n    Edge types: {edge_types}"
 
+        def str_weight_info(metrics):
+            if metrics.min == metrics.max:
+                if metrics.min == 1:
+                    return f"Weights: all 1.0 (default)"
+                else:
+                    return f"Weights: all {metrics.min}"
+            else:
+                return f"Weights: range=[{metrics.min}, {metrics.max}], mean={metrics.mean}, std={metrics.std}"
```
huonw, line 910: The test show that some of these can end up with a lot of sig-figs even for simple data; is it worth truncating them similar to how pandas does in its DataFrame view. We could either use a hard coded value, like:

```suggestion
                return f"Weights: range=[{metrics.min:.6}, {metrics.max:.6}], mean={metrics.mean:.6}, std={metrics.std:.6}"
```

or have the value customisable (either as a parameter or piggy back of pandas):

```python
                prec = pd.get_option("display.precision")
                return f"Weights: range=[{metrics.min:.{prec}}, {metrics.max:.{prec}}], mean={metrics.mean:.{prec}}, std={metrics.std:.{prec}}"
```

I'm not sure if this will work with integer weights, and so we may need to either explicitly cast the values, or use the `g` format specifier (to force it to always use the default floating point formatter), like, for the hard-coded option:

```suggestion
                return f"Weights: range=[{metrics.min:.6g}, {metrics.max:.6g}], mean={metrics.mean:.6g}, std={metrics.std:.6g}"
```

kjun9:
> Looks good, although did you miss my comment about the notebooks above?

Ah oops I did read it but forgot, I'll update them 👍

stellargraph/core/graph.py
kjun9: Seems like `graph._edges.weights` get a dtype of `object` if the graph was created with no edges.. looking into where to do the casting

huonw: This might be the same problem as https://github.com/stellargraph/stellargraph/pull/1220#discussion_r409221636

kjun9: I've added that one line change here since #1220 won't be landing for the RC

kjun9: actually this is no longer needed after #1284 🎉

huonw:
> I noticed that current behaviour is to show the edge type triples A-R->B and B-R->A separately in the info string even when the graph is undirected - should we be collapsing them?

Probably! Could you file an issue?

Issue #1172, opened 2020-03-30
Check stellargraph version in demo notebooks


huonw:
# Resolving the error or warning from a notebook

## StellarGraph library version is *older* than the notebook

The error from the notebook will be similar to:
```
This notebook requires StellarGraph version 1.0.0, but an older version 1.0.0rc1 is installed.
```

### Resolution

You will need to either:

1. update the installed version of StellarGraph (**recommended**):
  ```
  pip install --upgrade stellargraph
  ```
2. use a version of the notebook that matches the installed version of StellarGraph (see `stellargraph.__version__` in Python, or `pip freeze | grep stellargraph` on the command line).

  An appropriate version can be found online at https://github.com/stellargraph/stellargraph/tree/v$YOUR_VERSION_HERE/demos (replace `$YOUR_VERSION_HERE` with the version number, like `1.0.0`), or in a git checkout with `git checkout v$YOUR_VERSION_HERE` (like `git checkout v1.0.0`).

## StellarGraph library version is *newer* than the notebook

The warning from the notebook will be similar to:
```
This notebook is designed for an older StellarGraph version 0.11.0 and may not function correctly with the newer installed version 1.0.0.
```

### Resolution

Make sure the version of the notebook matches the version of Stellargraph (see point 2 in the previous section).

If you're simply working through the notebook as a demonstration, it's best to use the most recent version of StellarGraph, by downloading the most recent release version of the notebook from the GitHub master branch https://github.com/stellargraph/stellargraph/tree/master/demos . At the same time, it's worth checking that you also have the latest StellarGraph installed:
```
pip install --upgrade stellargraph
```

<details><summary>Old Description</summary>

(This is the description of the issue before we addressed in #1242)

Demo Jupyter notebooks for StellarGraph are designed and tested with specific versions of StellarGraph.  As the StellarGraph API evolves, notebooks are updated and tested on CI to confirm that they work correctly with the latest version of StellarGraph.

This presents a problem for our users who may download notebooks (from 'develop' branch or 'release'), and then be running them with an older version of StellarGraph.

Notebooks can now check that they are running with the correct version of StellarGraph, and report an error if the version of the StellarGraph library is older, or a warning if the version of StellarGraph library is newer.

</details>

pritamarnab:
In google colab I am using the following code

# install StellarGraph if running on Google Colab
import sys
if 'google.colab' in sys.modules:
  %pip install -q stellargraph[demos]==1.2.1

[...]

PR #1242, opened 2020-04-15
Feature/1172 notebook version check

timpitman:
We've had many support issues caused by users running notebooks with the incorrect version of StellarGraph.  This PR inserts a cell near the top of each demo notebook which checks the version of StellarGraph against the version that the notebook was created with.

See: #1172


RELEASE_PROCEDURE.md
```diff
@@ -19,6 +19,7 @@
        - Version bumping: Change version from “X.X.Xb” to “X.X.X”. E.g. version=”0.2.0b” to version=”0.2.0”
          - `stellargraph/version.py`
          - `meta.yaml`
+       - Update expected versions of demo notebooks: `scripts/format_notebooks.py -v -o demos/`
```
huonw, line 22: Since this is written down and will be copy pasted rather than retyped (I imagine), I think it makes sense to spell out the flags for clarity, rather than keep them short for writer-convenience:

```suggestion
       - Update expected versions of demo notebooks: `scripts/format_notebooks.py --check_version --overwrite demos/`
```

scripts/format_notebooks.py
```diff
@@ -181,6 +181,58 @@ def preprocess(self, nb, resources):
         return nb, resources
 
 
+class VersionCheckPreprocessor(preprocessors.Preprocessor):
+    metadata_tag = "VersionCheck"  # tag for added cells so that we can find them easily
+
+    try:
+        # determine the current stellargraph version by importing the library's version file
+        sys.path.append(
+            (
```
huonw, line 190: I don't think this extra set of parens is needed? Also, if that's not enough, some part of this (maybe the `join`?) could be split out into a separate variable to make this expression a bit less complicated and explode less with `black`:

```python
library_dir = os.path.join(os.path.dirname(__file__), os.path.pardir, "stellargraph")
sys.path.append(os.path.abspath(library_dir))
```

scripts/format_notebooks.py
```diff
@@ -181,6 +181,58 @@ def preprocess(self, nb, resources):
         return nb, resources
 
 
+class VersionCheckPreprocessor(preprocessors.Preprocessor):
+    metadata_tag = "VersionCheck"  # tag for added cells so that we can find them easily
+
+    try:
+        # determine the current stellargraph version by importing the library's version file
+        sys.path.append(
```
huonw, line 189: Maybe instead of modifying `sys.path` this could just do what `setup.py` does, and eval the `version.py` file directly:

https://github.com/stellargraph/stellargraph/blob/e12c5bedb6d99f55da630ad9c59526955d608fd2/setup.py#L74-L77

Thoughts?

scripts/format_notebooks.py
```diff
@@ -181,6 +181,58 @@ def preprocess(self, nb, resources):
...
+                        os.path.dirname(__file__), os.path.pardir, "stellargraph"
+                    )
+                )
+            )
+        )
+        import version as sgversion
+
+        stellargraph_version = sgversion.__version__
+    except ImportError as error:
+        print(f"ERROR: unable to import StellarGraph version: {error.message}")
```
huonw, line 202: It's a bit unusual to me that an error would just be printed, and execution would continue; what's the reasoning?

scripts/format_notebooks.py
```diff
@@ -181,6 +181,58 @@ def preprocess(self, nb, resources):
...
+        stellargraph_version = None
+
+    version_check_code = f"""\
+# verify that we're using the correct version of StellarGraph for this notebook
+import stellargraph as sg
+from packaging import version
+
+if version.parse(sg.__version__) < version.parse("{stellargraph_version}"):
+    raise ValueError(
+        f"The installed version {{sg.__version__}} of StellarGraph is older than the version {stellargraph_version} required for this notebook. Please update StellarGraph, eg: pip install --upgrade stellargraph"
```
huonw, line 212: A few minor things style things:

- e.g. (and i.e.) can be confusing, so it's often recommended to just spell them out (for instance, https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/eg)
- delimiting the `pip` invocation makes it a bit clearer where it starts and ends.

```suggestion
        f"The installed version {{sg.__version__}} of StellarGraph is older than the version {stellargraph_version} required for this notebook. Please update StellarGraph, for instance: 'pip install --upgrade stellargraph'"
```

scripts/format_notebooks.py
```diff
@@ -181,6 +181,58 @@ def preprocess(self, nb, resources):
...
+            )
+        )
+        import version as sgversion
+
+        stellargraph_version = sgversion.__version__
+    except ImportError as error:
+        print(f"ERROR: unable to import StellarGraph version: {error.message}")
+        stellargraph_version = None
+
+    version_check_code = f"""\
```
huonw, line 205: This might be a case where it's clearer to use `%`, because of the nested `f` string, e.g.:

```python
version_check_code = """\
...
if version.parse(sg.__version__) < version.parse("%(stellargraph_version)s"):
    raise ValueError(
         f"The installed version {sg.__version__} of StellarGraph is older than the version %(stellargraph_version)s required for this notebook. Please update StellarGraph, eg: pip install --upgrade stellargraph"
    )""" % {"stellargraph_version": stellargraph_version}
```

Totally up to you whether you go for this, though.

scripts/format_notebooks.py
huonw, line 212: On the topic of trade-offs, I think this message could possibly provide a bit more guidance: several instances of this problem seem to be users looking at develop notebooks, but using a recent release, meaning `pip install --upgrade stellargraph` may not actually fix the problem. Potentially we could do something like:

```python
if ...:
    raise ValueError(
        "This notebook requires StellarGraph version {stellargraph_version}, but found older version ({{sg.__version__}}) installed. Please check you're using notebooks from a matching git tag, or consider updating StellarGraph (for instance, 'pip install --upgrade stellargraph')"
    )
```

However, jamming everything into a single exception message is kinda annoying; I tried to rephrase the message above to make it shorter and thus make space for the extra explanation, but there's still more we could say (e.g. explaining how to get the git tag on the command line vs. GitHub, and detecting whether the required version is a prerelease (meaning `pip install` won't work) or a realised one (meaning it will)). The final formatting is also a bit suboptimal.

One way to handle this would be to do something like linking to an external document, where we can write as much as we want (and with better formatting). For instance, we could edit https://github.com/stellargraph/stellargraph/issues/1172 or add comments to it.

```python
if version.parse(sg.__version__) != version.parse("{stellargraph_version}"):
    raise ValueError(
        "This notebook requires StellarGraph version {stellargraph_version}, but found different version ({{sg.__version__}}) installed. Please see <https://github.com/stellargraph/stellargraph/issues/1172>."
    )
```

(I changed the `<` to ` !=` because the explanation on #1172 could talk about that case too.)

Another way would be to have a function like `sg.utils.check_version` that we call In each notebook like `sg.utils.check_version("{stellargraph_version}")`. Of course, older versions of `StellarGraph` won't have this function, so a user will get `AttributeError` instead of the useful message, but maybe we can manage that?

```python
import stellargraph as sg
try:
    sg.utils.check_version("{stellargraph_version}")
except AttributeError:
    raise ValueError("This notebook requires StellarGraph version {stellargraph_version}, but an older version is installed.  Please see <https://github.com/stellargraph/stellargraph/issues/1172>.")
```

scripts/format_notebooks.py
```diff
@@ -181,6 +181,58 @@ def preprocess(self, nb, resources):
...
+            cell
+            for cell in nb.cells
+            if self.metadata_tag not in cell["metadata"].get("tags", [])
+        ]
+        # find first (non-CloudRunner) code cell and insert before it
+        first_code_cell_id = next(
+            index
+            for index, cell in enumerate(nb.cells)
+            if cell.cell_type == "code"
+            and CloudRunnerPreprocessor.metadata_tag
```
huonw, line 227: We could factor out this tag checking logic, potentially?

E.g.

```python
class CloudRunnerPreprocessor(...):
    @staticmethod
    def has_tag(cell):
        return CloudRunnerPreprocessor.metadata_tag not in cell["metadata"].get("tags, [])
```

Or even just a helper function at the top level that can be used everywhere

```python
def tags(cell):
    return cell["metadata"].get("tags", [])

...

# in both CloudRunnerPreprocessor and VersionCheckPreprocessor
nb.cells = [
    cell for cell in nb.cells if self.metadata_tag not in tags(cell)
]
```

setup.py
```diff
@@ -45,7 +45,14 @@
...
 # Other demos do not have specific requirements
 EXTRAS_REQUIRES = {
-    "demos": ["numba", "jupyter", "seaborn", "rdflib", "mplleaflet==0.0.5"],
+    "demos": [
+        "numba",
+        "jupyter",
+        "seaborn",
+        "rdflib",
+        "mplleaflet==0.0.5",
+        "packaging",
```
huonw, line 54: Does this work with conda? Installing via conda doesn't have an easy way to get these requirements (compared to `stellargraph[demos]` for pip), so does this package exist automatically there?

scripts/format_notebooks.py
```diff
@@ -247,14 +299,20 @@ def preprocess(self, nb, resources):
...
     )
     parser.add_argument(
         "-r",
         "--run_cloud",
         action="store_true",
         help="Add or update cells that support running this notebook via cloud services",
     )
+    parser.add_argument(
+        "-v",
+        "--check_version",
```
huonw, line 312: Oh, one more thing; this name here is a bit peculiar, because it sounds more like a validation than something active (that is, it is a bit confusing with `--check`). Maybe it could be flipped to:

```suggestion
        "--version_checks",
```

or 

```suggestion
        "--version_validation",
```

scripts/format_notebooks.py
timpitman: Yes, I think this is more reliable if another stellargraph library is already on the python path 👍

scripts/format_notebooks.py
timpitman: changed now to the above code which is a little different; agree that it's better to just have the exception if the file isn't found

scripts/format_notebooks.py
timpitman: This is a good idea, and I think even more can be shared via a common class

setup.py
timpitman: interestingly, conda includes packaging by default but it can be forced out of an environment, so I've added it to our requirements (now for the library itself in pip and conda)

scripts/format_notebooks.py
timpitman: I think both are confusing! 🤷‍♂️

stellargraph/utils/version_validation.py
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+
+import warnings
+from packaging import version
+from ..version import __version__
+
+__all__ = ["validate_notebook_version"]
+
+
+def validate_notebook_version(notebook_version):
```
huonw, line 24: Is it worth having a test for this? Something like

```python
def test_current_version_older():
    with pytest.raises(ValueError, match=".* requires StellarGraph version 987.654.321, but an older version .*/1172"):
        validate_notebook_version("987.654.321")

def test_current_version_newer():
    with pytest.warns(DeprecationWarning, match=".* is designed for an older StellarGraph version 0.0.0 and may not function correctly with the newer .*/1172"):
        validate_notebook_version("0.0.0")
```

An optional extension would be:

```python
def test_current_version_vs_release():
    # compare a current pre-release version A.B.Cb0 (or similar) against A.B.C
    match = re.match(r"^(\d+\.\d+\.\d+)[^\d].*$")
    if not match:
        pytest.skip(f"can only test release versions for prereleases, not {stellargraph.__version__}")

    release_version = match[0]
    with pytest.raises(ValueError, match=".* requires StellarGraph version .* but an older version"):
        validate_notebook_version(relase_version)
```

If we were feeling very enthusiastic, we'd make `__version__` pluggable to be able to test `1.2.3b0` vs. `1.2.3` easily, but I think it's probably ok with just the above

demos/basics/loading-networkx.ipynb
```diff
@@ -60,6 +60,27 @@
   {
```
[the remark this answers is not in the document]
timpitman: yes, this was a mistake due to testing with that one!  have restored.

stellargraph/utils/version_validation.py
timpitman: Yes, worth having tests - I forgot!  I've made a simpler dev/release test using patch.

stellargraph/utils/version_validation.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Validate a notebook created for a specific version of StellarGraph.
+
+    Args:
+        notebook_version(str): the library version that the notebook was created for
+    """
+    version_stellargraph = version.parse(__version__)
```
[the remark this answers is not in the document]
timpitman: I'm not a big fan of re-implementing basic packages, but this avoids a dependency so worthwhile while it's relatively simple.

stellargraph/utils/version_validation.py
```diff
@@ -0,0 +1,81 @@
...
+    Args:
+        notebook_version(str): the library version that the notebook was created for
+    """
+    version_stellargraph = _parse(__version__)
+    version_notebook = _parse(notebook_version)
+    extra_info = (
+        "Please see <https://github.com/stellargraph/stellargraph/issues/1172>."
+    )
+    if version_stellargraph < version_notebook:
+        raise ValueError(
```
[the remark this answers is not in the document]
timpitman: This is a good idea, it makes a much better experience for the user.  I've gone a little further with the formatting so that the text and link is more obvious.

stellargraph/utils/version_validation.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    version_stellargraph = _parse(__version__)
+    version_notebook = _parse(notebook_version)
+    help_url = "https://github.com/stellargraph/stellargraph/issues/1172"
+
+    def _format_for_error(m):
+        return f"{m} Please see: <{help_url}>."
+
+    def _format_for_html(m):
+        return f'<div class="alert alert-block alert-danger">{m} Please see: <a href="{help_url}">{help_url}</a>.</div>'
```
huonw, line 84: This looks great 👍

Issue #1176, opened 2020-03-31
Make Deep Graph Infomax more extensible by not including per-model code


huonw:
### Description

Per https://github.com/stellargraph/stellargraph/pull/1162#pullrequestreview-381642755 and https://github.com/stellargraph/stellargraph/pull/1162#discussion_r398981400, our `DeepGraphInfomax` code currently requires custom `if isinstance(base_model, ...)` code for each sort of base model that can be trained with DGI, and similarly for `CorruptedGenerator`. If we "unwrap" this by moving it to the model and sequence classes, it will be easier to extend support for DGI (e.g. #1173), including for users defining their own models outside `StellarGraph`.

Issue #1173, opened 2020-03-30
Unsupervised HinSAGE via Deep Graph Infomax


huonw:
### Description

In #1162, we extend our `DeepGraphInfomax` type to support unsupervised training of GraphSAGE via mutual information. We should do the same for HinSAGE.

PR #1243, opened 2020-04-15
Generalise Deep Graph Infomax corruption to work with any generator

huonw:
This generalises `DeepGraphInfomax` (in particular, `CorruptedGenerator`) to theoretically work with any model. It does this by providing a `corrupt_index_groups` argument, which can be used to control which inputs get shuffled. It's a list of lists, in the form of a sequence of "groups", where each group is a sequence of input tensor indices that get concatenated and shuffled together.

This is a strict generalisation of full-batch shuffling (just shuffle the first tensor: `[[0]]`) and GraphSAGE shuffling (shuffle all tensors together `[[0, 1, ..., N - 1]]` where `N` is the number of input tensors). HinSAGE fits into the same framework, by having a group for each type of node (#1173).

For convenience and compatibility, the `corrupt_index_groups` argument can be set by default from the base generator. The `Generator` class now has an optional `default_corrupt_input_index_groups` method, which returns `None` by default, but can be overridden to return a list of appropriate groups for shuffling. This PR preserves the current `DeepGraphInfomax` functionality by overriding this for `FullBatchNodeGenerator`, `GraphSAGENodeGenerator` and `DirectedGraphSAGENodeGenerator`.

The `DeepGraphInfomax` class now has to take in the relevant `CorruptedGenerator` in order to work out what input tensors to duplicate. The existing `DeepGraphInfomax(model)` form is still supported for the current set of classes, but is deprecated.

This doesn't do the fully general form of having a "corrupter" class or function as in https://github.com/stellargraph/stellargraph/pull/1193#issuecomment-607628370. This is to avoid creating an abstraction that isn't an appropriate abstraction, because all of the current algorithms' corruptions are this style of shuffling, and so we don't have a reference for any other options. If/when we find other styles of corruption, we can generalise this fairly easily without breaking backwards compatibility (e.g. add a new argument or check the type of the existing one).

This also removes the checking and passing of `multiplicity`: it's now up to the users to pass in an appropriate generator/model for the task they're doing (_theoretically_ someone could do unsupervised link embedding training or something, with a link model).

See: #1176


stellargraph/mapper/corrupted.py
```diff
@@ -96,31 +133,31 @@ def __getitem__(self, index):
 
         inputs, _ = self.base_generator[index]
 
-        if isinstance(
-            self.base_generator, (FullBatchSequence, SparseFullBatchSequence)
-        ):
+        shuffled_feats = []
+        for group in self.corrupt_index_groups:
+            feats_orig = [inputs[idx] for idx in group]
```
kieranricardo, line 138: I don't there's a check on `idx` anywhere, this could trigger a confusing warning for the user. 

We could either wrap this in an exception here:

```python
try:
    feats_orig = [inputs[idx] for idx in group]
except KeyError as e:
    raise KeyError("something about invalid index in corrupt_index_groups")
```

Or maybe do a check in `__init__`:

```python
peeked_inputs, _ = self.base_generator[0]
try:
    feats_orig = [[peeked_inputs[idx] for idx in group] for group in corrupt_index_groups]
except KeyError as e:
    raise KeyError("something about invalid index in corrupt_index_groups")
```

kieranricardo, line 138: completely optional, but the code in this inner loop might be well suited to being in a private function `_merge_shuffle_split(feats)`

huonw: That's a good idea. I think it has to be a bit different because it's a "flat map", with multiple outputs per input group, not just a "map". Specifically, a pair of nested loops:

```python
        shuffled_feats = [
            corrupted
            for group in self.corrupt_index_groups
            for corrupted in corrupt_group(group)
        ]
```

Other than that, it does seem nicer.

huonw: I think the first one is better; the second one seems somewhat difficult to get "perfect", due to things like randomness (or other) problems with batch_num = 0 being requested twice, and handling of generators of length 0 (weird/useless, but still...).

stellargraph/mapper/corrupted.py
```diff
@@ -61,33 +104,28 @@ def flow(self, *args, **kwargs):
...
 
     Args:
         base_generator: the uncorrupted Sequence object.
+        corrupt_index_groups: the groups among which nodes will be shuffled (see :class:`CorruptGenerator` for more details)
+        num_batch_dims: the number of axes that are "batch" dimensions
     """
 
-    def __init__(self, base_generator):
-
+    def __init__(self, base_generator, corrupt_index_groups, num_batch_dims):
```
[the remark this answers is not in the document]
huonw: That's sensible 👍

Issue #1238, opened 2020-04-14
Return multiple arrays instead of zipping arrays into a list in `StellarGraph`


kieranricardo:
Some `StellarGraph` methods zip multiple numpy arrays and return them as one list, these could optimized to return seperate array.

These method are:

- `edges`
- `_transform_edges`
- `_adjacency_types` (maybe?)
- `_unique_type_triples` (maybe?)

huonw:
Relates to #718.

PR #1253, opened 2020-04-16
Ehancement/1238 no zip np array

kieranricardo:
This PR changes `StellarGraph._transform_edges` to return a tuple of `np.array`s instead of a list of tuples:

See #1238


kieranricardo:
yeah that's a fair point about the tupling of `sources` and `targets` in `.edges()`. How about combining these two in a 2D numpy array? I think most methods end up doing this anyway,.

Part of the reason why I went for seperate numpy arrays here is to make the transition to a tensorflow backend smoother. After #1220 im also hesistant to add even more code paths into `StellarGraph`!

ahh, yeah it is a breaking change. any thoughts on how to handle a change like this? hopefully it won't affect too many people; if people are only interacting with `StellarGraph` through our generators their code won't break. Also, I think its okay for a 1.0 release to have some breaking changes?

kjun9:
> How about combining these two in a 2D numpy array?

I think having named columns would be ideal.. otherwise I'm not sure how obvious it is to infer that the columns are "source" and "target" (I think it's obvious enough for our internal use, but if I'm talking more about some random user just creating a stellargraph and inspecting its edges). Maybe a named tuple if we really don't want to return a dataframe? 
```python
Edgelist = namedtuple("Edgelist", ["source", "target"])
```

> Part of the reason why I went for seperate numpy arrays here is to make the transition to a tensorflow backend smoother. After #1220 im also hesistant to add even more code paths into StellarGraph!

Hmm I see! Yeah seems like this PR is also bringing up similar discussion points around the "raw" stellargraph API vs user-facing stellargraph API, so we need to do some kind of middle-ground for now. I think `graph.edges` although used in stellargraph for internal ML things, it's also quite a basic fundamental method that a "graph" should expose, so I think we would expect people to use it outside of just letting our generators use it?

> ahh, yeah it is a breaking change. any thoughts on how to handle a change like this? hopefully it won't affect too many people; if people are only interacting with StellarGraph through our generators their code won't break. Also, I think its okay for a 1.0 release to have some breaking changes?

Yeah I guess we can just include some form of documentation either in the docstring or the release notes a small snippet to allow someone to transform the "new" edges to the shape of the "old" edges?

kieranricardo:
@kjun9 I'm going to limit this PR to just `StellarGraph._transform_edges` to get this in for the release. I think this is where most of the performance benefits are anyway because `graph.neighbors` is called multiple times per batch for many generators/random walks while `graph.edges` is usually only called once prior to making a generator.

kieranricardo:
I've made this new PR for _transform_edges https://github.com/stellargraph/stellargraph/pull/1286

huonw:
A risk with this sort of unconditional change (both here and #1286) is that immediately/automatically break any users of these methods, e.g. https://github.com/Fahimeh1983/celltypes/blob/master/BiasedDirectedWeightedWalk.py, and for these changes in particular, the breakage may be non-obvious. Specifically, basic operations like `for x in graph.edges()`, `len(graph.edges())` and `for x in graph.neighbors(...)` will still run, but they'll give wildly wrong/unexpected results and lead to either incorrect answers or downstream/non-obvious exceptions.

Is it worth trying to do some sort of migration assistance here?

huonw:
(Oh, sorry, I didn't read the discussion above that touched on that point. That's a bit embarrassing!)

stellargraph/mapper/full_batch_generators.py
```diff
@@ -175,6 +175,9 @@ def flow(self, node_ids, targets=None):
             and :meth:`predict`
 
         """
+        if (self.multiplicity == 2) and isinstance(node_ids, tuple):
+            node_ids = np.stack(node_ids[:2], axis=1)
```
kjun9, line 179: Hmm I think we could move this to `FullBatchLinkGenerator.flow` instead? Then we can improve the docstrings together:
```python
class FullBatchLinkGenerator(FullBatchGenerator):
    ...
    def flow(link_ids, targets=None):
        """
            link_ids: an EdgeList or an iterable of link ids specified as tuples of node ids
        """
        if isinstance(link_ids, tuple):
            link_ids = np.stack(link_ids[:2], axis=1)
        ...

```

There's also a line in the docstring for `FullBatchLinkGenerator` that should be updated:
```
Use the :meth:`flow` method supplying the links as a list of (src, dst) tuples
```

stellargraph/core/graph.py
```diff
@@ -37,6 +43,7 @@
 
 
 NeighbourWithWeight = namedtuple("NeighbourWithWeight", ["node", "weight"])
+EdgeList = namedtuple("EdgeList", ["source", "target", "type", "weight"])
```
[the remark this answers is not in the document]
kieranricardo: I like this!

kieranricardo:
Good point about breaking the API 😬I went for seperate function approach you suggested

I also added backwards compatibility into `FullBatchLinkGenerator` and `SampledLinkGenerator` to hande edges in the old format in case some users are preparing edges in a custom way. 

I also found our HinSAGE demo was preparing edges manually in the old format, so I updated this to provide edges in the 2D array format.

tests/mapper/test_full_batch_generators.py
```diff
@@ -356,32 +356,42 @@ def generator_flow(
         return A_dense, tind, y
 
     def test_generator_flow_notargets(self):
-        link_ids = list(self.G.edges())[:3]
+        sources, targets, _, _ = self.G.edge_arrays()
```
kjun9, line 359: If we're going with the approach of leaving `graph.edges` the same as it originally was and just adding a new `graph.edge_arrays` method, I think it might be better to just leave a lot of these tests as they originally were. In particular, all the tests that are testing `flow` methods it's probably more likely they'll continue to be called with `edges` in the original form?

stellargraph/core/graph.py
```diff
@@ -575,31 +589,33 @@ def _transform_edges(
...
-        return list(other_node_id)
-
-    def neighbors(
+    def neighbor_arrays(
         self, node: Any, include_edge_weight=False, edge_types=None
-    ) -> Iterable[Any]:
+    ) -> Iterable[np.ndarray]:
         """
         Obtains the collection of neighbouring nodes connected
         to the given node.
```
kjun9, line 609: I think the docstrings for these new methods should be slightly different from the existing methods (similar to how you've done `edge_arrays`), otherwise I'm not sure if it's obvious from just the names of the methods how they're different? Maybe something like
```
Obtains the collection of neighbouring nodes connected to the given node as a tuple of arrays (sources, targets, types, weights). ``types`` and ``weights`` will be None unless specified via the optional parameters.

...

Returns:
    (sources, targets, types, weights)
```
And similarly for `in_node_arrays`, `out_node_arrays`

stellargraph/core/graph.py
```diff
@@ -525,25 +535,29 @@ def edges(
...
+        """
+        Obtains the collection of edges in the graph.
+
+        Args:
+            include_edge_type (bool): A flag that indicates whether to return edge types.
+            include_edge_weight (bool): A flag that indicates whether to return edge weights.
+
+        Returns:
+            A tuple containing 1D arrays of the source and target nodes (sources, targets). Setting include_edge_type
+                and/or include_edge_weight to True will include arrays of edge types and/or edge weights in this tuple.
```
kjun9, line 556: I think this would be a bit more accurate, since we're always returning a 4-tuple in a particular order, so it'd be nice to spell out the order things are in:
```suggestion
        Obtains the collection of edges in the graph as a tuple of arrays (sources, targets, types, weights). ``types`` and ``weights`` will be `None` if the optional parameters are not specified.

        Args:
            include_edge_type (bool): A flag that indicates whether to return edge types.
            include_edge_weight (bool): A flag that indicates whether to return edge weights.

        Returns:
            A tuple containing 1D arrays of the source and target nodes (sources, targets, types, weights). Setting include_edge_type
                and/or include_edge_weight to True will include arrays of edge types and/or edge weights in this tuple, otherwise they will be set to ``None``.
```

stellargraph/core/graph.py
```diff
@@ -525,25 +535,29 @@ def edges(
...
-            )
-        else:
-            edges = list(zip(self._edges.sources, self._edges.targets))
-
+        edge_arrs = self.edge_arrays(include_edge_type, include_edge_weight)
+        edges = list(zip(*(arr for arr in edge_arrs[:3] if arr is not None)))
         if include_edge_weight:
-            return edges, self._edges.weights
+            return edges, edge_arrs[3]
+        return self._to_edges(edge_arrs)
```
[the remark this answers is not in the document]
kieranricardo: woops, i didn;t mean to leave that there!

kieranricardo:
> Were you going to revert the changes in the tests though?

I was thinking of leaving those ones as is. I changed the generator tests back to use `.edges()` because that's how they'll most likely be used. But with `test_graph.py` I think its okay to use  `edge_arrays` instead of `edges` in most tests and then just test `_to_edges()` seperately - what do you think?

kjun9:
@kieranricardo should I review this again?

kieranricardo:
@kjun9 I reckon its ready for re-review now - thanks! :)

PR #1254, opened 2020-04-16
Allow easy unsupervised HinSAGE training with Deep Graph Infomax

huonw:
This uses the generalised Deep Graph Infomax corruption support from #1243 to have HinSAGE specify a default way to corrupt. This means that HinSAGE can be easily trained unsupervised via Deep Graph Infomax, by just passing in the generator and model.

This is currently demonstrated running on Cora, and gives similar results to GraphSAGE as expected. Creating a real demo is future work: #1285.

Other possibilities:
- some sort of unsupervised link prediction task (e.g. using MovieLens), but this may be a bit awkward, and doesn't necessarily generalise that well to other heterogeneous algorithms that can be trained with DGI, like RGCN.
- a new dataset, like the ACM one from https://github.com/Jhy1993/HAN: 3 node types (3025 papers, 5835 authors, 56 subjects, with feature vectors of length 1830 on papers (I think)), and 2(?) edge types (paper-author and paper-subject)

I'm thinking we can leave this for future work, because either of these is a significantly harder and broader task than the change required here (e.g. working with a new dataset is always a little annoying).

See: #1173


stellargraph/mapper/sampled_node_generators.py
```diff
@@ -495,6 +496,14 @@ def sample_features(self, head_nodes, batch_num):
 
         return batch_feats
 
+    def default_corrupt_input_index_groups(self):
+        # every sample of a given node type can be grouped together
+        indices_per_nt = defaultdict(list)
+        for tensor_idx, (nt, _) in enumerate(self._sampling_schema[0]):
+            indices_per_nt[nt].append(tensor_idx)
+
+        return sorted(sorted(idx) for idx in indices_per_nt.values())
```
kieranricardo, line 505: I'm not sure but I think this list of lists is already sorted.

 ```suggestion
        return list(indices_per_nt.values())
```

It also seems like the order of both the groups and the elements within the groups are irrelevant anyway.

File: stellargraph/mapper/sampled_node_generators.py [hunk 2]
```diff
@@ -495,6 +496,14 @@ def sample_features(self, head_nodes, batch_num):
 
         return batch_feats
 
+    def default_corrupt_input_index_groups(self):
+        # every sample of a given node type can be grouped together
+        indices_per_nt = defaultdict(list)
+        for tensor_idx, (nt, _) in enumerate(self._sampling_schema[0]):
```
kieranricardo, line 502: not related to this PR (and so doesn't need to be fixed), but it seems unecessary that `self._sampling_schema[0]` is a list of lists. From what I can tell the outer list always has one element and we only access the first element of the outer list anyway.

stellargraph/layer/hinsage.py
```diff
@@ -581,6 +581,13 @@ def in_out_tensors(self):
             model output tensor(s) of shape (batch_size, layer_sizes[-1]).
 
         """
+
+        if multiplicity is not None and multiplicity != self.multiplicity:
```
kieranricardo, line 585: just curious, why can't we set a different `multiplicity` to `self.multiplicity` here.

huonw: Ah, this is unnecessary now. My original thought here was that this was the simplest way to have DGI support HinSAGE, given it passes in `multiplicity=1`, and it could be generalised in future. In particular, there's various support structures (like sampling-schema) that depend on the multiplicity and so would need to be recomputed if it was customised here.

However, #1243 was extended since this was written to have `DeepGraphInfomax` no longer pass a `multiplicity` argument, so I'll remove this. Good catch.

stellargraph/mapper/sampled_node_generators.py
huonw: I think it has two elements for link prediction, where each element represents the input tensor layout for each node of the edge.

stellargraph/mapper/sampled_node_generators.py
huonw: I agree it's irrelevant semantically, but it does change the final result (if the order of these elements changes, the order of the concatenation will change in `CorruptedSequence`, and so the final shuffled result will be different, even if seeds are set). Hence I wanted to have a "canonical" version, so there's no randomness or anything that might make things non-reproducible. 

I also think it's already sorted, with our current implementation of `_sampling_schema`/`GraphSchema.sampling_layout` but I thought this was a low effort and low cost way to ensure that it always stays sorted. I guess another option would be to just have it in a test, but that's somewhat less reliable in my mind.

huonw:
> If we do want to have a test for DGI × every algorithm, we end up with a whole bunch of duplication and test time. However, it probably is better as an "integration" test, so I'll add one.

I refactored it so we have basically one test for DGI + node classification, using a slightly more intricate parameterisation scheme.

stellargraph/mapper/sampled_node_generators.py
huonw: I added comment too, to try to explain/remind for future reference.

huonw:
I opened https://github.com/stellargraph/stellargraph/issues/1285 to track making a demo.

Issue #1239, opened 2020-04-14
AttributeError: 'StellarGraph' object has no attribute 'nvtxs'


davidxujiayang:
I'm trying to use metis to partition a graph following a similar approach as in the cluster GCN example in node classification. The relavant lines are

```
G, A, coords, data = build_graph_from_vtm(
    'data/airfoil/test_7.vtm', 'data/airfoil/polyMesh', ['p', 'U'], plot=False)
print(G.info())
number_of_clusters = 50
edgecuts, parts = metis.part_graph(G, number_of_clusters)
```
And the error outputs are (based on which I think I've generated the graph correctly):

> StellarGraph: Undirected multigraph
>  Nodes: 11092, Edges: 21254
> 
>  Node types:
>   default: [11092]
>     Features: float32 vector, length 2
>     Edge types: default-default->default
> 
>  Edge types:
>     default-default->default: [21254]
> Traceback (most recent call last):
>   File "stella_cfd.py", line 40, in <module>
>     edgecuts, parts = metis.part_graph(G, number_of_clusters)
>   File "/home/davidxu/anaconda3/envs/cfd/lib/python3.7/site-packages/metis.py", line 791, in part_graph
>     partition = (idx_t*graph.nvtxs.value)()
>   File "/home/davidxu/anaconda3/envs/cfd/lib/python3.7/site-packages/stellargraph/core/graph.py", line 436, in __getattr__
>     return super().__getattribute__(item)
> AttributeError: 'StellarGraph' object has no attribute 'nvtxs'

PantelisElinas:
Hi davidxujiayang,

thanks for using StellarGraph. 

I think we have an error in the notebook because when using Metis, `G` should be a `Networkx` object not `StellarGraph`. It is possible to convert a `StellarGraph` object to `Networkx` using the method `StellarGraph.to_networkx()` (see the docs at https://stellargraph.readthedocs.io/en/stable/api.html#stellargraph.StellarGraph.to_networkx). So, first convert your object to `Networkx` and then call `metis.part_graph(...)` to get the `clusters`. Remember that after you have the clusters, you need your `StellarGraph` object for machine learning. 

We will update the demo over the next few days to fix this error. In the meantime, let us know if you can't get the above to work.

Regards,

P.

davidxujiayang:
Hi Pantelis,

Thanks for your reply and suggestion. It works!

Best,
-David

Xu, Jiayang (David)
PhD Candidate
[...]

PR #1260, opened 2020-04-16
Fix bug using METIS for in ClusterGCN demo

PantelisElinas:
Updating the ClusterGCN demo notebook to work with METIS clustering algorithm now that the data are loaded directly into `StellarGraph` objects instead of `NetworkX`.

The Python `metis` (https://pypi.org/project/metis/) module used in this demo is not up to date with the latest version of `NetworkX` and looks like it is no longer updated as the bitbucket repo is gone. The solution in this pull request passes the graph as an adjacency list instead of `NetworkX` object. 

I think in the future we should either find a replacement for `metis` or "fork" it and keep it up to date ourselves. 

This update solves #1239


demos/node-classification/cluster-gcn/cluster-gcn-node-classification.ipynb
```diff
@@ -198,7 +198,7 @@
        " 'Theory'}"
```
huonw, line 198: <p>This could be made significantly faster by using the sparse adjacency matrix, converting to https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.lil_matrix.html#scipy.sparse.lil_matrix and reading the <code>rows</code> property:</p><p> </p><pre class="ql-syntax" spellcheck="false">lil_adj = G.to_adjacency_matrix().tolil()
adjlist = list(lil_adj.rows)
</pre><p>or</p><pre class="ql-syntax" spellcheck="false">lil_adj = G.to_adjacency_matrix().tolil()
adjlist = [tuple(neighbours) for neighbours in lil_adj.rows]
</pre><p> </p><p>This seems to give the exact same result, except the LIL form results in sorted lists, unlike the NetworkX form:</p><p> </p><pre class="ql-syntax" spellcheck="false">import stellargraph as sg
import networkx as nx

g_nx = nx.convert_node_labels_to_integers(G.to_networkx())
via_nx = [sorted(neighs) for neighs in nx.to_dict_of_lists(g_nx).values()]

via_lil = list(G.to_adjacency_matrix().tolil().rows)

via_nx == via_lil # True
</pre><p>For my claim of speed:</p><pre class="ql-syntax" spellcheck="false">%timeit -n 10 list(G.to_adjacency_matrix().tolil().rows)
# 5.16 ms ± 505 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit -n 10 list(nx.to_dict_of_lists(nx.convert_node_labels_to_integers(G.to_networkx())).values())
# 88.6 ms ± 1.59 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
</pre><p>(Using a <code>tuple(...)</code>  list comprehension instead of <code>list(...)</code>  doesn't make much difference)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1260/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

PantelisElinas: Ok, I couldn't figure out how to get the adjacency list representation directly from the `StellarGraph` object so I went via networkx because this is how `metis` handles the input as well. I updated to avoid the conversion to `NetworkX`. 

Maybe adding a method `StellarGraph.to_adjacency_list()` to wrap `G.to_adjacency_matrix().tolil()` would be helpful?

PantelisElinas:
> This looks good other than the (somewhat) optional improvement to avoid using NetworkX unnecessarily.
> 
> Also, it's unfortunate that this code path isn't tested. Another approach beyond the options you list might be switching to a different clustering algorithm, e.g. one in NetworkX (unfortunate that we'd have to convert) or something simple like label propagation build on top of `scipy.sparse` or `scipy.sparse.csgraph` (maybe there's library(s) for this already)? I think this would also resolve the issue of METIS requiring more than just a `pip install`?

When I first prepared this demo, I attempted to use the clustering algorithms available in `NetworkX` and those either run too slow or fast but computed clusters that were not as good as METIS (in the sense that performance on the classification task was not as good.) Also, the ClusterGCN paper uses METIS which explains why it is in the demo in the first place. I looked for other Python libraries that wrap the METIS implementation but one would not install on my laptop and the other had no documentation and was tailored to the user's narrow application on clustering meshes. I think if the `metis` module is taken off pypi, then we might just have to replace with a suboptimal `NetworkX` algorithm or entirely remove this option from the demo.

huonw:
Thanks for the background. I wrote up an issue at https://github.com/stellargraph/stellargraph/issues/1263 so we don't loose track of this work and the problem here.

PR #1265, opened 2020-04-17
Feature/1207 adj normalisation

PantelisElinas:
This updates `PaddedGraphGenerator` and `PaddedGraphSequence` with the option of using symmetric (GCN-like) or asymmetric (left multiply by inverse of degree matrix) normalisation of the degree matrix. Some graph classification algorithms such as DGCNN require the latter type normalization.

The default behaviour for the generator and sequence objects remain the same.

See #1207.

P.


tests/mapper/test_padded_graph_generator.py
```diff
@@ -163,3 +163,31 @@ def test_generator_flow_with_targets():
         assert batch[0][1].shape[0] == 1
         assert batch[0][2].shape[0] == 1
         assert batch[1].shape[0] == 1
+
+
+def test_generator_adj_normalisation():
```
kjun9, line 168: I think we could avoid writing pretty much the same test twice by using `pytest.parametrize` here
```python
@pytest.mark.parametrize("symmetric_normalization", [True, False])
def test_generator_adj_normalisation(symmetric_normalization):
    graph = example_graph(feature_size=4)

    generator = PaddedGraphGenerator(graphs=[graph])

    seq = generator.flow(graph_ilocs=[0], symmetric_normalization=symmetric_normalization)

    adj_norm_seq = seq.normalized_adjs[0].todense()

    adj = np.array(graph.to_adjacency_matrix().todense())
    inv_deg = np.diag(1.0 / adj.sum(axis=1))
    adj_norm = inv_deg.dot(adj)
    if symmetric_normalization:
        adj_norm = adj_norm.dot(inv_deg)

    assert np.allclose(adj_norm_seq, adj_norm)
```

PantelisElinas: I have updated the test as suggested.

PR #1274, opened 2020-04-19
Optimise RelationalFullBatchNodeGenerator construction (18× speedup, 560× smaller)

huonw:
This does four things to make preparing data for RGCN (in the form of creating a `RelationalFullBatchNodeGenerator`) faster, mostly by removing code due to the new StellarGraph having stronger requirements.

I benchmarked time in the same manner as #1272, on the AIFB dataset, running:

```python
import stellargraph as sg

G, affiliation = sg.datasets.AIFB().load()
%timeit -n 3 -r 5 sg.mapper.RelationalFullBatchNodeGenerator(G, sparse=True)
```

I benchmarked memory usage using pympler, subtracting out the size of the graph to show the incremental memory use over the raw `StellarGraph` data:

```python
from pympler.asizeof import asizeof
asizeof(sg.mapper.RelationalFullBatchNodeGenerator(G)) - asizeof(G)
```

| code | time (s) | total speedup (incremental) | memory (MiB) | memory reduction |
|---|---|---|---|---|
| original (`develop`) | 1.64 |  1.0× (1.0×) | 270 | 0% |
| fetch node features once | 1.25 | 1.3× (1.3×) | 270 | 0% |
| edge-type filtering in `to_adjacency_matrix`  | 0.35 | 4.7× (3.6×) | 270 | 0% |
| use `_get_index_for_nodes`, not manual `dict` | 0.26 | 6.3× (1.3×) | 269 | -0.4% |
| don't manually index `node_features`  | 0.09 |  18× (2.9×) | 0.483 | -99.8% |

Using `to_adjacency_matrix` means that the directedness of the graph now matters, so some tests have to be updated to use use `StellarDiGraph` instead of an undirected `StellarGraph`.

Two of these points require updates to `StellarGraph`:

- edge-type filtering in `to_adjacency_matrix`: expands that method to support only including edges of a single type
- don't manually index `node_features`: expands that method to support getting all features for a single type, which doesn't have to do any copying (this doesn't change behaviour, because the node features were in the right order to start with: doing `graph.node_features(graph.nodes())` ends up indexing the underlying array with `[0, 1, 2, ... , num_nodes - 1]`, which requires allocating new memory and copying, without actually changing the order or shape of the data)

The massive memory reduction for the last point is because the feature matrix is now shared with the underlying `StellarGraph` object and so isn't duplicated in memory. The final memory usage is thus dominated by the sparse adjacency matrices for each type, and so should be proportional to `O(number of edge types + number of edges)` (the first term models the constant cost of each sparse matrix). Previously, it included a `O(number of nodes * feature size)` term, which was very large for the AIFB dataset (`8285 nodes * 8285 features per node * 4 bytes per feature = 262MiB`).

Unfortunately, this still isn't enough to allow the notebook to run on CI, as it still hits out-of-memory errors (#819): https://buildkite.com/stellar/stellargraph-public/builds/3220

Some of these optimisations can potentially be applied to other locations in the library too (#1276, #1277).

See: #1272

### DIFF
--- a/stellargraph/mapper/full_batch_generators.py
+++ b/stellargraph/mapper/full_batch_generators.py
@@ -47,6 +47,7 @@ from . import (
 from ..core.graph import StellarGraph
 from ..core.utils import is_real_iterable
 from ..core.utils import GCN_Aadj_feats_op, PPNP_Aadj_feats_op
+from ..core.validation import comma_sep
 
 
 class FullBatchGenerator(Generator):
@@ -423,8 +424,8 @@ class RelationalFullBatchNodeGenerator(Generator):
         # extract node, feature, and edge type info from G
         node_types = list(G.node_types)
         if len(node_types) != 1:
-            raise ValuError(
-                f"{type(self).__name__}: expected one node type, found {num_node_types} types",
+            raise ValueError(
+                f"G: expected one node type, found {comma_sep(sorted(node_types))}",
             )
 
         self.features = G.node_features(node_type=node_types[0])
--- a/stellargraph/mapper/full_batch_generators.py
+++ b/stellargraph/mapper/full_batch_generators.py
@@ -421,7 +421,13 @@ class RelationalFullBatchNodeGenerator(Generator):
         G.check_graph_for_ml()
 
         # extract node, feature, and edge type info from G
-        self.features = G.node_features(G.nodes())
+        node_types = list(G.node_types)
+        if len(node_types) != 1:
+            raise ValuError(
+                f"{type(self).__name__}: expected one node type, found {num_node_types} types",
+            )
+
+        self.features = G.node_features(node_type=node_types[0])
 
         # create a list of adjacency matrices - one adj matrix for each edge type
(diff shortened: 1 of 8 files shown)


stellargraph/mapper/full_batch_generators.py
```diff
@@ -421,37 +421,22 @@ def __init__(self, G, name=None, sparse=True, transform=None):
         G.check_graph_for_ml()
 
         # extract node, feature, and edge type info from G
-        self.node_list = list(G.nodes())
-
-        self.features = G.node_features(self.node_list)
+        node_types = list(G.node_types)
+        if len(node_types) != 1:
+            raise ValuError(
```
kjun9, line 426:
```suggestion
            raise ValueError(
```

huonw: I guess I should have a test for this new branch 😄

huonw: Done (and reworded the message slightly).

Issue #1159, opened 2020-03-25
Upload the demo notebooks to ReadTheDocs for more control and SEO


huonw:
### Description

We currently expose our demos through git and GitHub only, meaning someone either:
- browses them starting at https://github.com/stellargraph/stellargraph/tree/develop/demos (and either just reads the static rendering or runs them with Binder/Colab #1119), 
- checks out the whole repo locally and starts up a jupyter lab themselves.

The second is great and we should support it forever, but the first is somewhat limiting, as it:
- requires us to be careful to set up links to point to the right place (e.g. linking into `master` versions, even from the `develop` branch's README, and using relative links to keep that version)
- makes us reliant on Github's functionality for indexing and search (including exposure to search engines like Google)
- gives us limited linking capability; it's not easy for us to have our documentation link to the right versions of the relevant demos (without just pointing to GitHub directly, which will go out of date for old version) or even vice versa

If we upload the demos to readthedocs, we can solve most of these (e.g. they'll be the right versions).

PR #1279, opened 2020-04-20
Add notebooks to docs

kjun9:
This is roughly working but the output is quite messy: https://stellargraph--1279.org.readthedocs.build/en/1279/demos/index.html

The script currently just takes the directory names containing demo notebooks and uses that as section headings, which is not ideal.

Notebooks also contain markdown headings which are inconsistently styled at the moment, which messes up the TOC tree.

Follow-ups:
* messy sections https://github.com/stellargraph/stellargraph/issues/1294
* binder/colab buttons https://github.com/stellargraph/stellargraph/issues/1293
* broken links https://github.com/stellargraph/stellargraph/issues/1292
* readme as indices https://github.com/stellargraph/stellargraph/issues/1298
* validate script in CI https://github.com/stellargraph/stellargraph/issues/1299

#1159


docs/demos/basics/index.txt
```diff
@@ -0,0 +1,9 @@
+basics
+====================================
+
+.. toctree::
```
timpitman, line 4: Is there a way to capitalise the contents headings to match the rest of the docs? (possibly just re-name the folders, unless there's a sphinx way)

scripts/create-nbsphinx-links.sh
```diff
@@ -0,0 +1,42 @@
...
+
+  # recurse through parent directories and create index.txt
+  nb_link_dir="$(dirname "$nb_link")"
+  while [ "$nb_link_dir" != "docs" ]; do
+    nb_link_parent_index="$nb_link_dir/index.txt"
+    if [[ ! -f "$nb_link_parent_index" ]]; then
+      echo "Creating '$nb_link_parent_index'"
+      mkdir -p "$nb_link_dir"
+      cat > "$nb_link_parent_index" << EOF
+$(basename "$nb_link_dir")
```
timpitman, line 17: Maybe just this can be capitalised? (ref my comment earlier in the PR?)

kjun9: Hmm yeah that's a reasonable idea! Seems like every directory is kebab case, except for `community_detection` - I guess I could put a formatting rule in here to replace `-`/`_` with space, and capitalise each word?

timpitman: I like the concept, but figuring out the sed is tricky!  Possibly awk more suited?

kjun9: BSD sed doesn't support `\U` to capitalise, so I'm trying something with `awk` now..

scripts/create-nbsphinx-links.sh
```diff
@@ -0,0 +1,47 @@
...
+      mkdir -p "$nb_link_dir"
+      cat > "$nb_link_parent_index" << EOF
+$section_title
+====================================
+
+.. toctree::
+    :titlesonly:
+    :glob:
+
+    */index
```
huonw, line 29: Why are both of these lines needed? Doesn't the `./*` glob include any `index` files?

kjun9: Ah yep seems like it 👍 a bit confused then how sphinx resolves the toc tree if it adds all the files from subdirectories in the glob recursively.. but the output seems correct at least

kjun9:
> I think there's two more follow up issues we could have:
> 
> * reusing the README.mds as the indices somehow (one approach would be to convert them into `README.rst`s and just have the `toctree` within them)
> * validating that the committed directory structure matches the one computed with the script on CI so there's no divergence? (Long term I guess it would better for it to be a Python script instead of a bash one, since that's more controlled/flexible.)

Thanks, that makes sense. I've created two more issues #1298 and #1299

Issue #1262, opened 2020-04-17
'custom_keras_layers' doesn't include all layers


huonw:
### Describe the bug

Per https://github.com/stellargraph/stellargraph/issues/1261, we have `custom_keras_layers` at the top level to use with `custom_objects` when deserialising a Keras model. Unfortunately this doesn't list all of our custom layers, so serialising and deserialising some models will fail.

https://github.com/stellargraph/stellargraph/blob/af85a986fa39edaed83987c7c3f7e536c2100fa2/stellargraph/__init__.py#L56-L72

### To Reproduce

This program finds all members of submodules of `sg.layers` that are subclasses of `Layer` and compares it to the `sg.custom_keras_layers` dictionary:

```python
import stellargraph as sg
import inspect
import tensorflow as tf

all_classes = [
    obj 
    for _, mod in inspect.getmembers(sg.layer) if inspect.ismodule(mod) 
    for _, obj in inspect.getmembers(mod) if inspect.isclass(obj)
]

all_layers = {
    layer
    for layer in all_classes 
    if issubclass(layer, tf.keras.layers.Layer) and "stellargraph" in inspect.getfile(layer)
}

all_layers.difference(sg.custom_keras_layers.values())
```

### Observed behavior

```
{stellargraph.layer.deep_graph_infomax.DGIDiscriminator,
 stellargraph.layer.deep_graph_infomax.DGIReadout,
 stellargraph.layer.graph_classification.GraphClassificationConvolution,
 stellargraph.layer.graphsage.GraphSAGEAggregator,
 stellargraph.layer.knowledge_graph.ComplExScore,
 stellargraph.layer.knowledge_graph.DistMultScore,
 stellargraph.layer.preprocessing_layer.GraphPreProcessingLayer,
 stellargraph.layer.preprocessing_layer.SymmetricGraphPreProcessingLayer,
 stellargraph.layer.watch_your_step.AttentiveWalk}
```

### Expected behavior

An empty set.

### Environment

Operating system: `Darwin-18.6.0-x86_64-i386-64bit`
Python version:
```
3.6.9 (default, Jul 10 2019, 12:25:55) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)]
```
Package versions:
<details>

```
absl-py==0.9.0
appnope==0.1.0
astor==0.8.1
attrs==19.3.0
backcall==0.1.0
bleach==3.1.4
boto==2.49.0
[...]

PR #1280, opened 2020-04-20
Include all layers in `custom_keras_layers`

huonw:
This adds all of the layers defined by StellarGraph to the `custom_keras_layers` dictionary, which can be used for deserialising models created with StellarGraph.

It does so by listing the layers in the source and iterating over them to create a dictionary, rather than:

- trying to automatically get a list of layers defined by StellarGraph, because that seems difficult, and it would be easy for it to be silently and non-obviously adding extra (see discussion of CI below) or missing some
- defining the dictionary literally as done previously, because there's a risk of typos in the class names, which is eliminated by accessing the class's `__name__` property

I tried to add a test, but Python's `inspect` module gets confused on CI and thinks that various Tensorflow classes are defined by Stellargraph, e.g.:

- test file: https://github.com/stellargraph/stellargraph/blob/d6ecf2c1d3a8ce048d7bc1f17c4ca569bd843a65/tests/test_init.py
- test results: https://buildkite.com/stellar/stellargraph-public/builds/3866#d3e4f9c2-4bf6-4ec0-84c8-9ae9ca565d84/299-384 

In the interests of having this code landed, this doesn't include a test.

See: #1262


huonw:
I do not understand what's going here. I can't reproduce it locally, even using the docker image from CI. 🤷‍♂️

Issue #1216, opened 2020-04-07
Store adjacency lists as `RaggedTensor`s in `StellarGraph`


kieranricardo:
This issue is to re-write the `stellargraph` adjacency lists as `RaggedTensor`s,  and  re-write necessary functions to handle the new `RaggedTensor` format.

huonw:
As of #1296, we now store these as a manual/custom ragged tensor built on NumPy. Note the timing in https://github.com/stellargraph/stellargraph/pull/1296#issuecomment-616960234 : ragged tensors are significantly slower than numpy arrays. Related Tensorflow issues:

- https://github.com/tensorflow/tensorflow/issues/38263
- https://github.com/tensorflow/tensorflow/issues/39750

The value of this change would likely be having more things TF end-to-end (e.g. #1206), and thus potentially more complied/jitted. #1228 is related too.

PR #1291, opened 2020-04-21
Compute edge adjacency lists lazily (32× speedup, 10× smaller)

huonw:
Many algorithms do not use the adjacency lists (the `_edges..._dict` fields on `EdgeData` that map from node ID to the ilocs of any incident edges). For instance, algorithms like `GCN`, `GAT`, `ComplEx` and even `GCNSupervisedGraphClassification` use only the adjacency matrix, which is computed directly from the edge list (the source and target columns). They are mainly used by random walks and neighbourhood sampling.

These adjacency lists are currently a significant portion of the memory use of a `StellarGraph` instance, as well as a significant portion of the time to construct. (The exact trade-off may change with #1216.)

To validate this change, I compared this PR with `develop` (7a508c579d635615cdbb88bf411b7cb9c67a9e9e).
- For `StellarGraph(...)` construction, there's dramatic improvement in the memory usage (reductions of up to 80%) and time (speed-ups of up to 30×) when the adjacency lists aren't computed, and no little change when the adjacency are computed (that is, the total time or memory usage doesn't change).
- This potentially adds a bit more overhead to doing neighbourhood queries, because of the additional `if` statement, but there's not strong evidence either way. (If it is actually a problem, we could handle this with some self-modifying code, where the `_adj_lookup` starts as a version that unconditionally calls `_init_adj_lists` and then replaces `_adj_lookup` with a version that call that doesn't, so the first call initialises and the later ones don't even think about it.)

This also extends the benchmarks to compare the default construction (without adjacency lists) to one where the adjacency lists are computed.

See: #1295

### Creation

Microbenchmark (`test_allocation_benchmark_creation` and `test_benchmark_creation`, using the `...-None-1000-5000` parameters: 1000 nodes, no features, 5000 edges).

| code | median memory (KiB) | median time (ms) |
|---|---|---|
| before | 541 | 21.4 |
| after (no adjacency lists) | 70 | 12.7 |
| after (with adjacency lists) | 542 | 20.0 |

The times for "after with adjacency lists" and "before" are within error of each other, based on their standard deviations/overlaps. (The "with adjacency lists" benchmark calls the function to compute them, i.e. forcing them the adjacency lists to be computed and stored in the `StellarGraph`.)

Real datasets (https://gist.github.com/kieranricardo/536776d95bfbc56e8ab709f1d261cb50):

| dataset | size before (MiB) | size after (MiB) | reduction | time before (s) | time after (s) | speedup |
|---|---|---|---|---|---|---|
| Cora | 1.22 | 0.234 | -80% | 0.0279 | 0.00882 | 3.2× |
| FB15k | 23.1 | 7.3 | -68% | 1.36 | 0.1712 | 7.9× |
| reddit | 450 | 166 | -63% | 27.3 | 0.910 | 30× |

Note that the "after" sizes and times are for when the adjacency lists are not used. The memory and time still has to be spent if/when they are used, bringing the numbers back up to be the same as "before".

### Using the adjacency lists

To validate that this doesn't slow things down too much, here's some before/after comparisons for functionality that actually uses the adjacency lists.

Neighbour traversals with ilocs (`test_benchmark_get_neighbours[True]`), all times in microseconds:

| code | min | mean | stddev | median | iqr |
|---|---|---|---|---|---|
| before | 865 | 997 | 156 | 947 | 135 |
| after | 900 | 1041 | 128 | 996 | 159 |

Random walks (via `pytest -k 'benchmark and walk'`), all times are means ± stddev.

| code | uniform | biased | graphsage | directed graphsage | hinsage | biased weighted | metapath |
|---|---|---|---|---|---|---|---|
| before | 595 ± 117 | 696 ± 91 | 444 ± 83 | 1047 ± 137 | 419 ± 55 | 15593 ± 1216 | 2757 ± 371 |
| after | 696** ± 92 | 713 ± 126 | 459 ± 97 | 1113 ± 151 | 424 ± 85 | 16305 ± 1479 | 2794 ± 326 |

(Column name -> benchmark mapping: uniform `test_benchmark_uniformrandomwalk`, biased `test_benchmark_biasedrandomwalk`, graphsage `TestBreadthFirstWalk::test_benchmark_bfs_walk`, directed graphsage `TestDirectedBreadthFirstNeighbours::test_benchmark_bfs_walk`, hinsage `test_benchmark_sampledheterogeneousbreadthfirstwalk`, biased weighted `test_benchmark_biasedweightedrandomwalk`, metapath `test_benchmark_uniformrandommetapathwalk`.)

**Summary**: there's maybe some hints that this new form is slower, but there's also hints that it's faster. In any case, the error bars are large enough that this probably isn't making much difference.


kieranricardo:
following our slack discussion, I supsect that your gist above might not be measuring the memory of the edge lists and skewing the results!

I made a small edit to explicitly copy over the edge lists before measuring the memory: https://gist.github.com/kieranricardo/536776d95bfbc56e8ab709f1d261cb50

From this gist on FB15K the adj lists are still 60% of the total memory! So I think this PR is still worth while 👍

huonw:
I think there's an argument to be made either way:

1. the current behaviour measures the incremental memory used by `StellarGraph` over storing the raw dataset in memory, and so answers the question "how much _more_ memory do I need to use StellarGraph over traditional methods"
2. the new behaviour measures something closer to the total memory used for storing the `StellarGraph` long term

However, arguably 2 is better for general user-facing communication, as you say.

Rather than change things here, I'll just wait to #1220, because that changes the edge list format to not be able to reuse the arrays from the external DataFrame.

kieranricardo:
> Rather than change things here, I'll just wait to #1220, because that changes the edge list format to not be able to reuse the arrays from the external DataFrame.

I'm not 100% on this point, but `pd.concat(type_dfs)` in `ElementData.__init__` duplicates memory of the dataframes so I think we're currently not reusing the arrays from the external dataframe. 

Running this locally and watching my `python` memory usage on activity monitor is ~7Gb.

```
> dfs = [pd.DataFrame(np.ones((1000, 1000))) for _ in range(1000)]
```

Then when I run this my memory usage goes up to 15Gb.
```
> concated = pd.concat(df)
```

huonw:
I've updated this, and indeed as you suspected in https://github.com/stellargraph/stellargraph/pull/1291#issuecomment-620399157, the iloc work didn't change the behaviour with ownership. I've updated the PR description with the new numbers incorporating the iloc work, and using the version that includes the edge lists https://gist.github.com/kieranricardo/536776d95bfbc56e8ab709f1d261cb50.

PR #1309, opened 2020-04-21
Return figure in plot_history

LarsNeR:
### Description

The `sg.utils.plot_history` should return the figure `fig`.

### User Story

**As** a developer

**I want to** work on the figure that `sg.utils.plot_history` creates

**So that** I can add titles, export the figure, and so on, ...

This is my first open source PR ever. I was not sure if I should create an Issue for that first. If something is missing please tell me what to do.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

### DIFF
--- a/stellargraph/utils/history.py
+++ b/stellargraph/utils/history.py
@@ -37,7 +37,7 @@ def plot_history(history, individual_figsize=(7, 4), return_figure=False, **kwar
         kwargs: additional arguments to pass to :meth:`matplotlib.pyplot.subplots`
 
     Returns:
-        [matplotlib.figure.Figure]: The figure object with the plots if `return_figure=True`, None otherwise
+        :class:`matplotlib.figure.Figure`: The figure object with the plots if ``return_figure=True``, None otherwise
 
     """
 
--- a/stellargraph/utils/history.py
+++ b/stellargraph/utils/history.py
@@ -34,6 +34,10 @@ def plot_history(history, individual_figsize=(7, 4), **kwargs):
         history: the training history, as returned by :meth:`tf.keras.Model.fit`
         individual_figsize (tuple of numbers): the size of the plot for each metric
         kwargs: additional arguments to pass to :meth:`matplotlib.pyplot.subplots`
+
+    Returns:
+        [matplotlib.figure.Figure]: The figure object with the plots
+
     """
 
     # explicit colours are needed if there's multiple train or multiple validation series, because
@@ -88,3 +92,5 @@ def plot_history(history, individual_figsize=(7, 4), **kwargs):
 
     # minimise whitespace
     fig.tight_layout()
+
+    return fig
--- a/stellargraph/utils/history.py
+++ b/stellargraph/utils/history.py
@@ -20,7 +20,7 @@ import matplotlib.pyplot as plt
 __all__ = ["plot_history"]
 
 
-def plot_history(history, individual_figsize=(7, 4), **kwargs):
+def plot_history(history, individual_figsize=(7, 4), return_figure=False, **kwargs):
     """
     Plot the training history of one or more models.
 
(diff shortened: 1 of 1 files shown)


huonw:
> This is my first open source PR ever. I was not sure if I should create an Issue for that first. If something is missing please tell me what to do.

Welcome to open source @LarsNeR! We're glad that we can be the ones to help. 😄 What you've done here is just fine for StellarGraph; some projects prefer to have issues filed first, but we're happy either way.

---

For this particular change, I think this is unfortunately a bit tricky to get perfect: returning the plot seems to mean Jupyter renders it twice by default. For example, let's look at the [graphsage node classification notebook](https://github.com/stellargraph/stellargraph/blob/develop/demos/node-classification/graphsage/graphsage-cora-node-classification-example.ipynb); that version saved in the repo has just one plot for `acc` and one for `loss` (at `In[18]`), while the version run on CI with this change has two for each (at `In[19]`): https://nbviewer.jupyter.org/urls/buildkite.com/organizations/stellar/pipelines/stellargraph-public/builds/3334/jobs/f14f9d1b-19f8-44f9-9f91-fed4dc7b172e/artifacts/9add61dc-72ef-43a3-80b7-fdcc55189c30

It seems this is specifically caused by returning the `Figure` object. The smallest reproducer I can find is:

```python
from matplotlib import pyplot as plt

def f():
    fig, ax = plt.subplots()
    ax.scatter([1, 2], [3, 2])
    return fig

f()
```

image


This has come up elsewhere, like https://github.com/statsmodels/statsmodels/issues/1265. 

---

I see a few options:

1. leave `plot_history` as it is, and just close this PR (worst option!). Those who want more detailed control over their plot would then have to copy the `plot_history` code and make the changes there
2. have a flag that changes whether it is returned, so someone who wants to do more with the plot can do `plot_history(history, return_figure=True)`:
   ```python
   def plot_history(history, individual_figsize=(7, 4), return_figure=False, **kwargs):
       ...

       if return_figure:
           return fig
   ```
3. Have each of the existing demo notebooks save the result, so that the history figure isn't the result value of the cell, like `history_figure = plot_history(history)` instead of just `plot_history(history)` (an alternative way to do this is to add a semicolon like `plot_history(history);` but our [black](https://black.readthedocs.io/en/stable/) formatting within each notebook removes this)
4. brainstorm if there's some other way to achieve the same thing (I suspect not)?

I'm inclined towards number 2. Do you have any thoughts?

LarsNeR:
> Welcome to open source

Thanks!
_____
Oh yes that's right. I missed that. If the call is the last one in the cell it will be plotted twice. I'm also leaning towards number two. Even if it is only a utils method but then there is no behavioral change.

stellargraph/utils/history.py
```diff
@@ -33,7 +33,12 @@ def plot_history(history, individual_figsize=(7, 4), **kwargs):
     Args:
         history: the training history, as returned by :meth:`tf.keras.Model.fit`
         individual_figsize (tuple of numbers): the size of the plot for each metric
+        return_figure (bool): if True, then the figure object with the plots is returned, None otherwise.
         kwargs: additional arguments to pass to :meth:`matplotlib.pyplot.subplots`
+
+    Returns:
+        [matplotlib.figure.Figure]: The figure object with the plots if `return_figure=True`, None otherwise
```
huonw, line 40: Thanks for taking care to expand the docs, I think there's just a couple of minor tweak we can do here. Currently the [] brackets are appearing in the documentation which looks a bit peculiar, and could be misleading (someone might interpret it as meaning it returns a list `[...]` that contains a figure): https://stellargraph--1309.org.readthedocs.build/en/1309/api.html#stellargraph.utils.plot_history

image

Additionally, unlike markdown, a single ` doesn't mean code in reStructured Text (notice how the `return_figure=True` text is italicised); it has to be doubled.

Thus, I think we should adjust this to:

```suggestion
        :class:`matplotlib.figure.Figure`: The figure object with the plots if ``return_figure=True``, None otherwise
```

LarsNeR: Thanks, of course we should change that. The thing with the brackets I got from here: [group_aggregate](https://stellargraph--1309.org.readthedocs.build/en/1309/api.html#stellargraph.layer.MeanAggregator.group_aggregate). I thought they are used if the type is not built-in.

huonw: Ah, I think that's a bug there too. Thanks for noticing it! I opened https://github.com/stellargraph/stellargraph/issues/1364.

huonw:
Thanks for taking the time to submit a pull request and work through the little issues! You've now officially contributed code to an open source project 😄

Issue #1313, opened 2020-04-22
`pip install` in Colab notebooks doesn't handle release candidates properly


huonw:
### Describe the bug

Our jupyter notebooks install stellargraph on Google Colab via `%pip install stellargraph[demos]`. This picks up the latest stable release version. When we do release candidates, this won't work, and there'll be a mismatch between the notebooks on `master` (corresponding to the release candidate) and the version of StellarGraph that gets installed if someone runs them on Colab (the previous stable release).

(This also may cause confusion if someone has started working with a notebook, and then we do a release, and then installs in future will pull in this wrong version.)

PR #1314, opened 2020-04-22
Install specific version of SG on Colab, in notebooks

huonw:
This uses the versioning logic we've already got for the notebook version checks (#1242) to install a specific version of StellarGraph on Google Colab. This has a few benefits:
- it ensures that we can install exactly the version the notebook is designed for, even if there's been releases in the mean time
- it allows Colab to work with prereleases and release candidates, which pip won't install by default (https://pip.pypa.io/en/stable/reference/pip_install/#pre-release-versions)

To validate that this is correct syntax, I ran `%pip install stellargraph[demos]==0.4.0b0` in a jupyter notebook (using a random old pre-release version we've published). It worked to install that version.

See: #1313

### DIFF
--- a/RELEASE_PROCEDURE.md
+++ b/RELEASE_PROCEDURE.md
@@ -19,7 +19,7 @@
        - Version bumping: Change version from “X.X.Xb” to “X.X.X”. E.g. version=”0.2.0b” to version=”0.2.0”
          - `stellargraph/version.py`
          - `meta.yaml`
-       - Update expected versions of demo notebooks: `scripts/format_notebooks.py --version_validation --overwrite demos/`
+       - Update expected versions of demo notebooks: `scripts/format_notebooks.py --default --overwrite demos/`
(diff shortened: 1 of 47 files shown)


huonw:
> Would pip install --pre stellargraph also work? Just wondering if that might be better since we never publish any b versions.

That could work, but it doesn't give the first benefit, of working to install old versions. With colab, I can imagine someone starting with a notebook, and extending it to work with their own task/problem and coming back to it in future. If we've done a release in the interim, coming back to it in future may install an incompatible version and the notebook no longer runs, but with the `==<version>` it should run no matter what.

> I guess it technically doesn't matter since notebooks in develop never get run in colab

Yeah, that was my logic.

Shall we merge this now, and think about it more before the full release? I filed https://github.com/stellargraph/stellargraph/issues/1315 to keep track.

kjun9:
> Shall we merge this now

yep sounds good to me

Issue #1090, opened 2020-03-17
Graph Classification Tutorial for beginners


PantelisElinas:
### Description

Write an article for our Medium blog to explain graph classification for beginners.

### User Story
As a: Research Engineer
I want: to explain in an introductory tutorial for new users how to use StellarGraph for graph classification
so that: they can use StellarGraph to solve their graph classification problems.

### Done Checklist (Development)
- [ ] Article published on StellarGraph Medium blog

huonw:
Can we aim to have a notebook too, potentially as the source of truth? That will ensure it's always up to date (it'll be checked on CI) and make it easier for users to jump in and tweak things.

PantelisElinas:
We can add a notebook. We just need a bigger than MUTAG dataset so we can have a simpler and more detailed narrative than the upcoming demo. I have updated the description.

PR #1319, opened 2020-04-22
DGCNN demo with PROTEINS dataset

PantelisElinas:
Hi,

this is a draft pull request seeking feedback with regards to the organisation of the demo. 

I am still trying to reproduce the results in the paper; there are a number of parameters that are not given, e.g., regularisation type if any, batch size, etc., and might be important for achieving the published values. I list a few differences between the paper and the demo at the end of the notebook. Alternatively, there could be an error in our implementation of DGCNN. I'll keep investigating but if anyone sees anything that looks incorrect, let me know.

P.


huonw:
I like the structure of this demo. I feel that it's easier to understand how the pieces fit together than the demo in #1053 because it's got a progression from generator to model & additional layers and then training/evaluation, with interspersed explanations, rather than needing to put it into a couple of big cells for K-fold validation with restarts. That is, I feel it goes a reasonable distance towards #1090. 👍

demos/graph-classification/dgcnn-graph-classification.ipynb
```diff
@@ -0,0 +1,1061 @@
+{
```
huonw, line 1: <p>I think this may not be doing exactly the right thing, because the dimension of the predictions is <code>(batch_size, 13, 1)</code>. For instance, adding a cell before the <code>epochs = 300</code> one that does:</p><p> </p><pre class="ql-syntax" spellcheck="false">preds = model.predict(test_gen)
preds.shape
</pre><p>I see <code>(112, 13, 1)</code> , i.e. there's 13 predictions for each of the 112 test graphs. Is that appropriate for this problem?</p><p> </p><p>https://github.com/stellargraph/stellargraph/issues/1344 is relevant.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1319/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

PantelisElinas: This might be the reason for the flaky performance. 

The issue is the output of the `Conv1D` layer. I tried flattening the output before passing to the `Dense` layer but unfortunately the `tf.keras.layers.Flatten` layer outputs a tensor of size `[None, None]` which cannot be used as input to the `Dense` layer. 

An alternative approach that seems to work is to apply a pooling operation such as `x_out = tf.keras.layers.GlobalMaxPool1D()(x_out)` after the second `Conv1D` layer and before the `Dense` layers. This produces the desired outcome and performance closer to what is reported in the paper; however, I still need to verify the latter. I think that adding `Flatten` or another pooling layer after a `Conv1D` layer and before a `Dense` layer is common practice due to the former's output shape incompatibility with the latter's input.

PantelisElinas: @huonw 

I fixed the above problem by making a small update to the `SortPooling` layer such that when `flatten_output=True` the output tensor is reshaped to have the correct shape instead of `None`. This is possible since the output shape for this layer is known given the inputs. This change makes it possible to follow the `Conv1D` layer with a `Flatten` layer that correctly passes the data to the `Dense` layer avoiding the error I mentioned above.

Performance now is on par with the original paper although I will run a bigger experiment with repeated k-fold cross validation to verify. 

I think the demo is ready for a proper review.

P.

PantelisElinas:
This also tackles #1344

demos/graph-classification/dgcnn-graph-classification.ipynb
```diff
@@ -0,0 +1,761 @@
...
```
huonw, line 1: <p>This first cell will need to be split up into a cell with the title only and then a cell with the rest of the text. This is so the "Run" buttons are placed right after the title. This requirement is being noticed on CI https://buildkite.com/stellar/stellargraph-public/builds/3588#8165461b-3270-4d18-9fe3-e52abf74a0b2/181-204 , but it's not failing properly (https://github.com/stellargraph/stellargraph/issues/1409). (https://github.com/stellargraph/stellargraph/pull/1363 has more info on the specifics of the check.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1319/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

huonw, line 1: <p>Is this cell still relevant now that there's the <code>predictions.shape</code>  cell above, and now that we've got the Flatten? It seems a bit peculiar to run predictions on an untrained model and so could maybe ether be deleted or have some explanation?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1319/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

huonw, line 1: <p>It looks like this doesn't kick in; should we increase the epochs further to train to "completion"?</p><p> </p><p>In terms of the length of time to train, it might be noticeably reduced by https://github.com/stellargraph/stellargraph/pull/1381 which I just merged, and, unlike the other graph classification notebook, there's lots of progress updates for the user to tell that something is happening here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1319/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

huonw, line 1: <p>Tiny thing, but I think an "inline list" like this should be either separated by a comma or start with a colon:</p><p> </p><blockquote>The experimental evaluation scheme in [1] does not specify some important details such as the regularisation used for the neural network layers, if a bias term is included, the weight initialization method used, and the batch size.&nbsp;</blockquote><p> </p><blockquote>The experimental evaluation scheme in [1] does not specify some important details such as: the regularisation used for the neural network layers; if a bias term is included; the weight initialization method used; and the batch size.&nbsp;</blockquote><p> </p><p>It took me a little bit to understand the grouping in this sentence as is, as it seemed like "The experimental ... network layers" was one entire clause and "if a bias term is included" was another separate one, rather than "The experimental ... important details" and "the regularisation ... layers" being separate clauses.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1319/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

PantelisElinas: <p>It's not, I forgot to delete it. In fact, we don't need the other two cells above printing the shapes for predictions and x_out. I was trying to debug the model and forgot to delete them.</p>

PantelisElinas: <p>Early stopping doesn't have to kick in if the quantity monitored does not require it. It is a failsafe more than anything to prevent overfitting. Different runs will produce different results with regards to convergence depending on the data split and initial weight values so for a different run maybe early stopping will be used. </p><p> </p><p>It might be best that we follow the training scheme in the paper and train for 100 epochs with no early stopping and just report the performance for the model at the 100th epoch.</p><p> </p><p>So I guess we have to options (I'm sure there are more but keeping it simple):</p><ol><li>Remove early stopping run for 100 epochs as in the paper</li><li>Increase the number of epochs</li></ol><p> </p><p>Also, for all demos we set <code>verbosity=0</code>  for the final version. In the current version I have it set to 1 for illustration. Shall I leave it as 1 so the user received feedback during training or set it to 0 like the other notebooks?</p><p> </p>

huonw: Let me rephrase: it looks like the loss is still decreasing (and accuracy still increasing). My understanding is that early stopping is a low-effort way to avoid having to fine tune the number of epochs: just choose a large number, and let early stopping detect when the model looks like it's learnt as much as it can under the given configuration.

However, following the paper without early stopping (i.e. 1) is good too.

> Also, for all demos we set verbosity=0 for the final version. In the current version I have it set to 1 for illustration. Shall I leave it as 1 so the user received feedback during training or set it to 0 like the other notebooks?

I don't think this is quite the case any more:

```
$ cd demos
$ git grep verbose=0 | wc -l
      22
$ git grep verbose=1 | wc -l
      13
$ git grep verbose=2 | wc -l
      21
```

From quick glance over the actual results, I think all of the `verbose=0` are in code, all but one of the `verbose=1` results, and similarly all but one of the `verbose=2` results. This leaves 22 `verbose=0` calls vs. 32 `verbose>0 ` calls.

Given this, and given it's a much better user experience (although mildly unfortunate that it's 100 lines of automated output), I suggest leaving it.

PantelisElinas: Ok, so I'll just set `epochs` to 150 and leave early stopping with patience at 25 so it neither runs too long and still have a chance to find a good model. Would that be ok or do you prefer an even larger value for epochs?

PantelisElinas: Ok then, I changed it to follow the paper removing early stopping and training for exactly 100 epochs.

I think I have now addressed all your suggestions so let me know if anything else needs to be updated.

PantelisElinas:
> As a final thing, I think this should update the doc-string of `DeepGraphConvolutionalNeuralNetwork` to include an appropriate `Flatten` layer or else we'll be encouraging users to use code that hits #1344.

This was already done as part of #1408

huonw:
Ah, sorry, I didn't notice when I checked locally; I must've accidentally been on a branch that didn't include that commit. 👍

Issue #1303, opened 2020-04-21
METR-LA dataset download


habiba-h:
Currently, the METR-LA dataset has to be manually downloaded from:  [(https://github.com/lehaifeng/T-GCN/tree/master/data)] for the `gcn_lstm-LA.ipynb` demo notebook. 
This needs to be resolved in future

timpitman:
See https://github.com/stellargraph/stellargraph/pull/1325 for the download fix, however documentation and tests of the other code in the datasets class is still needed

PR #1325, opened 2020-04-23
automatically download the METR_LA dataset

timpitman:
This is a simple change to download the dataset straight from GitHub (basically the whole TGCN repo as a single ~30MB .zip)

There's still some work remaining to be completed in another future PR for #1303:
* doc stubs
* test of the load() and other custom functions


stellargraph/datasets/datasets.py
```diff
@@ -850,25 +850,23 @@ def load(self):
...
 class METR_LA(
     DatasetLoader,
     name="METR-LA",
     directory_name="METR-LA",
-    url="https://github.com/lehaifeng/T-GCN/tree/master/data",
-    url_archive_format="n/a",
-    expected_files=["los_speed.csv", "los_adj.csv"],
+    url="https://github.com/lehaifeng/T-GCN/archive/master.zip",
+    url_archive_format="zip",
+    url_archive_contains_directory=False,
```
huonw, line 860: We could change `directory_name` to `T-GCN-master` and then remove this?

```suggestion
```

If we ever use the `sz` data from that repo, we could have the two datasets share a single downloaded folder.

stellargraph/datasets/datasets.py
```diff
@@ -850,25 +850,23 @@ def load(self):
...
-@experimental(reason="the data isn't downloaded automatically", issues=[1303])
+@experimental(reason="tests and documentation missing", issues=[1303])
 class METR_LA(
     DatasetLoader,
     name="METR-LA",
     directory_name="METR-LA",
-    url="https://github.com/lehaifeng/T-GCN/tree/master/data",
-    url_archive_format="n/a",
-    expected_files=["los_speed.csv", "los_adj.csv"],
+    url="https://github.com/lehaifeng/T-GCN/archive/master.zip",
```
huonw, line 858: Do you think it's worth pining to a particular commit here, in case there's mutations in future?

```suggestion
    url="https://github.com/lehaifeng/T-GCN/archive/8427128f04157e6fd0b239a8734a468d923cd0c9.zip",
```

(Using the current most recent commit https://github.com/lehaifeng/T-GCN/commit/8427128f04157e6fd0b239a8734a468d923cd0c9)

It's true that we don't have any similar defence for the other datasets, but since it's easy to do here, maybe we should?

(This will likely require updating `expected_files` or `directory_name`.)

timpitman: This does make for a very ugly folder name!: `T-GCN-8427128f04157e6fd0b239a8734a468d923cd0c9`

stellargraph/datasets/datasets.py
timpitman: the disadvantage here is that the directory name in datasets is quite unintuitive

huonw: That's fair, but I'm not sure it is a problem? I think it's fine for us to treat the `stellargraph-datasets` folder as an "unsupported" black box, where the format is totally internal? (This reasoning applies for the commit-hash-in-directory name below too.)

timpitman:
> Just so I'm clear, are you planning to do the doc/testing in this PR?

**No** - that's left to be addressed in #1303

stellargraph/datasets/datasets.py
timpitman: I suppose ugly here is fine; most users won't even see it and it's not going to clash with another dataset!

Issue #1237, opened 2020-04-14
Have `use_ilocs=True` return edge/node type iloc too


kieranricardo:
Currently `use_ilocs` (from #1220) specifies that a `StellarGraph` method should accept and return node ilocs instead of node ids. This issue is to extend this functionality to handle node and edge types/type ilocs in the same way.

PR #1366, opened 2020-04-26
Add node/edge type iloc conversions to StellarGraph

huonw:
This extends the ilocs we expose from just node ilocs, to ilocs for node and edge types as well. This is helpful for algorithms that manipulate these types, such as:

- knowledge graphs models, which often compute embeddings per-type, and thus ilocs are great as indices
- HinSAGE and Metapath2Vec, which can, theoretically, do comparison and manipulation of integers rather than more complicated types like strings (potentially including using lists/arrays instead of dicts, for any mappings with node/edge types as keys).

See: #1237


huonw:
cc @kieranricardo

PR #1367, opened 2020-04-26
Use ilocs end-to-end for knowledge graph ranking

huonw:
With #1366, there are native and supported APIs for working with edge type ilocs. This allows us to use ilocs end-to-end in knowledge graph ranking. This both simplifies the code and makes it faster, e.g. ranking is performed in the DistMult notebook approximately as follows (the model doesn't need to be trained, because we don't care about the actual ranks, just the time to compute):

```python
import stellargraph as sg
import tensorflow as tf

def run(dataset):
    graph, train, test, valid = dataset.load()
    gen = sg.mapper.KGTripleGenerator(graph, batch_size=5000)
    test_seq = gen.flow(test)
    
    model = sg.layer.DistMult(gen, embedding_dimension=100)
    tf.keras.Model(*model.in_out_tensors()).compile("adam", loss="mse")
    %time model.rank_edges_against_all_nodes(test_seq, graph)

run(sg.datasets.WN18())
run(sg.datasets.FB15k())
```

Results:

| dataset | pre-ilocs f2a96aab (s) | develop f5e519be (s) | this PR (s) |
|---|---|---|---|
| WN18 | 4.99 | 5.87 | 3.80 |
| FB15k | 42.6 | 59.2 | 20.7 |

That is, it's somewhere between 1.5-3× faster for DistMult on these datasets.

### DIFF
--- a/stellargraph/layer/knowledge_graph.py
+++ b/stellargraph/layer/knowledge_graph.py
@@ -602,27 +602,18 @@ def _ranks_from_score_columns(
     # the filtered rank is the number of unknown elements scored higher, where an element is
     # known if the edge (s, r, n) (for modified-object) or (n, r, o) (for modified-subject)
     # exists in known_edges_graph.
-
-    # FIXME(#870): this would be better without external IDs <-> ilocs translation
-    unmodified_nodes = known_edges_graph._nodes.ids.from_iloc(unmodified_node_ilocs)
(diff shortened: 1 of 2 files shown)


huonw:
cc @kieranricardo

huonw:
Did you miss the table in the PR description?!

Issue #1343, opened 2020-04-24
Automatically infer node type arguments for graphs with only one node type


kieranricardo:
When creating a `StellarGraph` with only one node type the user can pass in the nodes without specifying a node type and `StellarGraph` automically sets a default node type. What this default node type is  or how to get it is not immediately apparent.

Using HinSAGE requires specifying a node type but in the case of heterogenous graph one node type this may not obvioius. This issue is allow HinSAGE to automatically infer the head node type as the defulat node type in cases where there is only one node type.

(StellarGraph 1.0.0rc1 dogfooding.)

huonw:
I think this is more general than HinSAGE, e.g. it applies to `node_features(node_type=...)`. In a homogeneous graph, `graph.node_features()` should work.

Also, this would resolve #1023.

Issue #1023, opened 2020-03-05
HinSAGENodeGenerator's `head_node_type` parameter is unintuitive when working with a graph with just a default node type.


kjun9:
### Description

A heterogeneous graph could contain multiple edge types, but only a single node type, in which case the user may simply be using the default node type. `graph.info` shows the node type as "default", but the user never explicitly creates a node type called "default". Then, when they try to create a `HinSAGENodeGenerator` with the graph, the `head_node_type` parameter can be a bit puzzling since they wouldn't remember ever creating any node types at all.

In the generator, we could consider some options like:
* infer the head node type if there is only one node type available in the graph
* make it more transparent to the user when nodes/edges are assigned the "default" type while creating the graph.
* update the docs to guide the user better

kjun9:
(cc @annitrolla )

huonw:
I think the only way this issue can come up is in a homogeneous graph `one_node_type_graph`, since that's the only time that the user has no reason to be thinking about node types. Thus, #1375 hopefully addresses this by making `HinSAGENodeGenerator(one_node_type_graph, batch_size=2, num_samples=[3, 4])` work.

PR #1375, opened 2020-04-27
Infer node types for homogeneous-node graphs with 'unique_node_type'

huonw:
This adds a `unique_node_type` method to `StellarGraph`. The API of this is slightly peculiar, as it either:

- returns the single node type in a graph with only one (a "homogeneous-node graph")
- throws a `ValueError` if it doesn't have exactly one node type, optionally using a custom error message

This API is what I think is the most reliable one, as it's not possible to accidentally forget to detect and fail on the case when there's not one node type. (An alternative API would be to return `None` in the failure case. This would make sense in some languages, but Python's non-static type system (and also the flexibility for what StellarGraph allows for node types) means that a caller could forget to detect this case, and so continue when they should really stop.)

This PR then uses this new functionality to:
- make `homogeneous_graph.node_features()` work
- make `HinSAGENodeGenerator(homogeneous_graph, batch_size=5, num_samples=[10, 5])` work, meaning it's now easier to use (#1023) and is a drop-in replacement for `GraphSAGE`
- simplify some error handling code in other generators, by calling `unique_node_type` instead of manual `if len(...) > 1` code

See: #1343

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -710,6 +710,31 @@ class StellarGraph:
         """
         return set(self._nodes.types.pandas_index)
 
+    def unique_node_type(self, message=None):
+        """
+        Return the unique node type, for a homogeneous-node graph.
+
+        Args:
+            message (str, optional): a custom message to use for the exception; this can use the
+                ``%(found)s`` placeholder to insert the real sequence of node types.
+
+        Returns:
+            If this graph has only one node type, this returns that node type, otherwise it raises a
+            ``ValueError`` exception.
+        """
+
+        all_types = self._nodes.types.pandas_index
+        if len(all_types) == 1:
+            return all_types[0]
+
+        found = comma_sep(all_types)
+        if message is None:
+            message = (
+                "Expected only one node type for 'unique_node_type', found: %(found)s"
+            )
+
+        raise ValueError(message % {"found": found})
+
     @property
     def edge_types(self):
         """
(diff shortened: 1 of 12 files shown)


stellargraph/core/graph.py
```diff
@@ -755,13 +780,19 @@ def node_features(self, nodes=None, node_type=None):
...
+          returns the features for all nodes of type ``some_type``, in the same order as
+          ``graph.nodes(node_type=some_type)``.
+        - the graph has only one node type, in which case ``graph.node_features()`` returns the
+          features for all nodes (equivalent to the previous case with ``some_type`` set to the
+          single node type of ``graph``)
+        - ``nodes`` is passed, in which case ``graph.node_features(nodes=some_node_ids)`` returns
+          the features for each node in ``some_node_ids`` in the same order as that input. Every
+          node in ``nodes`` must have the same type. If ``nodes`` is passed without specifying
+          ``node_type``, the node type of ``nodes`` will be inferred (passing ``node_type`` in
+          addition to ``nodes`` will therefore be faster).
```
kjun9, line 795: This description does make sense to me after reading carefully, but I wonder if it might be easier to parse if homogeneous vs heterogeneous graphs were explained separately? 

For instance, I think it would be nice if the user didn't need to care about the existence of the `node_type` parameter at all if their graph was homogeneous. Maybe something like:
```
For graphs with a single node type, use ``graph.node_features()`` or ``graph.node_features(nodes=some_node_ids)``.

For graphs with multiple node types, the ``node_type`` parameter should be provided:
- ``graph.node_features(node_type=some_type)`` to retrieve features of all nodes of type ``some_type``
- ``graph.node_features(nodes=some_node_ids, node_type=some_type)`` to retrieve features for each node in ``some_node_ids`` which should all be of type ``some_type``. 

The node type can be inferred even for graphs with multiple node types by using ``graph.node_features(nodes=some_node_ids)`` as long as all nodes in ``some_node_ids`` are of a single type, but will be slower than providing the node type explicitly.
```

huonw: I like it. I went with something slightly different: a list for each graph kind, with 2 and 3 elements respectively. What do you think?

huonw:
> The extra 'match' in the tests when expecting the ValueError are more specific but will mean work if the message is changed in future

FWIW, I think it's important we have at least something in the `match=...` or else it's easy for the tests to not actually be testing what they want to test, given how common `TypeError` and `ValueError` are.

stellargraph/core/graph.py
kjun9: Nice! Yeah having a list for both looks good to me

Issue #1202, opened 2020-04-06
Improve documentation for how to download and run demos


timpitman:
### Description

Raised in #1171, our documentation does not clearly explain that demos are not installed via Conda or Pip and must be downloaded separately:

>> You can start working with the examples immediately in Google Colab or Binder by clicking the  and  badges within each Jupyter notebook. You can also run them locally, after installing StellarGraph. One of the following might be appropriate:

We need clear instructions on downloading the demos (as opposed to just installing demo dependencies).

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
This came up in dogfooding too (https://github.com/stellargraph/stellargraph/issues/1333)

kjun9:
Note that this isn't quite fully resolved by #1377 (https://github.com/stellargraph/stellargraph/pull/1377#pullrequestreview-400801873)

We should definitely add the instructions in `demos/README` too, and think a bit about how to make it easier to download a demo while viewing a specific demo in github - discussion in this issue might be relevant https://github.com/stellargraph/stellargraph/issues/1390

PR #1377, opened 2020-04-27
Update readme with instructions for downloading demos

kjun9:
This adds some instructions in the top-level README for downloading our demos to run them locally.

See #1202

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,9 +80,9 @@ StellarGraph is built on [TensorFlow 2](https://tensorflow.org/) and its [Keras 
 
 You can start working with the examples immediately in Google Colab or Binder by clicking the ![](https://colab.research.google.com/assets/colab-badge.svg) and ![](https://mybinder.org/badge_logo.svg) badges within each Jupyter notebook.
 
-Alternatively, you can run download a local copy of the demos and run them using `jupyter`. The demos can be downloaded by cloning the `master` branch of this repo, or by using the [`svn` command-line tool](https://subversion.apache.org/) to download the specific directory:
+Alternatively, you can run download a local copy of the demos and run them using `jupyter`. The demos can be downloaded by cloning the `master` branch of this repo, or by using the `curl` command below:
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,7 +78,14 @@ StellarGraph is built on [TensorFlow 2](https://tensorflow.org/) and its [Keras
...
 
-You can start working with the examples immediately in Google Colab or Binder by clicking the ![](https://colab.research.google.com/assets/colab-badge.svg) and ![](https://mybinder.org/badge_logo.svg) badges within each Jupyter notebook. You can also run them locally, after installing StellarGraph. One of the following might be appropriate:
+You can start working with the examples immediately in Google Colab or Binder by clicking the ![](https://colab.research.google.com/assets/colab-badge.svg) and ![](https://mybinder.org/badge_logo.svg) badges within each Jupyter notebook.
+
+Alternatively, you can run download a local copy of the demos and run them using `jupyter`. The demos can be downloaded by cloning the `master` branch of this repo, or by using the [`svn` command-line tool](https://subversion.apache.org/) to download the specific directory:
+```bash
+svn export https://github.com/stellargraph/stellargraph/branches/master/demos
```
timpitman, line 85: wow, it feels like about 10 years since I've used subversion - amazing that this is the easiest way!

README.md
```diff
@@ -78,7 +78,14 @@ StellarGraph is built on [TensorFlow 2](https://tensorflow.org/) and its [Keras
...
+```
+
```
timpitman, line 87: For people without svn, we could also give instructions for curl, ie:
```
curl https://codeload.github.com/stellargraph/stellargraph/tar.gz/master | tar -xz --strip=1 stellargraph-master/demos
```

kjun9: Ah nice! I think this is much better actually, rather than "surprise" new users with an `svn` command, so I might just replace it

kjun9:
I've left a comment https://github.com/stellargraph/stellargraph/issues/1202#issuecomment-621002025 I'll make a follow-up PR

Issue #1338, opened 2020-04-23
Empty graph fails in PaddedGraphGenerator


huonw:
### Describe the bug

Passing in an empty graph (no nodes or edges) into the PaddedGraphGenerator results in an error like "k exceeds matrix dimension".

(StellarGraph 1.0.0rc1 dogfooding)

kieranricardo:
Currently no reproducable example

huonw:
The root cause of this is `sparse.diagonal()` fails if `sparse.shape == (0, 0)` (where `sparse` is a `scipy` sparse matrix https://docs.scipy.org/doc/scipy/reference/sparse.html). `StellarGraph.to_adjacency_matrix` hits this for an undirected graph, if the graph is empty or the `nodes` filtering list is empty.

PR #1380, opened 2020-04-27
Handle and test empty graphs in PaddedGraphGenerator

huonw:
This adds various special cases to handle completely empty graphs in `PaddedGraphGenerator`.

(It relies on #1381.)

See also #1395 that just disallows empty graphs like this.

See: #1338


huonw:
#1395 merged instead of this.

PR #1387, opened 2020-04-28
Update evaluation method

PantelisElinas:
This updates the demo to use repeated k-fold cross validation instead of a single k-fold cross validation with random re-starts and best model selection.

This simplifies the demo removing one of the loops (replaced by `RepeatedStratifiedKFold` that hides one of the loops) and there is no longer need to keep track of model performance and best model selection. The update also clarifies that it might run for a long time and instruct the user to update the `verbose` parameter to receive some feedback about training/validation process.

Results are essentially the same although there is a small decrease in estimated generalisation performance although I believe this is a more accurate estimate of how well the method performs.


demos/graph-classification/supervised-graph-classification.ipynb
```diff
@@ -93,7 +93,8 @@
     "from tensorflow.keras.layers import Dense\n",
```
huonw, line 93: <p>It's a bit peculiar to me that this "parametrises" the text by <code>n_repeats</code>,  but mentions <code>fold = 10</code>  literally. Maybe this could be:</p><p> </p><blockquote>for repeated <code>fold</code>-fold cross-validation; that is we do <code>fold</code>-fold validation&nbsp; <code>n_repeats</code>&nbsp;times [...] train and evaluate <code>fold</code>  graph classification models</blockquote><p> </p><p>Other options could be:</p><p> </p><blockquote>for repeated 10-fold cross-validation; that is we do 10-fold validation&nbsp;5&nbsp;times [...] train and evaluate 10 graph classification models</blockquote><p> </p><blockquote>for repeated 10-fold cross-validation (set by <code>fold = 10</code> ); that is we do 10-fold validation&nbsp;5&nbsp;times (set by <code>n_repeats = 5</code> ) [...] train and evaluate 10 graph classification models</blockquote><p> </p><p>This also applies to "Finally, we plot a histogram of the accuracy of all&nbsp;<code>n_repeats x 10</code>&nbsp;models trained" below.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1387/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

huonw, line 93: <p>Does it make sense to collect all the histories into a list and do an "ensemble" history plot? I think <code>sg.utils.plot_history</code>  can take a list of histories instead of a single one and plot them all together.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1387/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

huonw, line 93: <p>Maybe instead of having to update the 75.9 ± 6.3 number here every time the notebook is rerun (since this notebook is not reproducible), we could not be quite so specific, and just refer back to the place that computes the number. The paragraph starting with "Given the above ... " could be changed to something like</p><p> </p><blockquote>The performance results reported above likely have high variance because of the small dataset size.</blockquote>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1387/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

PantelisElinas: I'll have a look but the plot will look busy and noisy as there is much variance in performance based on the fold.

PantelisElinas: It's just that 10 folds is standard for cross validation but `n_repeats` is not so I think the user can set this to a value they find suitable especially if they need speed by setting it to a smaller value.

PantelisElinas: Yeah, as I thought plotting all the histories looks ugly and unhelpful. We best leave this bit for the DGCNN notebook that demonstrates performance on a single split of the data.

image

Issue #1360, opened 2020-04-24
Many sphinx warnings from demos


huonw:
### Describe the bug

Doing a sphinx build emits many warnings, a significant chunk of which are from the demos directory. (See #1154 for the general form that covers the non-demo warnings)

This is related to issues like: #1292, #1293, #1294, #1298 and #1305. Some or all of the warnings may be fixed by addressing those.

### To Reproduce

```shell
cd docs
pip install -r requirements.txt
make html
```

### Observed behavior

Many warnings from the `demos` directory (output filtered to just the ones containing `demos`):

```
.../docs/demos/basics/index.txt:4: WARNING: toctree glob pattern '*/index' didn't match any documents
.../docs/demos/basics/loading-networkx.nblink:1423: WARNING: File not found: 'demos/README.md'
.../docs/demos/basics/loading-pandas.nblink:4554: WARNING: File not found: 'demos/README.md'
.../docs/demos/basics/loading-saving-neo4j.nblink:2598: WARNING: File not found: 'demos/README.md'
.../docs/demos/basics/loading-saving-neo4j.nblink:2598: WARNING: File not found: 'demos/connector/neo4j/readme.md'
.../docs/demos/calibration/index.txt:4: WARNING: toctree glob pattern '*/index' didn't match any documents
.../docs/demos/community_detection/index.txt:4: WARNING: toctree glob pattern '*/index' didn't match any documents
.../docs/demos/connector/index.txt:4: WARNING: toctree glob pattern './*' didn't match any documents
.../docs/demos/connector/neo4j/index.txt:4: WARNING: toctree glob pattern '*/index' didn't match any documents
.../docs/demos/embeddings/deep-graph-infomax-cora.nblink:548: WARNING: Title level inconsistent:
.../docs/demos/embeddings/deep-graph-infomax-cora.nblink:597: WARNING: Title level inconsistent:
.../docs/demos/embeddings/index.txt:4: WARNING: toctree glob pattern '*/index' didn't match any documents
.../docs/demos/ensembles/index.txt:4: WARNING: toctree glob pattern '*/index' didn't match any documents
.../docs/demos/graph-classification/index.txt:4: WARNING: toctree glob pattern '*/index' didn't match any documents
.../docs/demos/graph-classification/supervised-graph-classification.nblink:219: WARNING: image file not readable: demos/graph-classification/graph_classification_architecture.png
.../docs/demos/index.txt:4: WARNING: toctree glob pattern './*' didn't match any documents
[...]

Issue #1298, opened 2020-04-21
Reuse demo README as the indices for the demos documentation in readthedocs


kjun9:
### Description

This potentially requires just rewriting the READMEs to be `rst` files instead of markdown, so that we can include a sphinx toctree as part of the README. 

Follow-up from #1279

Issue #1292, opened 2020-04-21
Demos in readthedocs have broken links


kjun9:
### Describe the bug

Follow-up from #1279

Notebooks were added to our readthedocs documentation by using `nbsphinx`, but seems like quite a number of links got broken. For example in `loading-saving-neo4j`:
![image.png](https://images.zenhubusercontent.com/5ad3f71c4b5806bc2bcf28ff/61807d53-1420-4d96-ae40-8afa1aee045e)

huonw:
I think there's two causes here:

- some of the links have nested markup, in particular code https://github.com/spatialaudio/nbsphinx/issues/301, e.g. https://stellargraph.readthedocs.io/en/v1.0.0rc1/demos/basics/loading-networkx.html#Loading-from-many-graph-formats,-via-NetworkX
  > A NetworkX graph can be created in many ways, including manually via `` `add_...`` methods <https://networkx.github.io/documentation/stable/tutorial.html#creating-a-graph>\`__, and by [reading from files in many formats](https://networkx.github.io/documentation/stable/reference/readwrite/index.html). In this tutorial, we’ll only use the first one.
- some files don't exist in the sphinx docs (yet), like the `../README.md` in the screenshot in the issue (#1298)

huonw:
Now that #1394 and #1401 have landed, I think most of these are fixed. Scrolling through some random notebooks, they all seem to look good.

Issue #1139, opened 2020-03-24
Some demo directories do not have README.md files


huonw:
### Describe the bug

Most of our demo subdirectories have README files that provide context (see also #1087). Some don't.

- [ ] [`connector`](https://github.com/stellargraph/stellargraph/tree/develop/demos/connector)
- [ ] [`graph-classification`](https://github.com/stellargraph/stellargraph/tree/develop/demos/graph-classification)
- [ ] [`interpretability`](https://github.com/stellargraph/stellargraph/tree/develop/demos/interpretability)
- [ ] [`use-cases`](https://github.com/stellargraph/stellargraph/tree/develop/demos/use-cases)

### To Reproduce

1. Run `find demos/ -maxdepth 2 | grep -v ipynb | grep -v '/.*/[a-z]' | sort` to see the list of all directories and their README files (if they have one)

### Observed behavior

```
demos
demos/README.md
demos/calibration
demos/calibration/README.md
demos/community_detection
demos/community_detection/README.md
demos/connector
demos/embeddings
demos/embeddings/README.md
demos/ensembles
demos/ensembles/README.md
demos/graph-classification
demos/interpretability
demos/link-prediction
demos/link-prediction/README.md
demos/node-classification
demos/node-classification/README.md
demos/use-cases
```
### Expected behavior

Every `demos/...` subdirectory should have a README

### Environment

Operating system: N/A

Python version: N/A

Package versions: StellarGraph develop at a7ef31acfa0863ff7bfb3e4152ecfb61392ff7d5

### Additional context

N/A

timpitman:
I think that only `use-cases` remains without an index, due to the work in #1200

huonw:
`connector` doesn't either (`connector/neo4j` does, though).

PR #1391, opened 2020-04-28
Convert demo READMEs to Sphinx index pages

huonw:
This moves us towards using Read the Docs as the main source of demos, by duplicating the indexing information from READMEs into the Sphinx docs.

We've decided to move to using Read the Docs like this because:

- it gives us more control over how things are shown (e.g. #1386)
- the display is faster
- it allows us to version properly, including with internal links to documentation of the same version as the demo

I used [m2r](https://miyakogi.github.io/m2r/index.html) to convert each of our demo READMEs to a reStructuredText `index.txt` file, and then fixed them up by hand. For each index, I've gone with:

```rst
Title
======

... description with subsections etc ...

Table of contents
------------------

... full ToC listing of subfolders/demos of this index ...
```

(Where the description is empty in some cases, for directories without READMEs: #1139.)

The thinking is we'll replace the README.md files with a much simplified set of information, potentially just a link to the corresponding Read the Docs. However, the READMEs are still useful as an index if someone has a copy of the demos locally in Jupyter lab. Thus, this PR also extends the `demo_table.py` script to render the demo indexing table to both a raw-HTML markdown version into `demos/README.md` and a rST `list-table` version into `docs/demos/index.txt`. This means we can still retain that demo table in an otherwise reduced/not-human-maintained `demos/README.md`.

Other than extending `demo_table.py`, this tries to do the minimal work to get this working. Future work:

- optimising the display of each of the tables (at the moment they scroll sideways a lot): #1418
- handling abbreviations in the main indexing table, because it doesn't seem easy to get hover text in rST: #1417
- validating links (likely covered by #1360)
- updating the demos to have relative links pointing to the index files, not README.mds (likely covered by #1292 and #1360)
- updating the README files to remove their content: #1419
- updating any links to the demos on GitHub to point to Read the Docs: #1420

Rendered demo landing page: https://stellargraph--1391.org.readthedocs.build/en/1391/demos/index.html

See: #1298


huonw:
> do we need longer names for the headings again?

Yeah, I'm going to open a follow up issue covering this and other bits from the PR description:

> - optimising the display of each of the tables (at the moment they scroll sideways a lot)
> - handling abbreviations in the main indexing table, because it doesn't seem easy to get hover text in rST
> - updating the README files to remove their content
> - updating any links to the demos on GitHub to point to Read the Docs

(The other future work there should be handled by other issues.)

huonw:
I opened #1417, #1418, #1419, #1420.

PR #1392, opened 2020-04-28
Add metapath2vec link prediction demo

kjun9:
WIP - this currently updates edge data and metapath2vec random walker quite a bit, which was required to get the demo to finish running within reasonable time frames. Still working on improving the performance of the model in the demo.

See #934


kjun9:
Closing this, I think I can open a new one independently of #1399 since random walks have generally gotten much faster since this began

PR #1395, opened 2020-04-29
Disallow empty graphs in PaddedGraphGenerator

huonw:
`PaddedGraphGenerator` is designed for graph classification, and an empty graphs is awkward and arguably meaningless to classify: 

- there's no input data at all, so nothing for a classification algorithm to work with
- all empty graphs are identical and all are classified the same
- mean pooling doesn't work (and having a mask of entirely `False`s is likely an edge case that other complicated pooling mechanisms don't explicitly consider/test for)

This PR thus resolves the problems `PaddedGraphGenerator` has with empty graphs by catching and disallowing them early. A user should then filter out any empty graphs before/while creating the generator.

This is future-proof: if it turns out to be useful to work with empty graphs here, we can always remove the restriction in future.

This is a simpler alternative to #1380. Is disallowing it like this PR the correct trade-off, instead of making it work like #1380 (which doesn't yet actually work)?

See: #1338

### DIFF
--- a/stellargraph/mapper/padded_graph_generator.py
+++ b/stellargraph/mapper/padded_graph_generator.py
@@ -46,6 +46,14 @@ class PaddedGraphGenerator(Generator):
                 raise TypeError(
                     f"graphs: expected every element to be a StellarGraph object, found {type(graph).__name__}."
                 )
+
+            if graph.number_of_nodes() == 0:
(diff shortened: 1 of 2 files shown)


kieranricardo:
I think this is probably the right tradeoff, I can't image there's any use for classifying an empty graph! And if there is, that could/should very easily be done with a rules based approach (`if graph is empty: class = 0` or something)

Issue #1349, opened 2020-04-24
Training DeepGraphInfomax on a full batch methods needs all nodes


huonw:
### Describe the bug

Using only a subset of the nodes (by doing anything other then `cor_gen.flow(graph.nodes())`) for training DGI on a full batch method fails due to dimensions not matching.

Approximate reproducer:

```python
fb_gen = FullBatchNodeGenerator(graph)
cor_gen = CorruptedGenerator(fb_gen)

base_model = GAT(...)
cor_model = DeepGraphInfomax(base_model, cor_gen)
x_in, x_out = dgi_model.in_out_tensors()
model = tf.keras.Model(inputs=x_in, outputs=x_out)
model.compile(loss=tf.nn.sigmoid_cross_entropy_with_logits, optimizer="Adam")

train_nodes, test_nodes = sklearn.model_selection.train_test_split(graph.nodes(), ...)

train_gen = cor_gen.flow(train_nodes)

model.train(train_gen)
```

This fails with an error like

```
InvalidArgumentError:  Incompatible shapes: [1,14294,2] vs. [1,17868,2]
	 [[node BroadcastGradientArgs_4 (defined at <ipython-input-125-fda1c863e1ff>:1) ]] [Op:__inference_distributed_function_10696]
```

Notably, 14294 happens to be the length of `train_nodes`, and 17868 happens to be the size of the full graph.

(StellarGraph 1.0.0rc1 dogfooding.)

PR #1415, opened 2020-04-30
DGI FullBatch methods fix

kieranricardo:
This PR fixes the bug that caused DGI to break on fullbatch when not all nodes were specified. The bug was cause by the fact that fullbatch generators pass in:

`(features, target_indicies, adj)`

And DGI infers the batch shape from the first input, `features` in this case, causing the wrong batch shape to be inferred for full batch methods. There are two fixes (that I can think of):

1. Specify which input to infer the batch shape from for each generator along with `num_batch_dims`
2. Take the minimum batch shape along all inputs

This PR makes the second change as a minimal fix - minimal code and test changes - to get things working quickly but might be too hacky. Thoughts?

See #1349


stellargraph/mapper/corrupted.py
```diff
@@ -171,7 +171,7 @@ def corrupt_group(group_idx, group):
         ]
 
         # create the appropriate labels
-        batch_size = inputs[0].shape[: self.num_batch_dims]
+        batch_size = min(inp.shape[: self.num_batch_dims] for inp in inputs)
```
huonw, line 174: Hm, so this works because it's picking up the output indices tensor with shape `(1, num_outputs)`, right?

Maybe we could expand the comment to discuss this briefly? E.g.

```python
# create the appropriate labels: this needs to match the number of outputs in the batch. 
# For full batch methods, this is the output indices tensor (batch shape = 1 × num_out), 
# not the features tensor (1 × num_nodes). Rather than hard-code, we assume that the
# smallest "batch shape" is the right one.
```

We could potentially also rename this like the following, to hammer home the point:

```suggestion
        output_batch_shape = min(inp.shape[: self.num_batch_dims] for inp in inputs)
```

kieranricardo: > Hm, so this works because it's picking up the output indices tensor with shape (1, num_outputs)

Yep that's right, and for SAGE algos all inputs have the same batch shape anyways so we're all good (for now...)

> Maybe we could expand the comment to discuss this briefly?

Yeah I think that's a could point, its a strange and subtle bit of code that will probably trip someone up in the future.

Issue #1370, opened 2020-04-27
DeepGraphConvolutionalNeuralNetwork class name doesn't match convention


huonw:
### Description

The  `DeepGraphConvolutionalNeuralNetwork` class name doesn't match the abbreviation convention we use for the rest of the library, such as `GCN` instead of `GraphConvolutionalNetwork`.

https://github.com/stellargraph/stellargraph/blob/ff62cc6355a7a17358dc29766ec3d3e935734745/stellargraph/layer/graph_classification.py#L186

It should probably be `DGCNN`, which both matches StellarGraph's convention and [the paper](https://www.cse.wustl.edu/~muhan/papers/AAAI_2018_DGCNN.pdf) (e.g. Section 2's title and Table 2).

PantelisElinas:
I suggest that renaming to DGCNN will be confusing as there is also the following method with the same acronym Dynamic Graph CNN,

see https://github.com/WangYueFt/dgcnn

huonw:
That's true. The "ConvolutionalNeuralNetwork" section of the name is rather long (especially the NeuralNetwork bit), as is `Graph`. Maybe `DeepGCNN` for our current one, and then the other can be `DynamicGCNN` if it gets implemented. #1421

PR #1422, opened 2020-04-30
Rename DeepGraphConvolutionalNeuralNetwork to DeepGraphCNN

huonw:
The [paper](https://www.cse.wustl.edu/~muhan/papers/AAAI_2018_DGCNN.pdf) that introduces our (currently-named) `DeepGraphConvolutionalNeuralNetwork` calls it "DGCNN". This abbreviation is nice and short and matches both the paper and the convention in the rest of the library (e.g. `GCN` instead of `GraphConvolutionalNetwork`). It also avoids redundancy like "Graph" in a library entirely designed around graphs, and expanding common terms-of-art like "Neural Network" and even ["CNN"](https://en.wikipedia.org/wiki/Convolutional_neural_network).

Unfortunately, the DGCNN acronym has a collision, with ["Dynamic Graph CNN"](https://github.com/WangYueFt/dgcnn). 

This PR renames the class to `DeepGraphCNN` to balance brevity and convention with specificity.

See: #1370


PantelisElinas:
I would suggest `DeepGraphCNN` because the `G` can also stand for `Generative`. Even though I am not aware of any methods call `Deep Generative CNN`, generative models are a hot topic in deep learning; I wouldn't be surprised if a method called like the above is introduced sooner than later.

huonw:
`DeepGraphCNN` sounds great given that reasoning 👍

PR #1437, opened 2020-05-01
increased atol and added pytest-repeat as test dependency

kieranricardo:
`test_complex` is flaky and sometimes fails on a floating comparison. This PR fixes this problem by increasing the `atol` from `1e-14` to `1e-6` - indicating that this was probably a numerical instability near zero issue.

This also adds in `pytest-repeat` as a test dependency to help diagnose this issue and similar ones in the future.

To verify the fix I ran:

```pytest tests -k test_complex --count 400```

Which initially failed four times. On this PR the above test pass.

See: #1025

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,6 @@ EXTRAS_REQUIRES = {
     "neo4j": ["py2neo"],
     "test": [
         "pytest==5.3.1",
-        "pytest-benchmark>=3.1",
         "pytest-cov>=2.6.0",
         "pytest-repeat>=0.8.0",
         "coverage>=4.4,<5.0",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by huonw** [comment]
> This also adds in pytest-repeat as a test dependency to help diagnose this issue and similar ones in the future.

I'd personally prefer for each of us to just `pip install` this locally as required, rather than add it on CI and so on. That said, it does that matter that much. What do you think?

**Comment by kieranricardo** [comment]
fair point, I didn't think about the extra CI time 👍

File: setup.py [hunk 1]
```diff
@@ -50,8 +50,8 @@
     "neo4j": ["py2neo"],
     "test": [
         "pytest==5.3.1",
-        "pytest-benchmark>=3.1",
         "pytest-cov>=2.6.0",
+        "pytest-repeat>=0.8.0",
```
huonw, line 54: I think you deleted the wrong thing 😄

kieranricardo: hm yeah i think we might still need benchmark 😂

PR #1438, opened 2020-05-01
Remove StellarGraphBase reference in FullBatchGenerator

### LINKED ISSUE #1436: FullBatchGenerator still references StellarGraphBase in the Docs
### Describe the bug

The docstring for both `FullBatchNodeGenerator` and `FullBatchLinkGenerator` reference `G`'s type to be a `StellarGraphBase`. This then appears in the ReadTheDocs documentation, making it confusing for the end-user.

readthedocs with StellarGraphBase

Only these 2 classes continue to reference `StellarGraphBase`; all others like UnsupervisedSampler, ClusterNodeGenerator etc. have transitioned to specifying StellarGraph as `type(G)`.

### To Reproduce

 Do a Ctrl+F search for `StellarGraphBase` on Read the Docs API page. Only 2 results, which correspond to the 2 aforementioned classes.

### Observed behavior

`type(G)` reads `StellarGraphBase`.

### Expected behavior

`type(G)` to read `StellarGraph` to be consistent with the rest of the documentation. Also, as far as I can see, `StellarGraphBase` has already been removed from the code base.

### Environment

StellarGraph develop HEAD 6da46b8

----

I can make a PR to fix this if the team agrees with the discussion above. Thanks.

thatlittleboy:
Resolves #1436 

This PR removes legacy StellarGraphBase reference as the type of G, a parameter of the `FullBatchNodeGenerator` and `FullBatchLinkGenerator` classes.

I also took the liberty to tidy up the docstring of these 2 classes a little bit. (missing punctuation)

### DIFF
--- a/stellargraph/mapper/full_batch_generators.py
+++ b/stellargraph/mapper/full_batch_generators.py
@@ -238,7 +238,7 @@ class FullBatchNodeGenerator(FullBatchGenerator):
     For more information, please see the GCN, GAT, PPNP/APPNP and SGC demos: `<https://github.com/stellargraph/stellargraph/blob/master/demos/>`_
 
     Args:
-        G (StellarGraphBase): a machine-learning StellarGraph-type graph
+        G (StellarGraph): a machine-learning StellarGraph-type graph
         name (str): an optional name of the generator
         method (str): Method to pre-process adjacency matrix. One of 'gcn' (default),
             'sgc', 'self_loops', or 'none'.
@@ -324,7 +324,7 @@ class FullBatchLinkGenerator(FullBatchGenerator):
     For more information, please see the GCN, GAT, PPNP/APPNP and SGC demos: `<https://github.com/stellargraph/stellargraph/blob/master/demos/>`_
 
     Args:
-        G (StellarGraphBase): a machine-learning StellarGraph-type graph
+        G (StellarGraph): a machine-learning StellarGraph-type graph
         name (str): an optional name of the generator
         method (str): Method to pre-process adjacency matrix. One of 'gcn' (default),
             'sgc', 'self_loops', or 'none'.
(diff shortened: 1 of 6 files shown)


stellargraph/mapper/full_batch_generators.py
```diff
@@ -157,8 +157,8 @@ def flow(self, node_ids, targets=None):
         Args:
             node_ids: an iterable of node ids for the nodes of interest
                 (e.g., training, validation, or test set nodes)
-            targets: a 1D or 2D array of numeric node targets with shape `(len(node_ids)`
-                or (len(node_ids), target_size)`
+            targets: a 1D or 2D array of numeric node targets with shape `(len(node_ids),)`
```
huonw, line 160: Thanks for noticing the bugs here.

Unfortunately this doc string was wrong originally in another way too: the doc strings are reStructuredText formatting, not Markdown, and so code/literal formatting has to be a double backtick (https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup):

```suggestion
            targets: a 1D or 2D array of numeric node targets with shape ``(len(node_ids),)``
```

and similarly on the next line and for both of the instances in `FullBatchLinkGenerator.flow`.

This can be observed by the *italics* rather than `fixed width` in https://stellargraph--1438.org.readthedocs.build/en/1438/api.html#stellargraph.mapper.FullBatchLinkGenerator.flow

thatlittleboy: Thanks -- sorry for missing the changes in `FullBatchNodeGenerator`. I've gone ahead and applied the double backticks to the rest of the classes that I could find (by searching for ``` `(len ```) in the code base.

huonw: Good idea. It looks like there's been quite a bit of copy paste 👍 

> sorry for missing the changes in FullBatchNodeGenerator

Not a problem at all 😄

huonw:
CI is fixed, so let's get this merged!

Issue #1441, opened 2020-05-02
Significant random walk performance regressions with ilocs


huonw:
### Describe the bug

Between 991db681bd435755 (#1267) and its parent f2a96aab34a21af, there was a significant performance regression in several random walk benchmarks:

- `UniformRandomWalk`
- `BiasedRandomWalk`/Node2Vec
- `BiasedWeightedRandomWalk`

This is in contrast to the dramatic improvements of some of the other algorithms, like:
- `BreadthFirstWalk`/GraphSAGE (1.2&times; speedup)
- `DirectedBreadthFirstWalk`/Directed GraphSAGE (1.1&times;)
- `SampledHeterogeneousBreadthFirstWalk`/HinSAGE (4&times;)
- `UniformRandomMetapathWalk`/Metapath2Vec (7.3&times;)

### To Reproduce

1. Go to each of the commits (`git checkout 991db681bd`) and run `pytest -k 'benchmark and walk' --benchmark-autosave`
2. Compare each of them with `pytest-benchmark compare`

#### UniformRandomWalk (2.5&times; slower)

```
                  Min                    Max                   Mean                StdDev                 Median                   IQR            Outliers         OPS            Rounds  Iterations
before     200.6821 (1.0)       1,620.4750 (1.87)        245.9843 (1.0)         51.2964 (1.0)         243.8300 (1.0)         38.8300 (1.0)       181;151  4,065.3007 (1.0)        2971           1
after      510.6980 (2.54)      2,051.6489 (2.36)        596.5518 (2.43)       104.0608 (2.03)        551.7174 (2.26)        99.1141 (2.55)       152;83  1,676.3004 (0.41)       1646           1
```

#### BiasedRandomWalk (2.1&times;)

```
                  Min                    Max                   Mean                StdDev                 Median                   IQR            Outliers         OPS            Rounds  Iterations
before     291.5789 (1.45)        868.2869 (1.0)         346.6916 (1.41)        63.5608 (1.24)        329.3765 (1.35)        60.2580 (1.55)      245;155  2,884.4083 (0.71)       2900           1
after      623.5219 (3.11)      1,661.3051 (1.91)        730.6545 (2.97)       111.8798 (2.18)        737.4294 (3.02)       125.8801 (3.24)       112;49  1,368.6359 (0.34)       1382           1
```

#### BiasedWeightedRandomWalk (9&times;)

```
[...]

PR #1444, opened 2020-05-03
Optimise weighted BiasedRandomWalk (4× speedup)

huonw:
The switch to using ilocs for edges and thus in the random walkers (#1267 991db681) resulted in some dramatic slowdowns in the `BiasedRandomWalk` code. This is mainly due to requiring more ID<->iloc conversions for some methods. Some of that overhead is unavoidable. This PR removes some of the avoidable overhead.

This increases the size of the `BiasedRandomWalk` benchmark, so that they're slightly more realistic without taking ridiculously long. In particular, it's rare that someone would do a random walk starting at a single node.

Given that, it then optimises the `BiasedRandomWalk` code, focusing on the weighted form, because of the large slowdown observed in the change to use ilocs (from f2a96aab to 991db681).

The following table reports the median times for the `` and `` benchmarks (as run by `pytest -k 'benchmark and biased'`), along with the speedups (vs. `develop`), and incremental speedups (vs. the previous row).

| change | unweighted (ms) | speedup (incr.) | weighted (ms) | speedup (incr.) |
|---|---|---|---|---|
| f2a96aab (pre-iloc), new benchmark | 8.58 | 2.0× (N/A) | 86.7 | 1.6× |
| `develop`, new benchmark | 17.5 | 1.0× (N/A) | 143 | 1.0× (N/A)  |
| numpy-ised weight validation | 17.5 | 1.0× (1.0×) | 75.0 | 1.9× (1.9×) |
| do `length` `neighbors` calls, not `length + 1` | 15.4 | 1.1× (1.1×) | 71.1 | 2.0× (1.05×) |
| move `def transition_probability` | 15.4 |  1.1× (1.0×) | 70.9 | 2.0× (1.0×) |
| use `include_edge_weights`, not `_edge_weights` | 15.4 |  1.1× (1.0×) | 36.8 | 3.9× (1.9×) |

That is, weighted random walks with the new form sees a speedup of almost 4× over the current state of `develop`, and 2.4× over the state before the iloc change. This recovers only a slight amount of the performance regression for unweighted walks.

Here's the numbers on a "real world" dataset (Cora), comparing this PR, with the state when ilocs were introduced and the state immediately before that. Measured using:

```python
import stellargraph as sg
rw = sg.data.BiasedRandomWalk(G, n=3, length=10, p=2, q=4)
%timeit -n 1 -r 3 rw.run(G.nodes())
%timeit -n 1 -r 3 rw.run(G.nodes(), weighted=True)
```

(NB. the weighted walks are equivalent to the unweighted ones, given they're all the default weight of 1, but it's still using the weighted code path.)

| code | unweighted (s) | weighted (s) |
|---|---|---|
| f2a96aab (pre-iloc)  | 1.36 ± 0.0233 | 11.7 ± 0.0724 |
| 991db681 (iloc) | 2.02 ± 0.0173 | 13 ± 0.427 |
| This PR | 2 ± 0.101 | 5.64 ± 0.179  |

The speedups/slowdowns are less dramatic here, because there's more real work, and so the constant overhead of the initial iloc conversion and the weight validation is less noticeable.

See: #1441

### DIFF
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -461,8 +461,7 @@ class BiasedRandomWalk(RandomWalk):
                     # select one of the neighbours using the
                     # appropriate transition probabilities
                     choice = naive_weighted_choices(
-                        rs,
-                        (transition_probability(nn) for nn in neighbours),
+                        rs, (transition_probability(nn) for nn in neighbours),
                     )
 
                     previous_node = current_node
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -34,7 +34,7 @@ from scipy.special import softmax
 from ..core.schema import GraphSchema
 from ..core.graph import StellarGraph
 from ..core.utils import is_real_iterable
-from ..core.validation import require_integer_in_range
+from ..core.validation import require_integer_in_range, comma_sep
 from ..random import random_state
 from abc import ABC, abstractmethod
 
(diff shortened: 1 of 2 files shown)


stellargraph/data/explorer.py
```diff
@@ -352,6 +352,32 @@ def __init__(
         self.p = p
         self.q = q
         self.weighted = weighted
+        self._checked_weights = False
+
+        if weighted:
+            self._check_weights_valid()
+
+    def _check_weights_valid(self):
```
kieranricardo, line 360: The develop version also checks that multi-edges have the same weight, is this missing here? (or did I miss it??)

stellargraph/data/explorer.py
```diff
@@ -386,32 +412,29 @@ def run(
...
-                        # multigraph with different weights on edges between same pair of nodes
-                        raise ValueError(
-                            f"{name}: expected all edges between two particular nodes to have the "
-                            f"same weight value, found {list(wts)}"
-                        )
+            self._check_weights_valid()
+
+            # calculate the appropriate unnormalised transition probability, given the history of
+            # the walk
+            def transition_probability(nn, current_node):
```
kieranricardo, line 419: `current_node` isn't used in this function

stellargraph/data/explorer.py
huonw: Ah, yes, it's a behaviour change. This now properly handles a multigraph, previously it would pretend it was a non-multigraph, by just getting the weight of the first edge between the source/target nodes. Now it handles each edge independently, effectively summing the weights.

It should be highlighted.

stellargraph/data/explorer.py
huonw: Ah, left over.

PR #1446, opened 2020-05-03
Return NumPy array in from_iloc

huonw:
Indexing Pandas types is slow, and indexing a `pandas.Index` is particularly slow, because it returns a whole `Index` object. This isn't generally useful for what we use `from_iloc` for, which is getting a sequence of node, edge or type IDs from some ilocs. The result is generally then used for things where having an `Index` isn't useful.

This makes the included microbenchmark of `from_iloc` (`test_benchmark_external_id_index_from_iloc`) significantly faster (times in microseconds):

| code | min | mean &pm; std | max | median &pm; IQR | 
|---|---|---|---|---|
| develop | 50.27 |  57.1 &pm; 9.7 | 224 | 52.9 &pm; 10 |
| this PR | 3.83 | 4.82 &pm; 3.1 | 64.6 | 4.25 &pm; 0.37 |

That is, a speed up of 12&times; going by the medians (ranging from 3.5-13&times; depending on which number one compares).

In #1267 991db681, random walks slowed down significantly compare to the previous commit (f2a96aab) due to more iloc conversions from storing edges as ilocs, see #1441.

In order to more accurately capture the effect of this change on those, the random walk benchmarks are increased to be slightly more realistic, with larger graphs and more head nodes. (#1444 does this change to the `BiasedRandomWalk` benchmarks, this does the rest.)

The `from_iloc` change thus has a significant impact on various benchmarks. The ones with the largest change are listed in the following table. The before/after columns are median times in microseconds &pm; IQR, and the speedup columns are comparing "after" with the specified before column.

| `test_benchmark_...` | before (pre-ilocs f2a96aab) | before (post-ilocs `develop`) | after | speedup vs. pre-ilocs | speedup vs. post-ilocs |
|---|---|---|---|---|--|
| `get_neighbours[False]`* | N/A | 16.1 &pm; 1.9 | 9.60 &pm; 1.3 | N/A | 1.7× |
| `biasedrandomwalk` | 7.79 &pm; 1.1 | 15.9 &pm; 2.1 | 8.12 &pm; 1.0 | 0.96× | 2.0× |
| `biasedweightedrandomwalk`^ | 83.7 &pm; 11 | 37.4 &pm; 5.7 | 29.2 &pm; 3.2 | 2.9× | 1.3× |
| `uniformrandommetapathwalk` | 235 &pm; 26 | 112 &pm; 11 | 39.4 &pm; 4.6 | 6.0× | 2.8× |
| `uniformrandomwalk` | 10.6 &pm; 1.3 | 31.8 &pm; 1.3 | 14.0 &pm; 1.4 | 0.76× | 2.3× |

\* `test_benchmark_get_neighbours[use_ilocs=False]` didn't exist before ilocs
^ `BiasedRandomWalk(..., weighted=True)` has independent optimisations applied in #1444, included here.

Some other benchmarks have smaller improvements (speedups up to ~1.1&times;), such as the GraphSAGE and HinSAGE random walkers.

In summary, this is a 1.3&times;-2.8&times; speedups on many random walk benchmarks, bringing the benchmarks with slowdowns in #1441 to be about what they were before ilocs were used for edges in #1267 991db681. It likely reduces memory use too, due to not needing the metadata of the `Index` type.

For a real world dataset, continuing the Cora table from #1444:

| code | unweighted (s) | weighted (s) |
|---|---|---|
| f2a96aab (pre-iloc)  | 1.36 ± 0.0233 | 11.7 ± 0.0724 |
| 991db681 (iloc) | 2.02 ± 0.0173 | 13 ± 0.427 |
| #1444 | 2 ± 0.101 | 5.64 ± 0.179  |
| This PR | 1.46 ± 0.0376 | 5.46 ± 0.0562 |

See: #1441

### DIFF
--- a/tests/core/test_element_data.py
+++ b/tests/core/test_element_data.py
@@ -46,9 +46,9 @@ def test_benchmark_external_id_index_from_iloc(benchmark):
     N = 1000
     SIZE = 100
     idx = ExternalIdIndex(np.arange(N))
+    x = np.random.randint(0, N, size=SIZE)
 
     def f():
-        x = np.random.randint(0, N, size=SIZE)
         idx.from_iloc(x)
 
     benchmark(f)
--- a/tests/core/test_element_data.py
+++ b/tests/core/test_element_data.py
@@ -42,18 +42,6 @@ def test_external_id_index_to_iloc(count, expected_missing):
     assert idx.to_iloc(["A"]) == expected_missing
 
 
-@pytest.mark.parametrize(
-    "dtype", [np.uint8, np.int, np.float32, np.float64, str, object]
-)
-def test_external_id_index_numpy_cache(dtype):
-    values = np.arange(10).astype(dtype)
-    idx = ExternalIdIndex(values)
-
-    # approximate validation that the cache of the underlying numpy array doesn't result in higher
-    # memory use: repeated to_numpy() calls should be all using the same memory
-    assert np.shares_memory(idx._index.to_numpy(), idx._index_numpy_cache)
-
-
 def test_benchmark_external_id_index_from_iloc(benchmark):
     N = 1000
     SIZE = 100
--- a/tests/core/test_element_data.py
+++ b/tests/core/test_element_data.py
@@ -40,3 +40,27 @@ def test_external_id_index_to_iloc(count, expected_missing):
 
     # missing value
     assert idx.to_iloc(["A"]) == expected_missing
+
+
+@pytest.mark.parametrize(
+    "dtype", [np.uint8, np.int, np.float32, np.float64, str, object]
+)
+def test_external_id_index_numpy_cache(dtype):
+    values = np.arange(10).astype(dtype)
+    idx = ExternalIdIndex(values)
+
+    # approximate validation that the cache of the underlying numpy array doesn't result in higher
+    # memory use: repeated to_numpy() calls should be all using the same memory
+    assert np.shares_memory(idx._index.to_numpy(), idx._index_numpy_cache)
+
+
+def test_benchmark_external_id_index_from_iloc(benchmark):
+    N = 1000
+    SIZE = 100
+    idx = ExternalIdIndex(np.arange(N))
+
+    def f():
+        x = np.random.randint(0, N, size=SIZE)
+        idx.from_iloc(x)
+
+    benchmark(f)
--- a/tests/data/test_metapath_walker.py
+++ b/tests/data/test_metapath_walker.py
@@ -19,6 +19,7 @@ import numpy as np
 import pytest
 from stellargraph.data.explorer import UniformRandomMetaPathWalk
 from stellargraph.core.graph import StellarGraph
+from ..test_utils.graphs import example_graph_random
 
 
 # FIXME (#535): Consider using graph fixtures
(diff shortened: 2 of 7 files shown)


tests/data/test_metapath_walker.py
```diff
@@ -316,13 +317,17 @@ def test_init_parameters(self):
...
-        nodes = ["0"]
+        # this should be made larger to be more realistic, when it is fast enough
+        nodes = np.arange(0, 5)
         n = 5
         length = 5
-        metapaths = [["s", "n", "n", "s"], ["n", "s", "n"], ["n", "n"]]
+        metapaths = [
+            ["n-0", "n-1", "n-1", "n-0"],
+            ["n-0", "n-1", "n-0"],
+            ["n-0", "n-0"],
```
kjun9, line 330: Does `example_graph_random` guarantee that edge types do exist that traverse the graph this way, or is it relying on `n_edges` to be large enough that the chance is very low that we miss one? I think otherwise the random walks end up stopping early

huonw: Hm, you're correct that it doesn't guarantee it, but I think it's rather likely across, with 500 total edges and 100 nodes across 2 node types.

Of course, this has to be qualified by only having 5 head nodes. The average degree of each node in the graph is 10, so there's ~50 edges touching the 5 starting `n-0` starting nodes.

Using the fact that the four possibilities of source/target types for each edge are all equally likely (`1/4`), I think the probabilities of having none of the three paths is low:

- `n-0`--`n-0`: is the chance of every edge having source or target `n-1`, which is `(3/4)**50` &approx; 5.7e-7.
- `n-0 -- n-1 -- n-0`: is the chance of no `n-0 -- n-1` edges (since if there's one, it can follow that edge forwards and backwards), which `(1/2)**50` &approx; 8.9e-16
- `n-0 -- n-1 -- n-1 -- n-0`: is likely going to be larger and seems more complicated to calculate (what's the chance of every `n-0 -- n-1` going to an `n-1` with no edges to other `n-1`), I guess I can work it out, but I don't think it's too important because it's probably also small?

One approach to be more sure would be increasing the `np.arange(0, 5)` to `np.arange(0, 10)` or something?

Thoughts?

kjun9: The numbers sound small enough it's probably fine to leave as is 👍 thanks for checking!

tests/core/test_element_data.py
```diff
@@ -40,3 +40,15 @@ def test_external_id_index_to_iloc(count, expected_missing):
...
     assert idx.to_iloc(["A"]) == expected_missing
+
+
+def test_benchmark_external_id_index_from_iloc(benchmark):
+    N = 1000
+    SIZE = 100
+    idx = ExternalIdIndex(np.arange(N))
+
+    def f():
+        x = np.random.randint(0, N, size=SIZE)
```
kieranricardo, line 51: i think the overhead of this is likely to be similar to `from_iloc` and probably under-estimates the speedup from this PR!

I think a test similar to https://github.com/stellargraph/stellargraph/pull/1267#discussion_r416327081 might be appropriate here. but imo this isn't a blocking concern

huonw: Err, great catch. The overhead of `randint` is actually dramatically more than `from_iloc`, for 

```python
x = np.random.randint(0, N, size=SIZE)
def f():
    idx.from_iloc(x)
```

The median &pm; iqr is 4.245 &pm; 0.369, i.e. almost 80% of the time is spend doing `randint`. I'm going to go with the simpler form here because the between-run variance doesn't seem too high... I guess we might be able to get away with the simple form in the `StellarGraph` benchmarks too (although we're more likely to have some sort of adaptive algorithms there, e.g. handling very high degree nodes or commonly query nodes).

To validate this, I ran `test_benchmark_external_id_index_from_iloc` several times



```
   Min       Max    Mean  StdDev  Median     IQR  Outliers  OPS (Kops/s)  Rounds  Iterations     
3.9730   76.9699  4.2392  1.6609  4.1041  0.0510  208;1425      235.8910   22239           1
3.7202   52.2460  4.0642  1.1524  4.0079  0.1050  173;619       246.0481   28684           1
3.7029   48.3089  4.1099  1.3056  4.0610  0.0831  145;4052      243.3169   24395           1
3.7120  105.8390  4.1295  2.0666  3.9020  0.1900  275;531       242.1584   27607           1
3.6780   61.1350  4.2788  0.9586  4.3069  0.1958  80;2290       233.7111   25250           1
3.7560   66.0201  4.9839  4.0769  4.1490  0.1751  614;1808      200.6445   16076           1
3.8252   64.5600  4.8152  3.1023  4.2450  0.3690  800;1118      207.6745   25286           1
```

That is a variation of about 10%, which doesn't seem much larger than the variation between runs even without this.

kieranricardo: woah, so we're looking at a ~50x speedup now?

huonw: No, 12&times; (~50ms -> ~4ms).

Issue #1420, opened 2020-04-30
Links to demos should point to Read the Docs


huonw:
### Describe the bug

We've got a few links that go to our demos on github. After #1391, these should be updated to point to Read the Docs.

```
$ git grep '(.*demos/'
CHANGELOG.md:- [More helpful indexing and guidance in demo READMEs](demos/)
CHANGELOG.md:  - The [demos READMEs](demos/) now contain more guidance and explanation to make it easier to find a relevant example [\#1200](https://github.com/stellargraph/stellargraph/pull/1200)
CHANGELOG.md:  - Notebooks are [now included in the API documentation](https://stellargraph.readthedocs.io/en/latest/demos/index.html), for more convenient access [\#1279](https://github.com/stellargraph/stellargraph/pull/1279)
CHANGELOG.md:  - [Example Jupyter notebooks](https://github.com/stellargraph/stellargraph/tree/master/demos/) can now be run directly in [Google Colab](https://colab.research.google.com) and [Binder](https://mybinder.org), providing an easy way to get started with StellarGraph - simply click the ![](https://colab.research.google.com/assets/colab-badge.svg) and ![](https://mybinder.org/badge_logo.svg) badges within each notebook. [\#1119](https://github.com/stellargraph/stellargraph/pull/1119).
CHANGELOG.md:  - The [new `demos/basics` directory](demos/basics) contains two notebooks demonstrating how to construct a `StellarGraph` object from Pandas, and from NetworkX [\#1074](https://github.com/stellargraph/stellargraph/pull/1074)
CHANGELOG.md:  - The [GCN node classification demo](demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb) now has more explanation, to serve as an introduction to graph machine learning using StellarGraph [\#1125](https://github.com/stellargraph/stellargraph/pull/1125)
README.md:[The numerous detailed and narrated examples](https://github.com/stellargraph/stellargraph/tree/master/demos/) are a good way to get started with StellarGraph. There is likely to be one that is similar to your data or your problem (if not, [let us know](#getting-help)).
README.md:- [Consult the examples](https://github.com/stellargraph/stellargraph/tree/master/demos/)
[...]

Issue #1402, opened 2020-04-29
Links to documentation in Read the Docs demo notebooks should stay within the same doc version


huonw:
### Description

Several of our notebooks have links to our documentation, such as 

https://github.com/stellargraph/stellargraph/blob/625a99b798ef73957426c74066405457ff31174a/demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb#L125

```markdown
[the `Cora` loader](https://stellargraph.readthedocs.io/en/stable/api.html#stellargraph.datasets.datasets.Cora)
```

When published on Read the Docs (https://stellargraph.readthedocs.io/en/latest/demos/node-classification/gcn/gcn-cora-node-classification-example.html#Loading-the-CORA-network, although the link is broken at the time of writing, fixed by #1394), this link should point to the same version of the Read the Docs. Specifically, when viewing the `develop` docs (`.../latest/...`), the link should stay on the `develop` docs, but currently it always goes to `.../stable/...`.

Once we've done several stable releases with notebooks, like this, it'll ensure that looking at 1.0.0 notebooks always links to 1.0.0 documentation, and similarly for 1.1.0, etc.

PR #1448, opened 2020-05-03
Point to Read the Docs, not Github, for demos

huonw:
This updates any links to our Github demos folder to instead point to Read the Docs. 

The links within documentation _could_ be ``:doc:`... <demos/index>` `` relative links (or similar), but these will not work outside of Sphinx/Read the Docs (e.g. in `help(ClusterNodeGenerator)`). As such, this instead uses the full URLs always. To get the benefits of relative links (e.g. consistent versioning, and working locally without jumping out to the internet), we can post-process them in future (#1402 #1404).

See: #1420

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,9 @@ StellarGraph is built on [TensorFlow 2](https://tensorflow.org/) and its [Keras 
 
 ## Getting Started
 
-[The numerous detailed and narrated examples](https://github.com/stellargraph/stellargraph/tree/master/demos/) are a good way to get started with StellarGraph. There is likely to be one that is similar to your data or your problem (if not, [let us know](#getting-help)).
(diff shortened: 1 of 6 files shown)


huonw:
One detail here is that these links don't actually work, until the release happens, because the current `stable` on RtD is 0.11.1 (the 1.0.0rc1 doesn't count as stable, according to RtD).

kjun9:
> One detail here is that these links don't actually work, until the release happens, because the current `stable` on RtD is 0.11.1 (the 1.0.0rc1 doesn't count as stable, according to RtD).

Hmm right. I guess this could be done in the release branch if we care about the links not being broken for a day? Any new person landing on github is probably very likely to try click the `demos` link in the main readme 🤔

Issue #1382, opened 2020-04-27
Docker image 'latest' tag is develop, not master


huonw:
### Describe the bug

The `stellargraph/stellargraph` docker image (https://hub.docker.com/repository/docker/stellargraph/stellargraph) is published from the `develop` branch to `latest`, meaning there's no releases for our stable `master` releases, and `docker run stellargraph/stellargraph:latest` uses the `develop` branch, not one of our releases. 

If we're publishing docker images, we should make sure we emphasise the stable ones more.

This is coordinated by:

https://github.com/stellargraph/stellargraph/blob/492c4e30d962a9e4ce73d37924ef6e4b1bb597ea/.buildkite/hooks/pre-command#L4-L16

https://github.com/stellargraph/stellargraph/blob/492c4e30d962a9e4ce73d37924ef6e4b1bb597ea/.buildkite/pipeline.yml#L241-L248

### Environment

StellarGraph 492c4e30d962a9e4ce73d37924ef6e4b1bb597ea

### Additional context

N/A

timpitman:
Does anyone even use our docker images?

huonw:
I think our CI will do at most two pulls per build: 

- one for the cache when building
- one for the push step, to get the image and repush it

https://github.com/stellargraph/stellargraph/blob/1cca6b6c6f4cd1fea7c6932918cba18d15eef6d4/.buildkite/pipeline.yml#L232-L260

| pipeline | builds |
|---|---|
| `stellargraph-public` | 3545 |
| `stellargraph` (now unused) | 1866 |
| **Total** | **5411** |

This implies that at most 10822 of the pulls are from our CI. I don't _think_ any of our other pipelines use this docker image. I don't know if 15k pulls is a lot.

Removing the image would certainly be one way to resolve this issue.

That said, now that we have `sg.datasets`, it could be an easy way for a docker-familiar user to get a Jupyter Lab environment started and running the demos.

yvonnegh:
If target market is data scientists, not sure if anyone is using our Docker image.
Metrics from docker hub are not useful to see what's happening.

Could discuss whether we stop publishing Docker image and then do it properly.

PR #1455, opened 2020-05-04
Stop publishing the docker image

huonw:
Per #1382, our docker image is currently only published from `develop`, meaning there's no stable releases. We don't particularly want our users using random points on `develop`, because this doesn't have the bug fixing and polish that happens just before each release.

Rather than try to get stable images published for the upcoming 1.0 release tomorrow, this just disables them for now.

Thoughts?

See: #1382


timpitman:
Happy to just see the docker image removed for now!

Issue #1337, opened 2020-04-23
PaddedGraphGenerator.flow could detect common error of list of StellarGraph


huonw:
### Describe the bug

PaddedGraphGenerator.flow expects a list/sequence of integers (of indices into the list of graphs passed to PaddedGraphGenerator). Users are likely to make a mistake and pass in a list of graphs themselves, because it's easy to do:

```python
train_graphs, test_graphs = model_selection.train_test_split(graphs, ...)
train_gen = generator.flow(train_graphs)
```

The correct thing is to do:

```python
train_graphs, test_graphs = model_selection.train_test_split(range(len(graphs)), ...)
train_gen = generator.flow(train_graphs)
```

When we detect a type error, we could check if it's a `StellarGraph` and give a more specific error.

(StellarGraph 1.0.0rc1 dogfooding)

kieranricardo:
Potentiall we could resolve this by making the generator accept a list of `StellarGraph`s

yvonnegh:
Options: Make a better error or not an error
Probably leaning towards not an error.

PR #1458, opened 2020-05-04
PaddedGraphGenerator.flow now also accepts a list of StellarGraphs

kieranricardo:
This PR enables PaddedGraphGenerator.flow to also accepts a list of StellarGraphs.

See: #1337

### DIFF
--- a/stellargraph/mapper/padded_graph_generator.py
+++ b/stellargraph/mapper/padded_graph_generator.py
@@ -91,8 +91,8 @@ class PaddedGraphGenerator(Generator):
         with the supplied graph indexes and targets.
 
         Args:
-            graph_ilocs (iterable): an iterable of graph indexes in self.graphs for the graphs of interest
-                (e.g., training, validation, or test set nodes).
+            graph_ilocs (iterable): an iterable of graph indexes in self.graphs or an iterable of `StellarGraph`s
+                for the graphs of interest (e.g., training, validation, or test set nodes).
             targets (2d array, optional): a 2D array of numeric graph targets with shape ``(len(graph_ilocs),
                 len(targets))``.
             symmetric_normalization (bool, optional): The type of normalization to be applied on the graph adjacency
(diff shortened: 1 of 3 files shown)


stellargraph/mapper/padded_graph_generator.py
```diff
@@ -131,8 +132,13 @@ def flow(
                 f"expected batch_size to be strictly positive integer, found {batch_size}"
             )
 
+        if isinstance(graph_ilocs[0], StellarGraph):
```
huonw, line 135: I think we need to do a lot more validation here; e.g. this allows a user to pass in an empty graph (#1338) or a graph without features or a heterogeneous graph.

One way to do this would be to factor out the loop (or just the loop body) from `__init__` and call it both there and here.

Given there's a variety of edge cases, I wonder if we should do the easy thing (emitting a better error) for 1.0 and do this more flexible version for 1.1?

kieranricardo: good point, I had a quick go at extending existing edge case tests to handle `.floaw` and I think they're all covered

stellargraph/mapper/padded_graph_generator.py
```diff
@@ -91,8 +95,8 @@ def flow(
         with the supplied graph indexes and targets.
 
         Args:
-            graph_ilocs (iterable): an iterable of graph indexes in self.graphs for the graphs of interest
-                (e.g., training, validation, or test set nodes).
+            graph_ilocs (iterable): an iterable of graph indexes in self.graphs or an iterable of `StellarGraph`s
```
huonw, line 98: Tiny thing, but we should consider renaming this parameter to `graphs`, so the error messages from `self._check_graphs` match.

```suggestion
            graphs (iterable): an iterable of graph indexes in self.graphs or an iterable of `StellarGraph`s
```

and all the other instances of `graph_ilocs` in this file.

kieranricardo: This breaks code specifying `graph_ilcs` as a kwarg but `PaddedGraphGenerator` wasn't in the last release so I think this should be fine.

Issue #1390, opened 2020-04-28
Add download links to demo notebooks on readthedocs


timpitman:
### Description

When looking at our demo notebooks inside documentation on readthedocs (added for #1279), it would be useful to be able to directly download the notebook.

As an example, see scikit-learn, eg: https://scikit-learn.org/stable/auto_examples/plot_changed_only_pprint_parameter.html#sphx-glr-auto-examples-plot-changed-only-pprint-parameter-py

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

timpitman:
We're using nbsphinx plugin to embed notebooks inside docs, and it doesn't appear to support this.

The example above from scikit-learn uses a different plugin https://sphinx-gallery.github.io/stable/index.html which actually bundles files for download, but doesn't support Jupiter notebooks as input.

kjun9:
Another approach might be to just include a link in the notebook that points to itself on github (i.e. update format notebook script to put in the link) and the user can do a "save as" on that?

timpitman:
Adding a link to the notebook itself is a good idea, however I had a look and one problem is that GitHub doesn't allow direct download - clicking such a link opens the notebook source in a browser rather than just downloading it, eg: https://raw.githubusercontent.com/stellargraph/stellargraph/master/demos/basics/loading-networkx.ipynb

kjun9:
Yeah I think we can't get around that unless we also host all our demos somewhere other than github. Maybe something like below is not terrible?:

*View the notebook or download via the "Save Link as..." option:* [loading-networkx.ipynb](https://raw.githubusercontent.com/stellargraph/stellargraph/master/demos/basics/loading-networkx.ipynb)

huonw:
I think the `ipynb` files are already included (maybe by nbsphinx or maybe by nbsphinx-link?) in the Sphinx output; for instance, https://stellargraph.readthedocs.io/en/latest/demos/basics/loading-networkx.ipynb looks correct to me.

We can then probably use use `nbsphinx_prolog` to insert a download button (https://nbsphinx.readthedocs.io/en/0.6.1/prolog-and-epilog.html). Something involving `docname`, I guess.

[...]

PR #1459, opened 2020-05-04
Update download link and include in demo readme

kjun9:
This adds the demo download instructions to the demo readme (both on github and readthedocs)

Follow-up to address #1202 raised in https://github.com/stellargraph/stellargraph/pull/1377#pullrequestreview-400801873

Download link per notebook still needs to be done #1390 (#1460 )

readthedocs link: https://stellargraph--1459.org.readthedocs.build/en/1459/demos/index.html#download-the-demos

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -45,7 +45,7 @@ The demo notebooks can be run without any installation of Python by using Binder
     <td>see RGCN</td>
     <td></td>
     <td></td>
-    <td>see T-GCN</td>
+    <td></td>
     <td>yes</td>
     <td><a href='node-classification/gcn/gcn-cora-node-classification-example.ipynb'>demo</a></td>
     <td><a href='link-prediction/gcn/cora-gcn-links-example.ipynb'>demo</a></td>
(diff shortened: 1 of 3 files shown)


docs/demos/index.txt
```diff
@@ -260,6 +260,21 @@ Find a demo for an algorithm
 
 See :doc:`the root README <../README>` or each algorithm's documentation for the relevant citation(s).
 
+
+Download the demos
+------------------
+
+You can run download a local copy of the demos using the :code:`curl` command below:
```
huonw, line 267: No change required, but for future reference, I think code/literal formatting can be more conveniently expressed with just double backticks?

```suggestion
You can run download a local copy of the demos using the ``curl`` command below:
```

(I guess it's slightly different, in that the `:code:` role ones end up with the `code` CSS class in addition to `docutils literal notranslate`. 🤷‍♂️ )

demos/README.md
```diff
@@ -280,3 +280,16 @@ The demo notebooks can be run without any installation of Python by using Binder
 <!-- DEMO TABLE MARKER -->
 
 See [the root README](../README.md) or each algorithm's documentation for the relevant citation(s).
+
+## Download the demos
```
huonw, line 284: FWIW, https://github.com/stellargraph/stellargraph/pull/1433 removes content like this from this README. Do you think this is still valuable to include even with a switch to RtD for most things? (#1433 is designed to allow custom content like this.)

timpitman: I think it's still useful to know how to download all of the demos; downloading each one separately from readthedocs is less convenient in many cases.

kjun9: Ah I see, if we're linking to RTD, I don't think "downloading demos" is a particularly important custom content to add. Should I remove it from this PR?

docs/demos/index.txt
kjun9: Ah thanks, yeah I'm still having to google every little thing I wanna do in rst

demos/README.md
kjun9: > I think it's still useful to know how to download all of the demos

It'll be included in the readthedocs `demos/index.rst` still, but the question is around whether we also want to include it in `demos/README` right?

timpitman: Aaah, I see.  After looking at the output, I think just the base README is enough.

huonw: > is that being fixed in a different PR @huonw?

Yup: https://github.com/stellargraph/stellargraph/pull/1448 (https://stellargraph--1448.org.readthedocs.build/en/1448/README.html#getting-started)

PR #1460, opened 2020-05-04
Add download links to demo notebooks on readthedocs (#1390)

timpitman:
Adding links to our html documentation to download the notebook source files - there's now a "Download this notebook" button at the top and bottom of each demo.

Our use of `.nblink` files for nbsphinx makes this a little more complicated than expected.

See: #1390

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -196,7 +196,7 @@ nbsphinx_prolog = r"""
 .. raw:: html
 
     <div>
-      <p><a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn btn-primary">Download this notebook</a></p>
+      <p><a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn">Download this notebook</a></p>
     </div>
 """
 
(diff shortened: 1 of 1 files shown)


timpitman:
> It's not obvious to me what part of the code is caused by nblink files? It looks like vaguely like what I would expect for having plain `ipynb`s directly in the `docs` directory?

The examples at https://nbsphinx.readthedocs.io/en/0.6.1/conf.py use a function `env.doc2path` to resolve document name to (somewhat relative) path, however that resolves to our .nblink paths.

huonw:
As in, the paths stored the `nblink` files, pointing to the top-level `demos/` directory, rather than the paths to the `nblink` files?

Also, standard reminder that an issue number in a PR title isn't auto linked by GitHub. Without this, there's no public evidence of progress for someone without zenhub looking at the issue (https://github.com/stellargraph/stellargraph/issues/1390#ref-pullrequest-611574617, the "37 seconds ago" in the screenshot is from when I just edited the reference in):

image

timpitman:
> As in, the paths stored the `nblink` files, pointing to the top-level `demos/` directory, rather than the paths to the `nblink` files?

it gives the path of the .nblink file; their examples assume that's the path of the .ipynb file.

huonw:
I think that's what we want? The path of the nblink file is the eventual location of the ipynb file?

Issue #1418, opened 2020-04-30
Read the Docs demo tables involve a lot of sideways scrolling


huonw:
### Describe the bug

In #1391, we add tables indexing our demos to the Read the Docs pages. These tables are wide and involve a lot of sideways scrolling. We should consider if we can avoid this.

huonw:
We can potentially use CSS to make the Read the Docs column wider.

PR #1461, opened 2020-05-04
Add custom CSS for cell wrapping and no-scrolling in demo table

huonw:
This enables word-wrapping in the demo tables (in all tables, in fact) so that they don't take up so much horizontal room. In addition, it allows the table to overflow the right margin of the docs, so that there's no scrolling required on large screens. (This tries to use a media query, so that only the table scrolls on a small/narrow screen, not the whole page.)

(This PR is a reopening of #1456 to resolves some weirdness I saw with `custom.css` not updating on RtD.)

See: #1418


huonw:
This landed in #1472.

PR #1399, opened 2020-04-29
Make metapath walk faster by grouping adjacency lists by node type

kjun9:
This mainly updates `EdgeData` to store an inner set of dicts within each adjacency list, so that we group adjacency elements by their node type. This could probably be useful for other heterogeneous algorithms that want to traverse the graph either for random walks or SAGE-type sampling - currently it's just been used for Metapath2Vec specifically. At its current state in this PR, the speedup for Metapath2vec looks pretty great, but `graph.neighbours` becomes quite a bit slower for homogeneous graphs

Benchmark against `develop` for metapath walk:
```
----------------------------------------------------------------------------------------------------------------- benchmark: 2 tests -----------------------------------------------------------------------------------------------------------------
Name (time in us)                                                   Min                    Max                   Mean                StdDev                 Median                   IQR            Outliers         OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_uniformrandommetapathwalk (NOW)                 409.0180 (1.0)       1,415.0450 (1.0)         508.5108 (1.0)         89.4222 (1.0)         480.2570 (1.0)         38.4945 (1.0)       126;171  1,966.5264 (1.0)        1349           1
test_benchmark_uniformrandommetapathwalk (0007_1cca6b6)     11,785.8720 (28.82)    20,027.5200 (14.15)    15,493.4845 (30.47)    1,442.1304 (16.13)    15,570.4430 (32.42)    1,697.6358 (44.10)        15;3     64.5433 (0.03)         61           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Things do get a bit slower for just getting neighbours without caring about types, and constructing the stellargraph object
```
---------------------------------------------------------------------------------------- benchmark 'StellarGraph neighbours': 2 tests ----------------------------------------------------------------------------------------
Name (time in us)                                     Min                   Max                  Mean              StdDev              Median                IQR            Outliers         OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_get_neighbours (0009_1cca6b6)     673.0350 (1.0)      1,753.2520 (1.0)        772.2845 (1.0)      112.2834 (1.0)      738.3375 (1.0)      46.6965 (1.0)       120;151  1,294.8595 (1.0)        1244           1
[...]

PR #1462, opened 2020-05-04
Remove utils directory

kjun9:
Still working out https://github.com/stellargraph/stellargraph/pull/1399 and https://github.com/stellargraph/stellargraph/pull/1392, but I think it might be best just remove the `utils` directory pre-emptively for 1.0 to make our demos look cleaner. 

This will be deleting:
* `demos/link-prediction/random-walks/README` which is outdated and won't be needed with the new structure in RTD?
* Node2Vec link prediction on Cora (already covered by the demo notebook that replaced the script)
* Metapath2Vec link prediction on BlogCatalog3 (should eventually be covered by https://github.com/stellargraph/stellargraph/pull/1392)

Part of #1466


huonw:
This helps with #1471 too. 👍

Issue #1330, opened 2020-04-23
Loading code in demos should link back to "here's how to load"


huonw:
### Describe the bug

The dataset module makes our demos cleaner, but it does make it harder for a user to work out how to load their own data. We should link to the demos for loading data to clarify this.

(StellarGraph 1.0.0rc1 dogfooding)

yvonnegh:
Hidden loading code in datasets class, need to point back to it.
Could be automated.
More notebook operations...

Is important - was consistent in dogfooding that users didn't know how to load data and had to work it out.

Could do manually in each notebook but preference is to do automatically.
Maybe when dataset load is called, that can print output back to notebook formatted with link to other notebooks to explain how to load data, and readTheDocs for loading function.
or
Tag loading cell and markdown cell before it with a link or description to appropriate destination.

PR #1464, opened 2020-05-04
Link to Loading from Pandas demo in notebooks that use datasets module

huonw:
This inserts a cell with a link to the Pandas loading demo before any cells tagged `DataLoading`. This PR thus also adds a whole pile of those tags (semi-automatically: searching for cells that contain both `datasets` & `load`, and other similar queries).

Example: https://stellargraph--1464.org.readthedocs.build/en/1464/demos/node-classification/gcn/gcn-cora-node-classification-example.html#Loading-the-CORA-network

See: #1330

### DIFF
--- a/demos/calibration/calibration-pubmed-link-prediction.ipynb
+++ b/demos/calibration/calibration-pubmed-link-prediction.ipynb
@@ -155,7 +155,11 @@
   {
    "cell_type": "code",
    "execution_count": 5,
-   "metadata": {},
+   "metadata": {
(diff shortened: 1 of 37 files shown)


huonw:
> I was wondering if the search could be automated in the script, but this looks good.

I considered this but decided that it's not too much of a burden to maintain manually. Because this is documentation, it seems to me that the risk of a random cell having `datasets` ... `load` in it and having this text automatically inserted in a confusing place is worse.

PR #1398, opened 2020-04-29
Switch Binder and Colab links in notebooks to markdown block quote

huonw:
The raw HTML table was not converting to reStructuredText through `nbsphinx` (which, I think, uses `jupyter nbconvert --to=rst`, which uses pandoc). The output lost the links + images (see more details about `pandoc` directly at the end of this PR).

This PR updates the links to be normal markdown links + images, which ensures they do get converted correctly. To distinguish these links from normal body text, these are formatted as a block-quote (`> ...`). This gives acceptable (not great) formatting on each of the major platforms one might be using to look at this:

- Jupyter locally:  image
- [Github](https://github.com/stellargraph/stellargraph/blob/2972637ec3b75c7ab7bc6a7a96697489b03f62fd/demos/basics/loading-networkx.ipynb) image
- [nbviewer](https://nbviewer.jupyter.org/github/stellargraph/stellargraph/blob/2972637ec3b75c7ab7bc6a7a96697489b03f62fd/demos/basics/loading-networkx.ipynb) image
- [Read the Docs](https://stellargraph--1398.org.readthedocs.build/en/1398/demos/basics/loading-networkx.html)
image

Note that both Github and nbviewer have CSS that forces images to be formatted as block elements, and so they're positioned separately (instead of inline elements as is the default).

We can potentially improve the Read the Docs formatting with custom CSS or some post-processing (similar to #1401).

See: #1293

<details><summary>Reproducing pandoc's behaviour</summary>

```shell
echo '<a href="https://example.com">Y</a> <img src="example.com" />' \
  | pandoc --from=markdown --to=rst
```

The output is just `Y` by itself, the link and image both completely disappear.

From [their docs](https://pandoc.org/MANUAL.html#raw-html)

> The raw HTML is passed through unchanged in HTML, S5, Slidy, Slideous, DZSlides, EPUB, Markdown, CommonMark, Emacs Org mode, and Textile output, and suppressed in other formats.

That doesn't include reStructuredText.

Strangely enough (and in contradiction), it does seem to pass through the various `<table>` tags. The converted rST for the original table is contains the text and the table-related tags, but nothing else:

```rst
.. raw:: html

   <table>

.. raw:: html

   <tr>

.. raw:: html

   <td>

Run the master version of this notebook:

.. raw:: html

   </td>

.. raw:: html

   <td>

.. raw:: html

   </td>

.. raw:: html

   <td>

.. raw:: html

   </td>

.. raw:: html

   </tr>

.. raw:: html

   </table>
```

</details>

PR #1470, opened 2020-05-04
Fix Binder and Colab links on Github (#1469)

### LINKED ISSUE #1469: Colab/Binder links don't work on Github
### Describe the bug

Our cloud runner links don't work on GitHub. E.g. https://github.com/stellargraph/stellargraph/blob/develop/demos/calibration/calibration-pubmed-link-prediction.ipynb

Clicking either button hits:

```
Refused to load https://colab.research.google.com/github/stellargraph/stellargraph/blob/master/demos/calibration/calibration-pubmed-link-prediction.ipynb because it does not appear in the frame-src directive of the Content Security Policy.
```

```
Refused to load https://mybinder.org/v2/gh/stellargraph/stellargraph/master?urlpath=lab/tree/demos/calibration/calibration-pubmed-link-prediction.ipynb because it does not appear in the frame-src directive of the Content Security Policy.
```

in Safari's web console.

### Environment

StellarGraph eb4d832c14e8bc8180efb96b134c892d868a890d
### Additional context

N/A

timpitman:
This fixes #1469 - Github links to cloud runners were broken due to https://github.com/stellargraph/stellargraph/pull/1398.

This PR reverts #1398 and adds cloud runner links to sphinx using prolog and epilog of nbsphinx.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -196,7 +196,12 @@ nbsphinx_prolog = r"""
 .. raw:: html
 
     <div>
-      <p><a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn">Download this notebook</a></p>
+      <p>
+      <a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn">Download this notebook</a>
+      <br>Run the master version of this notebook:
+      <a href="https://mybinder.org/v2/gh/stellargraph/stellargraph/master?urlpath=lab/tree/{{ env.docname }}.ipynb" alt="Open In Binder"><img src="https://mybinder.org/badge_logo.svg"/></a>
+      <a href="https://colab.research.google.com/github/stellargraph/stellargraph/blob/master/{{ env.docname }}.ipynb" alt="Open In Colab"><img src="https://colab.research.google.com/assets/colab-badge.svg"/></a>
+      </p>   
     </div>
 """
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -195,11 +195,12 @@ texinfo_documents = [
 nbsphinx_prolog = r"""
 .. raw:: html
 
-    <div>
-      <p><a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn">Download this notebook</a>
-      <br>Run the master version of this notebook:
-      <a href="https://mybinder.org/v2/gh/stellargraph/stellargraph/master?urlpath=lab/tree/{{ env.docname }}.ipynb" alt="Open In Binder"><img src="https://mybinder.org/badge_logo.svg"/></a>
-      <a href="https://colab.research.google.com/github/stellargraph/stellargraph/blob/master/{{ env.docname }}.ipynb" alt="Open In Colab"><img src="https://colab.research.google.com/assets/colab-badge.svg"/></a>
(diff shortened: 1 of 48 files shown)


docs/conf.py
```diff
@@ -196,7 +196,11 @@
 .. raw:: html
 
     <div>
-      <p><a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn">Download this notebook</a></p>
+      <p><a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn">Download this notebook</a>
+      <br>Run the master version of this notebook:
```
huonw, line 200: I wonder if this would be better all on one line, without the `<br>`?
```suggestion
      Run the master version of this notebook:
```

huonw, line 200: Also, it's not directly related to this PR, but I realised that "master version" probably doesn't mean much to a user, an alternative would be something like:

```suggestion
      <br>Run the latest release of this notebook:
```

Future work along this line would be changing it to "Run this notebook", by using something like nbsphinx approach. They use `env.config.release` to have each doc version page execute with exactly that git tag https://github.com/spatialaudio/nbsphinx/blob/0.6.1/doc/conf.py#L51 .

docs/conf.py
```diff
@@ -196,7 +196,11 @@
 .. raw:: html
 
     <div>
-      <p><a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn">Download this notebook</a></p>
+      <p><a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn">Download this notebook</a>
```
huonw, line 199: An alternative to the other two minor tweaks, would be to rephrase this to be one longer line, something like:

```html
<p>
  Execute this notebook: 
  <a href="{{ env.docname.rsplit('/', 1).pop() }}.ipynb" class="btn">Download locally</a>
  <a href="https://mybinder.org/v2/gh/stellargraph/stellargraph/master?urlpath=lab/tree/{{ env.docname }}.ipynb" alt="Open In Binder"><img src="https://mybinder.org/badge_logo.svg"/></a>
  <a href="https://colab.research.google.com/github/stellargraph/stellargraph/blob/master/{{ env.docname }}.ipynb" alt="Open In Colab"><img src="https://colab.research.google.com/assets/colab-badge.svg"/></a>
</p>
```

That looks something like:

image

docs/conf.py
```diff
@@ -196,7 +196,11 @@
 .. raw:: html
 
     <div>
```
huonw, line 198: Oh, one more thing; if we add
```suggestion
    <div class="admonition info">
```

Then these boxes are highlighted slightly more separately:

image

docs/conf.py
timpitman: I like this version grouping them

docs/conf.py
timpitman: "latest release" does read better - have adopted in the notebooks themselves; in sphinx I've adopted your suggestion below.

docs/conf.py
timpitman: I added this; can't decide if it's needed with the other change though?

huonw: I think it helps especially at the end at the end to avoid the text running into the action buttons.

Issue #1428, opened 2020-05-01
Nested demo folders are weird on read the docs


huonw:
### Describe the bug

Some of our demos are structured in subfolders, which leads to a slightly peculiar and complicated display on Read the Docs, e.g.: https://stellargraph.readthedocs.io/en/latest/demos/node-classification/index.html#table-of-contents

![image](https://user-images.githubusercontent.com/1203825/80775303-98a2f280-8ba2-11ea-847b-f7db1123ca55.png)

We could remove an extra layer of links by removing the subfolders.

This applies to:

- node classification
- link prediction
- interpretability

yvonnegh:
Maybe worth rearranging on ReadTheDocs before they start linking to them.

But divergence between ReadTheDocs and Git repo....

Note there are some links to Notebook CI blogpost..

PR #1471, opened 2020-05-04
Rearrange demos to have fewer folders

huonw:
Per #1428, having many subfolders makes the docs display ugly and harder to understand.

This also renames the demos to have a consistent and globally unique filename, like `<algorithm>-<extras>-<task>`. This makes it easy to find a demo when looking within a file browser (such as the left hand panel of Jupyter lab) because they're sorted by algorithm name. It also ensures that a bug report just mentioning a filename is enough to work out exactly which demo and what task they're doing.

This retains an existing placeholder for the GCN node classification demo, since we've linked to that in various places as a "get started with StellarGraph" demo. An alternative would be to just rely/hope that we update all those lines (#1482). The placeholder has the same name as the original, and so there's no renamed detected for `demos/node-classification/gcn/gcn-cora-node-classification-example.ipynb` to `demos/node-classification/gcn-node-classification.ipynb`. However, this is just a rename, with some of the internal links updated.

Docs view: https://stellargraph--1471.org.readthedocs.build/en/1471/demos/index.html#table-of-contents

See: #1428


timpitman:
Just looking at the draft; some of the "Neo4j" typos are being fixed in the other PR I think.  Also there's still "Node representation learning with Watch Your Step on Cora" mentioning a dataset?

huonw:
I'll fix those in the title PR, but you're right that this is a much easier way to check the titles.

timpitman:
Thinking about this more; could the duplicate GCN demo just be a symlink?

huonw:
> Thinking about this more; could the duplicate GCN demo just be a symlink?

That's a good idea, but I think it won't work, because `format_notebooks` will be expecting its runner badges to include the symlink path (but they'll be the path for the main notebook).

huonw:
This now builds without any (new) warnings, and a link checker run against our docs works except for some external links and various links to `https://stellargraph.readthedocs.io/en/stable/...`, because we've not yet published a stable release with these new docs. I've tried to manually verify them as much as possible.

PR #1472, opened 2020-05-04
Add custom CSS for cell wrapping in demo table

huonw:
This does two things to make our wide demo table slightly easier to manage and reduce the sideways scrolling:

- allows individual cell contents to wrap across multiple lines, making each column narrower
- makes all docs wider (a maximum width 1000px instead of 800px) so that the there's more of the table visible at a time for people with wide-enough screens (total maximum doc width is 1300px), without making the line length too long

The second point affects all of our documentation and so may need to be balanced.

Demo table: https://stellargraph--1472.org.readthedocs.build/en/1472/demos/index.html#find-a-demo-for-an-algorithm image

Node classification table: https://stellargraph--1472.org.readthedocs.build/en/1472/demos/node-classification/index.html#find-algorithms-and-demos-for-a-graph image

API docs (to gauge the affect of widening): https://stellargraph--1472.org.readthedocs.build/en/1472/api.html

Current equivalent develop links, for comparison:

- https://stellargraph.readthedocs.io/en/latest/demos/index.html#find-a-demo-for-an-algorithm
- https://stellargraph.readthedocs.io/en/latest/demos/node-classification/index.html
- https://stellargraph.readthedocs.io/en/latest/api.html

This is an alternative to #1461, which ensures the table doesn't require scrolling by allowing it to overhang.

See: #1418

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,14 @@
+/**** override table width restrictions *****/
+
+.wy-table-responsive table td, .wy-table-responsive table th {
+    /* cells can wrap to be multi-line */
+    white-space: normal;
+}
+
+/* table can overhang the normal text width */
+.wy-table-responsive table {
+    background: white;
+}
+.wy-table-responsive {
+    overflow: visible;
+}
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,3 +1,8 @@
+div.wy-nav-content {
+    /* make the main content wider, for wider tables */
+    max-width: 1200px;
+}
+
 /**** override table width restrictions *****/
 
 .wy-table-responsive table td, .wy-table-responsive table th {
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -4,11 +4,3 @@
     /* cells can wrap to be multi-line */
     white-space: normal;
 }
-
-/* table can overhang the normal text width */
-.wy-table-responsive table {
-    background: white;
-}
-.wy-table-responsive {
-    overflow: visible;
-}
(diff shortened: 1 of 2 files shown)


kjun9:
> It looks much better, though on my MacBook there's still some width to spare - could we make it wide enough to fit the table without scrolling?

I don't think we'd want to fill the entire width of a large screen though, since it makes it harder to read normal text

huonw:
Like @kjun9, I'm concerned about making the page being too wide for the main docs and the demos themselves, and thus making them harder to read.

From some basic googling, it seems the guidelines on line length for text are relatively short, e.g.:

- 9-12 or 12-15 words https://www.fonts.com/content/learning/fontology/level-2/text-typography/length-column-width
- 45-85 characters https://www.smashingmagazine.com/2014/09/balancing-line-length-font-size-responsive-web-design/
- 50-75 characters https://baymard.com/blog/line-length-readability

Based on some random paragraphs, our current develop is already pushing these limits, with ~80 characters per line and 17-18 words per line.

This PR bumps this up to ~125 characters and ~25 words per line, which both dramatically violates those guidelines. It's also noticeably harder for me to read.

You can experiment with this live in browser tools by adjusting the `max-width` property:

image


That said... this does seem to be in an awkward middle ground of almost-too-wide for text but still not wide enough for the full table. At least on my machine, increasing it slightly (to 1300px instead of 1200px) avoids scrolling entirely. Another option would be _decreasing_ it, so the scrolling is more purposeful for the main demo table and the main text is easier to read. 1000px seems to be wide enough for all the tables except the main demo table.

huonw:
I've gone with 1000px so the main text isn't too wide.

Issue #1269, opened 2020-04-19
Community detection demo is a downstream task using embeddings


huonw:
### Description

The "attacks clustering" community detection demo https://github.com/stellargraph/stellargraph/blob/develop/demos/community_detection/attacks_clustering_analysis.ipynb is "just" a downstream task of applying a clustering algorithm on top of node embeddings, and so may fit better into `demos/embeddings`.

Issue #1221, opened 2020-04-08
Move "use-case" demos to archive repo with pinned Stellargraph version


huonw:
### Description

TODO: fill out

timpitman:
I'd like to move out difficult notebooks, however Hateful Twitter is a great demo and I've added #1224 for an alternate solution for Kaggle notebooks.

PR #1477, opened 2020-05-04
Temporarily ignore community detection and use-case demos in Read the Docs

huonw:
These three demos are interesting but they're major chunks of code and aren't run on CI. This means they likely don't work.

For the 1.0 release, we can temporarily stop them being published in our docs so they're not being highlighted, and then spend more time working out what to do with them for the next release.

This relates to #1269 and #1221.

Resulting docs: https://stellargraph--1477.org.readthedocs.build/en/1477/demos/index.html#table-of-contents

Thoughts?


timpitman:
We really don't lose much by taking them out of the docs, given they're unmaintained I think this improves the quality of our rtd demos

Issue #1335, opened 2020-04-23
Supervised graph classification notebook training runs for a long time without progress


huonw:
### Describe the bug

The supervised graph classification notebook defaults to training 50 models (10 folds with 5 restarts per folds) for up to 500 epochs. It does this with `verbose=0` in the Keras `Model` method calls, and no other progress information (other than `[...]`'s from #826), so it's not obvious to the user that it's making progress.

(StellarGraph 1.0.0rc1 dogfooding)

kieranricardo:
Possible can use `tqdm` module

PantelisElinas:
I understand that it might run for a while although it uses early stopping and hardly ever trains for 500 epochs, but how is this a bug? Are you suggesting that we need to add a cautionary note about training time somewhere in the notebook?

huonw:
The training cell trains silently for many minutes. On my machine, it's about 10-15s/restart, and thus 60s per fold, meaning 10 minutes just to run the training cell without any feedback.

This is suboptimal for a few reasons:

- the notebook just takes a long time to run, making it harder for us to edit and much more annoying for a user to work with.
- silent work always leaves someone with a "is it really doing anything" question in their head. I know I'm always concerned when running documentation or examples (i.e. code that I didn't write and/or am otherwise unfamiliar with) that take a long time because I don't know if it's hanging because I've misconfigured my installation somehow (or something like that). This is what the "it's not obvious to the user that it's making progress" in the original issue description is referring to.

Between these, I think it's likely that a non-trivial portion of potential users will give up ("I let it run for 5 minutes and nothing happened") or come away with a poor impression. The StellarGraph dogfooder who I observed using this notebook got a little frustrated in practice, too.

One simple way to address at least the second point is @kieranricardo's comment: the [`tqdm`](https://github.com/tqdm/tqdm) library could be used to wrap the training loops like `tqdm.tqdm(model_selection.StratifiedKFold(...))` and `tqdm.tqdm(range(num_restarts))` (it even automatically supports nice rendering for [nested loops](https://github.com/tqdm/tqdm#nested-progress-bars)) and so give some feedback to the user. This will ensure they (a) understand that it's working, and (b) give them an estimate for how long it will take.

[...]

PR #1485, opened 2020-05-05
Supervised graph classification progress update

kieranricardo:
This adds a really basic progress update into the supervised graph classification demo. It prints out which fold/repeat is being trained and how many are left to go after every fold/repeat.

See: #1335


demos/graph-classification/supervised-graph-classification.ipynb
```diff
@@ -471,10 +521,12 @@
...
     "test_accs = []\n",
     "\n",
-    "for train_index, test_index in model_selection.RepeatedStratifiedKFold(\n",
+    "stratified_folds = model_selection.RepeatedStratifiedKFold(\n",
     "    n_splits=folds, n_repeats=n_repeats\n",
-    ").split(graph_labels, graph_labels):\n",
+    ").split(graph_labels, graph_labels)\n",
     "\n",
+    "for i, (train_index, test_index) in enumerate(stratified_folds):\n",
+    "    print(f\"Training and evaluating on fold {i+1} out of {folds * n_repeats}...\")\n",
```
huonw, line 529: If we were concerned about the verbosity here, we _could_ go for:

```python
if (i + 1) % 5 == 0:
    print(f"Training and evaluating on fold {i+1} out of {folds * n_repeats}...")
```

and so only get every 5th update, but that's probably not necessary and a little confusing (we'd have to move it to the end of the loop and rephrase the message or something)

PR #1509, opened 2020-05-05
add link to paper

kieranricardo:
This PR adds a reference to the watch your step paper into the demo.


kieranricardo:
I think this has already been done

huonw:
What hinted at this being done? I don't see it fixed on develop? https://github.com/stellargraph/stellargraph/blob/develop/demos/embeddings/watch-your-step-embeddings.ipynb

demos/embeddings/watch-your-step-embeddings.ipynb
```diff
@@ -4,7 +4,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Node representation learning with Watch Your Step"
+    "# Node representation learning with Watch Your Step\n",
+    "\n",
+    "This notebook demonstrates how to use the StellarGraph implementation of [Watch Your Step](https://arxiv.org/abs/1710.09599)."
```
huonw, line 9: This either needs to be a "summary", which is a block quote like

```suggestion
    "> This notebook demonstrates how to use the StellarGraph implementation of [Watch Your Step](https://arxiv.org/abs/1710.09599)."
```

Or it needs to be a separate cell, after the cloud runner (or just run `scripts/format_notebooks.py --default --overwrite demos/embeddings` after editing it).

Issue #1495, opened 2020-05-05
Node classification with Simplified Graph Convolutions (SGC) demo notebook on RTD always points to "stable" version of "Loading from Pandas" demo


kjun9:
### Describe the bug

There is a link in the notebook in the "Loading the Cora Network" section that points to the "Loading from pandas" demo, but it seems to navigate to `stable` regardless of whatever version of the docs you're currently looking at.

### To Reproduce

1. Go to https://stellargraph.readthedocs.io/en/latest/demos/node-classification/sgc-node-classification.html#Loading-the-CORA-network
2. Click on the link from the text "See the “Loading from Pandas” demo for details on how data can be loaded"
3. See that it navigates to https://stellargraph.readthedocs.io/en/stable/demos/basics/loading-pandas.html

### Observed behavior

It navigates to https://stellargraph.readthedocs.io/en/stable/demos/basics/loading-pandas.html

### Expected behavior

It should navigate to https://stellargraph.readthedocs.io/en/latest/demos/basics/loading-pandas.html

### Environment

browser: chrome

viewing `latest` stellargraph readthedocs

kjun9:
Actually this might apply to many other demos that also link to that notebook

huonw:
Yeah, this is true of all of them.

PR #1510, opened 2020-05-05
Use a relative link for the reference to the Pandas notebook

huonw:
This works better for someone editing locally, and ensures it's a correct link for the specific version of the documentation the user is looking at.

See: #1495


kjun9:
Seems like this one is missing a docs build entirely, could you trigger one?

Issue #1432, opened 2020-05-01
Build sphinx documentation in CI to make errors and warnings more visible


timpitman:
### Description

Our documentation is built on readthedocs, which means that warnings for a given PR aren't easily visible.  We could build the sphinx docs on buildkite.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1518, opened 2020-05-06
Run Sphinx on CI to catch errors/warnings

huonw:
This adds a CI step that runs Sphinx, and catches errors and warnings. This will help us keep our documentation rendering as good as possible.

It takes this approach rather than just setting `fail_on_warning: true` in the [`.readthedocs.yaml` configuration file](https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-fail-on-warning) because we can get clearer and better error messages and description.

Example output, with the problems from before #1360 was fixed: https://buildkite.com/stellar/stellargraph-public/builds/3994#annotation-sphinx-doc-build

image

See: #1432


docs/requirements.txt
```diff
@@ -4,3 +4,4 @@ nbsphinx==0.6.1
 nbsphinx-link==1.3.0
 # needed for Pygments highlighting in notebooks
 ipython==7.13.0
+sphinx_rtd_theme==0.4.3
```
timpitman, line 7: I'm not entirely sue why our existing doc requirements were pinned?  rtd itself appears to install sphinx-rtd-theme<0.5

huonw: Yeah, I don't quite know why either. I guess it ensures our docs don't break accidentally, but maybe it's suboptimal for this, because we won't automatically pick up improvements to the theme. However, I think these will overrule the RtDs defaults (that is, our requirements here will be the ones that are used for rendering the docs online).

For reference, it seems like RtD uses:

```
Pygments==2.3.1
setuptools==41.0.1
docutils==0.14
mock==1.0.1
pillow==5.4.1
alabaster>=0.7,<0.8,!=0.7.5
commonmark==0.8.1
recommonmark==0.5.0
sphinx<2
sphinx-rtd-theme<0.5
readthedocs-sphinx-ext<1.1
```

Notably, they're using a _very_ old Sphinx version there. Our CI is now installing `Sphinx==3.0.3`. Maybe we should add more things to our requirements to ensure CI and RtDs are more similar?

.buildkite/pipeline.yml
```diff
@@ -162,6 +162,18 @@ steps:
...
+    command: ".buildkite/steps/test-documentation.sh"
+    <<: *timeout
+    depends_on: "runner-3_6"
+    plugins:
+      <<: *plugins
+      docker-compose#v3.2.0:
+        <<: *compose-config
+        run: runner-3_6
+    agents:
+      queue: "t2medium"
```
TheDen, line 175: Could this run potentially run successfully on the default queue (t2micro)?

huonw: I suspect it might be able to, but it would likely require pulling the `runner-3_6` image. I've been trying to keep steps using those images on `t2medium` and `t2large` instances, to have more cache hits.

docs/requirements.txt
timpitman: I couldn't find any documentation on why rtd pre-installs specific old versions of packages, not sure that we want to add what they've already go in requirements.txt, as it may change?

.buildkite/pipeline.yml
TheDen: Fair enough 👍

docs/requirements.txt
huonw: I've added the exact requirements specified by RtD into this PR, and opened https://github.com/stellargraph/stellargraph/pull/1537 to upgrade the versions to the current latest ones. I've done it this way so that this PR (theoretically) doesn't change anything at all about how our docs are rendered, and is just focused on getting accurate CI, and then #1537 can be independently tested and the rendering verified.

PR #1539, opened 2020-05-08
Implement RotE and RotH

huonw:
This implements the RotationE and RotationH algorithms from [1], aka the RotE and RotH from [2]. After doing this I realised that [1] was probably a very-preprint, and it's now a preprint in [2]. The latter expands the algorithm slightly to include reflections in addition to rotations, but still explicitly discusses the rotation-only model (under a very slightly different name). The RotE/RotH algorithm performs similarly (or better than) the full AttE/AttH algorithm.

1. Chami I, Wolf A, Sala F, Ré C. Low-Dimensional Knowledge Graph Embeddings via Hyperbolic Rotations. http://stanford.edu/~fredsala/Hazy-HyperbolicKGs.pdf
2. Chami I, Wolf A, Juan D-C, Sala F, Ravi S, Ré C. Low-Dimensional Hyperbolic Knowledge Graph Embeddings. http://arxiv.org/abs/2005.00545

RotE is effectively a generalisation/combination of RotatE and TransE, and so future work could be combining them all to have a single shared/configurable model (including emergent models like hyperbolic TransE).

As somewhat of an implementation detail, this adds a `utils.hyperbolic` module with the various functions required for the hyperbolic RotH. This allows them to have some basic tests, as well as potentially making it easier to implement other hyperbolic models, e.g.:

- https://cs.stanford.edu/people/jure/pubs/hgcn-neurips19.pdf
- https://arxiv.org/abs/1910.12892

Related discussion of RotationH/RotH and RotationE/RotE, including a comparison to TransE, RotatE, etc.: https://nbviewer.jupyter.org/gist/huonw/52d585e571fde62e252c54fbe668faa3

Follow up: #1664 adding a demo.

See: #1540

### DIFF
--- a/stellargraph/layer/knowledge_graph.py
+++ b/stellargraph/layer/knowledge_graph.py
@@ -671,7 +671,7 @@ class RotatE(KGModel):
         )
 
 
-class RotationHEScoring(Layer, KGScore):
+class RotHEScoring(Layer, KGScore):
     def __init__(self, hyperbolic):
         self._hyperbolic = hyperbolic
         if self._hyperbolic:
@@ -795,7 +795,24 @@ class RotationHEScoring(Layer, KGScore):
         return mod_o_pred.numpy(), mod_s_pred.numpy()
 
 
-class RotationH(KGModel):
+@experimental(reason="demo is missing", issues=[1664])
+class RotH(KGModel):
+    """
+    Embedding layers and a RotH scoring layer that implement the RotH knowledge graph
+    embedding algorithm as in https://arxiv.org/pdf/2005.00545.pdf
+
+    Args:
+        generator (KGTripleGenerator): A generator of triples to feed into the model.
+
+        embedding_dimension (int): the dimension of the embeddings (that is, a vector in
+            ``R^embedding_dimension`` plus a bias in ``R`` is learnt for each node, along with a pair of
+            vectors in ``R^embedding_dimension`` and ``R^(embedding_dimension / 2)`` for each node
+            type). It must be even.
+
+        embeddings_initializer (str or func, optional): The initialiser to use for the embeddings.
+
+        embeddings_regularizer (str or func, optional): The regularizer to use for the embeddings.
+    """
     def __init__(
         self,
         generator,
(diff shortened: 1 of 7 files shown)


stellargraph/layer/knowledge_graph.py
```diff
@@ -670,6 +671,200 @@ def __init__(
...
+        else:
+            self.curvature_prime = None
+
+        super().build(input_shapes)
+
+    def _curvature(self, expand=1):
+        if not self._hyperbolic:
+            return tf.constant([0.0])
+
+        return tf.math.softplus(self.curvature_prime)
```
kjun9, line 724: should this check if `curvature_prime` is None? (i.e. `build` hasn't been called)

stellargraph/layer/knowledge_graph.py
```diff
@@ -670,6 +671,200 @@ def __init__(
...
+            self._add = poincare_ball_mobius_add
+            self._squared_distance = lambda c, v, w: tf.square(
+                poincare_ball_distance(c, v, w)
+            )
+        else:
+            self._convert = lambda _c, v: v
+            self._add = lambda _c, v, w: v + w
+            self._squared_distance = lambda _c, v, w: tf.reduce_sum(
+                tf.math.squared_difference(v, w), axis=-1
+            )
```
kjun9, line 688: Should we define `curvature_prime` in init?
```suggestion
            )
            self.curvature_prime = None
```

stellargraph/utils/hyperbolic.py
```diff
@@ -0,0 +1,151 @@
...
+def poincare_ball_mobius_add(c, x, y):
+    r"""
+    Möbius addition of ``x`` and ``y``, on the Poincaré ball with curvature ``-c``: :math:`\mathbf{x} \oplus^c \mathbf{y}`.
+
+    See Section 2 of [1] for more details.
+
+    [1] O.-E. Ganea, G. Bécigneul, and T. Hofmann, “Hyperbolic Neural Networks,” `arXiv:1805.09112 <http://arxiv.org/abs/1805.09112>`_, Jun. 2018.
+
+    Args:
+        c (tensorflow Tensor-like): the curvature of the hyperbolic space(s). Must be able to be
```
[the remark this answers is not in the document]
huonw: Yeah, NumPy arrays. Basically anything that is supported by https://www.tensorflow.org/api_docs/python/tf/convert_to_tensor, because I think that's what functions like `tf.tanh` (etc.) support as input.

stellargraph/layer/knowledge_graph.py
huonw: Sounds good.

stellargraph/layer/knowledge_graph.py
huonw: Our other model classes don't stub out their weights (e.g. GraphConvolution's `kernel` and `bias` etc.), and it seems Keras itself doesn't either, e.g. `Dense`: https://github.com/tensorflow/tensorflow/blob/1248bc9c61dd4a1e6677c9cc1c747b792765ae58/tensorflow/python/keras/layers/core.py#L1127-L1186 . I guess this could be a first 😄  Thoughts?

kjun9: Ah in that case, I'm happy for you to just leave it. I guess it's probably more important for member variables that are supposed to be accessed from outside the class.

stellargraph/layer/knowledge_graph.py
```diff
@@ -670,6 +671,201 @@ def __init__(
...
+
+    def build(self, input_shapes):
+        if self._hyperbolic:
+            self.curvature_prime = self.add_weight(shape=(1,), name="curvature_prime")
+        else:
+            self.curvature_prime = None
+
+        super().build(input_shapes)
+
+    def _curvature(self, expand=1):
```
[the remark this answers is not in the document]
huonw: Good catch.

Issue #1538, opened 2020-05-08
Allow running a specific git commit of a notebook in colab and possibly binder from readthedocs


timpitman:
### Description

Extend our readthedocs binder buttons to insert on develop and other builds a link to run the specific git hash of a notebook, rather than just latest release version.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1541, opened 2020-05-08
use git revision in readthedocs notebook links (#1538)

timpitman:
See #1538 - Our google colab and binder links inside readthedocs demo notebooks don't currently work on PR's, and when they do work for 'stable', 'latest' and release versions, always point to master.

This PR gets the git commit from readthedocs internal variables and uses that to link to the actual commit.  Readthedocs sets a variable `github_version` that gives:
* commit hash for PR's and 'stable' (eg: https://readthedocs.org/projects/stellargraph/builds/10970680/)
* develop for 'latest'. (eg: https://readthedocs.org/projects/stellargraph/builds/11012887/)
* the tag for specific release versions (eg: https://readthedocs.org/projects/stellargraph/builds/10970681/)

If rendering outside readthedocs (eg a local sphinx build), or for 'stable', we substitute this with 'master'.

All builds on readthedocs should now point to the correct demo notebooks for colab or binder.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -206,20 +206,14 @@ nbsphinx_prolog = r"""
     <div class="admonition info">
       <p>
         Execute this notebook:
-        <a href="https://mybinder.org/v2/gh/stellargraph/stellargraph/{{ env.config.release|e }}?urlpath=lab/tree/{{ env.docname }}.ipynb" alt="Open In Binder"><img src="https://mybinder.org/badge_logo.svg"/></a>
-        <a href="https://colab.research.google.com/github/stellargraph/stellargraph/blob/{{ env.config.release|e }}/{{ env.docname }}.ipynb" alt="Open In Colab"><img src="https://colab.research.google.com/assets/colab-badge.svg"/></a>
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -196,18 +196,26 @@
...
-
 # -- Extension configuration -------------------------------------------------
 
 # This is processed by Jinja2 and inserted before each notebook
+# We use internal readthedocs variables to get the git version if available. Note that this is undocumented, however it
+# is shown in our readthedocs build logs, and is generated from a template:
+# https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl
 nbsphinx_prolog = r"""
+{% if env.config.html_context.github_version is defined and env.config.html_context.current_version != "stable" %}
+    {% set github_version = env.config.html_context.github_version %}
```
huonw, line 207: Maybe we could call this something a little more specific, like

```suggestion
    {% set git_revision = env.config.html_context.github_version %}
```

where a revision seems to be what git calls value that can be a commit-or-branch-or-... (e.g. https://git-scm.com/docs/git-rev-parse#_specifying_revisions).

(This is totally minor and doesn't need to change.)

Issue #1544, opened 2020-05-11
Peak memory usage of StellarGraph construction is important, but unmeasured


huonw:
### Describe the bug

For memory-constrained systems, the peak memory usage of a program is relevant. This applies for large graphs, where a final graph of size 3GB might be fine on a system with 8GB of RAM, but spiking to 12GB of RAM temporarily probably isn't. Constructing a StellarGraph exhibits a temporary spike like this, where it uses much more memory than the final object. This is somewhat problematic, and so we should make it easy to measure. (This issue is only about measuring, not fixing.)

### To Reproduce

1. Run https://gist.github.com/huonw/deb88a9179cfb588f9d5fff59955f56d
2. Check the tests for any benchmarks that record memory use _peak_

E.g. the following test records the final memory usage only:

https://github.com/stellargraph/stellargraph/blob/fd9448610299fecfb14f94eaf79a3e370d013ac9/tests/core/test_graph.py#L787-L813

### Observed behavior

The notebook reports:

|        |   memory (StellarGraph) |   peak memory (StellarGraph) |   memory (edge data) |   peak memory (edge data) |   peak memory (above edge data) |
|:-------|------------------------:|-----------------------------:|---------------------:|--------------------------:|--------------------------------:|
| Cora   |               0.0939255 |                     0.390519 |            0.0873337 |                  0.178535 |                        0.303185 |
| FB15k  |               6.10444   |                    35.4983   |           14.8337    |                 23.8789   |                       20.6646   |
| reddit |             146.781     |                   595.343    |          177.111     |                354.227    |                      418.231    |

### Expected behavior

We should make it easy to track and control this peak memory usage.

### Environment

Operating system: `Darwin-18.6.0-x86_64-i386-64bit`
Python version:
```
3.6.9 (default, Jul 10 2019, 12:25:55) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)]
```
Package versions:
<details>

```
absl-py==0.9.0
appnope==0.1.0
astunparse==1.6.3
backcall==0.1.0
boto==2.49.0
boto3==1.13.5
botocore==1.16.5
cachetools==4.1.0
certifi==2020.4.5.1
chardet==3.0.4
cycler==0.10.0
decorator==4.4.2
docutils==0.15.2
gast==0.3.3
gensim==3.8.3
google-auth==1.14.2
google-auth-oauthlib==0.4.1
google-pasta==0.2.0
[...]

PR #1546, opened 2020-05-11
Add benchmarks recording the peak memory usage

huonw:
This extends our current `allocation_benchmark` fixture to allow tracing the peak memory usage of a function, not just the long-term memory usage of the return value. In particular, it uses https://docs.python.org/3/library/tracemalloc.html#tracemalloc.get_traced_memory to record the peak between calls to the new `peak` "timer" function.

Current results for the `test_allocation_benchmark_creation` ("graph") and `test_allocation_benchmark_creation_peak` ("peak") benchmarks:

| params                    |   graph (KiB) |   peak (KiB) |   ratio |
|:--------------------------|--------:|-------:|--------:|
| None-None-0-0             |    14.3 |   26.9 |     1.9 |
| None-100-0-0              |    15.4 |   30.5 |     2.0 |
| None-None-200-100         |    14.4 |   26.4 |     1.8 |
| undirected-None-200-100   |    30.4 |   48.6 |     1.6 |
| directed-None-200-100     |    41.4 |   66.5 |     1.6 |
| both-None-200-100         |    54.9 |   90.2 |     1.6 |
| None-100-200-100          |    52.6 |   79.2 |     1.5 |
| undirected-100-200-100    |    68.7 |   86.1 |     1.3 |
| directed-100-200-100      |    75.9 |  104.1 |     1.4 |
| both-100-200-100          |    98.9 |  123.0 |     1.2 |
| None-None-5000-1000       |    69.9 |  205.2 |     2.9 |
| undirected-None-5000-1000 |   239.8 |  571.3 |     2.4 |
| directed-None-5000-1000   |   393.3 |  824.9 |     2.1 |
| both-None-5000-1000       |   564.9 |  998.3 |     1.8 |
| None-100-5000-1000        |   456.1 |  694.4 |     1.5 |
| undirected-100-5000-1000  |   626.3 |  961.0 |     1.5 |
| directed-100-5000-1000    |   779.3 | 1210.8 |     1.6 |
| both-100-5000-1000        |   950.1 | 1383.1 |     1.5 |

Notes:

- Param order in each test name: `force_adj_lists`, `feature_size`, `num_nodes`, `num_edges`.
- Sorted by `num_nodes`, then `feature_size`, then `force_adj_lists`.

That is, the peak memory usage ranges from 1.2-2.9× higher than the memory usage of the final graph object.

See: #1544

### DIFF
--- a/tests/core/test_graph.py
+++ b/tests/core/test_graph.py
@@ -21,7 +21,7 @@ import pytest
 import random
 from stellargraph.core.graph import *
 from stellargraph.core.experimental import ExperimentalWarning
-from ..test_utils.alloc import snapshot, allocation_benchmark
+from ..test_utils.alloc import snapshot, peak, allocation_benchmark
 from ..test_utils.graphs import (
     example_graph_nx,
     example_graph,
@@ -784,7 +784,7 @@ def test_benchmark_creation(
     benchmark(f)
 
 
-@pytest.mark.benchmark(group="StellarGraph creation", timer=snapshot)
+@pytest.mark.benchmark(group="StellarGraph creation (size)", timer=snapshot)
 # various element counts, to give an indication of the relationship
 # between those and memory use (0,0 gives the overhead of the
 # StellarGraph object itself, without any data)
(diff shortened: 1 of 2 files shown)


tests/core/test_graph.py
```diff
@@ -813,6 +813,35 @@ def f():
...
 
+@pytest.mark.benchmark(group="StellarGraph creation (peak)", timer=peak)
+# various element counts, to give an indication of the relationship
+# between those and memory use (0,0 gives the overhead of the
+# StellarGraph object itself, without any data)
+@pytest.mark.parametrize("num_nodes,num_edges", [(0, 0), (100, 200), (1000, 5000)])
+# features or not, to capture their cost
+@pytest.mark.parametrize("feature_size", [None, 100])
+@pytest.mark.parametrize("force_adj_lists", [None, "directed", "undirected", "both"])
+def test_allocation_benchmark_creation_peak(
```
timpitman, line 824: This function and `test_allocation_benchmark_creation` above could both call a function with the same code instead of repeating it?

huonw:
> Just so I've got this right, because this stuff is pretty wacky and I couldn't find documentation for the BenchmarkFixture class, BenchmarkFixture in this case will do:

https://github.com/ionelmc/pytest-benchmark/blob/dafa578d034525475fe7a7e14b085315e15a14e2/src/pytest_benchmark/fixture.py#L89-L99

You're basically right, but there's a key detail, because the `peak` timer both gets the peak-since-last-`peak`-call and resets the memory tracking (i.e. resets the peak to 0)

1. start = timer() gets: peak memory measurement (unused), resets peak counter to 0 (used)
2. do some stuff
3. end = timer() gets: peak memory measurement (used), resets peak counter to 0 (unused)

tests/core/test_graph.py
huonw: Good point; it's a bit fiddly so I'll do it as a separate commit.

Issue #1327, opened 2020-04-23
Edge features: support storing edge features in StellarGraph


huonw:
### Description

In service of #1326, we need to store the features within the `StellarGraph` object.

PR #1296, opened 2020-04-21
Store adj dictionaries as one contiguous np array

kieranricardo:
This PR stored adj lists in one contiguous numpy array.

For each adj dict, this PR instead stores the edge ilocs in one contiguous numpy array sorted by the index (source/target node for in/out dicts). The edge ilocs associated with a node are then accessed by:

- mapping a `node_iloc` -> `start_index`, `stop_index`
- lookup up flat array: -> `flat[start_index:stop_index]`

**====== Memory**

Measuring the memory of FB15k in MiB for develop and this PR for different adj caches:

| Ajd Lists | Develop | PR |
| ----------- |:--------:| -----:|
| None           | 7.3          | 7.3     |
| Directed      | 14.2       | 11.9    |
| Undirected | 16.7        |  12.1   |

For directed graphs the adj lists are 1.4x smaller, and 1.2x smaller for undirected graphs. 

A nice property of this PR is that undirected adj lists, directed adj lists, and the edge lists now all use approximately the same amount of memory.  (this doesn't apply if the `node_ilocs` and `edge_ilocs` are different types).

**====== Neighbour lookup**

PR benchmark

```
--------------------------------------------------------------------------------------- benchmark 'StellarGraph neighbours': 2 tests --------------------------------------------------------------------------------------
Name (time in us)                               Min                   Max                  Mean              StdDev                Median                 IQR            Outliers         OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_get_neighbours[True]        649.1660 (1.0)      2,333.2520 (1.0)        727.6934 (1.0)       86.5430 (1.0)        692.0295 (1.0)       41.2615 (1.0)       198;236  1,374.2051 (1.0)        1352           1
test_benchmark_get_neighbours[False]     7,630.3470 (11.75)    8,945.7340 (3.83)     8,079.2179 (11.10)    256.9341 (2.97)     8,024.1330 (11.60)    275.9956 (6.69)         26;8    123.7744 (0.09)        116           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Develop benchmark

```
--------------------------------------------------------------------------------------- benchmark 'StellarGraph neighbours': 2 tests ---------------------------------------------------------------------------------------
Name (time in us)                               Min                    Max                  Mean              StdDev                Median                 IQR            Outliers         OPS            Rounds  Iterations
[...]

PR #1547, opened 2020-05-11
Add a notebook capturing StellarGraph class resource usage

huonw:
In many PRs, we've recorded the time and/or memory usage of several real-world datasets, such as #1220, #1284, #1291, #1535.

This was done in an ad-hoc way, passing around various versions of a notebook as gists, e.g. 
1. initial version: https://gist.github.com/huonw/630907bf02f4ff443bb3d83139bb32aa
2. version with better source/target memory usage: https://gist.github.com/kieranricardo/536776d95bfbc56e8ab709f1d261cb50
3. version with end-to-end memory usage via `tracemalloc`: https://gist.github.com/huonw/deb88a9179cfb588f9d5fff59955f56d

This PR expands the last version to something that's appropriate to add to the repo, so that it's easily accessible.

The notebook records both time and both long term and peak memory usage for constructing five graphs:

- Cora
- BlogCatalog3
- FB15k with and without edge types
- reddit (optional, if manually downloaded from http://snap.stanford.edu/graphsage/#datasets)

These have a variety of sizes and "shapes" (Cora and reddit are homogeneous, BlogCatalog3 has multiple node types and edge types, FB15k has multiple edge types), although there's no heterogeneous graph with node features.

The memory usage is measured with [the `tracemalloc` Python stdlib module](https://docs.python.org/3/library/tracemalloc.html). This counts every single allocation, including ones from NumPy. The notebook loads the data freshly from disk when tracing allocations, and thus captures every piece of memory that goes into creating the raw data and the `StellarGraph` object. After creating the `StellarGraph`, the notebook cleans up the input data, leaving only the graph instance. At various points, the peak memory usage and snapshots of the individual allocations are recorded. This allows computing various metrics:

- memory usage of the memory controlled by the `StellarGraph` instance itself (`memory (graph)`)
- memory usage of the _new_ memory controlled by the instance, that isn't shared with the raw data (`memory (graph, not shared with data)`)
- peak memory usage during construction of the graph (`peak memory (graph)`)
- peak memory usage during construction, ignoring the base memory usage of the nodes/edges input data (`peak memory (graph, ignoring data)`)
- base memory usage of input data (`memory (data)`)
- as a check, peak memory usage when loading the input data (`peak memory (data)`), if this is equal to `peak memory (graph)` it likely means that the peak memory usage during graph construction is less than loading data, and so the `peak memory (graph)` is not longer meaningful

All of this more accurate that using a retroactive analysis like [`asizeof` from `pympler`](https://pythonhosted.org/Pympler/asizeof.html#asizeof), because it correctly counts allocations from NumPy arrays once and only once. Pympler can miss some memory: for `asizeof(obj)`, suppose `obj` contains an array that is computed as a slice of another one (so `array.flags.owndata == False`), but doesn't contain the parent/base array (`array.base`). `asizeof` seems to count `array` as 0, and thus a potentially-large allocation isn't being counted. (Gist 2 above was attempting to fix this by inserting `np.copy`s at appropriate places, but this is a fragile/inaccurate battle.)

Current status:

image

This notebook is designed to be used by StellarGraph contributors, and so is data-dense without too much explanation, and is placed into a new `demos/internal-developers` directory. It is still added to `demos` and our RtD, so that it runs on CI and for consistency.

See: #1544

## Some questions

- could/should this be a script instead of a notebook? I chose a notebook because it presents the results nicely, makes it easy to explore them (e.g. add a new dataset or investigate the details of an existing one) and fits into our existing testing infrastructure smoothly, but a script might make more sense
- could/should this be actual tests/benchmarks, so that it is run and reported by `pytest-benchmark`? I don't think it can easily be, without being expensive, basically having to run tests multiple times (e.g. we currently have three copies of benchmarks for creation, recording time, long-term memory usage and peak memory usage independently, see #1546 for instance). Potentially there's a way to report multiple metrics from a single pytest-benchmark benchmark?
- should this allow for doing a comparison in addition to just recording the results? This could similar to how `pytest-benchmark` has `pytest ... --benchmark-compare` and `pytest-benchmark compare`. (This could be future work.)


huonw:
Specific example of the Pympler behaviour:

```python
from pympler.asizeof import asizeof

import numpy as np

a = np.arange(100000)
b = a[:]

print(asizeof(a), asizeof(b), asizeof((a, b))) # 800096, 96, 800256
```

That is, `b` is all the same data as `a`, and keeps that memory alive, but if `asizeof` doesn't see the `a` object itself, then the memory isn't counted.

timpitman:
> ## Some questions
> * could/should this be a script instead of a notebook? I chose a notebook because it presents the results nicely, makes it easy to explore them (e.g. add a new dataset or investigate the details of an existing one) and fits into our existing testing infrastructure smoothly, but a script might make more sense

I think it's easiest to keep this in a notebook, and running it in Buildkite means the results are available for any build.

> * could/should this be actual tests/benchmarks, so that it is run and reported by `pytest-benchmark`? I don't think it can easily be, without being expensive, basically having to run tests multiple times (e.g. we currently have three copies of benchmarks for creation, recording time, long-term memory usage and peak memory usage independently, see #1546 for instance). Potentially there's a way to report multiple metrics from a single pytest-benchmark benchmark?

see above, I think it's fine as a notebook and I don't know how to integrate with pytest-benchmarks

> * should this allow for doing a comparison in addition to just recording the results? This could similar to how `pytest-benchmark` has `pytest ... --benchmark-compare` and `pytest-benchmark compare`. (This could be future work.)

this would only really be needed if we were planning a lot of ongoing active development on StellarGraph class?

huonw:
> Not sure if it belongs with the other user-facing demos though?

Indeed. I put it within `demos/` so that it automatically runs on CI. I guess we could extend our CI slightly to support running notebooks in `scripts/` or something?

> this would only really be needed if we were planning a lot of ongoing active development on StellarGraph class?

I guess we're not, but there's a few active pieces of work like #1296 and #1556.

huonw:
Oh, there's also #1327.

demos/internal-developers/README.md
```diff
@@ -0,0 +1,13 @@
+# Development-targeted notebooks
```
timpitman, line 1: This could be renamed to make it clearer that this is for internal stellargraph development, eg 
```suggestion
# StellarGraph Internal Development
```

docs/demos/internal-developers/index.rst
```diff
@@ -0,0 +1,13 @@
+Development-targeted notebooks
```
timpitman, line 1: Currently this is in the middle of our demos contents; I think it should be right at the bottom so that it's clearer that it's not a normal demo

demos/internal-developers/graph-resource-usage.ipynb
```diff
@@ -0,0 +1,1058 @@
+{
```
timpitman, line 1: <p>A possible extension is to add reddit to our datasets code, but not provide automatic download?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1547/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

docs/demos/internal-developers/index.rst
huonw: I think putting at the end may require doing something slightly silly like naming the folder(s) `zzz-internal-developers`?

demos/internal-developers/graph-resource-usage.ipynb
huonw: That won't directly help here, because this needs to be manually reading the files, but it would help for potential future demos that might use it.

PR #1555, opened 2020-05-12
gcn_lstm graph convolution layer update

habiba-h:
Before GCN layer had fixed size hidden units, now along with the number of graph convolution layers, the output size of each layer is user defined as well.


stellargraph/layer/gcn_lstm.py
```diff
@@ -131,7 +133,7 @@ def build(self, input_shapes):
         """
         n_nodes = input_shapes[-1]
         t_steps = input_shapes[-2]
-        self.units = t_steps
+        # self.units = t_steps
```
kieranricardo, line 136: I think this commented out code can be removed 
```suggestion
```

stellargraph/layer/gcn_lstm.py
```diff
@@ -67,10 +68,11 @@ def __init__(
         if "input_shape" not in kwargs and input_dim is not None:
             kwargs["input_shape"] = (input_dim,)
 
-        self.units = (units,)
+        self.units = units  # (units,)
```
kieranricardo, line 71: i think the comment can be removed here as well
```suggestion
        self.units = units
```

stellargraph/layer/gcn_lstm.py
```diff
@@ -56,6 +56,7 @@ def __init__(
         activation=None,
         use_bias=True,
         input_dim=None,
+        final_layer=None,
```
kieranricardo, line 59: I think `final_layer` is unused here

stellargraph/layer/gcn_lstm.py
```diff
@@ -191,47 +193,43 @@ def call(self, features):
...
+           bias_regularizer (str or func, optional): The regulariser to use for the bias of each layer.
+           bias_constraint (str or func, optional): The constraint to use for the bias of each layer.
+        """
 
     def __init__(
         self,
         seq_len,
         adj,
-        gc_layers,
+        gc_layer_size,
```
kieranricardo, line 229: seeing as this list of layer sizes i think:
```suggestion
        gc_layer_sizes,
```

might make more sense. and the same below

stellargraph/layer/gcn_lstm.py
```diff
@@ -289,7 +288,9 @@ def __init__(
         for ii in range(n_gc_layers):
             self._layers.append(
                 FixedAdjacencyGraphConvolution(
-                    units=self.seq_len, A=self.adj, activation=self.gc_activations[ii]
+                    units=self.gc_layer_size[ii],
```
kieranricardo, line 291: the constraints, regularizers etc. should be passed in here

Issue #1524, opened 2020-05-06
Support features from NumPy array, for multidimensional features and speed


huonw:
### Description

In #1336, we noted that construction from a NumPy array is much faster than Pandas.

Also, we currently only support feature vectors, but feature matrices can be useful. For instance, for time series prediction, a multivariate time series associated with each node is best modelled as  a `(number of time series, number of points per series)` matrix (or transposed), rather than flattening it (see #1527).

Issue #1336, opened 2020-04-23
Constructing n m-sized StellarGraphs is much slower than one (n*m)-sized one


huonw:
### Describe the bug

For graph classification, one is often constructing many small graphs, e.g. 10k graphs of 4-30 nodes. Constructing all of these as `StellarGraph`s seems to much slower than constructing a single StellarGraph of 40k-300k nodes.

(StellarGraph 1.0.0rc1 dogfooding)

huonw:
This is likely going to be much improved by things like #1291.

kieranricardo:
I tested this by comparing creating:

1.  10k graphs with 100 nodes and 300 edges
2. 1 graph with 1mil nodes and 3mil edges

Develop vs latest #1291:

1. 9s -> 0.4s
2. 64s -> 58s

So the absolute construction times goes down for both cases, but the relative slowdown is about 1000 times worse. The 58s is just from creating 10k empty stellargraphs.

kieranricardo:
Ah looks like creating an empty pandas dataframe with three columns accounts for ~50% of the overhead of creating an empty stellargraph (2.6ms vs 5.7ms). I'm don't think there's much we can do about the overhead of creating a large number of python objects :(

kieranricardo:
^ all timing done on #1291

huonw:
One way would potentially be construction from NumPy more directly (and, with #1228, even TensorFlow tensors), and avoid some of the Pandas overhead.

kieranricardo:
good idea! i thought numpy would have the same draw backs but creating an `np.empty(shape=(0,3))` is ~2000x faster than pandas

PR #1556, opened 2020-05-12
Add IndexedArray and allow construction from it and Numpy array

huonw:
This adds a very basic data frame class, `IndexedArray`, which consists of:

- a (multidimensional) NumPy array `values`, where the first dimension corresponds to "rows" (i.e. `values[0, ...]` is the first row, `values[1, ...]` is the second, etc.)
- an index of row labels, `index`, with one per row, so `index[idx]` is the label for the `values[idx, ...]` row.

This class is generally much more limited than a Pandas DataFrame (`IndexedArray` basically exists to be a nicer-to-use `(values, index)` tuple), but also has _much_ less overhead. For instance, the following shows the cost of constructing the two classes with or without data:

```python
import stellargraph as sg
import pandas as pd
import numpy as np

# empty
%timeit -n 100 sg.IndexedArray()
# 2.69 µs ± 129 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
%timeit -n 100 pd.DataFrame()
# 365 µs ± 24.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# data
N = 10000
idx = np.random.randint(2**63 - 1, size=N)
data = np.random.rand(N, 100)

%timeit -n 100 sg.IndexedArray(data, index=idx)
# 1.14 µs ± 40.2 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
%timeit -n 100 pd.DataFrame(data, index=idx)
# 144 µs ± 4.12 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

By allowing construction from this `IndexedArray` type (and a NumPy array too, by assuming that the IDs are 0, 1, 2, ...), we can reduce the overhead of building a `StellarGraph` object (#1336). The rows correspond to nodes. Construction from a NumPy array/`IndexedArray` also allows for more precise control over memory usage, because the feature arrays are passed around directly, without copies.

In addition, and more importantly, it opens the possibility of multidimensional node features (#1524), where each node has an associated matrix or higher-rank object. For a specific example: multivariate time series are most naturally modelled as an array of shape `(num time steps, num variates)` (or vice versa).

`IndexedArray` supports multidimensional rows in this PR, by passing in a high-rank array. However, `StellarGraph` does not, because there's a few places that assume vector-per-node features. Changing these would be better handled independently of this core change because it may involve some deprecations and other changes (#1561).

General notes:

- This is a more limited form of http://xarray.pydata.org/ , but I decided that (a) it was probably good to not add another dependency unless required, and (b) its overhead is similar to Pandas:
  ```python
  import xarray as xr
  %timeit -n 100 xr.DataArray()
  # 17.7 µs ± 825 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
  %timeit -n 100 xr.DataArray(data, coords={"dim_0": idx})
  # 148 µs ± 12.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
  ```
- We should have a demo discussing this: #1559
- This doesn't think about how to handle edges via NumPy at all (so we've still got Pandas overhead in any `StellarGraph` construction), although it will be applicable to edge features (#1327).
- Rendered docs: https://stellargraph--1556.org.readthedocs.build/en/1556/api.html#stellargraph.RowFrame

See: #1524

## Performance

With the benchmarks from #1558:

Construction time (`test_benchmark_creation`), median, in ms:

| features | nodes | edges | Pandas |  IndexedArray | speedup |
|---|---|---|---|---|---|
| None | 0 | 0 | 3.77 | 1.28 | 3.0× |
| None | 20000 | 100000 | 5.36 | 2.90 | 1.8× |
| 100 | 20000 | 100000 | 17.3 | 2.86 | 6.0× |

(It's potentially unexpected that the `IndexedArray` `None-20000-100000` is slower than `100-20000-100000`, but there's nothing that dependencies on the feature size in the `IndexedArray` code path, and they're well within error range of each other.)

Long-term memory usage (`test_allocation_benchmark_creation`), median, in KB:

| features | nodes | edges | Pandas |  IndexedArray | change |
|---|---|---|---|---|---|
| None | 0 | 0 | 7.0 | 5.3 | -1.7 (-24%) |
| None | 1000 | 5000 | 64.1 | 58.5 | 5.6 (-9%) |
| 100 | 1000 | 5000 | 459.6 | 58.5  | -401.1 (-87%) |

This reflects the features being guaranteed-shared with the original array, meaning that having features vs. not having features are the same. The memory cost of the `StellarGraph` object is just the graph structure.

Peak memory usage during construction (`test_allocation_benchmark_creation_peak`), median, in KB:

| features | nodes | edges | Pandas |  IndexedArray | change |
|---|---|---|---|---|---|
| None | 0 | 0 | 9.0 | 7.2 | -1.8 (-20%) |
| None | 1000 | 5000 | 84.2 | 78.3 | -5.9 (-7%) |
| 100 | 1000 | 5000 | 703.5 | 78.3  | -625.2 (-89%) |

Real world datasets (using an extended form of the notebook from #1547), formatted like `pandas  → numpy`.

| dataset | time (s) | memory (MB) | memory not shared with data (MB) | peak memory (MB) |
|---|---|---|---|---|
| Cora | 0.020 → 0.0019 | 15.6 → 15.6 | 15.6 → 0.062 | 46.8 → 32.0 |
| Reddit | 3.01 → 0.55 | 712 → 712 | 712 → 151 | 1868 → 1682(*) |

(*) This is the same as the peak memory required for for loading the data, so the true peak for just creating the `StellarGraph` is likely lower.

In summary, construction from a NumPy array/`RowFrame` is much faster (10× on Cora, 5.5× on Reddit), uses less "new" memory (i.e. doesn't duplicate feature arrays) and has lower peak memory usage.

## Some questions

- This `IndexedArray` type is reexported at the top level, like `sg.RowFrame`; is this appropriate?
- ~The `RowFrame` name is potentially a bit peculiar. I considered: `NodeFrame`, but this would be inappropriate after #1327; `DataFrame`, but having to be specific about `sg.DataFrame` vs. `pd.DataFrame` would be endlessly confusing. Can anyone think of better options?~ Renamed to `IndexedArray`.


huonw:
> (*) This is the same as the peak memory required for for loading the data, so the true peak for just creating the StellarGraph is likely lower.

I've suggested a fix to Python to allow us to measure the StellarGraph peak: 

- https://mail.python.org/archives/list/python-ideas@python.org/thread/QDWI37A4TJXOYUKULGPY2GKD7IG2JNDC/ 
- https://bugs.python.org/issue40630

stellargraph/core/graph.py
```diff
@@ -182,17 +203,17 @@ class StellarGraph:
...
             have an ID taken from the index of the dataframe, and they have to be unique across all
-            types.  For nodes with no features, an appropriate DataFrame can be created with
-            ``pandas.DataFrame([], index=node_ids)``, where ``node_ids`` is a list of the node
+            types.  For nodes with no features, an appropriate value can be created with
+            ``RowFrame(index=node_ids)``, where ``node_ids`` is a list of the node
             IDs. If this is not passed, the nodes will be inferred from ``edges`` with no features
             for each node.
 
-        edges (DataFrame or dict of hashable to Pandas DataFrame, optional):
+        edges (Numpy array, RowFrame, DataFrame or dict of hashable to RowFrame or Pandas DataFrame, optional):
```
kieranricardo, line 216: from what I can tell, I don't think `edges` can be constructed from a `RowFrame` because it needs column names

stellargraph/core/convert.py
```diff
@@ -134,6 +141,26 @@ def select_column(old_name):
 
         return ids, columns, features
 
+    def _convert_rowframe(self, type_name, data):
+        assert isinstance(data, (RowFrame, np.ndarray))
+        if self.selected_columns:
+            raise ValueError(
+                f"{self.name(type_name)}: expected a Pandas DataFrame when selecting columns {comma_sep(self.selected_columns)}, found {type(data).__name__}"
```
kieranricardo, line 148: I think this error will be raised when constructing from a `RowFrame` of edges

stellargraph/core/row_frame.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+
+        values_len = values.shape[0]
+
+        if index is None:
+            index = range(values_len)
+
+        if values_len != index_len():
+            raise ValueError(
+                f"values: expected {index_len()} rows to match the index value(s), found {values_len} rows"
```
kieranricardo, line 79: I think this could slightly re-worded for clarity, maybe"

```suggestion
                f"values: expected the index length {index_len()} to match the first dimension of values, found {values_len} rows"
```

stellargraph/core/convert.py
huonw: Yes, that's the intention; what you getting at here?

stellargraph/core/graph.py
huonw: Oops, copy-paste mistake.

stellargraph/core/row_frame.py
huonw: I like that 👍

stellargraph/core/convert.py
kieranricardo: just referencing the first comment above ^

huonw: Ah, right.

huonw:
It's a great idea to optional conversion from xarray! I filed https://github.com/stellargraph/stellargraph/issues/1575.

huonw:
> Real world datasets (using an extended form of the notebook from #1547)

I realised that #1547 landed, but I forgot to add the updated notebook that this is referring to as I intended. I've now added it. I think the changes are pretty innocuous ( https://app.reviewnb.com/stellargraph/stellargraph/pull/1556/files/ for the pretty view) so unless someone jumps in with comments in the next hour or so, I'll merge.

Issue #1526, opened 2020-05-06
Add feature-slicing/time-series generator from StellarGraph for time-series/sequential data


huonw:
### Description

Using #1524, we should try to create a data generator that yields slices of features, for use with time series and sequence algorithms and data, like `GraphConvolutionLSTM`.

Currently `GraphConvolutionLSTM` works directly with adjacency matrices and feature matrices.

PR #1560, opened 2020-05-13
Unit tests gcn lstm 

habiba-h:
These are preliminary unit tests for `gcn_lstm`. They will change with the new changes in `gcn_lstm` model.


stellargraph/layer/gcn_lstm.py
```diff
@@ -187,7 +187,7 @@ def call(self, features):
         return output
 
 
-@experimental(reason="Lack of unit tests and some code refinement", issues=[1131, 1132])
+# @experimental(reason="Lack of unit tests and some code refinement", issues=[1131, 1132])
```
kieranricardo, line 190: I think this PR doesn't fully take this out of experimental. I think your other PR is needed, plus maybe some polish work (in another PR) to use  to use `kernel_constraint` in the layers etc.

tests/layer/test_gcn_lstm.py
```diff
@@ -0,0 +1,139 @@
...
+def test_gcn_lstm_activations():
+    fx, fy, a = get_timeseries_graph_data()
+
+    gcn_lstm_model = GraphConvolutionLSTM(
+        seq_len=fx.shape[-2], adj=a, gc_layers=5, lstm_layer_size=[8, 16, 32, 64],
+    )
+    assert gcn_lstm_model.gc_activations == ["relu", "relu", "relu", "relu", "relu"]
+    assert gcn_lstm_model.lstm_activations == ["tanh", "tanh", "tanh", "tanh"]
+
+    with pytest.raises(ValueError):
```
kieranricardo, line 84: Its handy to check that we've got the right error message, something like:

```suggestion
    with pytest.raises(ValueError, match="Invalid number of activations.* graph convolution layer"):
```

Should work. And something similar below.

stellargraph/layer/gcn_lstm.py
huonw: I think we should also change how the model is constructed and used, to be more similar to the other models here: #1526, #1564. This is likely to require significant changes to the interface and thus would be best while it's still `@experimental`.

Issue #1527, opened 2020-05-06
Multivariate time-series forecasting for GCN-LSTM


timpitman:
### Description

Can we use Keras's existing multi-variate LSTM?

### User Story

**As** a ...

**I want to** [accomplish some things]

**So that** [achieve a larger goal]

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
One approach here might be generalising the transposition for graph convolution to include some reshaping to run graph convolution on `time steps * variate dimension`:

https://github.com/stellargraph/stellargraph/blob/ee3bd69435ba0a7b630dd0ae566f095c99167e66/stellargraph/layer/gcn_lstm.py#L178

PR #1564, opened 2020-05-14
Add SlidingFeaturesNodeGenerator for time series/sequence data

huonw:
This allows the feature vectors in a `StellarGraph` object to behave like a time series. The generator yields batches of sliding windows along each node's features. For input features of shape `T × ...` (for each of `N` nodes), this yields tensors of shape `B × N × W × ...` where `B` is the batch size and `W` is some fixed window size.

As an example of the windows, consider `W = 2`:

```
first node features:  [0, 1, 2, 3, 4]
B elem 0, first node: [0, 1]
B elem 1, first node:    [1, 2]
B elem 2, first node:       [2, 3]
B elem 3, first node:          [3, 4]
```

For supervised training of time-series->self prediction (i.e. predicting the values of a time series based on previous observations), this offers a `target_distance` parameter, that selects an element a certain distance ahead of each window as a target observation, e.g., for `target_distance=2`, the `<...>` denotes the selected target observation:

```
first node features:  [0, 1, 2, 3, 4]
B elem 0, first node: [0, 1]   <3>
B elem 1, first node:    [1, 2]   <4>
```

The generator also accepts a parameter of [type `slice`](https://docs.python.org/3/library/functions.html#slice), that can be used for train/test splits. That is, to create a training generator, one creates a slice denoting which examples are for training, and choses a distance, and calls `flow`. For example, training for 2-ahead predictions on the 20 observations starting at the 10th would look like:  `generator.flow(slice(10, 30), target_distance=2)`.

The current target selection scheme only works for predicting all elements of a time series from itself. This doesn't work if one wanted to use observations from one time series to predict another, or one had a multivariate time-series (#1527, #1580) and only wanted to predict a single variate.

This also adds optional support for using the generator to `GraphConvolutionLSTM`, but doesn't switch the demo notebook to using it, to keep this PR more focused.

See: #1526


huonw:
This builds on #1579, but can be reviewed before that lands. (Even if #1579 doesn't land, this just needs to be updated to swap some dimensions.)

huonw:
This is currently a very naive implementation of the windowing, but one can do it smarter using strides, to be O(1). I am starting with this “obvious” form for ease of review of the whole architecture. I’m thinking we can upgrade to a fancier form in future as a dedicated PR focused on just that.

> The current target selection scheme only works for predicting all elements of a time series from itself. This doesn't work if one wanted to use observations from one time series to predict another, or one had a multivariate time-series (#1527, #1580) and only wanted to predict a single variate.

Some ideas I had:

- support taking a function to transform the target feature array into the target
- potentially, to make the above more useful support `target_distance` being a slice or list in addition to an integer (e.g. have a single model learn 5, 15 and 60 minute ahead speed predictions)
- support taking a whole extra `targets` array of (multivariate) time series values of shape A x T x B, where A and B are arbitrary (although A would usually be the number of nodes) and T is at least as large as the number of windows. The first window would have targets `targets[:, 0, ...]` and so on.

stellargraph/mapper/sliding.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+        self._features = G.node_features(node_type=node_type)
+
+        self.window_size = window_size
+        self._batch_size = batch_size
+
+    def num_batch_dims(self):
+        return 1
+
+    def flow(self, sequence_iloc_slice, target_distance=None):
```
kjun9, line 47: It would be good to paste in some of the explanation in the issue description as part of the docstring here for what the arguments are referring to.

In particular, I think it may not be obvious that the query length is calculated after adding the target distance, instead of before? I'm guessing in most cases you'd use this with `slice(None)` to use the whole sequence, but if someone wanted to use a particular slice, e.g. `slice(0, 10)` they might expect the training set to have 10 examples, instead of `10 - target_distance`. Thoughts?

stellargraph/mapper/sliding.py
```diff
@@ -0,0 +1,138 @@
...
+        # if there's not enough data to fill one window, there's a problem!
+        if self._num_windows <= 0:
+            if target_distance is None:
+                target_str = ""
+            else:
+                target_str = f" + target_distance={target_distance}"
+
+            total_sequence_samples = features.shape[1]
+            start, stop, step = sequence_iloc_slice.indices(total_sequence_samples)
+            assert step == 1
```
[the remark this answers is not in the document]
kjun9: What's the purpose of this assert? Hasn't it already been checked that `sequence_iloc_slice` has `step=1` on line 69, or is it more for readability?

stellargraph/layer/gcn_lstm.py
```diff
@@ -244,6 +247,19 @@ def __init__(
...
+                    f"generator: expected a SlidingFeaturesNodeGenerator, found {type(generator).__name__}"
+                )
+
+            if seq_len is not None or adj is not None:
+                raise ValueError(
+                    "expected only one of generator and (seq_len, adj) to be specified, found multiple"
+                )
+
+            adj = generator.graph.to_adjacency_matrix(weighted=True).todense()
+            seq_len = generator.window_size
```
kjun9, line 262: Given that this class is still in experimental, maybe this is a good opportunity to rename `seq_len` to be `window_size` for consistency?

stellargraph/mapper/sliding.py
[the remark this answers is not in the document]
huonw: Yeah, for readability: 

- it's an assert (rather than a real `throw ValueError(...)`) because it is just syncing with the code on line 69 (e.g. if that one is changed, this one should be too)
- it's communicating that `step` is being ignored on purpose, because it has an uninteresting value

I'll add a brief comment.

stellargraph/mapper/sliding.py
huonw: Oh, yeah, this whole class lacks documentation. I'll add.

stellargraph/layer/gcn_lstm.py
huonw: I'd prefer not to "contaminate" this PR too many with things that may require changes elsewhere (e.g. in the demo, and/or merge conflicts with #1560). In addition, I think there's a bit more work required here to make this class super clean, e.g.:

- switching the `generator` parameter to be more usable (making `seq_len` etc. optional)
- potentially even removing the `seq_len` parameters and just using `generator`

Issue #1549, opened 2020-05-12
Knowledge graph: generalise, increase flexibility and reduce duplication


huonw:
### Description

Currently knowledge graph support has a lot of duplication between the two similar `DistMult` and `ComplEx` models. With #1521 and #1539, we have some fairly different models in `RotatE` and `RotationE`/`RotationH`. Across these classes we have significant amounts of duplication, but a fairly good idea of what's "real" duplication and what is coincidental. We should leverage these to create core abstractions that make it easy to control the existing models and implement new ones.

huonw:
Extra refactoring: https://github.com/stellargraph/stellargraph/pull/1522#discussion_r424833521

PR #1572, opened 2020-05-18
Use interleaved real/imag embeddings for complex KG models

huonw:
Our complex models previously had separate `Embedding`s layers for the real and imaginary parts of their node or edge-type embeddings. This is somewhat convenient, but it's even more convenient to pass around a single buffer. This PR represents a complex embedding `[a, b, ...]` as `[Re(a), Im(a), Re(b), Im(b), ...]`.

This helps generalise the models (#1549, #1573), because all of the models have the same basic property of storing one embedding layer for nodes and one for edge types; it becomes the responsibility of the relevant scoring function to interpret the flat array as an embedding.

This unfortunately seems to make training ComplEx slightly slower (develop: 240ms/step on WN18 in the notebook, this PR: 270ms/step), but I think the benefits are worth it.

Another option would be having the real/imaginary tensors consecutive (e.g. `[Re(a), Re(b), ..., Im(a), Im(b), ...]`). The interleaved approach allows using `ndarray.view` to do a zero-copy coercion from there real array to a user-facing complex array.

TODO:

- [ ] Run this on a Colab or other GPU machine to validate the slowdown


kieranricardo:
What about just storing the embeddings as complex tensors? This would eliminate the overhead of the strided slice and copying the embeddings of each batch, and calling `embedding.numpy()` would yield the complex tensors so no there's no extra copying to yield a user facing complex array.

huonw:
Do you have a specific idea for the best way to do that? I don't think the `Embedding` layer allows for complex types, e.g.:

```
import tensorflow as tf
idx = tf.constant([[0, 1]])

c = tf.keras.layers.Embedding(10, 20, dtype=tf.complex64)
c(idx) # ValueError: Expected float or integer dtype, got <dtype: 'complex64'>.

i = tf.keras.layers.Embedding(10, 20, embeddings_initializer="zeros", dtype=tf.int32)
i(idx)  
# array([[[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
#         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]]],
#       dtype=int32)>
```

We'd have to do something like implement our own ComplexEmbedding layer, using `tf.nn.embedding_lookup` directly internally. I guess this is possible, but it seems kinda unfortunate to have to maintain more code.

It also doesn't generalise to a model like RotationE or RotationH, where relations end up with two separate chunks of embeddings. See https://nbviewer.jupyter.org/gist/huonw/52d585e571fde62e252c54fbe668faa3#RotationE-and-RotationH : The m_r are computed like RotatE, with a phase argument, while the r_r are complex numbers. This would be phrased in the scheme in this PR, as an embedding layer of size 3n, that gets split into (for instance) `0:n` for the phase of m_r, `n::2` for the real part of r_r and `n+1::2` for the imag part of r_r.

kieranricardo:
hmm yeah its probably not maintaining our embedding layer. Are you able to do some timing on GPU? I'm concerned that the extra copying of tensors here might lead to a more dramatic slowdown on GPU and consume an unnecessary amount of GPU memory.

Another approach might be to remove the `_complexify` step and perform the ops directly on the strided tensor slices. I think this will prevent any extra tensor copies being made and passsed around

huonw:
I don't think there's any copying of tensors here? The slicing is cheap, including strided slicing. I believe the `complex` function doesn't copy things either, but maintains separate real and imag tensors.

I initially had things done on the manual strided tensors and it was noticeably slower than using TF's native complex support.

kieranricardo:
I can't prove it with tracemalloc for some reason (maybe because im wrong?) but checkout this colab notebook[https://colab.research.google.com/drive/1L-hveMZqh-Y79G1OvGeopMvB9GzCsppe?usp=sharing] and watch the RAM usage. When it hits the `tf.complex` step memory RAM usage doubles.  I observe the same thing locally which makes me think that  `tf.complex` is copying the tensors around.

> I initially had things done on the manual strided tensors and it was noticeably slower than using TF's native complex support.

ah that's unfortunate. Out of curiosity have you tested and timed the consecutive/concatenated approach?

huonw:
> I can't prove it with tracemalloc for some reason (maybe because im wrong?) but checkout this colab notebook and watch the RAM usage. When it hits the tf.complex step memory RAM usage doubles. I observe the same thing locally which makes me think that tf.complex is copying the tensors around.

Hm, interesting experiment. I can reproduce the memory increase, but only for the first call to `tf.complex`. For instance, if I add two cells like:

```python
del complex_tnsr
```
```python
complex_tnsr = tf.complex(tnsr, tnsr)
```

I see the memory usage decrease after the `del`, and _not_ increase after the second `complex_tnsr` assignment.

Also, that'll be running in eager mode, whereas our layer code will not.

> Out of curiosity have you tested and timed the consecutive/concatenated approach?

I think so; I believe it was the same as the interleaved approach, both with `tf.complex` and with manually implemented operations.

kieranricardo:
I can reproduce your results `del ..`  on colab, but locally the second `tf.complex` still allocates a bunch of memory. So maybe the colab RAM indicator is not 100% reliable?

Good point about the eager vs graph mode though. I dunno what will happen in graph mode but i still think it'd be good to have results from GPU timing just to check the training time regression isn't more significant on GPU

huonw:
Good idea. I'll investigate that next week.

huonw:
I tested the ComplEx notebook on Colab's instances:

| runtime type | ms/step develop | ms/step this PR |
|---|---|---|
| CPU | 620 | 260 |
| GPU | 150 | 150 |

I can't reproduce such a dramatic CPU slowdown locally (e.g. I checked with TF 2.2, in case there was some difference between 2.1 and 2.2, but no change).

huonw:
According to tensorboard's profiling either interleaved or sequential both result in a much more time spent in various `StridedSliceGrad` operations. Thus, I think the best approach is to just go back to the drawing board here and do #1549 in a way that doesn't require lots of strided things, and thus doesn't regress performance.

PR #1573, opened 2020-05-18
Share more code between knowledge graph models

huonw:
This introduces `KGModel`, as a superclass to hold the common functionality across the different knowledge graph models. This functionality includes:

- argument validation
- the preprocessing in `in_out_tensors` and `__call__`
- the ranking code

This becomes entirely driven by the scoring layers, via the `KGScoreLayer` abstract superclass for them. This superclass has some moving parts too:

- creation of a pair of lists of `tf.keras.layers.Embedding` layers
- appropriate conversion of embedding layers into something user-facing (e.g. complex-ifying real arrays)
- doing the per-batch bulk scoring

After this PR, adding a new knowledge graph model basically just means defining a scoring layer with the appropriate functions, and so should be much simpler.

This builds on #1572.

See: #1549


huonw:
This builds on #1572, and so isn't ready to land directly, but it could be at least briefly reviewed now, to parallelise any feedback.

timpitman:
The abstraction looks good, though I wonder about KGScore - could that inherit from Layer in order to get a little more code shared between ComplExScore and DistMultScore?

huonw:
I initially inherited from `Layer` but decided against it, because I could imagine one having a complicated scoring "layer" that was actually a full model containing sublayers (e.g. there's algorithms like ConvE that do some sort of convolutions for scoring).

> I just have one question, does subclassing the scoring layers from KGScore and Layer intefer with loading/saving models?

No, I don't think so. AFAIK, the deserialisation either calls the models constructor or depickles the class instance; either way, the parent classes would get handled appropriately. Just so I can cover all bases, did you have specific concerns about the extra superclasses?

An empirical test:

```python
import stellargraph as sg
import tensorflow as tf
import numpy as np

G, _, test, _ = sg.datasets.WN18RR().load()

gen = sg.mapper.KGTripleGenerator(G, batch_size=100)

complex = sg.layer.ComplEx(gen, embedding_dimension=10)

model = tf.keras.Model(*complex.in_out_tensors())

original_predictions = model.predict(gen.flow(test))

model.save("/tmp/complex")
loaded_model = tf.keras.models.load_model("/tmp/complex")

loaded_model.summary()

loaded_predictions = loaded_model.predict(gen.flow(test))
np.array_equal(original_predictions, loaded_predictions) # True
```

`summary()` output:

```
Model: "model_4"
__________________________________________________________________________________________________
Layer (type)                    Output Shape         Param #     Connected to                     
==================================================================================================
input_13 (InputLayer)           [(None, 1)]          0                                            
__________________________________________________________________________________________________
input_14 (InputLayer)           [(None, 1)]          0                                            
__________________________________________________________________________________________________
input_15 (InputLayer)           [(None, 1)]          0                                            
__________________________________________________________________________________________________
embedding_10 (Embedding)        (None, 1, 20)        818860      input_13[0][0]                   
                                                                 input_15[0][0]                   
__________________________________________________________________________________________________
embedding_11 (Embedding)        (None, 1, 20)        220         input_14[0][0]                   
__________________________________________________________________________________________________
compl_ex_score_5 (ComplExScore) (None, 1)            0           embedding_10[0][0]               
                                                                 embedding_11[0][0]               
                                                                 embedding_10[1][0]               
==================================================================================================
Total params: 819,080
Trainable params: 819,080
Non-trainable params: 0
__________________________________________________________________________________________________
```

huonw:
#1572 regressed performance significantly, so that was closed. I've now rewritten this to use a different approach: store a list of `Embedding` layers for nodes and edge types, and pass around lists of those, tensors and NumPy arrays, rather than trying to pass around a single instance for nodes/edge types. This allows models that have multiple components to the embeddings for nodes or edge types (e.g. real and imaginary) to have a separate `Embedding` layer for each one.

This seems to have the same performance as `develop`, based on some testing on Google Colab (which is a significant improvement over https://github.com/stellargraph/stellargraph/pull/1572#issuecomment-634392263).

This change is visible in: https://github.com/stellargraph/stellargraph/pull/1573/files/17af8e8e2e867bfb5e5b1e371e54b3c1bdc690d5..9a34d0af312e202e0bebc7bb8cc81c8b61f116ca

huonw:
The changes above were mostly reverting the scoring functions (etc.) to be as it was before, so I'm going to say that this has been sufficiently reviewed and merge.

PR #1579, opened 2020-05-20
Change input data to (batch, nodes, timesteps) in GCN-LSTM

huonw:
Previously the `GraphConvolutionLSTM` model had its input in the order `(batch, timesteps, nodes)`, which makes a lot sense with a pure time-series focus, since the each node is effectively a separate sensor within the time step.

This PR switches to put the nodes first, like `(batch, nodes, timesteps)`, because this may be slightly cleaner for for graphs models. In particular, it matches up more closely with how a `StellarGraph` object will store time series data, with all of the information for each node contained within the "row" for that node (#1561).

The multivariate generalisation then has shape like `(batch, nodes, timesteps, variates)` (#1527).

This can/should wait on #1560 to land first, so that this PR can be appropriately validated.


huonw:
Hi @habiba-h, similar to #1580, I'd appreciate your input here. Does this change seem reasonable? Is there some key aspect of time series that means we should retain the old ordering?

huonw:
@kieranricardo I unrequested you, because you've already got a lot of other reviews from me, but if you're enthusiastic feel free to review (especially since you've reviewed the other GCN-LSTM work).

stellargraph/layer/gcn_lstm.py
```diff
@@ -129,8 +129,7 @@ def build(self, input_shapes):
             input_shapes (list of int): shapes of the layer's inputs (node features and adjacency matrix)
```
kjun9, line 129: Should this be something like
```suggestion
            input_shapes (list of int): shapes of the layer's inputs (batch dimension, number of nodes, number of time steps)
```
?

stellargraph/layer/gcn_lstm.py
```diff
@@ -164,7 +164,7 @@ def call(self, features):
         Applies the layer.
 
         Args:
-            inputs (ndarray): node features (size B x T x N), where B is the batch size, T is the
+            inputs (ndarray): node features (size B x N x T), where B is the batch size, T is the
```
kjun9, line 167:
```suggestion
            features (ndarray): node features (size B x N x T), where B is the batch size, T is the
```

stellargraph/layer/gcn_lstm.py
```diff
@@ -293,6 +296,8 @@ def __init__(
                 )
             )
 
+        self._layers.append(Permute((2, 1)))
```
kjun9, line 299: maybe a short comment here would be helpful
```suggestion
        # put time-sequence dimension last for LSTM layers
        self._layers.append(Permute((2, 1)))
```

stellargraph/layer/gcn_lstm.py
huonw: Ah, I think this was actually referring to the inputs, so I'll switch it to

```
input_shapes (list of int): shapes of the layer's inputs (the batches of node features)
```

PR #1580, opened 2020-05-20
Support multivariate time series in GCN-LSTM

huonw:
Work in progress towards multivariate -> multivariate prediction with GCN-LSTM.

This makes the internals of GCN-LSTM always work with a rank-4 tensor: batch B × nodes N × timesteps T × variates V. (The existing univariate code is handled by promoting the B × N × T input tensor to B × N × T × 1 and then collapsing it back to B × N × T on output.)

The extra variates dimension is handled differently for the two components of GCN-LSTM:

- for GCN, each variate is treated as an independent feature, so it's a graph convolution with node features of size TV
- for LSTM, each node is treated as independent variates, and so the V variates are included in that, so it's LSTM of sequences of feature dimension NV

This is performed by splitting the layers stored in the `GraphConvolutionLSTM` class into separate lists, and putting various data reshaping operations around them.

At the moment, the input generator means that the output dimension is the same as the input dimension. This means it's doing predictions for every variate of the original sequence, for every node, but it should be able to be generalised to any sized output.

Notes:

- Consider not requiring the same dimensional input and output (e.g. multivariate input to single-variate output, for using many sensors to predict a specific property) #1630
- Consider moving the decoder dense layer out of the model, to be more similar to the other GCN etc. layers, that allow the user to slot on arbitrary decoders #1630
- This depends on the generator in #1564, which itself depends on swapping the input tensor order in #1579

See: #1527


huonw:
Hi @habiba-h, if you get a chance, I'd appreciate if you could look over the PR description and let me know if what I've described seems like a reasonable approach to generalising to multivariate input series. (Feel free to also look over the code 😄 )

habiba-h:
Hi @huonw, the high level description looks good. 
Although in `Keras` `LSTM`, the inputs are of these dimensions: 
`inputs: A 3D tensor with shape [batch, timesteps, feature]`
So for anyone wanting to use it will find this unnatural. But if you find it reasonable, feel free to change it as you like

huonw:
Yeah, I agree it doesn't match the Keras convention, but with #1564, users will be able to use `StellarGraph` as input, and so won't have to deal with this as directly.

huonw:
As with #1564, this depends on other PRs but could be reviewed in parallel, to at least validate the approach is vaguely reasonable, and to discuss the TODOs, like whether they're important for this PR or could be done as later work.

huonw:
This is ready for review.

stellargraph/layer/gcn_lstm.py
```diff
@@ -299,51 +304,72 @@ def __init__(
...
+        for layer in self._lstm_layers:
+            h_layer = layer(h_layer)
+
+        h_layer = Dropout(self.dropout)(h_layer)
+        h_layer = self._decoder_layer(h_layer)
+
+        if self.multivariate_input:
+            # flatten things out to the multivariate shape
+            # shape B x N x V
+            h_layer = Reshape((self.n_nodes, self.variates))(h_layer)
```
kjun9, line 371: Thanks for all the comments here! Nice and clear to follow

Issue #1531, opened 2020-05-06
Generalise Cluster-GCN


huonw:
### Description

Cluster-GCN is a training method that can potentially be used for many algorithms, e.g. "Cluster-GAT" and "Cluster-APPNP". We should generalise.

This could potentially be in the form of allowing `GAT` and so on to accept a `ClusterNodeGenerator` in addition to the `FullBatchNodeGenerator`. (If we do this, we can fold `ClusterGCN`'s implementation of `GCN`.)

Issue #1222, opened 2020-04-08
Support batch size > 1 for sparse tensors in GCN


huonw:
### Description

In https://github.com/stellargraph/stellargraph/pull/1205 the `GraphConvolution` layer is generalised to behave like the old `GraphClassificationConvolution`; specifically, it supports batches of multiple graphs, for dense adjacency matrices. We could support batches of sparse adjacency matrices (that is, a rank-3 `SparseTensor`) too.

The missing detail is a `batch_dot` equivalent for sparse matrices. https://github.com/tensorflow/tensorflow/issues/9210#issuecomment-497889961 has one possible implementation, but there's some caveats in the discussion there.

PR #1585, opened 2020-05-21
Feature/1531 generalize cluster gcn

kieranricardo:
This PR generalizes the cluster training method to work for `GCN`, `GAT`, and `APPNP`. To do this `APPNPPropagation` and `GraphAttention` layers had to be changed (for the dense case) to handle adjacency matrices with a batch dimension of 1 to match the new format of `GraphConvolution`. 

I think it's now probably a good idea to deprecate `ClusterGCN` and re-write the demo to use `GCN`, `GAT`, and `APPNP`. I'll do this in a follow up PR :) 

See #1531

### DIFF
--- a/stellargraph/layer/appnp.py
+++ b/stellargraph/layer/appnp.py
@@ -250,15 +250,13 @@ class APPNP:
         # Copy required information from generator
         self.method = generator.method
         self.multiplicity = generator.multiplicity
-        self.n_nodes = (
-            generator.features.shape[0]
-            if isinstance(generator, FullBatchGenerator)
-            else None
-        )
         self.n_features = generator.features.shape[1]
-
-        # Check if the generator is producing a sparse matrix
         self.use_sparse = generator.use_sparse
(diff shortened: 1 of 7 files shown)


stellargraph/layer/appnp.py
```diff
@@ -137,19 +137,17 @@ def call(self, inputs):
...
-        features = K.squeeze(features, 0)
-        propagated_features = K.squeeze(propagated_features, 0)
-
         # Propagate the node features
         A = As[0]
-        output = (1 - self.teleport_probability) * K.dot(
-            A, propagated_features
-        ) + self.teleport_probability * features
+        if K.is_sparse(A):
+            propagated_features = K.squeeze(propagated_features, 0)
```
huonw, line 143: This could have a FIXME(#1222), since I think that this could be generalised in the same manner.

huonw: @kieranricardo did you ignore this on purpose? (Not a problem if you don't think it applies, but a reply at least would be appreciated...)

Issue #1591, opened 2020-05-25
Improve code re-use between Neo4JClusterGrenerator and ClusterGenerator


timpitman:
### Description

Identified in #1567, we could add a new (abstract?) base class used by both Neo4JStellarGraph and StellarGraph to share more functionality and reduce duplicated code.  This could reduce code duplication in samplers such as Neo4jClusterNodeGenerator too.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1594, opened 2020-05-26
Enhancement/1591 de duplicate neo4j cluster gcn

kieranricardo:
This PR removes `Neo4jClusterNodeGenerator` and allows `ClusterNodeGenerator` to accept a `Neo4StellarGraph`. This required a couple of changes:

- addition of `check_graph_for_ml` in `Neo4StellarGraph`
- addition of `unique_node_type` in `Neo4StellarGraph`
- only storing one feature in `ClusterNodeGenerator` instead of all features for memory efficiency

 It might be worth still having `Neo4jClusterNodeGenerator` class that thinly wraps `ClusterNodeGenerator` to match `Neo4JGraphSAGENodeGenerator` and keep our `Neo4j` API consistent. Thoughts? 

See #1591


stellargraph/connector/neo4j/graph.py
```diff
@@ -138,6 +139,50 @@ def to_adjacency_matrix(self, node_ids):
...
+
+        Returns:
+            If this graph has only one node type, this returns that node type, otherwise it raises a
+            ``ValueError`` exception.
+        """
+        node_type_query = f"""MATCH (n) RETURN distinct labels(n)"""
+        result = self.graph_db.run(node_type_query)
+        all_types = result.data()[0]["labels(n)"]
+        if len(all_types) == 1:
+            return all_types[0]
```
huonw, line 176: Is it a problem that the `labels` returns a list of individual labels (because Neo4j allows multiple labels per node), and so this is likely to be returning a list?

kieranricardo:
> In addition, these queries seem like they might be O(num nodes),

Yeah I think that's a good point. Assuming we're dealing with large graphs here maybe its best to avoid all the O(num nodes) and O(num edges) checks we usually perform in our Neo4j code paths.

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -113,15 +114,13 @@ def __init__(self, G, clusters=1, q=1, lam=0.1, name=None):
...
-        # Check if the graph has features
-        G.check_graph_for_ml()
-
         self.node_list = list(G.nodes())
 
-        # Check that there is only a single node type
-        _ = G.unique_node_type(
-            "G: expected a graph with a single node type, found a graph with node types: %(found)s"
-        )
+        # If the graph is in the StellaGraph format check that there
```
huonw, line 119:
```suggestion
        # If the graph is in the StellarGraph format check that there
```

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -113,15 +114,13 @@ def __init__(self, G, clusters=1, q=1, lam=0.1, name=None):
...
+        # is only a single node type and that the graph has features
+        if isinstance(G, StellarGraph):
+            G.check_graph_for_ml()
+            _ = G.unique_node_type()
```
huonw, line 123: This lost the more explanatory error message:

```suggestion
            _ = G.unique_node_type(
                "G: expected a graph with a single node type, found a graph with node types: %(found)s"
            )
```

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -113,15 +114,13 @@ def __init__(self, G, clusters=1, q=1, lam=0.1, name=None):
...
-
         self.node_list = list(G.nodes())
 
-        # Check that there is only a single node type
-        _ = G.unique_node_type(
-            "G: expected a graph with a single node type, found a graph with node types: %(found)s"
-        )
+        # If the graph is in the StellaGraph format check that there
+        # is only a single node type and that the graph has features
+        if isinstance(G, StellarGraph):
```
huonw, line 121: I wonder if we should do this special casing within StellarGraph and Neo4jStellarGraph. One approach might be:

- add a defaulted parameter `expensive_checks=False` to `check_graph_for_ml`
- `StellarGraph` does its normal checks for `expensive_checks=False` (and currently there's nothing extra when `expensive_checks=True`)
- `Neo4jStellarGraph` does nothing, or some O(1) work for `expensive_checks=False`, and O(graph size) work for `expensive_checks=True`

I'm not entirely sure of the best way to handle `unique_node_type` with `expensive_checks=False`, maybe something using `db.labels()` (https://neo4j.com/docs/operations-manual/current/reference/procedures/)?

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -144,7 +143,7 @@ def __init__(self, G, clusters=1, q=1, lam=0.1, name=None):
             print(f"{i} cluster has size {len(c)}")
 
         # Get the features for the nodes
-        self.features = G.node_features(self.node_list)
+        self.features = G.node_features(self.node_list[:1])
```
huonw, line 146: This change invalidates the comment, given its now just the node features of a single node and is more than a little confusing. Could you update the comment to discuss more specifically what's going on?

Potentially we could even change this to do what it "really means" rather than store a subset of the feature vectors (since storing a subset is apparently fine)?

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -62,7 +63,7 @@ class ClusterNodeGenerator(Generator):
 
     def __init__(self, G, clusters=1, q=1, lam=0.1, name=None):
 
-        if not isinstance(G, StellarGraph):
+        if not isinstance(G, (StellarGraph, Neo4jStellarGraph)):
```
huonw, line 66: We need a `_StellarGraphBase` base class or something to both more clearly define an API and also make these sort of checks easy 😄 

Is there an issue for this? If not, could you open one?

kieranricardo: ah i thought we had one but turns out we didn't - I filed #1620 😄

stellargraph/mapper/mini_batch_node_generators.py
kieranricardo: Oh yep I'll update the comment to what I previously had in Neo4jClusterGCN.

Yeah the proper fix would be to just store the feature shape, but it would also now require code changes to ClusterGCN, FullBatchNodeGenerator, FullBatchLinkGenerator, and all the fullbatch methods and didn't want to include in this PR :( I'll also file an issue and chuck in a fixme

stellargraph/connector/neo4j/graph.py
```diff
@@ -138,6 +139,48 @@ def to_adjacency_matrix(self, node_ids):
...
+        """
+        Checks if all properties required for machine learning training/inference are set up.
+        An error will be raised if the graph is not correctly setup.
+        """
+
+        if expensive_check:
+            num_nodes_with_feats_query = f"""
+                MATCH(n)
+                WHERE "features" in keys(n)
+                return COUNT(*)
```
huonw, line 152: This might get optimised to be equivalent, but this is probably more directly expressed as via https://neo4j.com/docs/cypher-manual/current/clauses/where/#property-existence-checking

```suggestion
                WHERE EXISTS(n.features)
                RETURN COUNT(*)
```

But it's probably even more efficient as:

```suggestion
                WHERE EXISTS(n.features)
                RETURN n LIMIT 1
```

and then checking `len(results.data())` or something. Since we only need to know that there's at least one node with features, rather than the exact count.

(Also, I've consistent-ified the capitalisation of the keywords.)

stellargraph/connector/neo4j/graph.py
kjun9: I'm not so sure that checking for the number of labels in the graph is the best way to check for node types since each node can have multiple labels, and so we don't have a good 1-to-1 mapping from Neo4j labels to stellargraph types.. 

I went with a different approach in https://github.com/stellargraph/stellargraph/pull/1595/files where we just assume that every node is of default type.

We would want to support heterogeneous graphs with Neo4j in the future, but might need a bit more logic to do the mapping (unique set of labels -> type?). So for now, I'd prefer not to raise an error here based on the existence of labels - thoughts @huonw ?

huonw: Sounds reasonable for now. Maybe we should have an issue covering Neo4j <-> SG types?

kjun9: I've filed https://github.com/stellargraph/stellargraph/issues/1625

Issue #1568, opened 2020-05-18
Improve Neo4J GraphSAGE sampler


timpitman:
### Description

We should evaluate the current GraphSAGE NEO4J sampler and improve it, eg:
* better Cypher sampling code
* not requiring the features to be in-memory

### User Story

**As** a ...

**I want to** [accomplish some things]

**So that** [achieve a larger goal]

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

huonw:
This may have some overlap with the `Neo4jStellarGraph` @kieranricardo introduces in #1567.

Issue #1592, opened 2020-05-25
Allow specifying ID and feature attributes when creating Neo4JStellarGraph


timpitman:
### Description

Currently these are hard-coded mapped to CORA - for general use, users should be able to specify.

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1595, opened 2020-05-26
Refactor Neo4j GraphSAGE components to use Neo4jStellarGraph

kjun9:
This updates Neo4j-GraphSAGE related classes to start using the new `Neo4jStellarGraph` class, instead of taking in a separate `neo4j_graphdb` instance along with a `StellarGraph`. 

Some particularly notable choices
* the neo4j samplers and generators don't extend the "normal" generator and graphwalk base classes anymore since those base classes explicitly type-check that the graph being passed in during construction is a `StellarGraph`, so a `Neo4jStellarGraph` will fail the type-check. If we end up going with the approach of adding an abstract stellargraph base class, that would allow these classes to extend those base classes again.
* there is an option to cache the features in memory by calling `graph.cache_nodes()` - this is currently being done in the demo notebooks, I think it's currently quite a lot slower without the caching for such a small graph.
* there's a few FIXME's due to `Neo4jStellarGraph` not supporting heterogeneous graphs or the creation of a graph schema.
* there's also a FIXME since some of the validation code in the sampler had to be removed because they were coming from subclassing the `GraphWalk` class which has the helper methods for validation. I considered moving those methods out to top-level methods but left it for now to not make the diff even larger.. but I can do that if that seems reasonable, otherwise could just wait until the abstract base stellargraph class is available.

The notebooks run for a while (more than the cell timeout set in CI) so I introduced parameters to reduce the number of epochs run on CI, but we didn't need this before on these notebooks, so something did make them slower despite caching the node features 🤔 maybe the use of `nodes.ID` instead of `id(nodes)`?

See #1568


stellargraph/connector/neo4j/mapper.py
```diff
@@ -266,27 +202,21 @@ def sample_features(self, head_nodes, batch_num):
...
-        node_type = self.head_node_types[0]
-
         max_hops = len(self.in_samples)
         max_slots = 2 ** (max_hops + 1) - 1
         features = [None] * max_slots  # flattened binary tree
 
         for slot in range(max_slots):
             nodes_in_slot = node_samples[slot]
-            features_for_slot = self.graph.node_features(nodes_in_slot, node_type)
+            features_for_slot = self.graph.node_features(nodes_in_slot)
```
kieranricardo, line 219: Feature lookups seem to be the most expensive op here, it might be worth doing something like:

```python
batch_nodes = np.concatenate(node_samples)
batch_features = self.graph.node_features(batch_nodes)
idx = 0
for slot in range(max_slots):
    features_for_slot = batch_features[idx: (idx+len(node_samples[slot])]
    idx += len(node_samples[slot])
```

stellargraph/connector/neo4j/graph.py
```diff
@@ -58,25 +66,94 @@ def nodes(self):
         result = self.graph_db.run(node_ids_query)
         return [row["node_id"] for row in result.data()]
 
-    def node_features(self, node_ids):
+    def cache_nodes(self, dtype="float32"):
```
kieranricardo, line 69: can you add a docstring for `cache_nodes`? I reckon this will be a really handy function

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -110,368 +108,46 @@
...
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "### Splitting the data"
+    "neo4j_graphdb = py2neo.Graph(host=default_host, port=None, user=None, password=None)\n",
+    "neo4j_sg = Neo4jStellarDiGraph(neo4j_graphdb)\n",
+    "neo4j_sg.cache_nodes()"
```
kieranricardo, line 113: I think we should chuck a comment here, explaining whats happening and to not do this for masive graphs!

stellargraph/connector/neo4j/mapper.py
kieranricardo: ^ hopefully that reduces some of the query overhead but i'm not sure whether it'll be noticeable 🤷‍♂️

kieranricardo:
@kjun9 how much slower are the notebooks running with `ID` instead of `id()`?

stellargraph/connector/neo4j/graph.py
```diff
@@ -58,25 +66,94 @@ def nodes(self):
...
+                # single node
+                valid = np.array([nodes is not None])
+                nodes = [nodes]
+
+        # None's should be filled with zeros in the feature matrix
+        features = np.zeros((len(nodes), self.node_feature_sizes()[self._node_type]))
+
+        # fill valid locs with features
+        feature_query = f"""
+            UNWIND $node_id_list AS node_id
```
huonw, line 104: Does this `UNWIND` cause problems in the `nodes is None` case too? Potentially we should substitute in everything except the `RETURN`?

stellargraph/connector/neo4j/graph.py
```diff
@@ -58,25 +66,94 @@ def nodes(self):
...
+            RETURN node.features as features
+            """
+        result = self.graph_db.run(feature_query, parameters={"node_id_list": nodes})
+        rows = result.data()
+
+        # this method currently doesn't handle any other invalid IDs. If there are other
+        # invalid ids, we should raise an error
+        if len(rows) != sum(valid):
+            raise ValueError(
+                "nodes: Found values that did not return any results from the database."
```
huonw, line 115: I wonder how easy it is for us to be specific about the values that we're missing?

stellargraph/connector/neo4j/graph.py
```diff
@@ -58,25 +66,94 @@ def nodes(self):
         result = self.graph_db.run(node_ids_query)
         return [row["node_id"] for row in result.data()]
 
-    def node_features(self, node_ids):
+    def cache_nodes(self, dtype="float32"):
+        nodes = self.nodes()
```
huonw, line 70: This ends up doing two O(num nodes) queries, in here to get the list of nodes, and then in `_node_features_from_db`. In addition, the one in `_node_features_from_db` is "unnecessarily" having to use `UNWIND` to look up each node in order. I wonder if it's better to try to do a single one, somehow? For instance, `_node_features_from_db(None)` could include the node IDs too (e.g. return an `IndexedArray` instead of a NumPy array directly).

stellargraph/connector/neo4j/graph.py
```diff
@@ -58,25 +66,94 @@ def nodes(self):
...
+            )
+        else:
+            return self._node_features_from_db(nodes)
+
+    def unique_node_type(self):
+        return self._node_type
+
+    def node_feature_sizes(self):
+        if self._node_feature_size is None:
+            # if feature size is unknown, take a random node's features
```
huonw, line 148: This is a little concerning to me, because it's pushing us towards having to handle cache invalidation and similar things like that. Maybe we could at least write down the assumptions a bit better, e.g. off the top of my head:

- all nodes have features as a single list (under the `features` property, but I see #1592)
- all nodes' features have the same size

stellargraph/core/graph.py
```diff
@@ -39,6 +39,70 @@
 NeighbourWithWeight = namedtuple("NeighbourWithWeight", ["node", "weight"])
 
 
+def _features(element_data, unique, name, ids, type, use_ilocs):
```
huonw, line 42: If this is going to move to the top level to be used by other files, maybe we could make the name a little more descriptive, to provide more context, e.g.

```suggestion
def extract_element_features(element_data, unique, name, ids, type, use_ilocs):
```

By _not_ including it in the `__all__` or the docs, it's still effectively private to the Stellargraph module.

stellargraph/connector/neo4j/graph.py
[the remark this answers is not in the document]
kjun9: Hmm I kind of wanted to avoid going beyond the scope of "make the Neo4j GraphSAGE related classes use Neo4jStellarGraph" as much as possible in this PR. Yeah I agree a smarter caching approach would be nice, but I created this function so that it lets us approximately maintain existing behaviour of having node features in memory. Thoughts?

stellargraph/connector/neo4j/graph.py
```diff
@@ -58,25 +66,94 @@ def nodes(self):
...
+    def _node_features_from_db(self, nodes):
+        if nodes is None:
+            where_clause = ""
+            valid = slice(None)
+        else:
+            where_clause = " WHERE node.ID = node_id"
+            if isinstance(nodes, np.ndarray):
+                valid = nodes != None
+                # we need to create a list of integers to run the neo4j query with
+                nodes = [int(node) for node in nodes]
```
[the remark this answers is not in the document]
kjun9: Ah oops forgot they don't have to be integers.. hmm okay so instead this needs to check if nodes are a numpy integer type and if so convert to int

stellargraph/connector/neo4j/graph.py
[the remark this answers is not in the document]
huonw: Sounds reasonable, with a name change 👍

stellargraph/connector/neo4j/graph.py
kjun9: Good catch!

stellargraph/connector/neo4j/graph.py
kjun9: Hmm yeah can do something like
```
WITH {{ID: node.ID, features: node.features}} AS node_data
RETURN node_data
```
then check the IDs being returned

kjun9:
> @kjun9 how much slower are the notebooks running with `ID` instead of `id()`?

I'll try to get some numbers after I've updated everything for the other review comments first, but I think it's significantly slower - I think each notebook took around 20 minutes to run on my laptop, whereas previously they both ran on CI without any additional parameters to reduce the number of epochs or anything

kjun9:
@kieranricardo 

Two epochs NOW:
```
Epoch 1/2
6/6 - 55s - loss: 1.8951 - acc: 0.2148 - val_loss: 1.7253 - val_acc: 0.3491
Epoch 2/2
6/6 - 55s - loss: 1.6599 - acc: 0.4259 - val_loss: 1.6003 - val_acc: 0.3831
CPU times: user 7.41 s, sys: 1 s, total: 8.41 s
Wall time: 1min 57s
```

Two epochs on DEVELOP:
```
Epoch 1/2
6/6 - 5s - loss: 1.8477 - acc: 0.2667 - val_loss: 1.7080 - val_acc: 0.3244
Epoch 2/2
6/6 - 4s - loss: 1.6145 - acc: 0.3889 - val_loss: 1.5580 - val_acc: 0.4077
CPU times: user 4.71 s, sys: 503 ms, total: 5.21 s
Wall time: 9.98 s
```

So yeah... MUCH slower, I'll try creating an index on the `ID` property, see if that helps

kjun9:
Okay performance is more roughly on par with the previous approach, after creating an index on the node IDs.
```
Epoch 1/20
6/6 - 6s - loss: 1.8489 - acc: 0.3000 - val_loss: 1.7169 - val_acc: 0.3733
Epoch 2/20
6/6 - 6s - loss: 1.6329 - acc: 0.4519 - val_loss: 1.5399 - val_acc: 0.5505
Epoch 3/20
6/6 - 5s - loss: 1.4546 - acc: 0.7185 - val_loss: 1.4103 - val_acc: 0.6981
Epoch 4/20
6/6 - 5s - loss: 1.3256 - acc: 0.8444 - val_loss: 1.2881 - val_acc: 0.7330
Epoch 5/20
6/6 - 5s - loss: 1.1793 - acc: 0.8704 - val_loss: 1.1824 - val_acc: 0.7732
Epoch 6/20
6/6 - 6s - loss: 1.0519 - acc: 0.9407 - val_loss: 1.0960 - val_acc: 0.7974
Epoch 7/20
6/6 - 6s - loss: 0.9147 - acc: 0.9444 - val_loss: 1.0275 - val_acc: 0.8060
Epoch 8/20
6/6 - 5s - loss: 0.8387 - acc: 0.9741 - val_loss: 0.9604 - val_acc: 0.8060
Epoch 9/20
6/6 - 5s - loss: 0.7444 - acc: 0.9519 - val_loss: 0.9007 - val_acc: 0.8068
Epoch 10/20
6/6 - 5s - loss: 0.6679 - acc: 0.9667 - val_loss: 0.8598 - val_acc: 0.8117
Epoch 11/20
6/6 - 5s - loss: 0.5972 - acc: 0.9741 - val_loss: 0.8273 - val_acc: 0.8130
```
but it seems that creating an index in Neo4j requires you to specify a label for the entity that you want to create an index on, and only queries that match for this specific label end up using the index, so we need the user to pass in this "node label" when constructing stellargraph.

(for reference)
* create index - https://neo4j.com/docs/cypher-manual/current/administration/indexes-for-search-performance/#administration-indexes-syntax
*  create uniqueness contrainst (which also creates an index) - https://neo4j.com/docs/cypher-manual/current/administration/constraints/#query-constraint-unique-nodes

stellargraph/connector/neo4j/mapper.py
kjun9: Thanks Kieran! I've updated it to use your suggestion, the benefits are probably gonna be more noticeable when the node features aren't cached though

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
kjun9: I've added a markdown cell above with a bit more description

stellargraph/connector/neo4j/graph.py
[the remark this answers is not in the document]
kjun9: Updated to call `node.item()` which should do the right thing?

stellargraph/connector/neo4j/graph.py
kjun9: I've written down the assumptions in the docstring - what do you think?

demos/connector/neo4j/load-cora-into-neo4j.ipynb
```diff
@@ -470,12 +470,28 @@
...
+  },
   {
    "cell_type": "code",
    "execution_count": 14,
    "metadata": {},
    "outputs": [],
-   "source": []
+   "source": [
+    "node_id_constraint = \"\"\"\n",
+    "    CREATE CONSTRAINT\n",
```
huonw, line 487: I wonder if we could potentially have `Neo4jStellarGraph` check for a constrain on the ID (and maybe emit a warning if it doesn't exist)?

stellargraph/connector/neo4j/graph.py
```diff
@@ -51,32 +71,139 @@ def nodes(self):
...
+        if nodes is None:
+            feature_query = f"""
+                {self._match_node()}
+                {return_node}
+                """
+            result = self.graph_db.run(feature_query)
+            rows = result.data()
+            return IndexedArray(
+                np.array([node["node_data"]["features"] for node in rows]),
+                index=[node["node_data"]["ID"] for node in rows],
```
huonw, line 114: Does iterating over this twice work correctly? I think for some databases, the result cursor is a single use object and can't be used again?

stellargraph/connector/neo4j/graph.py
```diff
@@ -51,32 +71,139 @@ def nodes(self):
...
+            )
+        else:
+            if isinstance(nodes, np.ndarray):
+                valid = nodes != None
+                # we need to create a list with python types to run the neo4j query with
+                nodes = [node.item() if node is not None else node for node in nodes]
```
huonw, line 120: Nice! The `item` method seems perfect for this!

stellargraph/connector/neo4j/graph.py
```diff
@@ -34,14 +39,29 @@ class Neo4jStellarGraph:
...
         self._is_directed = is_directed
+        self._node_feature_size = None
+        self._nodes = None
+
+        # FIXME: methods in this class currently only support homogeneous graphs with default node type
+        self._node_type = globalvar.NODE_TYPE_DEFAULT
+
+    def _match_node(self):
+        if self.node_label is not None:
+            return f"MATCH (node:{self.node_label})"
```
huonw, line 62: I think this is vulnerable to a Cypher injection attack, e.g. if someone sets `node_label = "x) ..."` they can run arbitrary Cypher in the `...`. (or something like that). py2neo offers https://py2neo.org/v4/cypher/index.html#py2neo.cypher.cypher_escape which we could use either on a case-by-case basis like here:

```suggestion
            return f"MATCH (node:{cypher_escape(self.node_label)})"
```

Or by storing the labelled escaped. For either case, it might be good to do some basic "Hungarian notation", e.g. have the raw one be `self.raw_node_label` and (if we store it) have the escaped one be `self.cypher_node_label` or `self.escaped_node_label` or something.

Thoughts?

stellargraph/connector/neo4j/graph.py
```diff
@@ -51,32 +71,139 @@ def nodes(self):
...
+    def node_feature_sizes(self):
+        """
+        Get the feature sizes for the node types in the graph.
+
+        This method obtains the feature size by sampling a random node from the graph. Currently
+        this class only supports a single default node type, and makes the following assumptions:
+
+        - all nodes have features as a single list
+
+        - all nodes' features have the same size
```
huonw, line 192:
```suggestion
        - all nodes' features have the same size
        
        - there's no mutations that change the size(s)
```

stellargraph/connector/neo4j/graph.py
huonw: One extra addition, but otherwise great!

stellargraph/connector/neo4j/graph.py
```diff
@@ -51,32 +71,139 @@ def nodes(self):
...
+            A dictionary of node type and integer feature size.
+        """
+        if self._node_feature_size is None:
+            # if feature size is unknown, take a random node's features
+            feature_query = f"""
+                {self._match_node()}
+                RETURN node.features as features LIMIT 1
+                """
+            result = self.graph_db.run(feature_query)
+            self._node_feature_size = len(result.data()[0]["features"])
```
huonw, line 204: Potentially we could optimise this slightly in two ways:
- by doing the length-ing in the database and so send an integer, not the potentially-large feature vectors over the network
- avoiding create the `list` and `dict` structures with `run`+`data` when we're fetching a single result, since Py2neo has a special case function for that operation https://py2neo.org/v4/database.html#py2neo.database.Transaction.evaluate (https://py2neo.org/v4/database.html#py2neo.database.Cursor says "For queries that are expected to return only a single value within a single record, use the evaluate() method. This will return the first value from the next record or None if neither the field nor the record are present")

```suggestion
                RETURN size(node.features) LIMIT 1
                """
            self._node_feature_size = self.graph_db.evaluate(feature_query)
```

stellargraph/connector/neo4j/mapper.py
```diff
@@ -169,23 +131,25 @@ def sample_features(self, head_nodes, batch_num):
...
             for that layer.
         """
-        nodes_per_hop = self.sampler.run(
-            self.neo4j_graphdb, nodes=head_nodes, n=1, n_size=self.num_samples
-        )
-        node_type = self.head_node_types[0]
+        nodes_per_hop = self.sampler.run(nodes=head_nodes, n=1, n_size=self.num_samples)
+
+        batch_nodes = np.concatenate(nodes_per_hop)
+        batch_features = self.graph.node_features(batch_nodes)
```
huonw, line 137: > this is currently being done in the demo notebooks, I think it's currently quite a lot slower without the caching for such a small graph.

You mentioned this in sprint close, but I don't recall if you mentioned some numbers (and, it'd be nice to write them down for reference). Do you happen to remember approximately what the slowdown was?

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -25,7 +25,7 @@
    "source": [
```
huonw, line 25: <blockquote>Note that this should be avoided for larger graphs.</blockquote><p> </p><p>Maybe "Note that this should be avoided for graphs where the node features are large" or something, since it's fine if most of the size is in the edges?</p><p> </p><p>(Also applies to undirected.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1595/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

huonw, line 25: <p>Strangely, this isn't rendering correctly on ReviewNB, but it is on Github and RtD, so I guess it's not a problem. As a double-check, I assume it renders correctly in a local Jupyter instance?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1595/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

huonw, line 25: <p>This demo would be a little more convincing if the labels were read from Neo4j too. I guess that's potentially a little awkward?</p><p> </p><p>(Also applies to undirected.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1595/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

demos/connector/neo4j/undirected-graphsage-on-cora-neo4j-example.ipynb
```diff
@@ -19,6 +19,15 @@
     "<table><tr><td>Run the latest release of this notebook:</td><td><a href=\"https://mybinder.org/v2/gh/stellargraph/stellargraph/master?urlpath=lab/tree/demos/connector/neo4j/undirected-graphsage-on-cora-neo4j-example.ipynb\" alt=\"Open In Binder\" target=\"_parent\"><img src=\"https://mybinder.org/badge_logo.svg\"/></a></td><td><a href=\"https://colab.research.google.com/github/stellargraph/stellargraph/blob/master/demos/connector/neo4j/undirected-graphsage-on-cora-neo4j-example.ipynb\" alt=\"Open In Colab\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\"/></a></td></tr></table>"
```
huonw, line 19: <p>Maybe not necessary any more?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1595/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='38'/>

demos/connector/neo4j/load-cora-into-neo4j.ipynb
kjun9: That's a good idea, yeah that should be possible with `CALL db.constraints`

stellargraph/connector/neo4j/graph.py
kjun9: Hmm right, the following seems to work without problem:
```python
query = f"""
MATCH (node:paper)
RETURN node.ID as id LIMIT 5
"""

result = neo4j_graphdb.run(query)
data = result.data()

for d in data:
    print(d)

for d in data:
    print(d)
```
```
{'id': 31336}
{'id': 1061127}
{'id': 1106406}
{'id': 13195}
{'id': 37879}
{'id': 31336}
{'id': 1061127}
{'id': 1106406}
{'id': 13195}
{'id': 37879}
```
and this code path should be running during `cache_all_nodes_in_memory`

stellargraph/connector/neo4j/graph.py
kjun9: Makes sense, although this means we need to import `py2neo` here, so we should move `py2neo` from extra requires to just our regular list of requirements?

huonw: I think it'd be better to avoid that, so people who aren't using neo4j don't need the dependencies. This also makes it easier for us to be sure that we're not accidentally trying to connect to Neo4j (e.g. on CI). I think it's fine to do an `import py2neo` scoped within this function (or `__init__`) similar to how `Neo4jBatchedNodeGenerator` handles it.

Potentially we should have a `def _py2neo(): ...` function in `neo4j/__init__.py` that manages the import and the error messages for us?

stellargraph/connector/neo4j/graph.py
huonw: Oh, right, it's the `data()` method, that does the conversion to a list.

demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.ipynb
kjun9: Yeah it's rendering correctly in my jupyter lab
image

but yeah I've also noticed sometimes ReviewNB doesn't render things correctly (I often see random new lines)

stellargraph/connector/neo4j/graph.py
kjun9: The import in `Neo4jBatchedNodeGenerator`'s been removed as part of this PR (don't need it anymore since it's handled in `Neo4jStellarGraph`), so I've just done the same thing in-line without putting it in `neo4j/__init__.py` - I guess we can move it as we need?

demos/connector/neo4j/load-cora-into-neo4j.ipynb
kjun9: There's now a bit of regex to find the correct constraint.. not 100% sure how reliable that is but I guess in worst case scenario the user sees a warning that doesn't apply. Thoughts?

stellargraph/connector/neo4j/mapper.py
kjun9: After commenting out the caching, I'm getting
```
Epoch 1/20
6/6 - 174s - loss: 1.8612 - acc: 0.2370 - val_loss: 1.7212 - val_acc: 0.3790
Epoch 2/20
6/6 - 172s - loss: 1.6545 - acc: 0.4926 - val_loss: 1.5768 - val_acc: 0.4413
```
compared to 6 seconds per epoch, so up to ~30 x slower with Cora.

Probably also quite heavily dependent on these parameters:
```
batch_size = 50
num_samples = [10, 5]
```

stellargraph/connector/neo4j/graph.py
```diff
@@ -36,18 +41,72 @@ class Neo4jStellarGraph:
...
+                "node_label: No node label provided. " + warn_msg,
+                RuntimeWarning,
+                stacklevel=2,
+            )
+        else:
+            constraint_query = """
+                CALL db.constraints
+                """
+            constraint_regex = re.compile(
+                f"^CONSTRAINT ON \( \w+:{node_label} \) ASSERT \(\w+.ID\) IS UNIQUE$"
```
huonw, line 75: The `\`s here might be working strangely because this isn't a raw string; it's somewhat of a coincidence that none of the characters after `\` here are meaningful escapes (like `\n`). Thus, maybe:

```suggestion
                rf"^CONSTRAINT ON \( \w+:{node_label} \) ASSERT \(\w+.ID\) IS UNIQUE$"
```

```suggestion
                f"^CONSTRAINT ON \\( \\w+:{node_label} \\) ASSERT \\(\\w+.ID\\) IS UNIQUE$"
```

kjun9:
> Just to double check, have you validated that the warning is emitted on a graph with an index but no constraint?

Yep, double-checked just then, with a constraint, you get both a constraint and an index, and with an index you only get an index, so we see the warning when it's just an index when running the graph construction cell in the notebook:
```python
neo4j_sg = Neo4jStellarDiGraph(neo4j_graphdb, node_label="paper")
neo4j_sg.cache_all_nodes_in_memory()
```
```
/Users/jun021/work/stellar/stellargraph/stellargraph/connector/neo4j/graph.py:317: ExperimentalWarning: Neo4jStellarGraph is experimental: the class is not tested (see: https://github.com/stellargraph/stellargraph/issues/1578). It may be difficult to use and may have major changes at any time.
  super().__init__(graph_db, node_label=node_label, is_directed=True)
/Users/jun021/work/stellar/stellargraph/stellargraph/core/experimental.py:89: RuntimeWarning: node_label: No uniqueness constraint found on IDs of nodes with label 'paper'. Queries may be slow to run and may experience unexpected behaviour if there are duplicate IDs. Please consider creating a uniqueness constraint on a label applied to all nodes in the graph.
  return func_to_wrap(*args, **kwargs)
```

huonw:
Awesome, thanks for checking 😄

stellargraph/connector/neo4j/graph.py
```diff
@@ -49,8 +50,50 @@ class Neo4jStellarGraph:
...
```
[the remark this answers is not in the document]
kjun9: Yeah seems like it send it back without the backticks, so don't need to escape here
```python
node_id_constraint = """
    CREATE CONSTRAINT
    ON (n:`my paper`)
    ASSERT n.ID IS UNIQUE
    """

tx = graph.begin(autocommit=True)
tx.evaluate(node_id_constraint)
graph.run("call db.constraints").data()
```
```
[{'name': 'constraint_5d2fd5d0',
  'description': 'CONSTRAINT ON ( my paper:my paper ) ASSERT (my paper.ID) IS UNIQUE'}]
```

Issue #1590, opened 2020-05-24
Acceleration for (Directed)GraphSAGELinkGenerator


guomk:
### Describe the bug

I have a directed graph of about 2e4 nodes and 5e6 edges. And I was trying to feed the `UnsupervisedSampler` to `DirectedGraphSAGELinkGenerator` but the process takes too long to finish. I would be nicer if this process could be accelerated and there should be some log to help track the process (Currently there is no way to tell how many samplings have been performed.)

### To Reproduce
I couldn't disclose the dataset but any large dataset should be able to reproduce this behavior. Specifically, this behavior happens when executing this code block.

```python
biased_walker = sg.data.BiasedRandomWalk(G, n=number_of_walks, length=length, p=2, q=.5)
unsupervised_samples = UnsupervisedSampler(G, nodes=nodes, walker=biased_walker)

generator = DirectedGraphSAGELinkGenerator(G, batch_size, in_samples=in_samples, out_samples=out_samples)
train_gen = generator.flow(unsupervised_samples)
```

### Observed behavior

The whole code block takes several hours to finish. Python kernel is shown as busy. No log output.
90+% GPU memory usage

Jupyter Lab log shows
> W tensorflow/core/kernels/data/generator_dataset_op.cc:103] Error occurred when finalizing GeneratorDataset iterator: Cancelled: Operation was cancelled

### Expected behavior

Faster execution with log.

### Environment

OS: Windows 10

Python version:  *3.7.7*

Package versions: *stellargraph==1.0.0, tensorflow==2.1.0*

Code running on RTX2070 Super (8GB)

### Additional context

N/A

kieranricardo:
Hi, thanks for filing an issue!

To help us reproduce your issues could you share your model paramaters. What is your `batch_size`, `in_samples`, `out_samples`, `number_of_walks`, and how many nodes are in `nodes`?

You also mention that your GPU usage is at 90+%, are you refferring to `Volatile GPU-Util` or `GPU Memory Usage`?

guomk:
Hi, Thanks for your reply.

Parameter setting
`batch_size=50`
`in_samples=[10, 5]`
`out_samples=[10, 5]`
`number_of_walks=32`
`len(nodes)=16555`

The GPU usage I was referring to is memory usage.

[...]

PR #1596, opened 2020-05-26
10x faster biased walker

kieranricardo:
This PR speeds up `BiasedRandomWalk` by about 10x using numpy optimizations.

See: #1590

### DIFF
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -438,16 +438,18 @@ class BiasedRandomWalk(RandomWalk):
 
                     # select one of the neighbours using the
                     # appropriate transition probabilities
-                    neighbours, weights = self.graph.neighbor_arrays(current_node, include_edge_weight=True)
-                    mask = (neighbours == previous_node)
+                    neighbours, weights = self.graph.neighbor_arrays(
+                        current_node, include_edge_weight=True
+                    )
+                    mask = neighbours == previous_node
                     weights[mask] *= ip
-                    mask |= (np.isin(neighbours, previous_node_neighbours))
+                    mask |= np.isin(neighbours, previous_node_neighbours)
                     weights[~mask] *= iq
 
-                    choice = naive_weighted_choices(rs, weights)
+                    probs = np.cumsum(weights) / weights.sum()
+                    choice = np.where(probs > np.random.random())[0][0]
                     previous_node = current_node
                     previous_node_neighbours = neighbours
-                    current_node = neighbours[choice]
                     current_node = neighbours[choice]
 
                     walk.append(current_node)
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -414,28 +414,6 @@ class BiasedRandomWalk(RandomWalk):
         if weighted:
             self._check_weights_valid()
 
-            # calculate the appropriate unnormalised transition probability, given the history of
-            # the walk
-            def transition_probability(nn):
-                nn, weight = nn
-
-                if nn == previous_node:  # d_tx = 0
-                    return ip * weight
-                elif any(nn == pn for pn, _ in previous_node_neighbours):  # d_tx = 1
-                    return weight
-                else:  # d_tx = 2
-                    return iq * weight
-
-        else:
-            # without weights
-            def transition_probability(nn):
-                if nn == previous_node:  ## d_tx = 0
-                    return ip
-                elif nn in previous_node_neighbours:  # d_tx = 1
-                    return 1.0
-                else:  # d_tx = 2
-                    return iq
-
         ip = 1.0 / p
         iq = 1.0 / q
 
@@ -460,15 +438,17 @@ class BiasedRandomWalk(RandomWalk):
 
                     # select one of the neighbours using the
                     # appropriate transition probabilities
-                    choice = naive_weighted_choices(
-                        rs, (transition_probability(nn) for nn in neighbours),
-                    )
+                    neighbours, weights = self.graph.neighbor_arrays(current_node, include_edge_weight=True)
+                    mask = (neighbours == previous_node)
+                    weights[mask] *= ip
+                    mask |= (np.isin(neighbours, previous_node_neighbours))
+                    weights[~mask] *= iq
 
+                    choice = naive_weighted_choices(rs, weights)
                     previous_node = current_node
                     previous_node_neighbours = neighbours
                     current_node = neighbours[choice]
-                    if weighted:
-                        current_node = current_node.node
+                    current_node = neighbours[choice]
 
                     walk.append(current_node)
 


stellargraph/data/explorer.py
```diff
@@ -451,24 +435,26 @@ def transition_probability(nn):
...
-                    # appropriate transition probabilities
-                    choice = naive_weighted_choices(
-                        rs, (transition_probability(nn) for nn in neighbours),
-                    )
+                    mask = neighbours == previous_node
+                    weights[mask] *= ip
+                    mask |= np.isin(neighbours, previous_node_neighbours)
+                    weights[~mask] *= iq
+
+                    probs = np.cumsum(weights) / weights.sum()
```
huonw, line 452: What about moving this to `naive_weighted_choices`, so that temporal random walks see the same speedup? (I think it already has numpy array input.)

stellargraph/data/explorer.py
```diff
@@ -451,24 +435,26 @@ def transition_probability(nn):
...
+                    choice = np.where(probs > rs.random())[0][0]
```
huonw, line 453: I think this can use http://docs.scipy.org/doc/numpy/reference/generated/numpy.searchsorted.html:

```suggestion
                    choice = np.searchsorted(probs, rs.random(), side="right")
```

Base on https://stackoverflow.com/a/25032853/1256624 this is significantly faster for large arrays, and only fractionally slower for small arrays:

| N | where | searchsorted |
|---|---|---|
| 3 | 2.1 | 2.7 |
| 10 | 2.2 | 2.6 |
| 100 | 2.5 | 2.8 |
| 1000 | 5.3 | 3.7 |
| 10000 | 26.3  | 11.5 |

Measured with:

```python
import numpy as np
N = 10000
aa = np.arange(-N,N)
%timeit -n 100000 np.searchsorted(aa, N/2)+1
%timeit -n 100000 np.where(aa>N/2)[0][0]
```

In particular, this is trading off a little bit of speed for better handling of "super-vertices" with many connections. What do you think?

stellargraph/data/explorer.py
```diff
@@ -414,30 +414,14 @@ def run(
...
-                    return 1.0
-                else:  # d_tx = 2
-                    return iq
-
-        ip = 1.0 / p
-        iq = 1.0 / q
+        if np.isinf(ip):
+            raise ZeroDivisionError("p: value is too small")
+        if np.isinf(iq):
+            raise ZeroDivisionError("q: value is too small")
```
huonw, line 424: I don't think this is necessary: the `1.0 / p` division still happens in Python-space, and so still throws an error. If it was `cast_func(1.0) / cast_func(p)` that might be different, but that's not the case.

```suggestion
```

stellargraph/data/explorer.py
```diff
@@ -451,24 +435,26 @@ def transition_probability(nn):
...
+                    if len(neighbours) == 0:
                         break
 
-                    # select one of the neighbours using the
-                    # appropriate transition probabilities
-                    choice = naive_weighted_choices(
-                        rs, (transition_probability(nn) for nn in neighbours),
-                    )
+                    mask = neighbours == previous_node
+                    weights[mask] *= ip
```
huonw, line 448: We're sure that this will never be mutating the weights inside the `EdgeData`, right?

kieranricardo: yep, looking up an array with a list/array indices always returns a copy even if the indices are sequeiential e.g.

```
import numpy as np

arr = np.arange(5)
copy = arr[[0, 1, 2, 4]]
copy[:] = 0
assert (copy == 0).all()
assert not (arr == 0).all()
```

will work

stellargraph/data/explorer.py
kieranricardo: oh that's a good idea! I think a random walk is more likely to end up at super-vetex so I reckon this trade-off is worth it :+!:

stellargraph/data/explorer.py
kieranricardo: I guess this `ZeroDivisionError` is slightly misleading. When `p` is a really small float `ip` is cast to `np.inf` e.g. 

```
q = 1e-100
iq = 1 / q
print(np.float32(iq)) # np.inf here
```

Maybe another error would more appropriate here because its not _technically_ a zero division error.

stellargraph/data/explorer.py
kieranricardo: good idea, it looks like that's the temporal random walks are the only other algorithm using this

stellargraph/data/explorer.py
```diff
@@ -303,26 +303,8 @@ def naive_weighted_choices(rs, weights):
...
-    # pick a place in the overall interval
-    x = rs.random() * running_total
-
-    # find the subinterval that contains the place, by looking for the
-    # first subinterval where the end is (strictly) after it
-    for idx, end in enumerate(subinterval_ends):
-        if x < end:
-            break
+    probs = np.cumsum(weights) / weights.sum()
+    idx = np.searchsorted(probs, rs.random(), side="right")
```
huonw, line 307: I think this is mildly suboptimal for a few reasons:

- it does two passes over the `weights` array, which (a) is probably slightly slower than it needs to be, and (b) may result in a cumulative sum > the final sum, so `probs[-1]` > 1.
- a division is much slower than a multiplication
- doing O(len(weights)) operations is slower than a single operation

I think we can address each of these, e.g.: an "ablation" study:

```python
import numpy as np
N = 10000
weights = np.random.rand(N).astype(np.float32)
rs = np.random.RandomState()

def sum_div():
    probs = np.cumsum(weights) / weights.sum()
    idx = np.searchsorted(probs, rs.random(), side="right")
    
def sum_div_inplace():
    probs = np.cumsum(weights) 
    probs /= weights.sum()
    idx = np.searchsorted(probs, rs.random(), side="right")

def last_div_inplace():
    probs = np.cumsum(weights)
    probs /= probs[-1]
    idx = np.searchsorted(probs, rs.random(), side="right")
    
def last_mul():
    probs = np.cumsum(weights)
    idx = np.searchsorted(probs, rs.random() * probs[-1], side="right")

%timeit -n 10000 sum_div()
# 44.2 µs ± 2.3 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit -n 10000 sum_div_inplace()
# 38.6 µs ± 1.4 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit -n 10000 last_div_inplace()
# 38.7 µs ± 1.83 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit -n 10000 last_mul()
# 36.2 µs ± 1.23 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

| N | time `sum_div` | time `last_mul` |
|---|---|---|
| 3 | 11.5 | 9.8 |
| 10 | 11.7 |  9.88 |
| 100 | 11.9 | 9.89 |
| 1000 | 15.6 | 12.4 |
| 10000 | 44.2 | 36.2 |

```suggestion
    probs = np.cumsum(weights)
    idx = np.searchsorted(probs, rs.random() * probs[-1], side="right")
```

stellargraph/data/explorer.py
```diff
@@ -451,24 +418,30 @@ def transition_probability(nn):
...
-                    )
+                    if weighted:
+                        neighbours, weights = self.graph.neighbor_arrays(
+                            current_node, include_edge_weight=True, use_ilocs=True
+                        )
+                    else:
+                        neighbours = self.graph.neighbor_arrays(
+                            current_node, use_ilocs=True
+                        )
+                        weights = np.ones(neighbours.shape, dtype=weight_dtype)
```
huonw, line 431: Does the 10x speedup still apply to this case?

stellargraph/data/explorer.py
huonw: That sounds like a good approach, but we could instead phrase it as something like "1/p should be representable by the data type fo the edge weights", and not hard code a bound.

stellargraph/data/explorer.py
kieranricardo: yep, in this PR `weighted=None` and `weighted=True` take the same amount of time. For `weighted=None` the `weights` array is just created here instead of in `neighbor_arrays`. 

Buuut, actually in develop `weighted=True` takes alot longer so its more like a 100x speedup.

Also casting the `iq` and `ip` instead of `weights` has made the `weighted=None` case more like a 20x speedup, and that's without your suggestion!

huonw: Nice!

Could you update the PR title (and thus merge commit message) and description as appropriate to reflect the new state?

stellargraph/data/explorer.py
huonw: @kieranricardo did you miss applying something along the lines of @kjun9's suggestion before re-requesting a review?

kieranricardo:
@huonw woops forgot to push to ValueError change sorry 😅

stellargraph/data/explorer.py
```diff
@@ -414,30 +399,21 @@ def run(
...
-                    return 1.0
-                else:  # d_tx = 2
-                    return iq
-
-        ip = 1.0 / p
-        iq = 1.0 / q
+        if np.isinf(ip):
+            raise ValueError(
+                f"p: value is too small. It must be possible to "
+                f"represent 1/p in {str(weight_dtype)}"
```
huonw, line 410: FWIW, I think it's harder to read to wrap this (but maybe your IDE is doing it automatically)?

Also, we can easily give the user a little more guidance by, e.g. including the actual value, so they can understand what went wrong more directly.

Finally, `f` string `{...}` formatting is already implicitly calling `str` on the value.

E.g.:

```suggestion
                f"p: value ({p}) is too small. It must be possible to represent 1/p in {weight_dtype}, but this value overflows to infinity."
```

stellargraph/data/explorer.py
```diff
@@ -414,30 +399,21 @@ def run(
...
+            )
+        if np.isinf(iq):
+            raise ValueError(
+                f"p: value is too small. It must be possible to "
+                f"represent 1/p in {str(weight_dtype)}"
```
huonw, line 415: This should say `q`, and including the changes above:

```suggestion
                f"q: value ({q}) is too small. It must be possible to represent 1/q in {weight_dtype}, but this value overflows to infinity."
```

Issue #1565, opened 2020-05-14
Use directed, weighted graph in GCN. 


davidxujiayang:
Hi,

In current preprocessing steps for GCN a symmetric adjacency matrix is built. I'm wondering does this mean that only undirected graph is supported? Can I use a weighted, directed graph, such that the edge between two nodes have two different weightes in the two directions? 

Thanks,
David

huonw:
Hi, @davidxujiayang. This is somewhat possible. One way to do it would be to use the `transform` parameter to [`FullBatchNodeGenerator`'s constructor](https://stellargraph.readthedocs.io/en/stable/api.html#stellargraph.mapper.FullBatchNodeGenerator).

This has some built-in options, but also allows a custom function. This function must take two arguments (`features` for the features array, `A` for the adjacency matrix) and return a tuple of two values (the new features array and adjacency matrix). This allows completely customising both parts of the data preparation.

For example, to keep a matrix directed:

```python
import stellargraph as sg

graph = sg.StellarDiGraph(...)

def transform(features, A):
    A_norm = ... # any normalisation of A
    return features, A_norm

generator = sg.mapper.FullBatchNodeGenerator(graph, transform=transform)
```

This could also be used to insert a new weighted adjacency matrix, in lieu of "real" support for weighted graphs:

```python
def transform(features, A):
    # replace the adjacency matrix with a weighted one
    A = graph.to_adjacency_matrix(weighted=True)

    A_norm = ... # any normalisation of A
    return features, A_norm
```

---

On the topic of making this more naturally supported, there's two parts:

1. weighted graphs
2. directed graphs

### Weighted

I think weighted graphs are relatively simple: if we allow passing the `weighted=True` parameter to `to_adjacency_matrix` in `FullBatchGenerator`, we can automatically support edge weights. If you're interested, we'd love to help you [open a pull request](https://guides.github.com/activities/forking/) to add this feature (if not, that's perfectly fine, and we'll do it some time over the next week or so). The code changes required are focused in the `FullBatchGenerator` class:

https://github.com/stellargraph/stellargraph/blob/e2086659d8b1c3921978cec6749a78699e6f522a/stellargraph/mapper/full_batch_generators.py#L56-L92

[...]

PR #1600, opened 2020-05-27
Add support for edge weights in full-graph methods

huonw:
This adds a `weighted` parameter that passes through to `to_adjacency_matrix` calls, to recognise/use edge weights.

This includes:

- GCN, GAT, APPNP, PPNP (via FullBatchNodeGenerator/FullBatchLinkGenerator and ClusterNodeGenerator)
- RGCN (via RelationalFullBatchNodeGenerator)
- graph classification including DGCNN (via PaddedGraphGenerator)

See: #1565


stellargraph/connector/neo4j/graph.py
```diff
@@ -87,11 +87,16 @@ def to_adjacency_matrix(self, node_ids):
...
                 comprising the subgraph. The adjacency matrix is
                 computed for this subgraph.
+            weighted (bool, optional): Must be false, at the moment.
 
         Returns:
              The weighted adjacency matrix.
         """
 
+        # this param is for compatibility with StellarGraph.to_adjacency_matrix
+        if weighted is not False:
```
kjun9, line 97: Could just be
```suggestion
        if weighted:
```

huonw: I did the `is` form so that it had to be explicitly `False` rather than passing in a `[]` or something... but that's probably unnecessary. 🤷‍♂️

PR #1603, opened 2020-05-28
Add support for edge weights in graphwave

huonw:


huonw:
Hm, don't review this yet... it previously wasn't testing with edge weights != 1, and doesn't work with such edge weights.

PR #1604, opened 2020-05-28
Add support for edge weights to Watch Your Step

huonw:
It may or may not be meaningful, but it's easy for Watch Your Step to support edge weights using a weighted adjacency matrix. This adds that option, so that a user can explore, as desired.

As part of doing this, this has to fix a mishandling of the sparse matrix shapes that this exposes: inside `AdjacencyPowerGenerator`, this algorithm constructs an adjacency sparse tensor (`Aadj_T`) and a "transition matrix" sparse tensor (`transition_matrix_T`) for the random walks that Watch Your Step is simulating. These tensors need to be manually constructed by passing in an array of indices and an array of data for the non-zero entries. These arrays need to have the same length, and the first row of indices corresponds to the position of the first element of data (and so on).

Previously, `AdjacencyPowerGenerator` was sharing the indices array between the two sparse tensors, but this is incorrect. The normalised transition matrix has non-zero elements at the same positions as the adjacency matrix, but the normalisation process can apparently result in them being stored in a different order. That is, the indices array for the transition matrix can be a shuffled version of the adjacency matrix. This means that using the transition matrix data with adjacency matrix indices gives an incorrect result. Without weights, this apparently wasn't observed by the test (i.e. the shuffled weight values must've been identical), but adding weights does flag it.

This relates to #1565.

See: #1615


tests/mapper/test_adjacency_generators.py
```diff
@@ -110,3 +110,22 @@ def test_partial_powers(barbell, num_powers):
         for j in range(num_powers):
             print(i, j)
             assert np.allclose(partial_powers[0, j, :], actual_powers[j][i, :])
+
+
+def test_weighted(barbell):
```
huonw, line 115: This is just a copy of test_partial_powers, with `to_adjacency_matrix(weighted=True)` (and similarly in `AdjacencyPowerGenerator`) instead of `to_adjacency_matrix`, and it seems like that's not quite right. Is there more to the algorithm than that?

Issue #1270, opened 2020-04-19
Some demo READMEs don't use the new structure


huonw:
### Description

In #1200/#1087, the demo READMEs were mostly rewritten to have a structure like https://github.com/stellargraph/stellargraph/blob/186b2e25c515c5938db16261546d70e8fbabfa42/demos/node-classification/README.md: 

- introduction to the task
- table of demos

Some demos were not:

- [ ] `calibration`
- [ ] `community_detection`
- [ ] `connector/neo4j`
- [ ] `ensembles`
- [ ] `use-cases` (#1139)

huonw:
The READMEs have been removed, but this still applies to the `index.rst`s in the `docs` folder, e.g.

- new form:
  - https://github.com/stellargraph/stellargraph/blob/6b9709b9/docs/demos/node-classification/index.rst
  - https://stellargraph.readthedocs.io/en/v1.0.0/demos/node-classification/index.html
- old form:
  - https://github.com/stellargraph/stellargraph/blob/6b9709b9/docs/demos/ensembles/index.rst
  - https://stellargraph.readthedocs.io/en/v1.0.0/demos/ensembles/index.html

PR #1608, opened 2020-05-28
Use a table to categorise the various "basics" demos

huonw:
A table is a little more structured, and makes it easy to include more info in an easily digestible way, like supported formats and a hint about performance.

I chose the subset of formats to list by searching for the relevant reading function on GitHub (e.g. [Pandas](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html): [`"pd.read_csv"`](https://github.com/search?q=%22pd.read_csv%22&type=Code), and [NetworkX](https://networkx.github.io/documentation/stable/reference/readwrite/index.html): [`"nx.read_gml"`](https://github.com/search?q=%22nx.read_gml%22&type=Code)) and including the ones that are most common, using this as a proxy for the formats that are encountered the most, in the wild. For NetworkX, I didn't include the edge list format (even though it was common), because that's better done via Pandas.

Caveats: 
- these searches are fairly noisy but it should still be correlated to the frequency of use. For example, it includes results from the repo itself and any duplicates of the source code, and will miss calls with different phrasings like `pandas.read_csv(...)` or `from networkx import read_gml; ... read_gml(...)`.
- these searches aren't limited to just people doing graph work but I'm not sure we can do much better. For example, it might be that people working with graphs use SQL much more than HDF5.

Rendered docs: https://stellargraph--1608.org.readthedocs.build/en/1608/demos/basics/index.html

See: #1270

### DIFF
--- a/docs/demos/basics/index.rst
+++ b/docs/demos/basics/index.rst
@@ -16,7 +16,7 @@ Find demos for a format
    * - :doc:`loading-pandas`
      - Anything `supported by Pandas <https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html>`__: CSV, TSV, Excel, JSON, SQL, HDF5, many more
      - Good
-     - Via Pandas and `scikit-learn <http://scikit-learn.github.io/stable>`__ and more
+     - Via Pandas, `scikit-learn <http://scikit-learn.github.io/stable>`__ and more
    * - :doc:`loading-networkx`
      - Anything `supported by NetworkX <https://networkx.github.io/documentation/stable/reference/readwrite/index.html>`__: Adjacency lists, GEXF, GML, GraphML, Shapefiles, many more
      - Poor
--- a/docs/demos/basics/index.rst
+++ b/docs/demos/basics/index.rst
@@ -12,15 +12,19 @@ Find demos for a format
    * - Demo
      - Data formats
      - Performance
+     - Data preprocessing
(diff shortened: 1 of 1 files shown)


docs/demos/basics/index.rst
```diff
@@ -1,15 +1,30 @@
...
+   :header-rows: 1
+
+   * - Demo
+     - Data formats
+     - Performance
+     - Data preprocessing
+   * - :doc:`loading-pandas`
+     - Anything `supported by Pandas <https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html>`__: CSV, TSV, Excel, JSON, SQL, HDF5, many more
+     - Good
+     - Via Pandas and `scikit-learn <http://scikit-learn.github.io/stable>`__ and more
```
timpitman, line 19: possibly a comma instead of and?  technically the demo doesn't show use of scikit-learn?

huonw: I'm assuming you mean:

```suggestion
     - Via Pandas, `scikit-learn <http://scikit-learn.github.io/stable>`__ and more
```

Also, the demo does use `train_test_split` from scikit-learn in the Cora section. Arguably this isn't quite preprocessing, but I think it's close enough:

```python
from sklearn import model_selection

cora_train, cora_test = model_selection.train_test_split(
    cora_subject, train_size=0.25, random_state=123
)
cora_train
```

Issue #1571, opened 2020-05-18
Neo4J clustering support for cluster GCN on Neo4J


timpitman:
### Description

Use Neo4j's 'data science library' for clustering, instead of the random clusters of the current Neo4jClusterNodeGenerator

### Done Checklist

- [ ] Produced code for required functionality
- [ ] Tests written and coverage checked
- [ ] Code review performed
- [ ] Documentation on Google Docs  (if applicable)
- [ ] Documentation in repo
- [ ] Version number reflects new status
- [ ] CHANGELOG.md updated
- [ ] Team demo

PR #1613, opened 2020-05-28
Feature/neo4j clustering

kieranricardo:
This uses Neo4j graphdatascience library to cluster a graph.

This also includes a very small bugfix in `ClusterNodeSequence` where an empty `cluster_target_indices` would be `np.float64` instead of `np.int64` causing an error when `cluster_target_indices` was being used for indexing later on.

See: #1571


stellargraph/connector/neo4j/graph.py
```diff
@@ -138,6 +138,32 @@ def to_adjacency_matrix(self, node_ids):
...
+        """
+        cluster_query = """
+            CALL gds.louvain.stream({
+                nodeQuery: 'MATCH (n) RETURN id(n) AS id',
+                relationshipQuery: 'MATCH (n)-->(m) RETURN id(n) AS source, id(m) AS target'
+            })
+            YIELD nodeId, communityId
+            RETURN gds.util.asNode(nodeId).ID AS node_id, communityId AS community_id
+        """
+        result = self.graph_db.run(cluster_query).data()
```
huonw, line 157: Since this function does a single pass over the result (even with the `collect`/aggregating form), I think the `data()` call here is unnecessary:

```suggestion
        result = self.graph_db.run(cluster_query)
```

Or, assuming the group-by form in a separate comment works:

```suggestion
        return [row["node_ids"] for row in self.graph_db.run(cluster_query)]
```

(Plus deleting the rest of the function.)

stellargraph/mapper/mini_batch_node_generators.py
```diff
@@ -341,7 +341,8 @@ def __getitem__(self, index):
             # Dictionary to store node indices for quicker node index lookups
             # The list of indices of the target nodes in self.node_list
             cluster_target_indices = np.array(
-                [self.target_node_lookup[n] for n in target_nodes_in_cluster]
+                [self.target_node_lookup[n] for n in target_nodes_in_cluster],
+                dtype=np.int64,
```
huonw, line 345: Why is this bug only revealed with this patch?

demos/connector/neo4j/cluster-gcn-on-cora-neo4j-example.ipynb
```diff
@@ -153,6 +153,24 @@
     "test_targets = target_encoding.transform(test_labels)"
```
huonw, line 153: <p>These tiny clusters are a bit peculiar, especially because <code>q=1</code>,  so they're being handled one at a time. This means our GCN model is spending a big chunk of time being trained on subgraphs of size 1 or 2.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1613/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

stellargraph/mapper/mini_batch_node_generators.py
kieranricardo: because the clusters returned by Louvain were pretty small and so some didn;t have any target nodes. I guess its just chance that this has never happened before + i think we've only every run cluster gcn on large clusters (relative graph size).

huonw: Ah makes sense. Maybe this could be simultaneously changed in a separate PR that adds a test (with the thinking being that PR would land first, but also that it doesn't particularly matter if it lands second)?

demos/connector/neo4j/cluster-gcn-on-cora-neo4j-example.ipynb
kieranricardo: yeah its unfortunate, also the way the `q` parameter is setup it has to exactly divide the number of clusters which is 173.... which is prime....

stellargraph/mapper/mini_batch_node_generators.py
kieranricardo: https://github.com/stellargraph/stellargraph/pull/1622

docker/stellargraph-neo4j/install-lib.sh
```diff
@@ -2,25 +2,31 @@
...
+    gds_sha="685edc293084ddb2376fd0c68621d8da346905338062baa581c809417efb366f4a9200348e1216e09fc8edbc8fe352adeaf2427785aca52d3e25cc995cdd6cc9"
     apoc_version="4.0.0.4"
     apoc_sha="edea2796a9cd97f1daaeb291b5d88747fc4801dc63691eba6997157b34bd8db3525420a289e80d15f495c1d2fdbe00f9c626a74459f861db76e77c1c3e29e73b"
     ;;
   3.5)
+    gds_version="1.1.1"
+    gds_sha="3add8fd44849ddc257903a3e11c8b617ee0150fdbf722b254b7da588df4383a5b2655b6a121525d66f5b6d7d767dea069dcd5966636775f7b0c7506d3f6233f9"
     apoc_version="3.5.0.9"
     apoc_sha="b4b1a8f8940ad250c17b45296459afcfff2eb1beee18b8c8a394e2b5a434183b924b34cb5b04fea1a371fbfb87a286228bd4ae814829e0ef99f9e190a764fa1d"
+
```
huonw, line 17:
```suggestion
```

docker/stellargraph-neo4j/install-lib.sh
```diff
@@ -2,25 +2,31 @@
...
     ;;
   *)
     echo "NEO4J_VERSION: unsupported version '${NEO4J_VERSION}'"
     ;;
 esac
 
-echo "APOC version ${apoc_version} (sha512 ${apoc_sha})"
-
+echo "--- downloading apoc jar for neo4j ${NEO4J_VERSION}"
```
huonw, line 24: The APOC version `echo` was included to provide more insight to debug if something goes wrong (given it's hard to debug retrospectively in some cases). It'd be good to keep and include something similar for gds:

```suggestion
echo "--- downloading apoc jar ${apoc_version} for neo4j ${NEO4J_VERSION}"
```

or

```suggestion
echo "--- downloading apoc jar for neo4j ${NEO4J_VERSION}"
echo "APOC version ${apoc_version} (sha512 ${apoc_sha})"
```

docker/stellargraph-neo4j/install-lib.sh
```diff
@@ -2,25 +2,31 @@
...
 wget --no-verbose "https://github.com/neo4j-contrib/neo4j-apoc-procedures/releases/download/${apoc_version}/apoc-${apoc_version}-all.jar" --directory-prefix plugins/
-
-# validate the downloaded jar
 echo "${apoc_sha}  ${PWD}/plugins/apoc-${apoc_version}-all.jar" | sha512sum -c
+
+echo "--- downloading gds jar for neo4j ${NEO4J_VERSION}"
```
huonw, line 28: For convenience:

```suggestion
echo "--- downloading gds jar ${gds_version} for neo4j ${NEO4J_VERSION}"
```

or

```suggestion
echo "--- downloading gds jar for neo4j ${NEO4J_VERSION}"
echo "GDS version ${gds_version} (sha512 ${gds_sha})"
```

PR #1508, opened 2020-05-05
Fixes for various minor demo issues

huonw:
- #1488 (Links to Cora dataset docs in loading-pandas don't work) `stellargraph.datasets.datasets` was switched to just `stellargraph.datasets` in #1157. https://stellargraph--1508.org.readthedocs.build/en/1508/demos/basics/loading-pandas.html#Real-data:-Homogeneous-graph-from-CSV-files
- #1489 (Neo4j still caps in notebook load-cora-into-neo4j.ipynb) replaced all instances of `Neo4J` with `Neo4j`, except for the ones in our class names, e.g. https://stellargraph--1508.org.readthedocs.build/en/1508/demos/connector/neo4j/directed-graphsage-on-cora-neo4j-example.html
- #1490 (Loading-pandas demo overlap doesn't have overlap) switched to the `overlap` dataframe. https://stellargraph--1508.org.readthedocs.build/en/1508/demos/basics/loading-pandas.html#Multiple-node-types
- #1491 (Bad link formatting in loading-saving-neo4j.ipynb) removed the extra `(` https://stellargraph--1508.org.readthedocs.build/en/1508/demos/basics/loading-saving-neo4j.html#Saving-predictions-into-Neo4j
- #1493 (Content from neo4j connector readme is lost) copied the content back in and converted to rST: https://stellargraph--1508.org.readthedocs.build/en/1508/demos/connector/neo4j/index.html
- #1494 (Heading not formatted properly in DGI demo) removed the extra indentation https://stellargraph--1508.org.readthedocs.build/en/1508/demos/embeddings/deep-graph-infomax-embeddings.html#Data-Generators
- #1501 (Attri2vec link prediction dataset description is formatted peculiarly) this switches to using a block quote `> ...` for the whole description, rather than leaving the code segments out of it https://stellargraph--1508.org.readthedocs.build/en/1508/demos/link-prediction/attri2vec-link-prediction.html#Loading-DBLP-network-data
- #1502 (Incorrect heading formatting in Node2Vec node classification demo) the `<a name="..."></a>` seemed to be confusing things, and it seemed like it didn't even work correctly in Jupyter Lab https://stellargraph--1508.org.readthedocs.build/en/1508/demos/node-classification/node2vec-node-classification.html#References

Issue #1502, opened 2020-05-05
Incorrect heading formatting in Node2Vec node classification demo


huonw:
### Describe the bug

https://stellargraph.readthedocs.io/en/latest/demos/node-classification/node2vec-node-classification.html

> Note: For clarity of exposition, this notebook forgoes the use of standard machine learning practices such as Node2Vec parameter tuning, node feature standarization, data splitting that handles class imbalance, classifier selection, and classifier tuning to maximize predictive accuracy. We leave such improvements to the reader.
>
> &nbsp;&nbsp; \#\#\# References

https://stellargraph.readthedocs.io/en/latest/demos/node-classification/node2vec-weighted-node-classification.html

PR #1614, opened 2020-05-29
Add metapath2vec link prediction notebook

kjun9:
This adds a demo for metapath2vec link prediction. 

The structure of the demo follows the Node2Vec link prediction one closely, and I think this might be able to land without #1399 now since random walks have gotten faster as a result of other work that's landed since.

Note that the `average` operator was left out - I tried to give an explanation for why in the notebook. Basically I think it's not a suitable operator when we are calculating embeddings twice, since the embeddings may get rotated when you calculate it for the second time. 

`average` happened to be the "winning" operator in this particular demo during training, and then performs very poorly on the test set, so that's how I ended up catching this - we might want to also remove the `average` operator in the node2vec link prediction demo for the same reason.

See #934


demos/link-prediction/metapath2vec-link-prediction.ipynb
```diff
@@ -0,0 +1,969 @@
+{
```
huonw, line 1: <p>Our other demos don't list the dataset, because typically a user doesn't care about the dataset of these "learn how to use an algorithm" demos. i.e. just "Link prediction with Metapath2Vec" would be fine IMO.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1614/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

huonw, line 1: <blockquote>... on a heterogeneous graph (with multiple node types) ... how to traverse the different types</blockquote><p> </p><p>Potentially "on a heterogeneous graph (with multiple node types) ... how to traverse the different node types" to be slightly more specific?</p><p> </p><p>Also, I don't think the "References" <code>&lt;a ...&gt;&lt;/a&gt;</code> thing quite works right, e.g. the <code>#refs</code> links on RtD don't jump to it, and the rendering is a bit off (the References text ends up being nested inside a <code>&lt;blockquote&gt;</code> ): <a href="https://stellargraph--1614.org.readthedocs.build/en/1614/demos/link-prediction/metapath2vec-link-prediction.html" target="_blank">https://stellargraph--1614.org.readthedocs.build/en/1614/demos/link-prediction/metapath2vec-link-prediction.html</a> (#1502 and #1508 and specifically <a href="https://github.com/stellargraph/stellargraph/pull/1508/commits/9faae4adbafbffa65183c116450f715bcd638686" target="_blank">https://github.com/stellargraph/stellargraph/pull/1508/commits/9faae4adbafbffa65183c116450f715bcd638686</a> are relevant)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1614/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

huonw, line 1: <p>I'm looking forward to <a href="https://github.com/stellargraph/stellargraph/issues/1454" target="_blank">https://github.com/stellargraph/stellargraph/issues/1454</a>!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1614/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

huonw, line 1: <p>Out of curiosity, how long do these cells take at the moment?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1614/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

huonw, line 1: <p>Also out of curiosity, I wonder what the performance of Node2Vec (i.e. just ignoring the node types entirely) on this dataset would be.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1614/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

huonw, line 1: <p>This cell and the next one seem to be missing their output.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1614/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

huonw, line 1: <p>Last bit of curiosity: why PCA for these? (I know it's what's used in the Node2Vec one too, so not doesn't need to change, just wondering.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1614/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

huonw, line 1: <p>It might be nice to add a little more content here so it's a little clearer and the user has some "next steps" if this one wasn't quite right:</p><p> </p><ul><li>it could have a Conclusion heading</li><li>some links to other relevant demos, e.g. HinSAGE for heterogeneity, the whole link prediction folder (potentially these links could go in the opening section too, so a user can jump to another more relevant demo if they decide this one isn't quite right). I tried to do something like this at the start and end of <a href="https://stellargraph.readthedocs.io/en/latest/demos/node-classification/gcn-node-classification.html" target="_blank">https://stellargraph.readthedocs.io/en/latest/demos/node-classification/gcn-node-classification.html</a>, if you're looking for more explanation of what I mean.</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1614/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

huonw:
> Note that the average operator was left out - I tried to give an explanation for why in the notebook. Basically I think it's not a suitable operator when we are calculating embeddings twice, since the embeddings may get rotated when you calculate it for the second time.

Interesting. Why does it only affect average? I agree that it shouldn't affect L1 or L2 given it's a rotation, but I think Hadamard would be, e.g.:

- [1, 0] * [0, 1] = [0, 0]
- rotating by 45 degrees: [x, x] * [-x, x] = [-0.5, 0.5] (where x = sqrt(1/2)).

> average happened to be the "winning" operator in this particular demo during training, and then performs very poorly on the test set, so that's how I ended up catching this - we might want to also remove the average operator in the node2vec link prediction demo for the same reason.

Sounds reasonable, if it's "incorrect". Do you have any intuition for why it seems to be working not-horribly on Node2Vec (although it is noticeably worse)?

kjun9:
> Interesting. Why does it only affect average? I agree that it shouldn't affect L1 or L2 given it's a rotation, but I think Hadamard would be, e.g.:

Hmm.. yeah I think you're right, maybe `hadamard` should be omitted too 🤔 but for some reason it doesn't seem to perform as poorly as `average` on the test set. Maybe because a 45 degree rotation is the worst case scenario for `hadamard`?
```
[1, 0] * [0, 1] = [0, 0]

# 90 degrees
[0, 1] * [-1, 0] = [0, 0]
```

> Sounds reasonable, if it's "incorrect". Do you have any intuition for why it seems to be working not-horribly on Node2Vec (although it is noticeably worse)?

The "winning" operator for Node2Vec happens to be `l1` so we're getting good results on the test set which only uses the best operator. Running on the test set in that notebook with `average` I'm getting:
```
ROC AUC score on test set using 'operator_avg': 0.5814755917159764
```
so roughly just as horrible

demos/link-prediction/metapath2vec-link-prediction.ipynb
kjun9: <p>This particular line took for me</p><pre class="ql-syntax" spellcheck="false">CPU times: user 6min 12s, sys: 2.82 s, total: 6min 15s
Wall time: 5min 43s
</pre>

kjun9: <p>Hmm yeah hard to say without actually running it, but given that there's 333983 user-friend-&gt;user edges, and only 14476 group-belongs-&gt;user edges, I guess it may not be as effective at finding users that are connected via groups, and instead predict friendships mostly based on other friendships?</p><p> </p>

kjun9: I've added something similar to the gcn node classification demo at the top and bottom

kjun9: Hmm yeah I don't have a good reason for it other than copying it from other demos - do you think tsne should be the "default" unless there's a good reason?

demos/link-prediction/metapath2vec-link-prediction.ipynb
```diff
@@ -0,0 +1,1017 @@
...
```
kjun9, line 1: <p>Unfortunately, the visualisation doesn't look as good with the <code>l2</code>  operator (this was also a problem when working on other similar link prediction notebooks if I remember correctly, might've been CTDNE)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1614/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

demos/link-prediction/metapath2vec-link-prediction.ipynb
kjun9: I've opened https://github.com/stellargraph/stellargraph/pull/1619

kjun9:
> The PCA plot is kinda epic/not particularly interpretable, but I think that's fine?

Hmm yeah, I think it's still useful to keep it there to demonstrate how to visualise the embeddings even though it didn't turn out to be very useful for this dataset

Issue #1534, opened 2020-05-07
Improve demo indexing for keras Node2Vec notebooks


kjun9:
### Description

With https://github.com/stellargraph/stellargraph/pull/536 we now have two different versions of the `node2vec` algorithm being demoed with notebooks - `gensim` vs stellargraph components. 

Currently: 
* the top-level demo summary table shows these as two separate algorithms, but this could be improved to show that they are in fact one algorithm (maybe should be in one row with multiple links in a cell?)
* they're missing from the task-specific demo tables, but should be included in a similar way to whatever approach is taken for the top-level demo table

huonw:
Another option would be to instead remove the `gensim` version?

daokunzhang:
> Another option would be to instead remove the `gensim` version?

Hi @huonw ,

I think the `genesim` version is still necessary. As it is built on the original word2vec C implementation, it is much faster than the keras implementation and it can handle much larger dataset. As our metapath2vec implementation and the link prediction with random walks demo still rely on the `gensim` word2vec, I think it is better to preserve a `gensim` node2vec demo, to make a consistently. 

Best regards,
Daokun

PR #1632, opened 2020-06-01
Improve demo indexing for keras Node2Vec notebooks (#1534)

timpitman:
see #1534

build docs for demo table:
  * https://stellargraph--1632.org.readthedocs.build/en/1632/demos/index.html

and demo node classification table:
  * https://stellargraph--1632.org.readthedocs.build/en/1632/demos/node-classification/index.html (here it gets a row for each, as the table is demo-centric)

### DIFF
--- a/scripts/demo_indexing.py
+++ b/scripts/demo_indexing.py
@@ -472,21 +472,18 @@ ALGORITHMS = [
         "Node2Vec",
         weighted=T(link="node-classification/node2vec-weighted-node-classification"),
         nc=[
+            T(text="via unsup.", details="via embedding vectors",),
             T(
-                text="via unsup. keras",
+                text="keras",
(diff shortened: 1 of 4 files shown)


scripts/demo_indexing.py
```diff
@@ -471,14 +471,27 @@ def via_rl(link=None):
     Algorithm(
         "Node2Vec",
         weighted=T(link="node-classification/node2vec-weighted-node-classification"),
-        nc=via_rl(link="node-classification/node2vec-node-classification"),
+        nc=[
+            T(
```
huonw, line 475: Just to reiterate/make concrete, what we discussed on the stand-up, we could potentially collapse this like:

```suggestion
            T("via", details="via embedding vectors"),
            T(
```

and then remove the `via`/`via embedding vectors` from the later ones, to maybe make the text easier to read.

Issue #1312, opened 2020-04-22
Graph.info shows directed edge types in an undirected graph


kjun9:
### Describe the bug

Current behaviour is to show the edge type triples `A-R->B` and `B-R->A` separately in the info string even when the graph is undirected. We should probably be collapsing them into one?

huonw:
Reproducing example (it would've be good to include at least something basic, so that we can understand specifically what this issue is about in future).

```python
import stellargraph as sg
import pandas as pd

nodes = {"A": pd.DataFrame(index=[0, 1]), "B": pd.DataFrame(index=[2])}
edges = pd.DataFrame({"source": [0, 2], "target": [2, 1]})
print(sg.StellarGraph(nodes, edges).info())
```

This graph has edges `0 (A) -> 2 (B)` and `2 (B) -> 1 (A)`.

### Observed output

```
StellarGraph: Undirected multigraph
 Nodes: 3, Edges: 2

 Node types:
  A: [2]
    Features: none
    Edge types: A-default->B
  B: [1]
    Features: none
    Edge types: B-default->A

 Edge types:
    B-default->A: [1]
    A-default->B: [1]
```


### Expected output

Something like:

```
StellarGraph: Undirected multigraph
 Nodes: 3, Edges: 2

 Node types:
  A: [2]
    Features: none
    Edge types: A-default->B
  B: [1]
    Features: none
    Edge types: B-default->A

 Edge types:
    A-default->B: [2]
```

### Versions

Stellargraph bcde23a37531256321853f44a233e8b48d8fc455

PR #1650, opened 2020-06-02
Sort edge type triple for undirected graph info

kjun9:
This adds a check when collecting metrics for edge type triples for graph info, so that if the graph is undirected, we sort each type triple. 

~I think the edge types in the "Node types" section might still be incorrect~ NVM seems it was already doing the right thing. So in the "Node types" section, the arrow is always pointing outward from the node type.

See #1312

### DIFF
--- a/stellargraph/core/graph.py
+++ b/stellargraph/core/graph.py
@@ -1344,6 +1344,14 @@ class StellarGraph:
                 "weight": self._edges.weights,
             }
         )
+
+        # if graph is undirected and edges exist, we want to sort the triple
+        if not self.is_directed() and len(df):
+            sorted_types = df.apply(
+                lambda r: sorted([r.src_ty, r.tgt_ty]) + [r.rel_ty], axis=1
+            )
+            df[["src_ty", "tgt_ty", "rel_ty"]] = pd.DataFrame(sorted_types.tolist())
+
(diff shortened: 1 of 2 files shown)


stellargraph/core/graph.py
```diff
@@ -1344,6 +1344,14 @@ def _edge_metrics_by_type_triple(self, metrics):
                 "weight": self._edges.weights,
             }
         )
+
+        # if graph is undirected and edges exist, we want to sort the triple
+        if not self.is_directed() and len(df):
+            sorted_types = df.apply(
+                lambda r: sorted([r.src_ty, r.tgt_ty]) + [r.rel_ty], axis=1
+            )
+            df[["src_ty", "tgt_ty", "rel_ty"]] = pd.DataFrame(sorted_types.tolist())
```
huonw, line 1353: Since this isn't touching the `rel_ty` column, I think this can be done with less Python iteration and less pandas.Series construction via NumPy:

```suggestion
            sorted_types = df[["src_ty", "tgt_ty"]].to_numpy()
            sorted_types.sort(axis=1)
            df[["src_ty", "tgt_ty"]] = sorted_types
```

This form might even allow removing the `len(df)` condition, although that's less important.

PR #1658, opened 2020-06-02
Link prediction comparison demo between Node2Vec, Attri2Vec, GraphSAGE and GCN

daokunzhang:
Following @huonw 's advice, I add a link prediction demo to compare the link prediction permanence on the Cora dataset under the same edge train-test-split setting.

The results show that Attri2Vec performs best, while GraphSAGE and Node2Vec perform comparably.


huonw:
A few brief comments:

- I'm curious about the approach here: algorithms like Attri2Vec and GraphSAGE can natively do link predicting (indeed, that's how the unsupervised embeddings are being trained). Could we instead use that native support for those algorithms, and embeddings only for Node2Vec (etc.)? What do you think? Maybe this would stop the GraphSAGE code taking 4 minutes per epoch, which is quite slow?
- What about including GCN and/or GAT to make this a "complete" demo?
- Maybe we could call it something like `homogeneous-comparison-link-prediction.ipynb`:
  - `homogeneous`, because we could have other demos that do a comparison on a more complicated (heterogeneous) dataset
  - moving `link-prediction` to the end to match the style of the other demos in this folder.
- It'd be great to have a section at the end that summarises the results similar to ["Overall results" section in the Deep Graph Infomax demo](https://stellargraph.readthedocs.io/en/latest/demos/embeddings/deep-graph-infomax-embeddings.html#Overall-results). For this example, it could be done by having a column for each algorithm:
    > ## Node2Vec link prediction results
    >
    > ```python
    > node2vec_results = get_result(node2vec_embedding, "Node2Vec")
    > node2vec_results
    > ```
    > 
    > (... the same for for each algorithm ...)
    > 
    > ## Overall results
    > 
    > ```python
    > pd.DataFrame(
    >     {
    >         "Node2Vec": node2vec_results["ROC AUC score"],
    >         "Attri2Vec": attri2vec_results["ROC AUC score"],
    >         "GraphSAGE": graphsage_results["ROC AUC score"],
    >     },
    > )
    > ```

Otherwise, this looks pretty awesome!

daokunzhang:
Hi @huonw ,

Thanks for your feedback and valuable comments. The following is my response to your comments:

> * I'm curious about the approach here: algorithms like Attri2Vec and GraphSAGE can natively do link predicting (indeed, that's how the unsupervised embeddings are being trained). Could we instead use that native support for those algorithms, and embeddings only for Node2Vec (etc.)? What do you think? Maybe this would stop the GraphSAGE code taking 4 minutes per epoch, which is quite slow?

Indeed, Attri2Vec and GraphSAGE use the ``link_classification`` module to learn node embeddings. However, the employed ``link_classification function`` has different meanings compared with the link prediction task we are to address. The link classification used by Attri2Vec and GraphSAGE in embedding learning is performed on the ``target, context`` pairs collected from short random walks with the objective of preserving the high-order proximity. The ``target, context`` pairs may not be connected in the original networks, and cannot act as the accurate supervision for link classification. The GraphSAGE link prediction demo `` graphsage-link-prediction.ipynb`` performs link prediction in an end-to-end manner on by constructing link embedding vectors on the GraphSAGE embeddings of connected nodes. However, the learned embedding only capture the similarity between directly connected nodes while fails to capture the similarity between nodes connected through 2 or more hops. As is shown in the demo notebook, the link prediction performance on the test set is not satisfactory (with 0.7970 accuracy). In this demo, we aim to leverage the informative node embeddings learned by different algorithms, use the link operator to form link embeddings and perform link prediction on the link embeddings through training a link classification classifier.

For fair comparison, I set the embedding dimension of GraphSAGE to 128, which is consistent with Attri2Vec and Node2Vec. To make sure it is well trained, I use a larger number of random walks and use more epochs, which results the long running time of GraphSAGE.

> * What about including GCN and/or GAT to make this a "complete" demo?

In this demo, we mainly investigate the performance of unsupervised embedding methods in link prediction. However, GCN and GAT require node labels as supervised information to learn informative node embeddings. Though the unsupervised GCN can also be used to predict links as is shown in the demo notebook ``gcn-link-prediction.ipynb``, its performance is not satisfactory as only the embedding similarity between connected nodes is preserved. 

> * Maybe we could call it something like `homogeneous-comparison-link-prediction.ipynb`:
>   
>   * `homogeneous`, because we could have other demos that do a comparison on a more complicated (heterogeneous) dataset
>   * moving `link-prediction` to the end to match the style of the other demos in this folder.

Thanks! I will rename the name of this notebook to `homogeneous-comparison-link-prediction.ipynb`.

> * It'd be great to have a section at the end that summarises the results similar to ["Overall results" section in the Deep Graph Infomax demo](https://stellargraph.readthedocs.io/en/latest/demos/embeddings/deep-graph-infomax-embeddings.html#Overall-results). For this example, it could be done by having a column for each algorithm:
>   > ## Node2Vec link prediction results
>   > ```python
>   > node2vec_results = get_result(node2vec_embedding, "Node2Vec")
>   > node2vec_results
>   > ```
>   > 
>   > 
>   > (... the same for for each algorithm ...)
>   > ## Overall results
>   > ```python
>   > pd.DataFrame(
>   >     {
>   >         "Node2Vec": node2vec_results["ROC AUC score"],
>   >         "Attri2Vec": attri2vec_results["ROC AUC score"],
>   >         "GraphSAGE": graphsage_results["ROC AUC score"],
>   >     },
>   > )
>   > ```
> 
> Otherwise, this looks pretty awesome!

Good idea! I will summarize the results of different algorithms into one data frame to show them in a more awesome way.

Best regards!
@daokunzhang

huonw:
> However, the learned embedding only capture the similarity between directly connected nodes while fails to capture the similarity between nodes connected through 2 or more hops. . In this demo, we aim to leverage the informative node embeddings learned by different algorithms, use the link operator to form link embeddings and perform link prediction on the link embeddings through training a link classifier.

Ah, okay, so you're focused on doing semi-supervised link prediction via embeddings, rather than supervised link prediction. This isn't a problem, but we should be conveying that. It seems like this demo could be framed as a comparison of different ways to compute embeddings for random walks, evaluated through link prediction:

- Node2Vec: do random walks to sample nodes, and use Word2Vec to compute vectors for each node
- GraphSAGE: do random walks to sample nodes, and use GraphSAGE to compute vectors for each node
- Attri2Vec: do random walks to sample nodes, and use Attri2Vec to compute vectors for each node

Have you considered other approaches to doing this in a semi-supervised manner? For example: pre-train the models with `UnsupervisedSampler` and then fine-tune the link prediction model itself on the labelled data, so that the weights can be trained end-to-end for the labelled data?

> As is shown in the demo notebook, the link prediction performance on the test set is not satisfactory (with 0.7970 accuracy)

It seems perfectly fine for me to have a demo include algorithms that perform poorly, and have a semi-supervised approach that improves the performance. That seems like a great thing to have in a demo notebook, in fact, although maybe this one would end up too large. 

For example, we have https://stellargraph.readthedocs.io/en/latest/demos/node-classification/gcn-deep-graph-infomax-fine-tuning-node-classification.html that uses Deep Graph Infomax to do semi-supervised node classification.

> In this demo, we mainly investigate the performance of unsupervised embedding methods in link prediction. GCN and GAT require node labels as supervised information to learn informative node embeddings. Though the unsupervised GCN can also be used to predict links as is shown in the demo notebook gcn-link-prediction.ipynb, its performance is not satisfactory as only the embedding similarity between connected nodes is preserved.

If you're wanting to keep this focused on semi-supervised link prediction, I think GCN and GAT can be trained via link prediction with `UnsupervisedSampler` in exactly the same way as `GraphSAGE`?

---

Given the above, it seems like there's a few options here:

1. include comparisons to fully-supervised approaches
2. give the demo a new title (and file name) and otherwise emphasise that this is using the expensive `UnsupervisedSampler` approach to train models in a semi-supervised manner, rather than the "native" link prediction built in to each of the algorithms, for example, "Comparison of link prediction via node embeddings from Node2Vec, Attri2Vec and GraphSAGE" or something
3. something else?

What do you think?

daokunzhang:
Hi @huonw ,

> If you're wanting to keep this focused on semi-supervised link prediction, I think GCN and GAT can be trained via link prediction with `UnsupervisedSampler` in exactly the same way as `GraphSAGE`?

I tend to focus on the semi-supervised link prediction setting. I think it deserves to try to use the `UnsupervisedSampler` to train GCN and GAT in the same manner of training GraphSAGE. Our title can be "Comparison of link prediction via node embeddings learned from random walks". In the current version, Node2Vec is based on Gensim implementation, and it does not use `UnsupervisedSampler`. Do you think it is necessary to use the Keras implementation of Node2Vec to make all algorithms consistent in the use of `UnsupervisedSampler`. In this way, our title can be changed into "Comparison of link prediction via node embeddings learned with UnsupervisedSampler".

huonw:
> I think it deserves to try to use the UnsupervisedSampler to train GCN and GAT in the same manner of training GraphSAGE.

Sounds good!

>  In the current version, Node2Vec is based on Gensim implementation, and it does not use UnsupervisedSampler. Do you think it is necessary to use the Keras implementation of Node2Vec to make all algorithms consistent in the use of UnsupervisedSampler? In this way, our title can be changed into "Comparison of link prediction via node embeddings learned with UnsupervisedSampler".

I think "learned from random walks" and the Gensim one is perfectly fine by me!

As a minor detail, I wonder if it's worth standardising on Node2Vec random walks for the `UnsupervisedSampler` ones (passing a `BiasedRandomWalk` value to the `walker` parameter of the constructor), instead of using the default `UniformRandomWalk`. (docs for convenience: https://stellargraph.readthedocs.io/en/stable/api.html#stellargraph.data.UnsupervisedSampler )

daokunzhang:
Thanks @huonw ,

I will try to finish the updated demo notebook today. 

> As a minor detail, I wonder if it's worth standardising on Node2Vec random walks for the `UnsupervisedSampler` ones (passing a `BiasedRandomWalk` value to the `walker` parameter of the constructor), instead of using the default `UniformRandomWalk`. (docs for convenience: https://stellargraph.readthedocs.io/en/stable/api.html#stellargraph.data.UnsupervisedSampler )

I think it is improper to use the `UnsupervisedSampler` to generate Gensim Node2Vec random walks. The `UnsupervisedSampler` will generate a batch of positive and negative node pairs, while the Gensim Node2Vec requires a set of random walk sequences as input to learn node embeddings. Actually, Gensim Node2Vec has already included the procedure to collect `target, context` node pairs with a fixed window size.

huonw:
I mean use Node2Vec-style random walks for the methods like Attri2Vec, GraphSAGE and GAT that use `UnsupervisedSampler`. That is, `BiasedRandomWalk` not `UniformRandomWalk`.

daokunzhang:
> I mean use Node2Vec-style random walks for the methods like Attri2Vec, GraphSAGE and GAT that use `UnsupervisedSampler`. That is, `BiasedRandomWalk` not `UniformRandomWalk`.

Got it. I will try to unify the random walk to `BiasedRandomWalk`.

daokunzhang:
Hi @huonw ,

I have updated the link prediction comparison demo notebook. However, I cannot pass the spell check in the Buildkite check. 

In the updated demo, I added the unsupervised GCN trained by the `UnsupervisedSampler` as a new baseline. Training GCN with the `UnsupervisedSampler` is a bit complicated. The `FullBatchLinkGenerator` does not support `UnsupervisedSampler`, I have to iterates the batches of node pairs generated `UnsupervisedSampler` and train the model with the `keras.model.train_on_batch`  function. 

Training GAT with the `UnsupervisedSampler` is more complicated. It is also very hard to generate node embeddings from node list with the trained GAT link model. As the higher time complexity, training unsupervised GAT with a large number node pairs generated by random walks will be much more time-consuming. Considering GAT and GCN are both graph convolutional networks and GCN is enough to represent graph convolutional networks, I do not add GAT in the link prediction comparison demo.

stellargraph/mapper/full_batch_generators.py
```diff
@@ -160,6 +160,8 @@ def flow(self, node_ids, targets=None):
                 (e.g., training, validation, or test set nodes)
             targets: a 1D or 2D array of numeric node targets with shape ``(len(node_ids),)``
                 or ``(len(node_ids), target_size)``
+            use_ilocs (bool): if Ture, node_ids are represented by ilocs,
```
huonw, line 163: The spellcheck step is highlighting the "Ture" here (and in the `flow` below):

```suggestion
            use_ilocs (bool): if True, node_ids are represented by ilocs,
```

daokunzhang: Thanks @huonw , to make the `FullBatchLinkGenerator` able to handle the node ilocs generated by the `UnsupervisedSampler`, I add an argument `use_ilocs` to the `FullBatchLinkGenerator.flow` function.

demos/link-prediction/homogeneous-comparison-link-prediction.ipynb
```diff
@@ -0,0 +1,1515 @@
+{
```
huonw, line 1: <p>The <code>&lt;a name=""&gt;</code>  links don't work on Read the Docs, which is our main venue for people to read the demos. You can explore this at https://stellargraph--1658.org.readthedocs.build/en/1658/demos/link-prediction/homogeneous-comparison-link-prediction.html .</p><p> </p><p>The HTML has two problems there:</p><p> </p><ul><li>the <code>#ref1</code>  (etc.) links don't go anywhere and are confusing</li><li>the HTML of the references themselves comes out funny: they're wrapped in a <code>&lt;blockquote&gt;</code>,  when they should just be normal text (it's why they're indented in the rendering on Read the Docs above, but not in the rendering in Jupyter)</li></ul><p> </p><p>I think it's fine to just leave them as non-links, and people can scroll manually if they absolutely want the citation (since this document is short enough and I imagine the citations will be relatively rarely followed). This means removing the  <code>[...](#...)</code> around each of the <code>[1]</code>  citations, as well as the <code>&lt;a ...&gt;&lt;/a&gt;</code>  pieces.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

huonw, line 1: <p>Maybe the construction of the <code>BiasedRandomWalk</code>  walker could be put into a function so that it doesn't have to be copied into each of the <code>_embedding</code>  functions, and the explanation of the <code>p</code>  parameter (and all the others for node2vec) can be placed in one place. For example, a section that has a title like "Node2Vec random walking" with the explanation of the parameters and a cell like:</p><p> </p><pre class="ql-syntax" spellcheck="false">def create_node2vec_walker():
    p = 1.0
    q = 1.0
    ...
    
    return BiasedRandomWalk(graph, n=num_walks, length=walk_length, p=p, q=q)
</pre><p> </p><p>Finally, I think with p=1 and q=1, "biased" random walk is equivalent to a uniform random walk. Did you choose those hyper parameters by just copying from https://stellargraph.readthedocs.io/en/latest/demos/link-prediction/node2vec-link-prediction.html ? (It's fine if so.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

huonw, line 1: <p>Could you file an issue about expanding GCN to work with UnsupervisedSampler directly, so that we can eventually remove the loop in this cell?</p><p> </p><p>Also, why does this use the <code>LinkEmbedding</code>  class while the other models use <code>link_classification</code>  instead?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

huonw, line 1: <p>@kjun9 found in https://github.com/stellargraph/stellargraph/pull/1614 that the average operator and even the Hadamard one probably was meaningless for these sort of algorithms, because the embeddings can end up rotated. What do you think?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

huonw, line 1: <p>Maybe this function could be renamed to something like <code>train_and_evaluate</code>  since that describes the task a little better than just "get result".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

huonw, line 1: <p>Out of curiosity, approximately how long do these cells take to run?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

huonw, line 1: <p>Is it worth training these for longer? It looks like the accuracy of this one in particular has not plateaued?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

huonw, line 1: <p>It's interesting to me that all of these algorithms are all very similar to the base Node2Vec algorithm.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

huonw, line 1: <p>Could you add a <code>## Conclusion</code> heading to this cell?</p><p> </p><p>It might also be good to link to the demos for the specific algorithms, like:</p><p> </p><pre class="ql-syntax" spellcheck="false">For more information about the link prediction process, all of these algorithms have specific demos with more details:

- [Node2Vec](node2vec-link-prediction.ipynb)
- ...
</pre><p> </p><p>This gives a user some "next" steps to follow if they get to the end of this demo and are interested in further things.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='42'/>

daokunzhang: <p>Thanks! I will remove the citation commands.</p>

daokunzhang: <p>I think it is fine to define a function to generate biased random walks, which can used by all embedding learning modules.</p><p> </p><p>For the setting of the value of p and q, I just follow the demo <a href="https://stellargraph.readthedocs.io/en/latest/demos/link-prediction/node2vec-link-prediction.html" target="_blank">https://stellargraph.readthedocs.io/en/latest/demos/link-prediction/node2vec-link-prediction.html</a>&nbsp;</p><p>In this demo notebook, we just give an example about how to use the validation set to choose the best link operator. It can also be used to choose the hyper parameters like p and q. Here, we only provide a default value for p and q, and leave room for uses to explore properer values.</p><p> </p>

daokunzhang: <p>I will create an issue to make GCN able to work with UnsupervisedSampler.</p><p> </p><p>Here, I use <code>LinkEmbedding</code>  rather than <code>link_classification</code> for the reason that <code>link_classification</code>  cannot output the prediction with correct shape. The input <code>x_out</code>  has a different shape, compared with other embedding learning functions.</p><p> </p>

daokunzhang: <p>Good idea! I will rename it to <code>train_and_evaluate</code>.</p>

daokunzhang: <p>From my observation, <code>average</code> operator indeed performs poorly compared with other operators. I agree that embedding can be rotated with the average operator. However, I think <code>Hadamard</code> is useful to measure node embedding difference. At least, <code>Hadamard</code> can tell us whether each dimensions of two embedding vectors share the same sign or not, i.e., the product of two positive or two negative figures is positive while the product of one positive and one negative figures is negative. For the embedding vectors generated by the <code>relu</code>  activation like GCN, Hadamard can indicate wether two embedding vectors have overlapped non-zero dimensions and which dimensions are overlapped. From my observation, <code>Hadamard</code> performs competitively compared with <code>operator_l1</code> and <code>operator_l2</code> and even performs better than them in some cases.</p>

daokunzhang: <p>I takes no more than 5 minutes. Do you think it is necessary to test the running time and show it?</p>

daokunzhang: <p>In my first trial, I use 6 epochs. That has plateaued binary_accuracy. I will change it back.</p>

daokunzhang: <p>Atti2Vec is better, while GraphSAGE and GCN has comparable performance with Node2Vec. Graph convolution or neighbourhood aggregation benefits the node classification task, which largely depends on node attributes for the Cora dataset. In comparison, it is hard for them to boost the link prediction performance, which is dominated by the structure information.</p>

daokunzhang: <p>I ever tried to give a conclusion. However, due to the randomness of edge train-test split, the results of each run is different. In most cases, Attri2Vec performs best, while the performance ranks of Node2Vec, GraphSAGE and GCN change significantly in different runs. It is hard to draw a consistent conclusion. Do you think it deserves to fix the random seed for splitting the train/test edge sets?</p><p> </p><p> </p><p>I think it is helpful to instruct readers to refer to the link prediction demos for the specific algorithms.</p>

huonw: > However, due to the randomness of edge train-test split, the results of each run is different. In most cases, Attri2Vec performs best, while the performance ranks of Node2Vec, GraphSAGE and GCN change significantly in different runs. It is hard to draw a consistent conclusion. Do you think it deserves to fix the random seed for splitting the train/test edge sets?

I don't understand? I was just meaning adding the heading to the existing cell without changing the content?

daokunzhang: <p>Sorry! I misunderstood. I though you were to ask me to add an ## Conclusion section to discuss the experimental results. </p><p> </p><p>It is fine to just add a a&nbsp;<code>## Conclusion</code>&nbsp;heading.</p><p> </p>

huonw: No, I don't think so. I was just curious 😄

daokunzhang:
Hi @huonw ,

The demo notebook has been updated. You can have a check.

daokunzhang:
> I think we can address both of these in the same manner, at least partially: create a top-level cell (e.g. just before the "biased random walk" section) that defines the number of epochs to train for as an `epochs = ...` variable, and use this for each of the methods, instead of defining a new `epochs` variable inside each `..._embedding` function.

I think it is OK to define a global epoch variable to train and change its value when we run CI.

> There's also another inconsistency thing, that doesn't necessarily need to change: GCN and GraphSAGE are trained on much fewer input walks than Attri2Vec, with `num_walks = 1` instead of `num_walks = 4`. Is this unfairly penalising them?

Yeah, it is a bit unfair. However, after I set the `num_walks` to 4 for GCN and GraphSAGE, their performance does not change, while consumes four-fold time. When I set the value of `num_walks` to 1 consistently, Attri2Vec will deliver unsatisfactory performance.

demos/link-prediction/homogeneous-comparison-link-prediction.ipynb
```diff
@@ -0,0 +1,1634 @@
...
```
huonw, line 1: <p>You need to use a single cell with all of the parameters, as the new parameters are inserted after the first one, and thus the later ones override it. You can see this in the notebook that was run on CI: https://nbviewer.jupyter.org/urls/buildkite.com/organizations/stellar/pipelines/stellargraph-public/builds/4791/jobs/292b124f-0618-4cab-9b0a-93762e76243e/artifacts/929ec060-09dd-4294-abf7-c8b79fd6c07b#Suggested-Hyperparameters</p><p> </p><p>The inserted cell is the one with the <code># Parameters</code>  comment, and the three later cells win and override those customisations.</p><p> </p><p>(I think the failure on CI is due to running out of memory; hopefully that will be addressed by getting these parameters to be overridden correctly!)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/stellargraph/stellargraph/pull/1658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

daokunzhang: Thanks @huonw ! I ever try to use a single cell to make the variables overwritten by the notebook-parameters.yml. However, one the batch_size is overwritten to 5, it will consume more time. In the new update, I only make the `epochs` variable be overwritten. I set different `walk_number` values for different algorithms. With the overwritten the `epochs`, the demo will consume much less time.

daokunzhang: Thanks @huonw , I will change the cell carefully to make the parameter correctly overwritten by notebook-parameters.yml.

huonw: To ensure that `batch_size = 50` "wins", that cell will need to be after the `epoch` parameters one, see the ordering in: https://nbviewer.jupyter.org/urls/buildkite.com/organizations/stellar/pipelines/stellargraph-public/builds/4798/jobs/3253fa74-8196-4f0e-a480-5e22a206f48d/artifacts/54d6f390-7c39-4d63-9033-ca62d3466431#Parameter-Settings

daokunzhang: Yeah, I have changed the order.

daokunzhang:
Hi @huonw ，

To make the Node2Vec learning process consistent with Attri2Vec, GraphSAGE and GCN. I replace the Gensim Word2Vec implementation with the Keras Node2Vec implementation. Then we have the same set of hyper parameters: `walk_number`, `walk_length`, `batch_size` and `epochs`. Before we run the embedding learning algorithm, I add new section `Parameter Setting` to introduce the parameter settings of different algorithms. In this section I set the value of `epochs` to 6 and make it able to be overwritten by `notebook-parameters.yml` when performing the build kite check.

PR #1659, opened 2020-06-03
Use papermill --report-mode for easier debugging on CI

huonw:


huonw:
Hm, this doesn't seem to do what I hoped (specify the cell contents when executing). 🤷‍♂️

Issue #462, opened 2019-07-28
Can weighted graphs be used inGraphSAGE?


ericbroda:
I have graphs that have weights that reflect the relative importance of some relationships compared to others.  In other words, some links between nodes have lower weights (low importance) and others have higher weights (high importance).

This could be very useful for clustering and perhaps even node classification in situations where weights are available.

Is there a way to have GraphSAGE take this into account?

adocherty:
Hi Eric,

Currently edge weights are not supported in GraphSAGE. We are looking into possible ways to take edge weights into account. We will update this issue when we do.

Best regards,
Andrew

colinwxl:
I am also interested in GraphSAGE on the weighted graph. In my opinion, it should consider both the `sampling` procedure and the `link-prediction` part.

mimisen-boop:
> This could be very useful for clustering and perhaps even node classification in situations where weights are available.

Could you specify more about ideas that might be useful with node classification?

huonw:
We've now merged #1667, which adds support for edge weights in GraphSAGE sampling. Edge weights can be used for training a link prediction model too.

patchy631:
> We've now merged #1667, which adds support for edge weights in GraphSAGE sampling. Edge weights can be used for training a link prediction model too.

How can i use the edge weights as one of the parameters for link prediction?
Right now I'm trying to make some changes to sampled_link_generator.py, so that with every batch i also append the corresponding edge weight as a parameter. Plz help if a better way exists.

adocherty:
Hi,

I'm not clear as to what you are interested in doing with the edge weights.  Could you tell us more about how you envisage the edge weight to be used as a parameter for link prediction?

patchy631:
Hi @adocherty
I'm working with a road network graph. Each node in my stellar graph is a single digital road element, it's neighbouring nodes would be the roads directly connected to it in reality. So on the edge i have number of vehicles going from a road to it's neighbour connected via this edge. So for doing link prediction I need to inculcate this traffic data present on the edge as one of the parameter for downstream classification.

So link data is like:  (node1_data)-----edge_data-----(node2_data)

Thanks

PR #1667, opened 2020-06-10
Add support for edge weights to GraphSAGE sampling

huonw:
This expands GraphSAGE (undirected and directed) to support weighted sampling, where edges with higher weights are taken proportionally more often.

For example, suppose there's there's 4 edges from node A:

| source | target | weight |
|---|---|---|
| A | B | 0 |
| A | C | 1 |
| A | D | 2 |
| A | D | 3 |

An unweighed walk starting at A will choose each of the edges with equal propability and so end up on B, C or D in proportion 1:1:2 (edge counts). A weighted walk will choose the edges proportional to the weights, so end up on the vertices in proportion 0:1:5 (sum of edge weight). (Worth specifically highlighting: a weighted walk will never chose the A-B edge because it has weight 0.)

See: #462

### DIFF
--- a/stellargraph/data/explorer.py
+++ b/stellargraph/data/explorer.py
@@ -721,7 +721,7 @@ class SampledHeterogeneousBreadthFirstWalk(GraphWalk):
                 q.extend([(node, node_type, 0)])
 
                 # add the root node to the walks
-                walk.append([node])
+                walk.append(np.array([node]))
                 while len(q) > 0:
                     # remove the top element in the queue and pop the item from the front of the list
                     frontier = q.pop(0)
@@ -746,7 +746,7 @@ class SampledHeterogeneousBreadthFirstWalk(GraphWalk):
                                 samples = rs.choice(neigh_et, size=n_size[depth - 1])
                             else:  # this doesn't happen anymore, see the comment above
                                 _size = n_size[depth - 1]
-                                samples = [-1] * _size
+                                samples = np.full(_size, -1)
 
                             walk.append(samples)
                             q.extend(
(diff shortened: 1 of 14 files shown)


tests/test_utils/graphs.py
```diff
@@ -375,3 +375,35 @@ def weighted_hin():
...
+    edges = pd.DataFrame(
+        {
+            "source": [0, 0, 1, 1, 2, 2],
+            "target": [1, 2, 3, 4, 5, 6],
+            "weight": [1.0, 5, 20, 1, 1, 0],
+        }
+    )
+    cls = StellarDiGraph if is_directed else StellarGraph
+
+    def check_occurrence(sequence):
```
kjun9, line 398: nice way of checking 👍

stellargraph/data/explorer.py
```diff
@@ -233,6 +231,35 @@ def _check_sizes(self, n_size):
...
+            neighbours = neigh_func(
+                cur_node, use_ilocs=True, include_edge_weight=weighted
+            )
+
+            if weighted:
+                neighbours, weights = neighbours
+        else:
+            neighbours = []
+
+        if len(neighbours) == 0:
```
kjun9, line 251: I wonder if we should do something like:
```python
if len([w for w in weights if w != 0]) == 0:
```
so that we still have `-1`s propagated if all neighbours have zero weight?

huonw: Hm, that's a good point. That particular form is likely to be expensive because it's manually iterating the array, but it's probably worth doing something with https://docs.scipy.org/doc/numpy/reference/generated/numpy.count_nonzero.html or using the final value of the `cumsum` in `naive_weighted_choice`. I'll probably do it in the latter, because then it applies to weighted Node2Vec too (which currently doesn't handle this case).

huonw: Done, and fixed/tested in Node2Vec and CTDNE as appropriate as well.

PR #1679, opened 2020-06-12
Use np.testing.assert_array_equal, not assert np.array_equal or np.allclose

huonw:
The output of np.testing.assert_array_equal is much more informative than the output of `assert np.array_equal`, because it includes things like the number of different elements and their absolute and relative differences, e.g.:

```python
>>> assert np.array_equal([1, 2], [3, 2])
...
AssertionError
>>> np.testing.assert_array_equal([1, 2], [3, 2])
...
AssertionError: 
Arrays are not equal

Mismatch: 50%
Max absolute difference: 2
Max relative difference: 0.66666667
 x: array([1, 2])
 y: array([3, 2])
```

The same is true of `np.testing.assert_allclose` vs. `np.allclose`. This makes our test failure output both:

- more informative, because there's a hint about how wrong a failure is
- smaller, for the reproducibility tests (which previously included all of the models' weight NumPy arrays, in their entirety)

### DIFF
--- a/tests/layer/test_watch_your_step.py
+++ b/tests/layer/test_watch_your_step.py
@@ -40,7 +40,7 @@ def test_AttentiveWalk():
 
     output = att_wlk(random_partial_powers).numpy()
 
-    np.testing.assert_allclose(output, random_partial_powers.mean(axis=1))
+    np.testing.assert_allclose(output, random_partial_powers.mean(axis=1), rtol=1e-5, atol=1e-8)
 
 
 def test_WatchYourStep_init(barbell):
--- a/tests/layer/test_watch_your_step.py
+++ b/tests/layer/test_watch_your_step.py
@@ -40,7 +40,7 @@ def test_AttentiveWalk():
 
     output = att_wlk(random_partial_powers).numpy()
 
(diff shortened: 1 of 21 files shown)


tests/layer/test_watch_your_step.py
```diff
@@ -40,7 +40,9 @@ def test_AttentiveWalk():
 
     output = att_wlk(random_partial_powers).numpy()
 
-    assert np.allclose(output, random_partial_powers.mean(axis=1))
+    np.testing.assert_allclose(
+        output, random_partial_powers.mean(axis=1), rtol=1e-5, atol=1e-8
```
timpitman, line 44: does this result in slightly different behaviour with the thresholds?

huonw: This change here is actually matching the thresholds: annoyingly `allclose` and `assert_allclose` have different defaults for their parameters:

| function | `rtol` | `atol` |
|---|---|---|
| [`allclose`](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html) | 1e-5 | 1e-8 |
| [`assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html)  | 1e-7 | 0 |

PR #1686, opened 2020-06-17
Implement self-adversarial negative sampling knowledge graph loss

huonw:
This implements a slightly different loss function for knowledge graph algorithms. It's so called "self-adversarial", because it uses the knowledge graph's algorithm computed scores to give different weights to the negative sampled edges. The intuition is that some edges end up being "obviously" negative as the model trains, and so it's better to focus more attention on the highly-scored negative edges and ignore the very low scoring ones.

Recall that knowledge graph algorithms typically work by taking in a set of true edges, and, for each one, computing a series of corruptions or negative edges by changing the source or target nodes (leaving the other node and the edge type unchanged). This self-adversarial scoring works by grouping the negative edges by their original "parent" edge, and computing a softmax over the scores within that group. That is, for each true edge `(s, r, o)`, the scores of the negative samples `(s', r, o)` and `(s, r, o')` are compared against each other to compute the weights for that set of negative samples. Weights for negative edges created from other true edges (even in the same batch) are computed independently.

This approach gives a small, but, apparently, noticeable improvement in the performance of knowledge graph algorithms.

This is equations (5) and (6) from [1], and that paper includes an comparison of different scoring techniques in Table 7 (screenshots below, for ease of reference).

[1] Z. Sun, Z.-H. Deng, J.-Y. Nie, and J. Tang, “RotatE: Knowledge Graph Embedding by Relational Rotation in Complex Space,” [arXiv:1902.10197](http://arxiv.org/abs/1902.10197), Feb. 2019.

See: #1708

image

image

### DIFF
--- a/stellargraph/losses.py
+++ b/stellargraph/losses.py
@@ -29,7 +29,7 @@ def graph_log_likelihood(batch_adj, wys_output):
     """
     Computes the graph log likelihood loss function as in https://arxiv.org/abs/1710.09599.
 
-    This is different to most keras loss functions in that it doesn't directly compare predicted values to expected
+    This is different to most Keras loss functions in that it doesn't directly compare predicted values to expected
     values. It uses `wys_output` which contains the dot products of embeddings and expected random walks,
     and part of the adjacency matrix `batch_adj` to calculate how well the node embeddings capture the graph
     structure in some sense.
--- a/stellargraph/losses.py
+++ b/stellargraph/losses.py
@@ -14,6 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+__all__ = [
+    "graph_log_likelihood",
+    "SelfAdversarialNegativeSampling",
+]
+
 import tensorflow as tf
 
 from .core.experimental import experimental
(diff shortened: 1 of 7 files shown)


stellargraph/losses.py
```diff
@@ -49,3 +54,49 @@ def graph_log_likelihood(batch_adj, wys_output):
     loss = tf.math.reduce_sum(tf.abs(matrix))
 
     return tf.expand_dims(loss, 0)
+
+
+class SelfAdversarialNegativeSampling(tf.keras.losses.Loss):
+    """
```
timpitman, line 60: did you see the implementation of this at?: https://github.com/DeepGraphLearning/KnowledgeGraphEmbedding/blob/2442e8b9bb1a40e44edabcb3cc4f145ac24b1c7b/codes/model.py#L251

huonw: Yeah, I did briefly look at it. I think their non-Keras approach makes their code simpler for this particular example, because they're having to implement the full training step, so they have more detailed control over how the original sampling connects to the loss function. It also means it doesn't translate that directly into our implementation here.

That said, looking at it a second time, I realised that the implementation here isn't _quite_ right: this PR doesn't do the equivalent of the `detach` https://github.com/DeepGraphLearning/KnowledgeGraphEmbedding/blob/2442e8b9bb1a40e44edabcb3cc4f145ac24b1c7b/codes/model.py#L269-L272

I'll add a call to https://www.tensorflow.org/api_docs/python/tf/stop_gradient which should be equivalent.

Issue #1687, opened 2020-06-18
Migrate CI to a free service


huonw:
### Description

We may have to migrate off out Buildkite AWS agent scaling account.

Some alternative options:

| Name | Free compute | Notes |
|---|---|---|
| Buildkite | [Unclear](https://buildkite.com/pricing) | Current provider, works great |
| Travis CI | [5 concurrent](https://www.travis-ci.com/plans) | Kinda clunky, IME |
| Circle CI | [1 concurrent](https://circleci.com/pricing/) | No experience |
| GitHub Actions | [20 concurrent](https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits) | No experience, well integrated into GitHub, Rust (with high CI demand) [switched to it](https://blog.rust-lang.org/inside-rust/2020/04/07/update-on-the-github-actions-evaluation.html) |
| Azure Pipelines | [1 concurrent](https://azure.microsoft.com/en-us/pricing/details/devops/azure-devops-services/) | No experience, Rust previously used it but migrated to Actions |
| Drone CI | [Unclear, maybe 1 concurrent](https://drone.io/enterprise/opensource/#features) | No experience, but heard of it, vaguely |
| AppVeyor | [1 concurrent](https://www.appveyor.com/pricing/) | No experience, but heard of it used for Windows along side Travis (Linux) |

Other lists/comparisons:

- https://github.com/ligurio/awesome-ci

### Checklist for migration

The steps from `.buildkite/pipeline.yml` https://github.com/stellargraph/stellargraph/blob/3867a66b5df85b2b3bb33cd324b95829552008e2/.buildkite/pipeline.yml

- [x] (N/A) :docker::python: build 3.6 runner
- [x] (N/A) :docker::python: build 3.7 runner
- [x] (N/A) :docker::python: build 3.8 runner
- [x] (#1688) :python: 3.6
- [x] (#1688) :python: 3.7
- [x] (#1688) :python: 3.8
- [x] (#1692) :python::book: test notebooks %n
- [x] (#1694) :python::book: test neo4j notebooks against neo4j 4.0
- [x] (#1694) :python::book: test neo4j notebooks against neo4j 3.5
- [x] (#1690) :python-black: format
- [x] (#1697) :python-black::jupyter: check notebook format
- [x] (#1697) :jupyter::book: check notebook text
- [x] (#1717) :book: check demo indexing
- [x] (#1717) :book: check documentation
- [x] (#1715) :shell: format
[...]

PR #1688, opened 2020-06-18
Add basic GitHub Actions CI workflow, running pytest tests

huonw:
This is an MVP of using GitHub Actions for CI instead of Buildkite. This change is designed to reduce our costs and our DevOps requirements.

It works successfully for that:

- The build is running, on Python 3.6, 3.7 and 3.8, in parallel: https://github.com/stellargraph/stellargraph/actions/runs/139107467
- The JUnit XML files are being uploaded: (see pytest-results-... in link above)
- Codecov results are being uploaded: https://codecov.io/gh/stellargraph/stellargraph/commit/6470d09d5cdb07620a352544b6e865724f4254f8/build

This is quite easy because there's native support (or actions) for:
- build matrices, with multiple options such as different Python versions
- dependency caching (so we don't have to do the prebuilding of docker images like we do on Buildkite... for now) (I haven't investigated the improvement from this, yet, though)
- attaching artefacts
- uploading codecov results

Things missing compared to Buildkite:

- clean annotation of test failures?
- everything other than pytest tests (notebooks, documentation, formatting, etc.)

These can be done as follow up work.

See: #1687

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,14 +43,14 @@ jobs:
           # just runs them once (as a test)
           py.test -ra tests/ --doctest-modules \
               --cov=stellargraph --cov-report=xml \
-              -p no:cacheprovider --junitxml="junit-${{ matrix.python-version }}.xml" \
+              -p no:cacheprovider --junitxml="pytest-results-${{ matrix.python-version }}.xml" \
               --benchmark-disable
 
       - name: Upload pytest test results
         uses: actions/upload-artifact@v1
         with:
           name: pytest-results-${{ matrix.python-version }}
-          path: junit-${{ matrix.python-version }}.xml
+          path: pytest-results-${{ matrix.python-version }}.xml
         if: ${{ always() }}
 
       - name: Upload coverage to Codecov
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,43 @@
+---
+
+name: Continuous Integration
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[test,demos]
+
+      - name: Run pytest tests
+        run: |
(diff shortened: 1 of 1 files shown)


huonw:
This seems to be working vaguely successfully:

- The build is running, on Python 3.6, 3.7 and 3.8, in parallel: https://github.com/stellargraph/stellargraph/actions/runs/139107467
- The JUnit XML files are being uploaded: (see pytest-results-... in link above)
- Codecov results are being uploaded: https://codecov.io/gh/stellargraph/stellargraph/commit/6470d09d5cdb07620a352544b6e865724f4254f8/build

This is quite easy because there's native support (or actions) for:
- build matrices, with multiple options such as different Python versions
- dependency caching (so we don't have to do the prebuilding of docker images like we do on Buildkite... for now)
- attaching artefacts
- uploading codecov results

huonw:
> dependency caching (so we don't have to do the prebuilding of docker images like we do on Buildkite... for now) (I haven't investigated the improvement from this, yet, though)

| build | restore cache step | dep installation step | total  before "Run pytest tests" |
|---|---|---|---|
| [with cache](https://github.com/stellargraph/stellargraph/pull/1688/checks?check_run_id=782973387) | 0:36 | 2:31 | 3:14 |
| [no cache](https://github.com/stellargraph/stellargraph/pull/1688/checks?check_run_id=782732519) | 0:01 | 3:47 | 4:01 |

That is, without any GitHub Actions-specific optimisations beyond copy-pasting the caching code from https://github.com/actions/cache/blob/master/examples.md#simple-example, the time to get to "running" is ~3 minutes.

This isn't far off the time to get to running on Buildkite, which probably ranges from ~1 minute to 3.5 minutes or more (depending on docker image cache hits), e.g. for https://buildkite.com/stellar/stellargraph-public/builds/4794:

| version | build runner step | time until running `script.sh` in 3.x step | total before `script.sh` |
|---|---|---|---|
| 3.6 | 0:44 (hit) | 1:51 (miss) | 2:35 |
| 3.7 | 1:37 (miss) | 0:22 (hit) | 1:59 |
| 3.8 | 1:42 (miss) | 0:43 (hit) | 2:25 |

The hit/miss refers to whether the docker images being pulled from docker hub already existed on the agent (a cache hit) or if the hundreds-of-megabyte images had to be downloaded (a miss), taking about a minute.

The "build runner" steps had per-layer cache hits, in that were rebuilding an image from cache without having to recompute any layers (and thus the push is fast). However, the overall image still has to exist on the agent, and so without an image cache hit, the majority of the time is in pulling that with very little time spent actually building the image or repushing it. (When the builder already has the relevant image locally, the pull step is fast, as in 3.6).

The "time until running ..." refers to everything before "Running /bin/sh ..." in the "3.x" steps. This includes potentially having to pull the runner image from the "build 3.x runner" step from docker hub. This is really fast if the build agent already has the image locally, but slow otherwise.

huonw:
The merge into develop seems to be running tests correctly: https://github.com/stellargraph/stellargraph/actions/runs/139262212

Issue #1251, opened 2020-04-16
Full batch models fail to save with "If specifying TensorSpec names for nested structures, either zero or all names have to be specified"


huonw:
### Describe the bug

As reported in https://community.stellargraph.io/t/valueerror-if-specifying-tensorspec-names-for-nested-structures-either-zero-or-all-names-have-to-be-specified/97, saving GAT or any other full batch model seems to fail.

### To Reproduce

```python
import stellargraph as sg
import tensorflow as tf

def test(x):
    print(f"### {type(x).__name__}")
    print("```")
    model = tf.keras.Model(*x.in_out_tensors())
    try:
        tf.keras.models.save_model(model, "/tmp")
    except Exception as e:
        print(f"tf.keras.models.save_model: {e!r}")
        
    try:
        tf.saved_model.save(model, "/tmp")
    except Exception as e:
        print(f"tf.saved_model.save: {e!r}")

    try:
        model.save("/tmp")
    except Exception as e:
        print(f"model.save: {e!r}")
    print("```")


G, _ = sg.datasets.Cora().load()

fb_gen = sg.mapper.FullBatchNodeGenerator(G)
test(sg.layer.GAT([4], fb_gen))

test(sg.layer.GCN([4], fb_gen))

test(sg.layer.PPNP([4], fb_gen, ["relu"]))
test(sg.layer.APPNP([4], fb_gen, ["relu"]))

gs_gen = sg.mapper.GraphSAGENodeGenerator(G, 5, [2])
test(sg.layer.GraphSAGE([4], gs_gen))

hs_gen = sg.mapper.HinSAGENodeGenerator(G, 5, [2], head_node_type="paper")
test(sg.layer.HinSAGE([4], hs_gen))

cg_gen = sg.mapper.ClusterNodeGenerator(G)
test(sg.layer.ClusterGCN([4], ["relu"], cg_gen))

av_gen = sg.mapper.Attri2VecNodeGenerator(G, 10)
test(sg.layer.Attri2Vec([4], av_gen))

rg_gen = sg.mapper.RelationalFullBatchNodeGenerator(G)
test(sg.layer.RGCN([4], rg_gen))

kg_gen = sg.mapper.KGTripleGenerator(G, 10)
test(sg.layer.ComplEx(kg_gen, 10))
test(sg.layer.DistMult(kg_gen, 10))

pg_gen = sg.mapper.PaddedGraphGenerator([G])
[...]

PR #1689, opened 2020-06-18
Test saving sparse full batch methods against tf-nightly

huonw:
The upstream tensorflow issue(s) (https://github.com/tensorflow/tensorflow/issues/38465, https://github.com/tensorflow/tensorflow/issues/40373) that stopped us being able to save our sparse full batch models seems to be fixed in `tf-nightly`, based on the minimal reproducer we provided there. This PR is working out whether it works for the full code.

See: #1251


huonw:
This does seem to work in that the un-xfailed tests here pass (the failure looks like TensorFlow changed some error messages and/or validation about a particular edge case that the test is exercising).

PR #1691, opened 2020-06-18
Add error annotations in black GitHub Actions action

huonw:
This uses the `::error file=...::...` command https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message to add annotations onto each file that's misformatted. This is a more consistent experience for black, matching our other linters. It is also arguably better and easier to understand, because it involves less digging through logs.

In particular, files that have incorrect formatting now have a "Check" comment added to the first line. This is visible as a summary report when clicking on "details" of the black step (for instance, https://github.com/stellargraph/stellargraph/pull/1691/checks?check_run_id=794668066 ) and also directly attached to each file under the "Files changed" tab:

image

See: #1687


huonw:
The changes to `setup.py` are just to demonstrate the check, they'll be entirely reverted before landing.

.github/workflows/ci.yml
```diff
@@ -211,3 +197,29 @@ jobs:
...
+          echo ::add-path::$HOME/.local/bin
+
+      # run black and annotate incorrect files with reviewdog
+      - name: Install black
+        run: |
+          python3 -m pip install --upgrade pip
+          python3 -m pip install black==19.10b0
+      - name: Check Python formatting
+        env:
+          REVIEWDOG_GITHUB_API_TOKEN: ${{ github.token }}
```
timpitman, line 217: does this need reviewdog rather than just using github actions error output?

huonw: I guess we could do the `sed` here on the output of `black` directly to get the right structure. That is probably cleaner. I'll give it a go 👍

huonw: Yeah, much simpler, with less installation and so on.

PR #1692, opened 2020-06-19
Run notebooks on GitHub Actions

huonw:
This runs all the basic Jupyter notebooks on Github Actions, in parallel. It does this in the simplest way to get parallelism (and also, the simplest way to get a "loop" in GitHub Actions configuration), with a large job matrix: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix .

The matrix has to be manually specified, so a new script (`scripts/ci/ci_workflow_notebook_list.py`) is added, that checks that the matrix of notebooks matches the notebooks on disk.

This script is better if it is simple, and it's better if the configuration all in one place in the `ci.yml` file. Thus, the notebooks that aren't run in the default configuration (such as Neo4j notebooks #1694, and notebooks that fail #818 #819) are still listed and then excluded with `exclude`: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix

See: #1687


scripts/ci/run_notebook.sh
```diff
@@ -0,0 +1,12 @@
...
+set -euo pipefail
+
+notebook="$1"
+# papermill will replace parameters on some notebooks to make them run faster in CI
+papermill --execution-timeout=600 \
+  --parameters_file "./.buildkite/notebook-parameters.yml" --log-output \
+  "$notebook" "$notebook"
+
+notebook_name="$(basename "$notebook")"
+echo "::set-output name=notebook_name::$notebook_name"
```
timpitman, line 12: I used "environment" rather than "output" in my PR - this looks better, will update mine 👍

PR #1694, opened 2020-06-19
Run Neo4j notebooks on GitHub Actions

huonw:
This expands on #1692 to also run the notebooks that use Neo4j on our new GitHub Actions CI.

It does this by using docker-compose to manually run the Neo4j image in the background. This is a bit clunky, but is done in the interests of getting something working now. In particular, I couldn't work out how to use the native support for "services" for a job (https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices) to build the image from a local dockerfile.

See: #1687


scripts/ci/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+---
```
timpitman, line 1: could this file should have a more descriptive name than just "docker-compose.yml"?

huonw: Potentially, but maybe the `ci` directory name is enough?

huonw: I'll rename to "actions-docker-compose.yml" since it's having to be used with the `-f` argument anyway and so `docker-compose`'s default of looking for `./docker-compose.yml` doesn't help.

PR #1697, opened 2020-06-22
Check notebook formatting and text on Github Actions CI

timpitman:
See #1687 

This adds stages for:
* check-notebook-format
* check-notebook-text

Files producing errors are annotated - a run with a break is shown here: https://github.com/stellargraph/stellargraph/actions/runs/144396650

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,6 @@
 ---
-name: Continuous Integration
+# Continuous Integration is too long for the GitHub view
+name: CI
 
 "on":
   push:
@@ -9,11 +10,15 @@ name: Continuous Integration
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8]
+        os: [ubuntu-latest, windows-latest]
+
+    runs-on: ${{ matrix.os }}
+
+    env:
+      JUNIT_NAME: pytest-results-${{ matrix.python-version }}-${{ matrix.os }}
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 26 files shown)


.github/workflows/ci.yml
```diff
@@ -316,3 +316,38 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+
+      - name: Install requirements
+        run: |
+          pip install nbconvert
+          pip install nbformat
+          pip install black
```
huonw, line 333: I think this can be one line, and, more importantly, I think the black version should be fixed, so that a black upgrade doesn't randomly change our formatting (different to our real dependencies, because this is just an internal tool).

```suggestion
          pip install nbconvert nbformat black==19.10b0
```

The one-line-ing could be applied to the checker too.

Finally, the GitHub Actions Python guide suggests upgrading pip as the first step: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#installing-dependencies

scripts/format_notebooks.py
```diff
@@ -520,7 +520,7 @@ def preprocess(self, nb, resources):
                     check_failed.append(str(file_loc))
 
                     if on_ci:
-                        # CI doesn't provide enough state to diagnose a peculiar or
+                        # Buildkite CI doesn't provide enough state to diagnose a peculiar or
```
huonw, line 523:
```suggestion
                        # CI doesn't provide enough state to diagnose a peculiar or
```

Github Actions doesn't provide enough state either.

scripts/notebook_text_checker.py
```diff
@@ -517,6 +517,18 @@ def file_list(path, errors):
...
+            def github_file_list(path, errors):
+                whole_list = "%0A".join(error.replace("\n", "%0A") for error in errors)
+                return (
+                    f"::error file={path}::Notebook failed text check:%0A{whole_list}"
+                )
+
+            annotation_output = "\n".join(
+                github_file_list(path, errors) for path, errors in all_errors
+            )
+            print(annotation_output)
```
huonw, line 531: I think this can be slightly simplified to a single loop that prints line-by-line?

```python
for path, errors in all_errors:
    whole_list = ...
    message = f"..."
    print(f"::error file={path}::{message}")
```

scripts/notebook_text_checker.py
```diff
@@ -517,6 +517,18 @@ def file_list(path, errors):
 
 This check can be run locally, via `{command}`."""
 
+        if "GITHUB_ACTIONS" in os.environ:
+
+            def github_file_list(path, errors):
+                whole_list = "%0A".join(error.replace("\n", "%0A") for error in errors)
```
huonw, line 523: Another option here would be to make it more "human-facing" for writing, and then do a single escaping step at the end; something like:

```suggestion
                whole_list = "\n".join(error for error in errors)
                message = f"""\
Notebook failed text check

{whole_list}"""
                escaped = message.replace("\n", "%0A")
```

This also makes it easier to incorporate more text, e.g. the `command`, if you think that's sensible.

This is totally optional and ties into the suggestion below about doing one loop.

huonw:
The CI errors seemed to be mechanical (reformatting `ci.yml` with prettier, and undeleted the accidentally(?)-deleted `scripts/ci/check-copyright-headers.sh` file), so I've done them and pushed, so hopefully it's ready for you to merge tomorrow morning without needing to wait for CI.

PR #1706, opened 2020-06-22
Test notebooks on Windows too

huonw:
Extension to #1696


huonw:
This involves far too many check to reasonably run on CI for the moment, but we've verified that it works at the moment, so that is good enough for now.

PR #1707, opened 2020-06-22
Annotate pytest failures with GitHub Actions

huonw:
This replaces the JUnit test summary annotation used on Buildkite with annotations on individual tests about the failures. In particular, a custom script walks through the JUnit XML file to find any failures or errors, and adds a summary of the failures to each test via the `::error file=...::...` command https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message 

For reference, a JUnit XML file emitted by pytest looks something like:

```xml
<?xml version="1.0" ?>
<testsuites>
	<testsuite errors="4" failures="18" hostname="fv-az28" name="pytest" skipped="12" tests="890" time="916.014" timestamp="2020-06-22T01:56:26.278020">
		<testcase classname="tests.test_aaa_on_gpu" name="test_on_gpu_when_requested" time="0.000">
			<error message="STELLARGRAPH_MUST_USE_GPU is not set to 1, so a GPU does not have to be used" type="pytest.skip">tests\test_aaa_on_gpu.py:26: STELLARGRAPH_MUST_USE_GPU is not set to 1, so a GPU does not have to be used</error>
		</testcase>
		<testcase classname="tests.test_calibration" name="test_temperature_scaling_bad_input_type" time="3.927"/>
		<testcase classname="tests.test_calibration" name="test_temperature_scaling_fit_predict" time="11.547">
			<system-out>Using Early Stopping based on performance evaluated on given validation set.
</system-out>
		</testcase>
[...]
		<testcase classname="tests.datasets.test_datasets" name="test_cora_load[True-True-True]" time="0.798">
			<failure message="AssertionError: assert dtype('int64') == int  +  where dtype('int64') = Int64Index([  31336, 1061127, 1106406,   13195,   37879, 1126012, 1107140,\n            1102850,   31349, 1106418,\n    ...227, 1131180, 1128974, 1128975, 1128977,\n            1128978,  117328,   24043],\n           dtype='int64', length=2485).dtype  +    where Int64Index([  31336, 1061127, 1106406,   13195,   37879, 1126012, 1107140,\n            1102850,   31349, 1106418,\n    ...227, 1131180, 1128974, 1128975, 1128977,\n            1128978,  117328,   24043],\n           dtype='int64', length=2485) = &lt;bound method StellarGraph.nodes of &lt;stellargraph.core.graph.StellarDiGraph object at 0x000001EC160766D8&gt;&gt;()  +      where &lt;bound method StellarGraph.nodes of &lt;stellargraph.core.graph.StellarDiGraph object at 0x000001EC160766D8&gt;&gt; = &lt;stellargraph.core.graph.StellarDiGraph object at 0x000001EC160766D8&gt;.nodes">is_directed = True, largest_cc_only = True, subject_as_feature = True

    @pytest.mark.parametrize(&quot;is_directed&quot;, [False, True])
    @pytest.mark.parametrize(&quot;largest_cc_only&quot;, [False, True])
[...]    
&gt;       assert g.nodes().dtype == int
E       AssertionError: assert dtype('int64') == int
E        +  where dtype('int64') = Int64Index([  31336, 1061127, 1106406,   13195,   37879, 1126012, 1107140,\n            1102850,   31349, 1106418,\n    ...227, 1131180, 1128974, 1128975, 1128977,\n            1128978,  117328,   24043],\n           dtype='int64', length=2485).dtype
E        +    where Int64Index([  31336, 1061127, 1106406,   13195,   37879, 1126012, 1107140,\n            1102850,   31349, 1106418,\n    ...227, 1131180, 1128974, 1128975, 1128977,\n            1128978,  117328,   24043],\n           dtype='int64', length=2485) = &lt;bound method StellarGraph.nodes of &lt;stellargraph.core.graph.StellarDiGraph object at 0x000001EC160766D8&gt;&gt;()
E        +      where &lt;bound method StellarGraph.nodes of &lt;stellargraph.core.graph.StellarDiGraph object at 0x000001EC160766D8&gt;&gt; = &lt;stellargraph.core.graph.StellarDiGraph object at 0x000001EC160766D8&gt;.nodes

tests\datasets\test_datasets.py:194: AssertionError</failure>
		</testcase>
[...]
		<testcase classname="tests.utils.test_hyperbolic" name="test_poincare_ball_exp_specialisation" time="0.000">
			<error message="test setup failure">@pytest.fixture
    def seeded():
&gt;       seed = np.random.randint(2 ** 32)
[...]
&gt;   ???
E   ValueError: high is out of bounds for int32

_bounded_integers.pyx:1343: ValueError</error>
		</testcase>
[...]
	</testsuite>
</testsuites>
```

Note that `<testcase>` errors and failures are indicated by `<failure message=...>...</failure>` and `<error message=...>...</error>` child elements.

The JUnit files emitted by pytest don't include the test location at all, so we have to estimate them, by searching the stack trace (text contents of the `<failure>`/`<error>` element) for the first match for a `<file>:<line>:` pattern. This should generally be StellarGraph code, and will usually be the call or `assert` within the test function that had the error.

This is an optional extra, and it only triggers when tests fail, so this heuristic/approximate approach are okay. The code detects failures and highlight them, so that things can be fixed in future.

For instance, for https://github.com/stellargraph/stellargraph/pull/1707/checks?check_run_id=798011967 :

image

The annotations also are included for files that aren't modified. (Unfortunately, it seems there's more bugs here, because only a small number of the total number of failure annotations are actually listed.) For instance, if a change to code breaks an existing test:

image

(It's unclear to me why the cards are listed twice, even for a single OS/Python-version pair. I think it's a GitHub bug, because it's definitely only in the logs once: https://github.com/stellargraph/stellargraph/pull/1707/checks?check_run_id=798011967#step:8:31 )

Finally, as warned by codeclimate, the `xml.etree.ElementTree` module is apparently unsafe, for untrusted XML data. This doesn't concern us particularly, because (a) the JUnit XML files are being generated by us and so are trusted, and (b) the problems of an attack aren't new (if someone is able to attack this part of the CI, they can already run arbitrary code in it).

See: #1687

### DIFF
--- a/scripts/ci/junit_to_github_checks.py
+++ b/scripts/ci/junit_to_github_checks.py
@@ -17,43 +17,13 @@
 import argparse
 from collections import defaultdict
 import os
+import re
 import sys
 import textwrap
 import xml.etree.ElementTree as ET
 
 
-def deduce_file_position(testcase, base):
-    name = testcase.get("name")
-    try:
-        index = name.index("[")
-    except:
-        # no parameters
-        base_name = name
-    else:
-        base_name = name[:index]
-
-    classname = testcase.get("classname")
-    filename = classname.replace(".", "/") + ".py"
-
-    open_filename = os.path.join(base, filename)
-    try:
-        with open(open_filename) as f:
-            contents = f.read()
-    except:
-        raise ValueError(
-            f"failed to read '{filename}', as deduced from failing test '{name}' in '{classname}'"
-        )
-
-    # estimate the location of the test by doing a string search
-    try:
-        index = contents.index(f"def {base_name}")
-    except:
-        raise ValueError(f"could not find failing test '{base_name}' in '{filename}'")
-
-    # number of newlines since the start of the file = zero-based line, GitHub uses one-based lines
-    line = contents.count("\n", 0, index) + 1
-
-    return filename, line
+FILE_LINE = re.compile(r"([a-z0-9_/\\.]*):([0-9]*):")
 
 
 def main():
@@ -63,8 +33,6 @@ def main():
     parser.add_argument("file", type=argparse.FileType("r"), default="-", nargs="?")
     args = parser.parse_args()
 
-    base_directory = os.path.join(os.path.dirname(__file__), "../..")
-
     tree = ET.parse(args.file)
     root = tree.getroot()
 
@@ -72,13 +40,19 @@ def main():
     for testcase in root.findall(".//testcase"):
         for child in testcase:
             if child.tag in ("error", "failure"):
-                try:
-                    filename, line = deduce_file_position(testcase, base_directory)
-                except ValueError as e:
-                    invalid.append(e)
+                classname = testcase.get("classname")
+                name = testcase.get("name")
+
+                match = FILE_LINE.search(child.text)
+                if match is None:
+                    invalid.append(
+                        f"output of test '{name}' in '{classname}' does not contain match for filename & line regex /{FILE_LINE.pattern}/"
+                    )
                     continue
 
-                name = testcase.get("name")
+                filename = match[1]
+                line = int(match[2])
+
                 base_message = child.get("message").replace("\\n", "\n")
                 indented = textwrap.indent(base_message, "    ")
 
--- a/scripts/ci/junit_to_github_checks.py
+++ b/scripts/ci/junit_to_github_checks.py
@@ -38,13 +38,15 @@ def deduce_file_position(testcase, base):
     with open(os.path.join(base, filename)) as f:
         contents = f.read()
 
-    # estimate the line of the test by doing a string search
+    # estimate the location of the test by doing a string search
     try:
         index = contents.index(f"def {base_name}")
     except:
         raise ValueError(f"count not find failing test '{base_name}' in {filename}")
 
-    line = contents.count("\n", index) + 1
+    # number of newlines since the start of the file = zero-based line, GitHub uses one-based lines
+    line = contents.count("\n", 0, index) + 1
+
     return filename, line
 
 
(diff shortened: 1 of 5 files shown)


huonw:
Only the changes to `ci.yml` and the new `scripts/ci/junit_to_github_checks.py` file are real. The example changes to `gcn.py` and `test_graph.py` will be reverted before landing.

scripts/ci/junit_to_github_checks.py
```diff
@@ -0,0 +1,81 @@
...
+    parser.add_argument("file", type=argparse.FileType("r"), default="-", nargs="?")
+    args = parser.parse_args()
+
+    tree = ET.parse(args.file)
+    root = tree.getroot()
+
+    invalid = []
+    for testcase in root.findall(".//testcase"):
+        for child in testcase:
+            if child.tag in ("error", "failure"):
```
timpitman, line 42: does this skip annotating warnings?

huonw:
> There are many duplicate annotations in the sample test failure - a future extension could be to group and filter duplicates

The duplicates are either different parameterisations of the test (which are vaguely useful to repeat?), or the seeming GitHub bugs where a single annotation is shown more than once.

scripts/ci/junit_to_github_checks.py
huonw: Yeah it doesn't. Unfortunately, I don't think warnings are captured as a separate thing. For instance, the summary of the tags and their attributes and the counts (`[...]`) in a random Pytest file I have lying around is:

```
<testsuites> [1] 
  <testsuite> [1] errors, failures, hostname, name, skipped, tests, time, timestamp
    <testcase> [890] classname, name, time
      <skipped> [12] message, type
      <system-out> [138] 
      <failure> [18] message
      <system-err> [31] 
      <error> [4] message
``

Also, at the moment we end up with a lot of warnings from tests, so we'd have a lot of noise here.

Issue #1711, opened 2020-06-23
40k instances of DeprecationWarning: tostring() is deprecated. Use tobytes() instead. in Github Actions logs


huonw:
### Describe the bug

Tests runs on GitHub Actions result in thousands of lines of logging, related to a single deprecation warning.

Details:

| Actions | Buildkite | Actions status | commit |
|---|---|---|---|
| https://github.com/stellargraph/stellargraph/runs/783114964?check_suite_focus=true#step:6:88 | https://buildkite.com/stellar/stellargraph-public/builds/4814#3e6b3710-da57-4f89-bd4f-410a8cf6c628/262-340 | good | https://github.com/stellargraph/stellargraph/commit/3165a5281004c88c556a67e7a2d0e4eb59c8d427 (last good) |
| https://github.com/stellargraph/stellargraph/runs/793566010?check_suite_focus=true#step:6:39880 | https://buildkite.com/stellar/stellargraph-public/builds/4886#e1f2642d-36ed-4dce-9fa6-3ef999115437/262-340 | bad | https://github.com/stellargraph/stellargraph/commit/aa2accf6ecc61347e71cd64d6b16218a7444d239 (first bad)
| https://github.com/stellargraph/stellargraph/runs/797744297?check_suite_focus=true#step:6:39881 | https://buildkite.com/stellar/stellargraph-public/builds/4962#8f167b08-b3a5-49e5-947b-81c3bffeaca9/262-340 | bad | https://github.com/stellargraph/stellargraph/commit/ba95b0ea5212799b39278341c5c450bfef393632 (current develop) |

### To Reproduce

???

### Observed behavior

```
=============================== warnings summary ===============================
...
tests/test_ensemble.py::test_ensemble_init_parameters
tests/test_ensemble.py::test_ensemble_init_parameters
tests/test_ensemble.py::test_ensemble_init_parameters
...
tests/reproducibility/test_graphsage.py::test_link_prediction[False]
tests/reproducibility/test_graphsage.py::test_link_prediction[False]
tests/reproducibility/test_graphsage.py::test_link_prediction[False]
tests/reproducibility/test_graphsage.py::test_link_prediction[False]
  /opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/site-packages/tensorflow/python/framework/tensor_util.py:523: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
    tensor_proto.tensor_content = nparray.tostring()
```

### Expected behavior

[...]

PR #1716, opened 2020-06-23
Copy environment variables from Buildkite to GitHub Actions

huonw:
See: #1711

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
 
     env:
       JUNIT_NAME: pytest-results-${{ matrix.python-version }}-${{ matrix.os }}
-      PYTHONDONTWRITEBYTECODE: 1
       # make sure that no new uses of the legacy constructor are added (see also: test in
       # test_graph.py, filterwarnings in pytest.ini)
       PYTHONWARNINGS: "error:Constructing a StellarGraph:DeprecationWarning"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,9 @@ jobs:
           python -m pip install --upgrade pip
           pip install .[test,demos]
 
+      - name: List dependency versions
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,6 +19,10 @@ jobs:
 
     env:
       JUNIT_NAME: pytest-results-${{ matrix.python-version }}-${{ matrix.os }}
+      PYTHONDONTWRITEBYTECODE: 1
+      # make sure that no new uses of the legacy constructor are added (see also: test in
+      # test_graph.py, filterwarnings in pytest.ini)
+      PYTHONWARNINGS: "error:Constructing a StellarGraph:DeprecationWarning"
```
timpitman, line 25: I don't think this is needed now as it's in purest config?

huonw: Sort of... it's been in the pytest config since the start; I now recall that having this as an environment variable was designed to catch problems in notebooks which don't use pytest.ini (see the description of https://github.com/stellargraph/stellargraph/pull/1027 for more discussion). And... I suspect it is the cause of the difference between Actions and Buildkite in #1711, given the build here _didn't_ have the 40k lines of logging: https://github.com/stellargraph/stellargraph/pull/1716/checks?check_run_id=798025608#step:7:91

I'll quickly remove the `PYTHONDONTWRITEBYTECODE: 1` to be sure it's `PYTHONWARNINGS` that's affecting it. (My guess is that it is, by overriding the default settings that silence the warning, or something.)

huonw: Yep, it looks just specifying the non-default `PYTHONWARNINGS` here is enough to silence the warning-spew: https://github.com/stellargraph/stellargraph/pull/1716/checks?check_run_id=798135852#step:7:90

huonw:
Closing this investigatory PR in favour of a more surgical fix: https://github.com/stellargraph/stellargraph/pull/1724

PR #1717, opened 2020-06-23
Github Actions CI add checks for documentation and demo indices

timpitman:
See #1687 

This PR adds the CI stages for:
* demo index checking
* documentation checking


docs/spelling_wordlist.txt
```diff
@@ -298,3 +298,5 @@ visualize
 visualizing
 weren
 Éric
+Neo
+laplacian
```
huonw, line 302: This is a proper noun and so should be capitalised:

```suggestion
Laplacian
```

Issue #1703, opened 2020-06-22
API docs and demos should have more inter-links


huonw:
### Description

Our API docs and our demos are somewhat separate, e.g. https://stellargraph.readthedocs.io/en/v1.1.0/api.html#stellargraph.StellarGraph.from_networkx doesn't link to the NetworkX demo.

There should be more links, to make it easier for people to jump between and find example code.

PR #1718, opened 2020-06-23
Add more "see also" to API documentation, particularly to demos

huonw:
This goes through our documentation and adds more links to other relevant things. It does this by adding [a `.. seealso::` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso) to many classes/functions, and adds some semi-structured content to each.

(Note: the Sphinx docs say that the `seealso` contents should be [a definition list](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#definition-lists), but this (a) looks bad, and (b) isn't even [followed by Sphinx itself](https://github.com/sphinx-doc/sphinx/blob/3.x/doc/usage/restructuredtext/directives.rst#directives) ("Refer to the ...").)

In particular:

- for each model (e.g. GCN), it lists:
  - any demos using the model
  - the data generator(s) to use with the model
  - any other related models (e.g. RGCN as a heterogeneous generalisation of GCN)
  - example: https://stellargraph--1718.org.readthedocs.build/en/1718/api.html#graphsage
- for each generator, it lists:
  - the models that can use the generator
  - some demos that use the generator (not a full list, like for the models)
  - any other related generators (e.g. the link generator for node models)
  - example: https://stellargraph--1718.org.readthedocs.build/en/1718/api.html#stellargraph.mapper.GraphSAGENodeGenerator
- for the random walks (like `BiasedRandomWalk`), it lists:
  - any demos using the random walker
  - the other random walks, including `UnsupervisedSampler`
  - example: https://stellargraph--1718.org.readthedocs.build/en/1718/api.html#stellargraph.data.UniformRandomWalk
- for datasets, it lists:
  - some tips about loading data
  - the source link (that was already there, and is just moved into the `.. seealso`)
  - example: https://stellargraph--1718.org.readthedocs.build/en/1718/api.html#stellargraph.datasets.Cora
- for a few other things, it follows approximately this pattern.

As part of doing this, it makes heavy use of [the `:class:` role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-class) for cross-referencing. Since a lot of the cross references are to other modules (e.g. `stellargraph.layer.GCN` referencing `stellargraph.mapper.FullBatchNodeGenerator`), it uses the auto-search dotted syntax, like ``:class:`.FullBatchNodeGenerator` `` (see discussion at the end of the linked "Cross-referencing Python objects" section for more details about this). This change is actually applied across all of our existing docs, because there were several existing instances of `:class:` cross-references that failed to become links (for instance, the "StellarGraph" in the "Returns" of `Cora.load`: [before](https://stellargraph.readthedocs.io/en/v1.1.0/api.html#stellargraph.datasets.Cora.load), [after](https://stellargraph--1718.org.readthedocs.build/en/1718/api.html#stellargraph.datasets.Cora.load)).

See: #1703


huonw:
Note to self: this needs updating for #1725 (but I think that's already been flagged by the merge-conflicts).

PR #1726, opened 2020-06-24
Compute the cache dir with pip, for Windows

huonw:


huonw:
We decided to go with #1719, because not having a cache is faster. 🤷‍♂️

PR #1728, opened 2020-06-25
Override doc order for `stellargraph` top-level docs

huonw:


huonw:
This results in all of the `StellarGraph` class methods being shuffled, so closing.

PR #1732, opened 2020-06-25
Add Actions job to serve as the 'required' pre-merge check on GitHub

huonw:
This PR adds a single CI Actions job `all-jobs-passed` that represents the status of the whole CI workflow run. In particular, if the whole run passes, the `all-jobs-passed` job will pass, and if any step fails, the `all-jobs-passed` will be skipped and thus not run. This allows us to have a single status check to mark as required, instead of having to mark every single testing and notebook (and all the others) manually, in a way that cannot be validated automatically (see #1739 for more discussion of this).

It works by having the `all-jobs-passed` step have [a `needs` dependency](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) on every other step. This means that the `all-jobs-passed` step can only run if all the other jobs pass.

Maintaining this list will be error prone, so the existing `scripts/ci/ci_workflow_notebook_list.py` script is adapted to check this `needs` list in addition to checking the notebook matrix. (The script was renamed to `scripts/ci/validate_ci_workflow.py` and edited, but the changes are significant enough that git is treating this as just a delete/add, not a rename.)

Validation:

- all jobs passed, and so did the `all-jobs-passed` job: https://github.com/stellargraph/stellargraph/pull/1732/checks?check_run_id=817171302
- a job failed, and the `all-jobs-passed` job was skipped: https://github.com/stellargraph/stellargraph/pull/1732/checks?check_run_id=817050460

See: #1739


scripts/ci/validate_ci_workflow.py
```diff
@@ -0,0 +1,133 @@
...
+import glob
+from collections import Counter
+
+NOTEBOOK_MARKER = "# MARKER: list of all notebooks"
+NOTEBOOK_JOB = "notebooks"
+
+ALL_JOBS_PASSED_MARKER = "# MARKER: list of all jobs"
+ALL_JOBS_PASSED_JOB = "all-jobs-passed"
+
+WORKFLOW = ".github/workflows/ci.yml"
```
TheDen, line 28: This could potentially be generated via relative paths, if it's required to be run outside the folder

huonw: Good idea, I'll use `__file__` and `os.path` to make this be relative to the script, rather than the PWD of the user who executes it.

huonw:
Just to be sure this is working as desired, here's some builds from this PR:

- all jobs passed, and so did the `all-jobs-passed` job: https://github.com/stellargraph/stellargraph/pull/1732/checks?check_run_id=817171302
- a job failed, and the `all-jobs-passed` job was skipped: https://github.com/stellargraph/stellargraph/pull/1732/checks?check_run_id=817050460

huonw:
I've now marked the job here as required:

image

It seems to work, e.g. while waiting for a build to finish:

image

PR #1742, opened 2020-06-29
Test against pre-release versions

huonw:
Of particular note is [TensorFlow 2.3.0-rc0](https://github.com/tensorflow/tensorflow/releases/tag/v2.3.0-rc0), which includes the fix for #1251.


huonw:
Failures:

- #1743
- #1749
- The Conda step still uses TF 2.2.0, and so fails because the xfails for #1251 are removed

PR #1744, opened 2020-06-29
Run interpretability notebooks on CI

huonw:
The switch to Github Actions seems to allow us to run the interpretability notebooks on CI, which we couldn't do with Buildkite easily.

This relates to #819.


huonw:
> This is a nice bonus of actions! Do the containers used supply more ram?

Not entirely sure: we were running on t2.large EC2 instances before (2 CPUs, 8GB RAM):

https://github.com/stellargraph/stellargraph/blob/d4f100591dadc83b1c093fd53a84e21714eff694/.buildkite/pipeline.yml#L126

while Github Actions apparently offers 2 CPUs and only 7GB RAM: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

Issue #1746, opened 2020-06-29
Notebooks fails on Github Actions do not upload artefact


huonw:
### Describe the bug

Our CI tries to upload notebooks as artefacts, to make it easy for a human to understand exactly what went right/wrong. This is currently not working properly: if the notebook fails, the `notebook_name` output used for the uploading step is not set.

### To Reproduce

1. Do a CI run where a notebook fails during execution, e.g. https://github.com/stellargraph/stellargraph/pull/1744/checks?check_run_id=817171162

### Observed behavior

https://github.com/stellargraph/stellargraph/pull/1744/checks?check_run_id=817171162

The "Upload artifact" step fails with:

```
##[error]Value cannot be null. (Parameter 'name')
```

### Expected behavior

The "Upload artifact" step should always run successfully.

### Environment

Operating system: Ubuntu

Python version: 3.6

Package versions: StellarGraph develop: https://github.com/stellargraph/stellargraph/commit/3836ce6271e0220265a3c334164df365d0b7cca9

### Additional context

N/A

PR #1748, opened 2020-06-29
Use a single artifact for all notebooks on CI

huonw:
An alternative exploration for #1746. 

This probably isn't what we want to do long term, because it would probably interfere with adding links to rendered forms via https://nbviewer.jupyter.org .


huonw:
This does result in much cleaner output, with three zip files (no neo4j, neo4j 3.5 and neo4j 4.0) instead of one per notebook:

- This PR (complete output):
  image

- Current develop (truncated):
  image

Issue #1780, opened 2020-07-28
Sparse GAT isn't tested properly


huonw:
### Describe the bug

GAT tries to uise subclassing to share tests between dense and sparse models, but accidentally defines a function instead of a subclass:

https://github.com/stellargraph/stellargraph/blob/ec132647e5cf43ff683e3f2e72e18ac6daa98202/tests/layer/test_graph_attention.py#L625-L627

### To Reproduce

1. Look at `test_graph_attention.py`

### Observed behavior

The sparse tests define a new function that is ignored.

### Expected behavior

The sparse tests should declare a subclass, like `class TestGATsparse(Test_GAT):`

### Environment

Operating system: all

Python version: all

Package versions: StellarGraph ec132647e5cf43ff683e3f2e72e18ac6daa98202

### Additional context

N/A

youph:
Ops

PR #1779, opened 2020-07-28
Un-xfail full-batch saving tests fixed by TensorFlow 2.3

huonw:
[TensorFlow 2.3.0](https://github.com/tensorflow/tensorflow/releases/tag/v2.3.0) was released over the last day. This release includes the fix for https://github.com/tensorflow/tensorflow/issues/38465 (which we filed a duplicate of at https://github.com/tensorflow/tensorflow/issues/40373) which is the underlying issue behind #1251.

As such, we can remove the xfail markings from tests involving the saving of full-batch models like APPNP, GCN and RGCN.

Fixes: #1251

PR #1781, opened 2020-07-28
Test sparse GAT properly

huonw:
This replaces the `TestGATsparse` function with a class, that has the correct subclass behaviour to pick up the tests from the `Test_GAT` subclass. This requires adjusting some tests, but otherwise the code seems to work.

Fixes: #1780


adocherty (approved):
Looks good to me, after fixing the formatting 👍

huonw:
Thanks for the review.

Status: this depends on #1779 so that the test of saving and loading the (sparse) model can work (with TF 2.3), and that seems to be waiting on TF 2.3 being available in conda (https://github.com/stellargraph/stellargraph/pull/1779#issuecomment-664747491).

Issue #1784, opened 2020-08-09
ImportError: cannot import name 'rng_integers' from 'scipy._lib._util'


ZihanChen1995:
### Describe the bug

Hi! Thank you for this great library. I'm trying to install Stellargraph on Ubuntu 18.04. However, after I successfully installed Stellargraph via conda ([follow this instruction](https://stellargraph.readthedocs.io/en/stable/README.html#installation)), when I try to import Stellargraph, this problem happened:

```
ImportError: cannot import name 'rng_integers' from 'scipy._lib._util' (/home/zihan/anaconda3/envs/TF2/lib/python3.7/site-packages/scipy/_lib/_util.py)
```

To avoid the possible problem coming from `scipy`, I reinstalled it, and made sure it successfully installed in this environment:

```
conda list -n TF2

scipy                     1.4.1                    pypi_0    pypi
stellargraph              1.2.1                      py_0    stellargraph
tensorflow                2.2.0           mkl_py37h6e9ce2d_0  
```

Hence I'm quite confused with why this problem happens. Could you please help me with this issue? Thank you a lot in advance!

### Environment

Operating system: Ubuntu 18.04

Python version: 3.7.7

Package versions: stellargraph==1.2.1, tensorflow==2.2.0

***

More detailed error description:
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-11-855dad28d1e4> in <module>
----> 1 import stellargraph as sg
      2 #from stellargraph.data import EdgeSplitter
      3 #from stellargraph.mapper import FullBatchLinkGenerator
      4 #from stellargraph.layer import GCN, LinkEmbedding
      5 # import scipy

~/anaconda3/envs/TF2/lib/python3.7/site-packages/stellargraph/__init__.py in <module>
     37 
     38 # Import modules
---> 39 from stellargraph import (
     40     data,
     41     calibration,

~/anaconda3/envs/TF2/lib/python3.7/site-packages/stellargraph/data/__init__.py in <module>
     21 
     22 # Expose the stellargraph.data classes:
---> 23 from .explorer import *
     24 from .edge_splitter import *
     25 from .node_splitter import *

~/anaconda3/envs/TF2/lib/python3.7/site-packages/stellargraph/data/explorer.py in <module>
     29 import warnings
     30 from collections import defaultdict, deque
---> 31 from scipy import stats
     32 from scipy.special import softmax
     33 

[...]

PR #1785, opened 2020-08-09
Use scipy 1.4.1 exactly

huonw:
This is a test for #1784, which sees some issues that may be connected to scipy==1.4.1.


huonw:
This didn't seem to be the cause of #1784, closing.

Issue #1789, opened 2020-08-14
GraphSage Algorithm Aggregation and Weight order error


aj111000:
### Describe the bug

This isn't specifically a bug, but rather an implementation error. The order that the GraphSage Algorithm is performing the Aggregation and Weight multiplication operations is incorrect.

In the initial paper, the (pre-activation) output for vertex v at layer k is the weight matrix W_k multiplied by the output for vertex v from layer k-1 concatenated with the aggregated feature vector. 

However, stellargraph actually does this in the opposite order, first applying the weights to both the previous output vector and aggregated feature vector and then concatenating the outputs. 

This is a pretty substantial problem, because it means that the first half of the output for each layer is only based on the feature vector and the second half of the output is only based on the aggregated vector

For example, let's say that the feature vector x and aggregated vector y are both 5x1 and we want our output to be 2x1. In the standard graphsage algorithm, our weight matrix would be of the form W_k=[A B; C D] where A, B, C, and D are all 1x5 matrices, and the output would be W_k*[x;y]=[A*x+B*y; C*x+D*y], so both x and y impact both entries in the output. However, in the current implementation, we instead have two 5x1 weights X and Y and our output is [X*x; Y*y], meaning that y has no influence over the first half of the output and x has no influence over the second half of the output. 

(Note: I know that the implementation is actually a row vector multiplied by a matrix and not a matrix multiplied by a column vector, I was just using it for clarification)

I don't want this to seem like I'm being critical, you guys have made an incredible piece of software, and I only noticed this because I've been having to do a lot of work with some of the lower level functions.

### To Reproduce

Just look at the GraphSage Aggregator class

### Observed behavior

See description

### Expected behavior

See description

### Environment

N/A

### Additional context

N/A

aj111000:
There's also a relatively easy fix that I can see. If you redefined the weights so that each weight is nxm (n for the size of the input and m for the size of the output), and then used an add operation instead of a concatenate operation, then that would solve the problem. That would be the equivalent of writing [A B; C D][x; y] = [A; C]x + [B; D]y, which is both equivalent to the original paper and just as modular as the current implementation.

huonw:
Thanks for filing an issue. I'm not in a position to dig into the details right now, but we really appreciate that you've done so. If you feel like it, we'd be more than happy to merge a pull request with a fix, otherwise, I will try to look at this in future, but i can't guarantee when!

aj111000:
[...]

PR #1790, opened 2020-08-21
Fix for graphsage weight concatenation issue

aj111000:
I went into depth about the issue in https://github.com/stellargraph/stellargraph/issues/1789, so go there if you want to see more of the gory details, but I'll try to give a summary here.

Essentially, in the original graphsage paper, they first concatenate the previous layer's output with the aggregated feature vector and then applies a fully connected layer to the whole thing. However, in the current stellargraph implementation, weights are first applied to each vector and then the results are concatenated. 

There are two main issues with this implementation. First, the network has about half as many parameters as it should, which leads to a lack of expressiveness. Secondly, and much more importantly, it means the first half of the output for each layer is ONLY based on the feature vector, and the second half is ONLY based on the aggregated vector, meaning that the network lacks the ability to gain any information about their interplay (at least within one layer).

I fixed this in a very simple way. First of all, I changed the output sizes of all the weights from outputdim/2 (plus some remainder terms) to just being outputdim. Next, I replaced the concatenation operation with a sum operation, which is the equivalent of saying that if A=[X,Y] and v = [x;y], then Av=[X,Y][x;y]=Xx+Yy, which is completely equivalent

I made no other changes, so it should still work with all other uses of the graphsage model

### DIFF
--- a/stellargraph/layer/graphsage.py
+++ b/stellargraph/layer/graphsage.py
@@ -147,15 +147,11 @@ class GraphSAGEAggregator(Layer):
                 "There must be at least one input with a non-zero neighbourhood dimension"
             )
 
-        # Calculate the dimensionality of each group, and put remainder into the first group
-        # with non-zero dimensions, which should be the head node group.
-        group_output_dim = self.output_dim // num_groups
-        remainder_dim = self.output_dim - num_groups * group_output_dim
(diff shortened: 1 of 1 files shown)


aj111000:
I'm going to fix the whitespace error and resubmit

PR #1792, opened 2020-08-21
Fix for the graphsage feedforward algorithm

aj111000:
I went into depth about the issue in #1789, so go there if you want to see more of the gory details, but I'll try to give a summary here.

Essentially, in the original graphsage paper, they first concatenate the previous layer's output with the aggregated feature vector and then applies a fully connected layer to the whole thing. However, in the current stellargraph implementation, weights are first applied to each vector and then the results are concatenated.

There are two main issues with this implementation. First, the network has about half as many parameters as it should, which leads to a lack of expressiveness. Secondly, and much more importantly, it means the first half of the output for each layer is ONLY based on the feature vector, and the second half is ONLY based on the aggregated vector, meaning that the network lacks the ability to gain any information about their interplay (at least within one layer).

I fixed this in a very simple way. First of all, I changed the output sizes of all the weights from outputdim/2 (plus some remainder terms) to just being outputdim. Next, I replaced the concatenation operation with a sum operation, which is the equivalent of saying that if A=[X,Y] and v = [x;y], then Av=[X,Y][x;y]=Xx+Yy, which is completely equivalent

I made no other changes, so it should still work with all other uses of the graphsage model

As a part of that, I had to change some of the tests that were related to how graphsage models were initialized (i.e., the ones that test the weight sizes for an initialized graphsage model)


huonw:
Thanks for the pull request and thanks for waiting. This looks good, although I'm following up with the authors of this implementation to see if there's some historical reason for this approach.

aj111000:
Alright, sounds good! Thanks a lot

PR #1833, opened 2020-12-20
Update gcn-lstm-time-series.ipynb

andife:
typos


huonw (approved):
Thank you very much for reading over this in fine detail too. CI should be fixed with #1840 and #1842 but I'm not sure if it's going to apply to this PR properly. I've retried the build, but may have to push some extra commits too.

PR #1842, opened 2021-01-14
Add fixed IDs to each notebook cell

huonw:
nbformat recently released 5.1.0, which includes randomly generated cell IDs in every cell: https://nbformat.readthedocs.io/en/latest/changelog.html#id2

> - Implemented CellIds from [JEP-62](https://github.com/jupyter/enhancement-proposals/blob/master/62-cell-id/cell-id.md)

This thus requires us to update our notebooks, or else our CI validation of their formatting fails (https://github.com/stellargraph/stellargraph/runs/1699039930?check_suite_focus=true). Using the default mode doesn't work, because they seem to be randomly generated on every run (even after saving them once), and thus the 'formatting' changes each time `format_notebook.py` is run. This PR thus implements a basic hashing scheme to give fixed IDs for a cell:

- hash the cell's source code with SHA256 and use the first 8 hex digits (e.g. `abcd1234`)
- add a counter if this is not the first cell with the given source code in the notebook (e.g. `abcd1234-1`)

This is idempotent: reformatting a notebook without changes will give the same result. Thus, it can be easily used as a check on CI.


scripts/format_notebooks.py
```diff
@@ -289,6 +291,50 @@ def preprocess(self, nb, resources):
...
+        return out_nb, out_resources
+
+    def preprocess_cell(self, cell, resources, cell_index):
+        cell = copy.deepcopy(cell)
+        source = cell["source"]
+
+        # the core of the ID is a hash of the source code: this will be the same across every time a
+        # given version of the notebook is reformatted.
+        hasher = hashlib.sha256()
+        hasher.update(source.encode())
```
adocherty: Won't this be different when code is reformatted?

scripts/format_notebooks.py
```diff
@@ -289,6 +291,50 @@ def preprocess(self, nb, resources):
...
+        assert len(ids) == len(set(ids))
+
+        return out_nb, out_resources
+
+    def preprocess_cell(self, cell, resources, cell_index):
+        cell = copy.deepcopy(cell)
+        source = cell["source"]
+
+        # the core of the ID is a hash of the source code: this will be the same across every time a
+        # given version of the notebook is reformatted.
```
adocherty: This comment is a little confusing - you mean this will generate the same IDs every time it's run on the same input notebook, not that it's idempotent when the notebook is reformatted?

adocherty (approved):
I'm glad they have moved to add unique IDs for cells!  The code looks good to me, just a minor thing that a comment is ambiguous.

Is there any reason why you are going with hashes, rather than something really simple like an integer count (with no need for collision checking) such the following option in the JEP?

https://github.com/jupyter/enhancement-proposals/blob/master/62-cell-id/cell-id.md#option-a-strings-from-an-integer-counter

scripts/format_notebooks.py
huonw: It's true that it depends on the exact source (including spacing etc.) but this runs after `FormatCodeCellPreprocessor` and so the input code here will always be formatted appropriately.

scripts/format_notebooks.py
huonw: This preprocessor is individually idempotent because it's not editing the source, and `format_notebooks.py --default` as a whole is idempotent because each cell's source is formatted before computing these ids, and so it's using the same source as it would when run a second time (where that second time wouldn't have to reformat individual cells).

huonw (commented):
> Is there any reason why you are going with hashes, rather than something really simple like an integer count (with no need for collision checking) such the following option in the JEP?

It's true that's an option. I decided that this was slightly more resilient, allowing cells to be inserted/deleted without changing the ids of the unchanged cells (except for the edge case when they have the same source as something else). On the other hand, this hashing version changes unnecessarily if the source is edited but the notebook structure is otherwise left unchanged.

I'll switch to the simpler form that you suggest 👍

PR #1853, opened 2021-01-28
Fix minor bugs in docs

nalinbhardwaj:
Fixes GCN link prediction demo to use test_gen for test_flow generation. If I understand everything correctly, this change doesn't actually impact runtime, but prevents any confusion since `test_gen` is otherwise an unused variable in the demo right now.

Fixes typo in `GraphSAGELinkGenerator` in docstring which propagates to ReadDocs.


huonw (approved):
Great catches. Thank you for taking the time to create a pull request!

PR #1866, opened 2021-02-09
Improve efficiency of neighborhood sampling query

m0baxter:
Improve efficiency of neighborhood sampling query when the central node has many neighbors. The orignal query made 2 * ( number of neighbors ) database hits, the new version makes ( number of neighbors ) + num_samples database hits  (less repeated ids in the sampled nodes).

In the particular case I tested on `node_label = "entity"` `id_property = "id"` and the graph is undirected ( `direction_arrow = "--"` ). Also, `num_samples` in these examples is set to 25.

The two version of the query are essentially:

old:
```
MATCH (cur_node:entity) WHERE cur_node.id = $some_node_id
MATCH (cur_node){direction_arrow}(neighbors)

// put all neighbours into a list and pull ids:
WITH CASE collect(neighbors.id) WHEN [] THEN [null] ELSE collect(neighbors.id) END AS in_neighbors_list

// pick random neighbours with replacement:
WITH apoc.coll.randomItems(in_neighbors_list, 25, True) AS in_samples_list

RETURN in_samples_list
```

new:
```
// Pull neighbours of given node:
MATCH (cur_node:entity) WHERE cur_node.id = $some_node_id
MATCH (cur_node)--(neighbors)

// put all neighbours into a list:
WITH CASE collect(neighbors) WHEN [] THEN [null] ELSE collect(neighbors) END AS in_neighbors_list

// pick random neighbours with replacement:
WITH apoc.coll.randomItems(in_neighbors_list, 25, True) AS sampled

// unpack sampled node list:
unwind sampled as nn

// collect ids of sampled nodes:
WITH CASE collect( nn.id ) WHEN [] THEN sampled ELSE collect( nn.id ) END AS in_samples_list

RETURN in_samples_list
```
In a case where the `some_node_id` has ~1,700,000 neighbours the profiler says:

old:
```
Cypher version: CYPHER 4.2, planner: COST, runtime: INTERPRETED. 3364764 total db hits in 2250 ms.
```

new:
```
Cypher version: CYPHER 4.2, planner: COST, runtime: INTERPRETED. 1682409 total db hits in 1480 ms.
```

An example where the node has no neighbours:

old:
```
Cypher version: CYPHER 4.2, planner: COST, runtime: INTERPRETED. 4 total db hits in 86 ms.
```

new:
```
Cypher version: CYPHER 4.2, planner: COST, runtime: INTERPRETED. 4 total db hits in 0 ms.
```

### DIFF
--- a/stellargraph/connector/neo4j/sampler.py
+++ b/stellargraph/connector/neo4j/sampler.py
@@ -58,7 +58,7 @@ def _bfs_neighbor_query(sampling_direction, id_property, node_label=None):
 
             // pick random nodes with replacement
             WITH apoc.coll.randomItems(in_neighbors_list, $num_samples, True) AS sampled
-            
+
             // pull the ids of the sampled nodes only
             unwind sampled as nn
             WITH CASE collect(nn.{id_property}) WHEN [] THEN sampled ELSE collect(nn.{id_property}) END AS in_samples_list
@@ -66,7 +66,7 @@ def _bfs_neighbor_query(sampling_direction, id_property, node_label=None):
             RETURN in_samples_list',
             {{ node_id: node_id, num_samples: $num_samples  }}) YIELD value
 
-        RETURN apoc.coll.flatten(collect(value.in_samples_list)) AS sampled
+        RETURN apoc.coll.flatten(collect(value.in_samples_list)) AS next_sample
         """
 
 
--- a/stellargraph/connector/neo4j/sampler.py
+++ b/stellargraph/connector/neo4j/sampler.py
@@ -53,16 +53,20 @@ def _bfs_neighbor_query(sampling_direction, id_property, node_label=None):
             // find the neighbors
             MATCH (cur_node){direction_arrow}(neighbors)
 
-            // put all ids into a list
-            WITH CASE collect(neighbors.{id_property}) WHEN [] THEN [null] ELSE collect(neighbors.{id_property}) END AS in_neighbors_list
+            // collect neighbors in a list
+            WITH CASE collect(neighbors) WHEN [] THEN [null] ELSE collect(neighbors) END AS in_neighbors_list
 
             // pick random nodes with replacement
-            WITH apoc.coll.randomItems(in_neighbors_list, $num_samples, True) AS in_samples_list
+            WITH apoc.coll.randomItems(in_neighbors_list, $num_samples, True) AS sampled
+            
+            // pull the ids of the sampled nodes only
+            unwind sampled as nn
+            WITH CASE collect(nn.{id_property}) WHEN [] THEN sampled ELSE collect(nn.{id_property}) END AS in_samples_list
 
             RETURN in_samples_list',
             {{ node_id: node_id, num_samples: $num_samples  }}) YIELD value
 
-        RETURN apoc.coll.flatten(collect(value.in_samples_list)) as next_samples
+        RETURN apoc.coll.flatten(collect(value.in_samples_list)) AS sampled
         """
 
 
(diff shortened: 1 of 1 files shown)


stellargraph/connector/neo4j/sampler.py
```diff
@@ -53,16 +53,20 @@ def _bfs_neighbor_query(sampling_direction, id_property, node_label=None):
...
-            WITH CASE collect(neighbors.{id_property}) WHEN [] THEN [null] ELSE collect(neighbors.{id_property}) END AS in_neighbors_list
+            // collect neighbors in a list
+            WITH CASE collect(neighbors) WHEN [] THEN [null] ELSE collect(neighbors) END AS in_neighbors_list
 
             // pick random nodes with replacement
-            WITH apoc.coll.randomItems(in_neighbors_list, $num_samples, True) AS in_samples_list
+            WITH apoc.coll.randomItems(in_neighbors_list, $num_samples, True) AS sampled
+
+            // pull the ids of the sampled nodes only
+            unwind sampled as nn
```
huonw: Totally minor, but (as you note with the `as` -> `AS` below), this code is defaulting to capitalising the keywords:

```suggestion
            UNWIND sampled AS nn
```

stellargraph/connector/neo4j/sampler.py
```diff
@@ -53,16 +53,20 @@ def _bfs_neighbor_query(sampling_direction, id_property, node_label=None):
...
+            WITH CASE collect(nn.{id_property}) WHEN [] THEN sampled ELSE collect(nn.{id_property}) END AS in_samples_list
```
huonw, line 65: I've lost the context for this code, so could you remind me why this needs to do the same `WITH CASE ...` code as the `neighbors` one above? Does something like the following work?

```suggestion
            WITH collect(nn.{id_property}) AS in_samples_list
```

Or even

```suggestion
            RETURN collect(nn.{id_property}) AS in_samples_list
```

and delete the `RETURN in_samples_list` below.

huonw (changes requested):
This looks great, and that's a big benefit for such a minor change. Thank you for taking the time to investigate and submit a pull request!

I've got a minor adjustment, and a question, but otherwise I think this looks pretty good 👍

stellargraph/connector/neo4j/sampler.py
m0baxter: the problem with this is that when everything in `sampled` is null this will return a singleton list `[ null ]` rather than what you need down stream which is a list that is `num_samples` long. I realize it looks a little convoluted but I couldn't think of a way to write the statement in a cleaner way.

huonw: Ah, cool. I didn't realise that [collect()](https://neo4j.com/docs/cypher-manual/current/functions/aggregating/#functions-collect) ignores `null`, which makes this all make sense!

Arguably one could very clean it up with an additional `WITH`, but I don't think that's necessary (this also optionally applies to the `collect(neighbors)` code above).

```suggestion
            WITH collect(nn.{id_property}) AS in_samples_list
            // collect ignores nulls, so re-handle the no-neighbours case to ensure we get [null, null, ...] output
            WITH CASE in_samples_list WHEN [] THEN sampled ELSE in_samples_list END AS in_samples_list
```

huonw (approved):
Looks great; thanks for responding so quickly.

I've left an additional optional suggestion, feel free to add it (or not), or else I'll just merge tomorrow if I don't hear from you. It's not important enough to delay landing this :)

m0baxter:
Seems like that last change broke something. It seems like `sampled` is somehow going out of scope. I just reverted it to the older version.

huonw:
Ah, sorry about that. Thanks again!

PR #1880, opened 2021-02-24
Fixing test flakiness

sleepy-owl:
The test `test_poincare_ball_distance_self` is flaky. This PR proposes a fix for this issue.

To find a solution, I collected samples of value of `d` in the assertion from several test executions and computed the tail distribution. I computed the extreme percentiles to check how high can the relative differences be.

```
0.9:: 1e-07
0.99:: 1e-06
0.9999:: 1e-06
```

For this fix I chose the 99.99th percentile which indicates the relative difference can go up to `1e-06` instead of the default `1e-07` for `assert_allclose`. I think setting the bound using the statistical evaluation might be a good way to ensure the test is not flaky.

Do you guys think this makes sense? Please let me know if this looks good or if you have any other suggestions. Also, here I assume there are no bugs in the code under test.

### DIFF
--- a/tests/utils/test_hyperbolic.py
+++ b/tests/utils/test_hyperbolic.py
@@ -69,7 +69,7 @@ def test_poincare_ball_distance_self(seeded):
 
         d = poincare_ball_distance(c, vs, vs)
         assert d.shape == vs.shape[:-1]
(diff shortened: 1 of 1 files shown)


huonw (approved):
It's unfortunate that this is flaky, but this seems like a good solution. Thank you.

PR #1986, opened 2021-10-11
Fix flaky test_normalized_laplacian in tests.core.test_utils

sgrigory:
Closes #1160 

Test `test_normalized_laplacian` in `tests.core.test_utils` is flaky, this PR fixes it.

## Reason for flakiness
The test was failing because it assumed the sum of elements of normalized asymmetric Laplacian should be 1. In fact this sum is equal to _the number of degree 0 nodes_. Most of the time the graphs generated by `example_graph_random` had one such node and the current test passed. However, sometimes there are 2 or more of them, and then the test failed

## Why the new test works
The new test checks that sum of elements of normalized asymmetric Laplacian is equal to the number of degree 0 nodes. 
Let's show that this always holds. Indeed, Laplacian is computed as `L = I - D * A`, where `A` is the adjacency matrix and `D` is diagonal matrix whose elements are 1 over row sums of `A` https://github.com/stellargraph/stellargraph/blob/e46af15c2dbcaa9ae6328dd1ec62c6f9ef7f5467/stellargraph/core/utils.py#L152

If the graph doesn't have degree 0 nodes, all rows in `A` have non-zero sums, so sums of each row in `D * A` is equal to 1, and the sum of elements in `L` is equal to zero.
If, however, a certain node is degree 0, the corresponding row contains only zeros. Then the corresponding element of `D` is `inf`, but multiplied by `A` is computed to 0 in https://github.com/stellargraph/stellargraph/blob/e46af15c2dbcaa9ae6328dd1ec62c6f9ef7f5467/stellargraph/core/utils.py#L135
So the corresponding row of `D * A` sums to 0 instead of 1. 

Thus each degree 0 node adds 1 to the sum of elements of `L`

## Tests
The following code runs without errors on the PR branch (and fails on `develop`)
```
from tests.core.test_utils import example_graph_random, test_normalized_laplacian
for i in range(10000):
    random_graph = example_graph_random()
    test_normalized_laplacian(random_graph)
```

PR #1987, opened 2021-10-16
Fix flaky test_normalize_adj in tests.core.test_utils

### LINKED ISSUE #1678: Flaky test: test_normalize_adj in tests.core.test_utils
### Describe the bug

This test sometimes fails in a flaky manner:

- https://buildkite.com/stellar/stellargraph-public/builds/4237 (twice)
- https://buildkite.com/stellar/stellargraph-public/builds/4148
- https://buildkite.com/stellar/stellargraph-public/builds/3967
- https://buildkite.com/stellar/stellargraph-public/builds/3837

```
E       assert 5 == 4.0 ± 4.0e-01
E        +  where 4.0 ± 4.0e-01 = (4.0, 0.1)
E        +    where  = pytest.approx
E        +    and   4.0 = ()
E        +      where  = matrix([[0.22222222, 0.44444444, 0.33333333, 0.        , 0.        ,\n         0.        ],\n        [0.5       , 0.125 ...    ,\n         0.        ],\n        [0.        , 0.        , 0.        , 0.        , 0.        ,\n         0.        ]]).sum

tests/core/test_utils.py:44: AssertionError
```

Commit of most recent failure: https://github.com/stellargraph/stellargraph/commit/ee3bd69435ba0a7b630dd0ae566f095c99167e66

This has failed since other fixes for this test: #996 (#997).

sgrigory:
Closes #1678 

This PR fixes `test_normalize_adj` in `tests.core.test_utils` which was failing in a flaky way. 

The test was occasionally failing in two places:
- Comparing the top eigenvalue of symmetrically normalized adjacency matrix with 1. See, for example, https://buildkite.com/stellar/stellargraph-public/builds/3837
- Checking the sum of elements of non-symmetrically normalized adjacency matrix. See, for example, https://buildkite.com/stellar/stellargraph-public/builds/3967

The first point of failure seems to be the question of precision when calculating eigenvalues. The current error threshold equal to 1e-7 is too low. I ran 100000 iterations of the test and got 14963 occasions when the error is > 1e-8 and 33 occasions when the error is > 1e-7. By setting the threshold to 1e-5 we would make it virtually impossible for the test to fail. 

The second point of failure is similar to https://github.com/stellargraph/stellargraph/pull/1986. The sum of elements of the normalized adjacency matrix should be compared to the number of nodes of positive degree (each row contributes 1 except for those corresponding to degree 0 nodes). For graphs generated with `example_graph_random` this is equal to 5 most of the time, but not always -  that's why the test was failing

I also remove unused import in `stellargraph/data/epgm.py`, which is not in the dependencies and caused many CI jobs to fail - let me know if you would prefer to keep it and just add to the dependencies.

## Tests

The following code runs without errors:
```
from tests.core.test_utils import example_graph_random, test_normalize_adj
for i in range(100000):
    random_graph = example_graph_random()
    test_normalize_adj(random_graph)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -339,6 +339,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: s-weigand/setup-conda@v1
+        with:
+          python-version: 3.8
 
       - name: Install conda build tools
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -339,8 +339,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: s-weigand/setup-conda@v1
-        with:
-          python-version: 3.8
 
       - name: Install conda build tools
         run: |
--- a/stellargraph/data/epgm.py
+++ b/stellargraph/data/epgm.py
@@ -21,7 +21,6 @@ import os
 import json
 import uuid
 import numpy as np
-import chardet
 import scipy.sparse as sp
 
 import multiprocessing
--- a/tests/core/test_utils.py
+++ b/tests/core/test_utils.py
@@ -184,14 +184,13 @@ def test_smart_array_concatenate_broadcast():
         _check_smart_concatenate([a0, c43])
 
 
-@flaky_xfail_mark(AssertionError, 1678)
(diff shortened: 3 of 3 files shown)


huonw (approved):
Looks good, thank you. Deleting the import is fine.

